PR #212, opened 2026-02-18
Integrate config-service with novu-bridge for template resolution

Lokendra-egov:
- Fix ConfigServiceClient to correctly parse resolve response (resolved.entry.value path and RequestInfo casing)
- Add seed script for NOTIF_TEMPLATE_MAP config entry with templateKey, twilioContentSid, paramOrder, and templateBody
- Update novu-bridge application.properties with correct service URLs
- Disable preference check until user-preferences-service is wired up

### DIFF
--- a/backend/digit-config-service/src/main/java/org/egov/config/web/model/RequestInfo.java
+++ b/backend/digit-config-service/src/main/java/org/egov/config/web/model/RequestInfo.java
@@ -6,6 +6,8 @@ import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 
+import java.util.List;
+
 @Data
 @AllArgsConstructor
 @NoArgsConstructor
@@ -39,6 +41,9 @@ public class RequestInfo {
     @JsonProperty("userInfo")
     private UserInfo userInfo;
 
+    @JsonProperty("plainAccessRequest")
+    private Object plainAccessRequest;
+
     @Data
     @AllArgsConstructor
     @NoArgsConstructor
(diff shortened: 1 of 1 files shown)


backend/digit-config-service/src/main/java/org/egov/config/repository/ConfigEntryRepository.java
```diff
@@ -0,0 +1,108 @@
...
+@RequiredArgsConstructor
+public class ConfigEntryRepository {
+
+    private final JdbcTemplate jdbcTemplate;
+    private final ConfigEntryQueryBuilder queryBuilder;
+    private final ConfigEntryRowMapper rowMapper;
+    private final ObjectMapper objectMapper;
+
+    public void save(ConfigEntry entry) {
+        String sql = "INSERT INTO config_entry (id, config_code, module, event_type, channel, tenant_id, locale, enabled, " +
```
pradeepkumarcm-egov: Need to change table name. Follow the standard used in digit. Append eg_config as table name.

backend/digit-config-service/src/main/java/org/egov/config/repository/querybuilder/ConfigEntryQueryBuilder.java
```diff
@@ -0,0 +1,149 @@
...
+
+import org.springframework.stereotype.Component;
+
+import java.util.List;
+
+@Component
+public class ConfigEntryQueryBuilder {
+
+    private static final String BASE_SELECT = "SELECT id, config_code, module, event_type, channel, " +
+            "tenant_id, locale, enabled, \"value\", revision, " +
```
pradeepkumarcm-egov: “value” is currently used as the column name. Please confirm whether this column should be renamed. If “value” conflicts with any system variable or reserved keyword, it should be changed to configValue. @Lokendra-egov

PR #235, opened 2026-03-06
removed twilio hardcode as provider and added encryption and decryption in config-service

nitish-egov:
## Release Notes

* **New Features**
  * Sensitive configuration data is now automatically encrypted and decrypted based on schema definitions
  * Messaging system now dynamically resolves providers per tenant and channel with support for provider-specific credentials

* **Configuration**
  * Added crypto service integration with configurable endpoints for encryption and decryption operations

### DIFF
--- a/backend/digit-config-service/src/main/java/org/egov/config/utils/EncryptionDecryptionUtil.java
+++ b/backend/digit-config-service/src/main/java/org/egov/config/utils/EncryptionDecryptionUtil.java
@@ -119,6 +119,7 @@ public class EncryptionDecryptionUtil {
             
             // For config service, use null RequestInfo as we don't need user-specific decryption logic
             // The schema-based security fields determine what needs to be decrypted
+            @SuppressWarnings("unchecked")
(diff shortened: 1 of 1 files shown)


backend/digit-config-service/src/main/java/org/egov/config/service/enrichment/ConfigDataEnricher.java
```diff
@@ -73,4 +99,74 @@ private String resolveUserId(RequestInfo requestInfo) {
...
+                return data;
+            }
+
+            // Encrypt the values
+            JsonNode encryptedValues = cryptoClient.encryptValues(tenantId, valuesToEncrypt);
+            
+            // Replace original values with encrypted ones using correct field order
+            return replaceWithEncryptedValuesInOrder(data, fieldOrder, encryptedValues);
+
+        } catch (Exception e) {
```
vinothrallapalli-eGov: Generic exceptions should be avoided. Use specific exception types so that the error can be handled and understood more accurately.

PR #302, opened 2026-04-06
integrated the complaintSource in the PGR while creating

priyanshu-egov:


backend/pgr-services/src/main/java/org/egov/pgr/web/models/Service.java
```diff
@@ -76,7 +76,6 @@ public class Service   {
         @JsonProperty("applicationStatus")
         private String applicationStatus = null;
 
-        @NotNull
```
vinothrallapalli-eGov, line 79: Source will be mandatory, so revert this changes

vinothrallapalli-eGov (approved):

PR #403, opened 2026-04-20
 Refactor complaint filtering and add boundary level localizations

priyanshu-egov:
- Update complaint type filtering to use menuPath instead of department field
- Change serviceCode references from code to serviceCode for consistency
- Add validation requiring minimum two boundary levels in hierarchy loader
- Add comprehensive test coverage for hierarchy loading with localization verification


Hari-egov (approved):
Approved

PR #407, opened 2026-04-20
fix: boundary level config not applied when MDMS data missing (#406)

### LINKED ISSUE #406: Nai Pepea: CMS: CSR Employee unable to create complaint
**Describe the bug**
CSR Employee unable to create complaint .

**To Reproduce**
- Login with CSR user 
- Go to Create Complaint 
- Fill the form and submit 
Actual:
- Boundary data not displayed in UI 
- "LOWEST_LEVEL_CONFIG_NOT_PRESENT" error displayed 
- Upon submit, this error is displayed "ES_COMMON_PLEASE_SELECT_ALL_BOUNDARY_LEVELS"

https://github.com/user-attachments/assets/f04a57ce-437a-4625-915a-7eb94bdfb71f

ChakshuGautam:
## Summary

Fixes #406 in egovernments/Citizen-Complaint-Resolution-System

The `BoundaryFilter` component in the employee Create Complaint form shows `LOWEST_LEVEL_CONFIG_NOT_PRESENT` instead of rendering boundary dropdowns. This happens in deployments where the `CMS-BOUNDARY.HierarchySchema` MDMS data does not exist.

## Root Cause

In `createComplaintForm.js`, the boundary config block was gated behind `hierarchyData` being truthy:

```js
if (field.key === "boundaryComponent" && hierarchyData) {
```

When `CMS-BOUNDARY.HierarchySchema` MDMS data doesn't exist (common in new/simple deployments that rely on `globalConfigs` for boundary configuration), `hierarchyData` is `null`. This causes the entire config block to be skipped, meaning `lowestLevel`, `highestLevel`, and `hierarchyType` are never set on the field populators. The `BoundaryFilter` component then renders the error message `LOWEST_LEVEL_CONFIG_NOT_PRESENT`.

## Fix

1. **Remove the `&& hierarchyData` guard** from the `if` condition so the config block always runs for `boundaryComponent` fields.
2. **Use optional chaining** (`hierarchyData?.lowestHierarchy` instead of `hierarchyData.lowestHierarchy`) so that when `hierarchyData` is null, it gracefully falls through to the `globalConfigs` values.

This ensures the fallback chain always works: MDMS data -> globalConfigs -> hardcoded defaults ("Ward"/"City"/"ADMIN").

### Changed file
- `frontend/micro-ui/web/micro-ui-internals/packages/modules/pgr/src/pages/employee/CreateComplaint/createComplaintForm.js`

## Test Plan

- [ ] Deploy on a tenant **without** `CMS-BOUNDARY.HierarchySchema` MDMS data configured
- [ ] Navigate to Employee > Create Complaint
- [ ] Verify boundary dropdowns render correctly using `globalConfigs` values (PGR_BOUNDARY_LOWEST_LEVEL / PGR_BOUNDARY_HIGHEST_LEVEL)
- [ ] Deploy on a tenant **with** `CMS-BOUNDARY.HierarchySchema` MDMS data configured
- [ ] Verify boundary dropdowns still render correctly using MDMS values (no regression)
- [ ] Verify the boundary hierarchy type is applied correctly in both scenarios


ChakshuGautam:
Closing — base branch drift caused 243 files in diff. Will recreate with clean branch from develop.

PR #566, opened 2026-05-08
Develop Merge From CCRS

Hari-egov:
Develop Merge From CCRS


Hari-egov (commented):
Hey @vinothrallapalli-eGov 
can you review this PR

frontend/micro-ui/web/micro-ui-internals/packages/modules/pgr/src/components/MobileNumberWithPrefix.js
```diff
@@ -0,0 +1,235 @@
...
+
+  return (
+    <div style={{ maxWidth: "37.5rem", width: "100%" }}>
+      <div
+        style={{
+          display: "flex",
+          alignItems: "center",
+          border: localError ? "2px solid #d4351c" : "1px solid #464646",
+          overflow: "hidden",
+          backgroundColor: "#FFFFFF",
```
vinothrallapalli-eGov, line 176: Coloring and theming should use from CSS, try to avoid inline CSS

vinothrallapalli-eGov (changes requested):
## Code Review — PR #566 (Develop Merge From CCRS)

### Overview

This PR bumps the PGR module from `1.0.26` → `1.0.32` and delivers several bug fixes across 6 versions:
- A new `MobileNumberWithPrefix` component to replace ad-hoc mobile input with a standardised, MDMS-driven one with country code prefix dropdown
- `countryCode` now correctly flows through to the create-complaint API payload (both citizen and employee paths)
- Inbox search now properly includes `countryCode` in search criteria
- Employee complaint details: citizen-only workflow actions (RATE, REOPEN, COMMENT) are hidden for SUPERUSER accounts
- `boundaryHierarchy` now safely handles JSON-stringified values
- Dependency bumps: `digit-ui-react-components` 1.8.21→1.9.3, `digit-ui-module-core` 1.9.21→1.9.23, `hrms` 1.9.8→1.9.10, `cms` 1.0.25→1.0.32
- axios pinned to `0.27.2` with IE11 transpilation via webpack

---

### Critical Issues

**1. React Hook called outside a component — Rules of Hooks violation**

`createComplaintForm.js:577`:
```js
updatedData.countryCode = window?.Digit?.Hooks?.pgr?.useMobileValidation?.().validationRules?.prefix || "+91";
```
This is inside an `onFormValueChange` callback (not a component or custom hook). Calling a hook inside a regular function violates the Rules of Hooks and will either silently return stale/incorrect data or crash unpredictably. The `countryCodeRef.current` pattern introduced elsewhere in the same file should be used here instead — it is already the right solution.

---

### Medium Issues

**2. `window.__PGR_INBOX_COUNTRY_CODE__` as shared state**

Using a window global to communicate country code between `MobileNumberWithPrefix` and `UICustomizations` is fragile:
- It leaks across page navigations (SPA context)
- It breaks if two instances of the component exist on the same page
- It is invisible to React's render cycle

A React ref or context passed through the form config (like `countryCodeRef` used in `createComplaintForm.js`) is the correct pattern — that approach is already used for the employee create flow.

**3. `MobileNumberWithPrefix.js` uses ES5-style `var` throughout**

The entire component is written with `var` destructuring of `useState` results:
```js
var _prefix = useState(validationRules.prefix || "+91");
var selectedPrefix = _prefix[0];
```
The rest of the codebase uses `const`/`let` and array destructuring. This is inconsistent and harder to read.


**4. `readOnly` vs `disabled` inconsistency on the mobile input**

In `MobileNumberWithPrefix.js`:
```jsx
<select disabled={disable} ...>
<input readOnly={disable} ...>
```
`readOnly` still allows focus and form submission; `disabled` prevents both. This asymmetry means a "disabled" state behaves differently for the two sub-fields, which can confuse form library integrations. Use `disabled` on both.

---

### Minor Issues

**5. CHANGELOG.md has a duplicate header**

Lines 61–62 repeat the `# Changelog / All notable changes...` header that already exists at the top of the file. The file now has two identical headers.

**6. `.env.sample` deleted with no explanation**

Removing the sample environment file makes developer onboarding harder. If intentional (moved to docs or superseded), please note it in the PR description.

**7. `useEffect` with dynamic key in dependency array** (`MobileNumberWithPrefix.js`)

```js
}, [formData[fieldKey], formData.countryCode]);
```
Dynamic property access in dependency arrays (`formData[fieldKey]`) causes ESLint `exhaustive-deps` warnings and can produce stale closure issues when `fieldKey` itself changes.

**8. `getFormAccessors` prop on FormComposerV2** (`createComplaintForm.js`)

```jsx
getFormAccessors={({ getValues }) => { getValuesRef.current = getValues; }}
```
If `FormComposerV2` does not officially expose this prop it will silently no-op. The `getValues` capture via `onFormValueChange` (already done a few lines above) is the reliable path — this extra accessor appears redundant.

**9. No-op ES6 shorthand → explicit conversions** (`FormExplorer.js`, `fetchDataAndSetParams.js`, `utils/index.js`)

```js
// Before: boundaryHierarchy,
// After:  boundaryHierarchy: boundaryHierarchy,
```
These are functionally identical. They add noise to the diff without any change in behaviour.

---

### What's Done Well

- The `countryCodeRef` pattern in `createComplaintForm.js` is a clean way to bypass react-hook-form's unregistered field limitation
- Converting `tenantId` to a getter in `CreateComplaintConfig.js` is an elegant fix for the stale value bug
- `useMobileValidation.js` improvements (filtering by `fieldType === "mobile"`, using `isDefault` flag) are correct and well-reasoned
- The `isCitizenOnlyAction` helper in `PGRDetails.js` is clean and the SUPERUSER action filtering logic is correct
- `boundaryHierarchy` JSON-parse safety wrappers are defensive and backward-compatible

---

**Must fix before merge:** Item #1 (Rules of Hooks violation). Items #2–#5 are strongly recommended. Items #6–#10 can be addressed in a follow-up.

frontend/micro-ui/web/micro-ui-internals/packages/modules/pgr/src/components/MobileNumberWithPrefix.js
Hari-egov: @vinothrallapalli-eGov 
Its Custom component,
have no dependency with others

PR #577, opened 2026-05-13
prepend country code to mobile number before publishing SMS

Lokendra-egov:


pradeepkumarcm-egov (approved):

pradeepkumarcm-egov (commented):
Check null condition in line 106,111 lines. If empUser is not null then use mobile number and country code.

Issue #567, opened 2026-05-09
[Bug] Ansible deployment fails at docker compose up


subhashini-egov:
**Describe the bug**
Ansible deployment fails at docker compose up with a non-zero return code. See stack below. However, services are up, Gatus dashboard is up etc..Postman collection fails to run. 

```
TASK [Start DIGIT stack] ********************************************************************************************************************************************
[ERROR]: Task failed: Module failed: non-zero return code
Origin: /Users/subha/Code/Chakshu_Forks/Citizen-Complaint-Resolution-System/ansible/playbook-deploy.yml:196:7

194         chdir: "{{ digit_dir }}"
195
196     - name: Start DIGIT stack
          ^ column 7

[...]

PR #598, opened 2026-05-26
fix(ddh): idempotent MDMS seed — prevent DataSecurity double-seed crashing egov-user

KDwevedi:
## Summary
Makes `default-data-handler`'s MDMS seed **idempotent**. `MdmsBulkLoader.loadAllMdmsData` blind-created every bundled record with no existence check, so on a `db_fast_path` deploy (or any re-converge) — where the data already exists — it **duplicates** every record.

For `DataSecurity.SecurityPolicy` this is fatal: a duplicate policy crashes egov-user's encryption client on startup —
```
java.lang.IllegalStateException: Duplicate key [Attribute@.., Attribute@..]
  at EncryptionPolicyConfiguration.initializeEncryptionPolicyAttributesMapFromMdms (Collectors.toMap, no merge fn)
```
— so `egov-user` exits(1) → `dependency failed to start` → `docker compose up -d` returns rc=1 → the Ansible **`Start DIGIT stack`** task fails and the deploy aborts. mdms-v2 does not reject the duplicates because the dump-inserted rows bypass its uniqueness check.

## Root cause evidence (clean `develop` deploy on a dev box)
```
pg DataSecurity.SecurityPolicy:  total=64, distinct models=32   ← each policy duplicated 2x
   2025-12-18 | 32   ← db_fast_path dump baseline
   <deploy day> | 32 ← default-data-handler re-seeded the same 32
statea (not the active state tenant): 32, single cluster
```

## Fix
Add a per-`schemaCode` idempotency guard to `MdmsBulkLoader`: before seeding a bundle, search the target tenant for existing rows of that `schemaCode`; if any exist, skip the bundle. Seed bundles are fixed/all-or-nothing, so any existing row means it was already seeded. **Fail-open** — if the pre-check errors, fall through and seed as before. This also makes re-converges (the nightly dev-deploy cycle) safe.

## Validation
- ✅ Image builds clean (`build_default_data_handler` → `default-data-handler:local`).
- ⏳ Behavioral: fresh `db_fast_path` deploy on a dev box — expect `DataSecurity.SecurityPolicy` to stay at **32** (not 64) and `egov-user` healthy / `PLAY RECAP failed=0`. (Evidence to follow in a comment.)

Found while validating #567 on a fresh machine (the `user-proxy.conf` mount fix holds; this was the *next* blocker).

### DIFF
--- a/utilities/default-data-handler/src/main/java/org/egov/handler/util/MdmsBulkLoader.java
+++ b/utilities/default-data-handler/src/main/java/org/egov/handler/util/MdmsBulkLoader.java
@@ -6,6 +6,9 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.egov.common.contract.request.RequestInfo;
 import org.egov.handler.config.ServiceConfiguration;
+import org.egov.handler.web.models.MdmsCriteriaReqV2;
+import org.egov.handler.web.models.MdmsCriteriaV2;
+import org.egov.handler.web.models.MdmsResponseV2;
 import org.egov.tracer.model.CustomException;
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.support.PathMatchingResourcePatternResolver;
@@ -27,6 +30,7 @@ public class MdmsBulkLoader {
     private final ObjectMapper objectMapper;
     private final RestTemplate restTemplate;
     private final ServiceConfiguration serviceConfig;
+    private final MdmsV2Util mdmsV2Util;
 
 
     public void loadAllMdmsData(String tenantId, RequestInfo requestInfo, String mdmsDataPath) {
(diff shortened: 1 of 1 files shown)


KDwevedi:
## ✅ Behavioral validation — green

Fresh `db_fast_path` deploy on a dev box (clean wipe → deploy with the fixed image):

```
PLAY RECAP: ok=90 changed=23 failed=0
DataSecurity.SecurityPolicy:  pg|32  statea|32   (was 64 before the fix — no longer doubled)
egov-user:  digit-egov-user-1 Up (healthy)       (previously exited(1) on the duplicate-key crash)
```

DDH log shows the guard skipping every already-present bundle on the db_fast_path tenant:
```
MdmsBulkLoader : MDMS data already present for schemaCode DataSecurity.SecurityPolicy in tenant pg — skipping seed (idempotent)
... (same for ACCESSCONTROL-*, common-masters.*, the other DataSecurity.* schemas)
```

No regression to the `user-proxy.conf` mount fix (0 occurrences). The `Start DIGIT stack` task that previously failed at compose-up now completes; the full deploy reaches `failed=0`.

vinothrallapalli-eGov (approved):

Issue #478, opened 2026-04-28
Nai Pepea: Configurator UI: Complaint creation validation issue


Gurjeet-eGov:
**Describe the bug**
Complaint creation validation issue.

**To Reproduce**
1. Login to configurator
2. Go to create complaint 
3. Enter details in UI
Choose locality as County
Enter invalid pincode

**Expected behavior**
- User shall not be able to raise complaint at County or Sub County level, or we need to allow the same in Digit UI.
- No validation message is displayed for Postal Code
- Complaint create form shall be cleared automatically once a complaint is successfully registered
- 

https://github.com/user-attachments/assets/ae297ae7-75fb-4c65-a429-2cbe38eb0b51

KDwevedi:
Fix in theflywheel/digit-ui-esbuild#69 covering all three sub-bugs:

1. **Postal code** — pattern hardcoded the Indian 6-digit `/^[1-9][0-9]{5}$/` and a generic 'required' error. Switched to Kenya 5-digit `/^[0-9]{5}$/` (leading zeros allowed for codes like Nairobi GPO 00100), made the field optional since the boundary picker already pins to a Ward, and added a pattern-specific error key `CS_COMPLAINT_POSTALCODE_INVALID_ERROR` seeded en_IN + sw_KE.

2. **Ward-only submit** — `PGRBoundaryComponent` was emitting any level (County, Sub-County, Ward). Added an `isBoundaryLeaf()` guard to the employee CSR submit and the citizen `isFieldValid('SelectedBoundary')` check, so County / Sub-County selections are blocked with a toast (`CS_COMPLAINT_BOUNDARY_LEAF_REQUIRED`, also seeded).

3. **Form clear on success** — `clearSessionFormData()` only emptied sessionStorage; react-hook-form's in-memory values lingered. Captured FormComposerV2's `reset` through `onFormValueChange` into a ref and call it on success before navigating. Also removed a duplicate `clearSessionFormData` that was firing on every onSuccess (including `ResponseInfo.Errors`), so soft errors no longer nuke the operator's data.

Verified on naipepea: source patched on /opt/digit-ui-esbuild, esbuild HMR rebuilt, localization upserted + cache-busted.

KDwevedi:
Ready for re-test on naipepea. Two PRs in theflywheel/digit-ui-esbuild:

[...]

Issue #477, opened 2026-04-28
Nai Pepea: Citizen: Unable to create complaint due to locality issue


Gurjeet-eGov:
Unable to create complaint due to locality issue.

**To Reproduce**
1. Go to File a Complaint
2. Select complaint type and sub type
3. Select a location on map
4. Remove existing pre-filled pin code
5. Select only "Ward" Complaint location ui 

**Actual behavior**
- If user clicks "Next" without select any locality details, then ui displays "Ward" as the only required field. User is able to select the "Ward" directly without first selecting County or SubCounty
- The ward displays complete list but some last values are hidden behind UI footer element 

https://github.com/user-attachments/assets/aaedb60f-5ff9-4091-a7d5-74944ddb5043

Gurjeet-eGov:
Resolved 

https://github.com/user-attachments/assets/a1d9701a-6077-4621-ae71-61eca7292762

Issue #469, opened 2026-04-27
Nai Pepea: Citizen: During new complaint registration, unable to proceed "Pin Complaint Location" step


Gurjeet-eGov:
- Login as citizen user
- Go to File a Complaint
- Enter Complaint Details
- At pin location, do not select anything and just click next
- Default postal code will be 40476, UI will not allow to proceed with given pin
- Go back a step
- Select a different pin on map and click next

UI will not allow to proceed further with selected valid pin location 

https://github.com/user-attachments/assets/ac6ae11b-8774-4b64-8f51-aa6984a5e93d
https://github.com/user-attachments/assets/ac8ac0aa-0136-4da8-b90f-0097391a1125

PR #600, opened 2026-05-26
fix(pgr-create): tag boundary with isLeaf so ward enforcement fires; submit-time postal validation (#478)

KDwevedi:
## Summary
Closes the still-open BLOCKER on #478 — QA's 2026-05-19 re-test still showed both sub-bugs:
> "user is able to create complaint at Nairobi City County level"
> "user is able to create complaint with random pincode"

## Root cause
The earlier PRs (theflywheel/digit-ui-esbuild #69 / #72) added an `isBoundaryLeaf` guard based on `selectedBoundary.children.length === 0`. The check **looked** right but failed in practice:

`BoundaryDropdown` does `data.find((n) => n.code === code)` to map a value back to a tree node. Whether `.children` survives on the returned reference depends on how the option list was hydrated upstream — in the manual-selection path it generally doesn't, so `Array.isArray(sb.children)` returned `false` and the predicate `!Array.isArray(sb.children) || sb.children.length === 0` collapsed to `true` for *every* pick (County, Sub-County, Ward all looked like leaves). Submission therefore passed at any level. Confirmed against the live naipepea bundle: both the function and the toast key are present, just toothless.

Postal: the config declares `validation.pattern: /^[0-9]{5}$/` on a `type:"number"` field — the pattern rule doesn't reliably fire on number inputs, and the map-pincode overwrite in `onSubmit` was running unguarded.

## Fix
**Trust hierarchy depth instead of a fragile children array.**

| File | Change |
|---|---|
| `digit-ui-esbuild/products/pgr/src/components/BoundaryComponent.js` | Both emit paths (manual selection + map-driven auto-fill) now attach `isLeaf: selectedBoundary.boundaryType === boundaryHierarchy[boundaryHierarchy.length - 1]` to the SelectedBoundary payload. Additive — every existing consumer ignores the extra field. |
| `digit-ui-esbuild/products/pgr/src/pages/citizen/Create/FormExplorer.js` | `isFieldValid` for `SelectedBoundary` trusts `sb.isLeaf` when present; falls back to the children heuristic for session-cached values from before the tag. Also a belt-and-suspender leaf check + a submit-time `^[0-9]{5}$` postal pattern check in `onSubmit`. |
| `digit-ui-esbuild/products/pgr/src/pages/employee/CreateComplaint/createComplaintForm.js` | `isBoundaryLeaf` uses the same `isLeaf`/fallback shape. Adds the same submit-time postal pattern check. |

## Net effect on the QA repro
- Pick County → `Next`: blocked by `isFieldValid` (boundaryType ≠ Ward → `isLeaf=false`); toast `CS_COMPLAINT_BOUNDARY_LEAF_REQUIRED`.
- Pick Ward → flows through cleanly.
- Postal blank → accepted (field is optional).
- Postal `00100` (5 digits) → accepted.
- Postal `999` / `123456` / anything that doesn't match `^[0-9]{5}$` → rejected with `CS_COMPLAINT_POSTALCODE_INVALID_ERROR`.

Both toast keys are already seeded `en_IN` + `sw_KE` on naipepea (`rainmaker-pgr`) from earlier work.

## Workflow note
Per the move to land digit-ui-esbuild fixes under egovernments/CCRS develop, this PR patches the vendored `digit-ui-esbuild/`. Vendored files were sha-matched to theflywheel main before edits, so the diff would apply identically there if the standalone repo were still the target.

Refs: #478, #469 (related — boundary cascade), #477 (related)

### DIFF
--- a/digit-ui-esbuild/products/pgr/src/pages/citizen/Create/FormExplorer.js
+++ b/digit-ui-esbuild/products/pgr/src/pages/citizen/Create/FormExplorer.js
@@ -287,8 +287,15 @@ const FormExplorer = () => {
     // egovernments/CCRS#478 — postal validation message.
     if (merged?.postalCode != null && String(merged.postalCode).trim().length > 0) {
       const pc = String(merged.postalCode).trim();
-      if (!/^[0-9]{5}$/.test(pc)) {
-        setToast({ label: t("CS_COMPLAINT_POSTALCODE_INVALID_ERROR"), type: "error" });
+      // Postal-code shape is per-country. Read from globalConfigs
+      // `CORE_POSTAL_CONFIGS` so each tenant can pin their own pattern
+      // (Kenya 5 digits, India 6, UK alnum, US 5/5+4, …). Falls back to
+      // the legacy hard default when the host hasn't configured it.
+      const postalCfg = window?.globalConfigs?.getConfig?.("CORE_POSTAL_CONFIGS") || {};
+      const postalPattern = postalCfg.postalCodePattern || "^[0-9]{5}$";
+      const postalErrorKey = postalCfg.postalCodeErrorMessage || "CS_COMPLAINT_POSTALCODE_INVALID_ERROR";
+      if (!new RegExp(postalPattern).test(pc)) {
+        setToast({ label: t(postalErrorKey), type: "error" });
         return;
       }
     }
--- a/digit-ui-esbuild/products/pgr/src/pages/citizen/Create/FormExplorer.js
+++ b/digit-ui-esbuild/products/pgr/src/pages/citizen/Create/FormExplorer.js
@@ -234,12 +234,22 @@ const FormExplorer = () => {
       case "SelectedBoundary":
         // Tolerate the legacy SelectAddress shape so any in-flight
         // session data created before this change still validates.
-        // Also require the picked boundary to be a leaf (no children) so
-        // a citizen can't file at the County or Sub-County level — the
-        // PGR routing and SLA both key off the Ward (closes
-        // egovernments/CCRS#478 — locality validation, citizen path).
+        // Also require the picked boundary to be a leaf so a citizen
+        // can't file at the County or Sub-County level — PGR routing
+        // and SLA both key off the Ward (closes egovernments/CCRS#478
+        // — locality validation, citizen path).
+        //
+        // PGRBoundaryComponent now tags the emitted node with `isLeaf`
+        // (true only when boundaryType matches the deepest hierarchy
+        // level). Trust the tag when present — the earlier check
+        // relied on `sb.children` being preserved on the picked node,
+        // but BoundaryDropdown's `data.find()` doesn't reliably keep
+        // the children array, so County-level picks were silently
+        // passing as "leaves". Fall back to the children heuristic
+        // for older session-cached values that predate the tag.
         if (data?.SelectedBoundary?.code) {
           const sb = data.SelectedBoundary;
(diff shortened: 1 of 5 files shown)


digit-ui-esbuild/products/pgr/src/pages/citizen/Create/FormExplorer.js
```diff
@@ -269,6 +279,37 @@ const FormExplorer = () => {
...
 
+    // Postal pattern check — Kenya is 5 digits. The field is optional
+    // (the boundary picker already pins to a Ward, and many citizens
+    // don't know the postal code), so only enforce format when filled.
+    // The config has a `validation.pattern` on a `type:"number"` field
+    // which doesn't reliably fire — explicit submit-time guard. Closes
+    // egovernments/CCRS#478 — postal validation message.
+    if (merged?.postalCode != null && String(merged.postalCode).trim().length > 0) {
+      const pc = String(merged.postalCode).trim();
+      if (!/^[0-9]{5}$/.test(pc)) {
```
vinothrallapalli-eGov: @KDwevedi Do not hard code with 5 digits instead make configurable where each country has own length and starting digits.

digit-ui-esbuild/products/pgr/src/pages/employee/CreateComplaint/createComplaintForm.js
```diff
@@ -292,6 +300,22 @@ const CreateComplaintForm = ({
...
       return;
     }
+    // Postal pattern check — Kenya is 5 digits. Optional field; only
+    // enforce format when filled. The config-level `validation.pattern`
+    // on a `type:"number"` field doesn't reliably fire, so do it
+    // explicitly here. Closes egovernments/CCRS#478 — postal validation
+    // message, CSR path.
+    if (_data?.postalCode != null && String(_data.postalCode).trim().length > 0) {
+      const pc = String(_data.postalCode).trim();
+      if (!/^[0-9]{5}$/.test(pc)) {
```
vinothrallapalli-eGov: @KDwevedi Do not hard code with 5 digits instead make configurable where each country has own length and starting digits.

vinothrallapalli-eGov (approved):

KDwevedi:
@vinothrallapalli-eGov — addressed in `d114318`. Mirrored the existing `core_mobile_configs` pattern so postal validation is tenant-configurable:

- New host_vars block `core_postal_configs` ([_example.yml](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/develop/local-setup/ansible/inventory/host_vars/_example.yml)) — `postalCodePattern`, `postalCodeLength`, `postalCodeErrorMessage`. Documented with Kenya / India / UK / US example shapes.
- New `corePostalConfigs` var + `CORE_POSTAL_CONFIGS` branch in `globalConfigs.js.j2` (same place `coreMobileConfigs` lives).
- Both submit-time checks (`FormExplorer.js` citizen path + `createComplaintForm.js` employee path) now read pattern + error-key from `globalConfigs.getConfig("CORE_POSTAL_CONFIGS")` with a fallback to the legacy `^[0-9]{5}$` + `CS_COMPLAINT_POSTALCODE_INVALID_ERROR` so existing tenants without the new host_vars block are unaffected.

Form-config-level hardcode in `CreateComplaintConfig.js:156` (the `pattern: /^[0-9]{5}$/` on the input field) is also Kenya-only, but PR body already documents that the `type:"number"` field doesn't reliably enforce it (which is why the submit-time guard exists). Leaving that for a small separate refactor — out of scope for this review feedback.

PR #611, opened 2026-05-27
fix(playbook): guard insecure-registries daemon.json write; clear Mac/Linux defaults in maputo example

KDwevedi:
**The bug.** The `Configure Docker insecure-registries for VPC registry` task unconditionally writes `/etc/docker/daemon.json` with `{"insecure-registries": ["{{ docker_registry }}"]}` and then restarts docker. When `docker_registry` is a public TLS host like `ghcr.io/<org>` (which is the default `maputo.yml.example` sets!), the value is invalid — Docker rejects path-bearing or TLS hostnames in `insecure-registries` and refuses to start, taking the whole running stack down on the next deploy.

**Caught fresh on ovh-cloud-dev** (Ubuntu 25.04, OVH VPS), running `./deploy.sh maputo`:
```
TASK [Restart docker if daemon.json changed]
fatal: [maputo]: FAILED! Unable to restart service docker

# /etc/docker/daemon.json had just been written with:
{"insecure-registries": ["ghcr.io/chakshugautam"]}

# journalctl -xeu docker.service:
failed to start daemon: insecure registry ghcr.io/chakshugautam is not valid:
invalid host "ghcr.io/chakshugautam"
```
All 33 containers stopped.

**The fix.**
- New host_var `insecure_registries` (list). Playbook only writes `daemon.json` when the list is non-empty; the restart-on-change task is also guarded against the case where the copy task was skipped.
- `group_vars/digit.yml` defaults `insecure_registries` to `[docker_registry]`, preserving the historical Hetzner-VPC behaviour for bomet/naipepea — zero behaviour change there.
- `maputo.yml.example` overrides to `insecure_registries: []` (because its `docker_registry` points at public ghcr.io) and drops the Mac-only `deploy_become: false` default in favour of a clear comment — the playbook's own default is `true`, which is the right value on every Linux target.
- `_example.yml` documents the new field with the path-vs-host gotcha.

**After.** `./deploy.sh maputo` on a Linux box with a public registry runs clean past the docker config step instead of bricking the daemon. Validated end-to-end on ovh-cloud-dev (the same machine that hit the bug) — past from-source builds, into compose-up.

### DIFF
--- a/local-setup/ansible/inventory/group_vars/digit.yml
+++ b/local-setup/ansible/inventory/group_vars/digit.yml
@@ -113,12 +113,21 @@ upstream_minio_port: 19000
 # ──────────────────────────────────────────────────────────────────────
 
 # Where pre-built custom DIGIT images live. The egov-ci box in the
-# Hetzner VPC runs an insecure registry at 10.0.0.4:5000 — the playbook
-# writes this value into /etc/docker/daemon.json as an insecure-registry
-# entry on the target. Operators outside the VPC should override this
-# in host_vars (or set it to an empty string + flip enable_mcp: false
-# if they don't need digit-mcp at all).
+# Hetzner VPC runs an insecure registry at 10.0.0.4:5000. Operators
+# outside the VPC should override this in host_vars (e.g. to a public
+# registry like ghcr.io/<org>) or set it to an empty string + flip
+# enable_mcp: false if they don't need digit-mcp at all.
 docker_registry: 10.0.0.4:5000
+
+# Hosts to add to Docker daemon.json `insecure-registries`. Defaults to
+# the docker_registry above (matches the historical Hetzner-VPC behaviour
+# of writing 10.0.0.4:5000 as insecure). Override to `[]` when using a
+# public TLS registry like ghcr.io — Docker rejects a path-bearing or
+# TLS hostname in `insecure-registries` and refuses to start, taking the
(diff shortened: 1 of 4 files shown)


vinothrallapalli-eGov (approved):

KDwevedi:
**Validated end-to-end on Linux:** fresh-volumes `./deploy.sh maputo` on ovh-cloud-dev (Ubuntu 25.04, 6 cores, 11 GB) just landed:

```
PLAY RECAP: ok=90 changed=20 unreachable=0 failed=0 skipped=84  | 5m02s
```

- 36 containers up, 29 healthy, 0 unhealthy/restarting
- All historically-crashing JVMs healthy (egov-user, workflow-v2, enc-service, mdms, localization)
- ADMIN/eGov@123 login via Kong → access_token OK
- digit-ui + globalConfigs.js → HTTP 200
- Peak memory 8.4 GB / 11 GB (3.0 GB free) — fits even on small VMs

Without this patch (verified in the same session): the deploy bricks docker at the `Restart docker if daemon.json changed` task when `docker_registry` is any public TLS hostname (ghcr.io/\<org>, registry.preview.egov.theflywheel.in/\<org>), taking the whole stack down. With this patch, daemon.json is left untouched and the deploy proceeds cleanly through from-source builds → compose-up → OpenBao init → healthy stack.

Issue #445, opened 2026-04-23
Nai Pepea:  In Edit Profile Screen, Upload Profile Image feature not working


pradeepkumarcm-egov:
**Describe the bug**

In the Edit Profile section, there is an option to upload a profile image. However, when attempting to upload a JPEG image, the upload does not work.

**To Reproduce** 
 
**Screenshots**
 
<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/50c61819-8eb5-4714-bafb-a56409f355b1" />
 
**Additional context**

KDwevedi:
Fixed in [theflywheel/digit-ui-esbuild#29](https://github.com/theflywheel/digit-ui-esbuild/pull/29) (merge commit `500b4fa`). Deployed to https://naipepea.digit.org/digit-ui/.

**Root cause.** `packages/modules/core/src/pages/citizen/Home/ImageUpload/UploadDrawer.js` line 18 rejected any file ≥ 1 MB (`file.size >= 1000000`) with a `MAXIMUM_UPLOAD_SIZE_EXCEEDED` toast. Modern phone-camera JPEGs are routinely 2–5 MB, so the check fired before the upload ever attempted and the only feedback was a single toast — readers (reasonably) interpreted it as "upload is broken".

**Fix.** Raise the ceiling to 5 MB (DIGIT filestore's default allowlist ceiling for image modules), named as a constant for the next person to tune.

**Validate.** Employee → profile → upload a 2 MB JPEG → avatar updates. Upload a 6 MB file → see the size-exceeded toast as expected.

KDwevedi:
Bundle-verified on naipepea — `MAX_PROFILE_IMAGE_BYTES` constant and the new 5 MB threshold are in the deployed digit-ui bundle. **🟡 TO VALIDATE** by uploading a typical 2–3 MB phone-camera JPEG on `/user/profile`.

Gurjeet-eGov:
Employee edit profile ui crashing
<img width="1363" height="655" alt="Image" src="https://github.com/user-attachments/assets/95b8b597-06e0-42b5-9877-f86e0eb19e60" />

KDwevedi:
Crash fix shipped in #672 — null-safes the 6 `validationConfig?.<field>.test()` accessors so a missing tenant regex no longer throws on Edit Profile mount/keystroke. Upload-image symptom not covered here, will follow up separately once the crash unmasks it on a reproducer tenant.

[...]

PR #672, opened 2026-05-30
fix(profile): null-safe validationConfig regex access in UserProfile (#445)

KDwevedi:
## Summary

Closes the **Edit Profile crash** half of #445 ("Edit Profile: Upload Profile Image feature not working").

When `validationConfig` is fetched asynchronously, or a tenant config omits a field (e.g. tenant ships a `name` regex but no `password` regex), the inner regex is undefined and
```js
validationConfig?.password.test(value)
```
throws `Cannot read properties of undefined (reading 'test')` on first keystroke / mount, taking the whole Edit Profile screen down.

Six call sites in `UserProfile.js` had the outer optional-chain but not the inner. This PR adds the inner `?.` so a missing regex returns `undefined` (and the surrounding `!` guard treats it as invalid input) instead of throwing.

## What changed

`digit-ui-esbuild/packages/modules/core/src/pages/citizen/Home/UserProfile.js` — 6 lines:

| Line | Before | After |
|---:|---|---|
| 468 | `validationConfig?.password.test(value)` | `validationConfig?.password?.test(value)` |
| 483 | `validationConfig?.password.test(value)` | `validationConfig?.password?.test(value)` |
| 499 | `validationConfig?.password.test(value)` | `validationConfig?.password?.test(value)` |
| 531 | `validationConfig?.name.test(name)` | `validationConfig?.name?.test(name)` |
| 538 | `validationConfig?.mobileNumber.test(mobileNumber)` | `validationConfig?.mobileNumber?.test(mobileNumber)` |
| 575 | `validationConfig?.password.test(trimmedNewPassword) && !validationConfig?.password.test(trimmedConfirmPassword)` | `validationConfig?.password?.test(trimmedNewPassword) && !validationConfig?.password?.test(trimmedConfirmPassword)` |

Lines 406 and 442 already had the inner `?.`; this PR brings the rest of the file to parity.

## Scope

This PR addresses the **crash** symptom (the screenshot in the issue where the page goes blank on edit). The **"Upload Profile Image not working"** symptom (file picker / preview) is *not* covered here — that lives in `UploadDrawer.js` and the `useUploadFiles` hook and will need a separate dig once we can reproduce on a tenant where the crash no longer masks it.

## Test plan

- [x] Patch verified green on local dev rebuild
- [ ] Confirm Edit Profile screen renders for both citizen & employee on a tenant where `validationConfig.password` is undefined (e.g. ke / bomet)
- [ ] Confirm Save still rejects invalid `name` / `mobileNumber` when the regex IS defined (no regression to validation behavior)
- [ ] Re-open issue if upload symptom remains after this fix lands

### DIFF
--- a/digit-ui-esbuild/packages/modules/core/src/pages/citizen/Home/UserProfile.js
+++ b/digit-ui-esbuild/packages/modules/core/src/pages/citizen/Home/UserProfile.js
@@ -465,7 +465,7 @@ const UserProfile = ({ stateCode, userType, cityDetails }) => {
     // hid it because it wasn't bound to state. Mirror the sibling
     // handlers (`setUserNewPassword`, `setUserConfirmPassword`).
     setCurrentPassword(value);
-    if (!validationConfig?.password.test(value)) {
+    if (!validationConfig?.password?.test(value)) {
       setErrors({
         ...errors,
         currentPassword: {
@@ -480,7 +480,7 @@ const UserProfile = ({ stateCode, userType, cityDetails }) => {
 
   const setUserNewPassword = (value) => {
     setNewPassword(value);
-    if (!validationConfig?.password.test(value)) {
+    if (!validationConfig?.password?.test(value)) {
       setErrors({
         ...errors,
         newPassword: {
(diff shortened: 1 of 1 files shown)


KDwevedi:

Issue #555, opened 2026-05-07
[Bug] Uploaded file thumbnail is not visible on UI


Gurjeet-eGov:
- Uploaded file thumbnail is not visible on UI .
- Complaint created with attachment, the image is not displayed in complaints detail page of citizen and employee ui 

https://github.com/user-attachments/assets/56d1a2c2-38ab-4d9a-97dd-6815690865a9

<img width="1036" height="647" alt="Image" src="https://github.com/user-attachments/assets/c5d50076-8628-4f76-b7f8-304d3175580a" />

<img width="1036" height="647" alt="Image" src="https://github.com/user-attachments/assets/40d20925-f7b0-4936-b0ba-5d276d7fc46a" />

<img width="838" height="652" alt="Image" src="https://github.com/user-attachments/assets/74346a77-dfd0-4342-b442-187cea9dae35" />

prtkjakhar:
UI parser fixed in https://github.com/theflywheel/digit-ui-esbuild/pull/101 (merged + deployed to https://naipepea.digit.org), but image loading still requires a server-side URL fix — backend / nginx work needed @ChakshuGautam @KDwevedi.

**UI side (fixed):** `ComplaintPhotos.js` was iterating `Object.keys(res.data)` skipping `fileStoreIds` / `responseInfo` looking for sibling URL entries that never exist. The actual Filefetch response shape is `{ fileStoreIds: [{ id, url: "url1,url2-large,url3-medium,url4-small,…" }], responseInfo }`. Rewritten to walk `fileStoreIds[].url`, take the first segment as the full image, prefer a "small" variant for the thumb. Verified on a complaint with attachments — the thumbnail `<img>` element is now rendered (was 0 elements before) with `alt="issue thumbnail"` and the correct src.

**Server side (still broken):** the rendered `src` points at `http://minio:9000/egov-rainmaker/ke.nairobi/property-upload/…`. `minio:9000` is an internal Docker hostname not reachable from the browser, so the image fails to load (`naturalWidth: 0`). The filestore service is returning internal URLs that the public nginx config isn't rewriting to a publicly-reachable path — please point the filestore at an HTTPS-fronted MinIO endpoint (or rewrite `http://minio:9000/` → `https://naipepea.digit.org/file-store/` in the response, depending on your preferred pattern).

Once the URL is browser-reachable, the UI fix above will render the thumbnails as expected.

KDwevedi:
[...]

PR #673, opened 2026-05-30
fix(upload): surface filestore rejection as toast on complaint create (#555)

KDwevedi:
## Summary

Closes the **silent-upload-rejection** half of #555 ("Uploaded file thumbnail is not visible on UI") — specifically Gurjeet's follow-up: *"When user uploads invalid file format, the ui does not display any validation message"* with raw `{ Errors: [...] }` JSON shown.

The `ImageUploadHandler` atom (used by citizen complaint create `SelectImages`, `ReopenComplaint/UploadPhoto`, etc.) called `Digit.UploadServices.Filestorage` without a try/catch. When the filestore rejected the file (typical case: PDF/SVG under a tenant whose `ALLOWED_FORMATS_MAP` excludes them, see [[reference_filestore_quirks]]), the rejection bubbled as an unhandled promise: no toast, no preview update, leaving the user with no UI indication that anything had failed.

## What changed

`digit-ui-esbuild/packages/react-components/src/atoms/ImageUploadHandler.js`:

1. **Wrap `uploadImage` in try/catch.** On error, pull the API's `Errors[0].message` (or `.message` fallback) and surface it through the existing `Toast` already wired to `error` state.
2. **Localize the existing "File is too large" string.** Was a hard-coded English literal; now goes through `t("CS_FILE_TOO_LARGE")` with English fallback.
3. **Add the missing `useTranslation()` hook call.** The `import { useTranslation } from "react-i18next"` was present (line 2) but the hook was never invoked, so the file shipped with a dead import.

## Scope

This PR closes the **error-surfacing** symptom. The complaint detail page rendering symptom in #555 ("complaint created with attachment, the image is not displayed in complaints detail page") was already fixed in 37850e84 (`ComplaintPhotos.js` parser fix). Will verify both are visibly resolved on the next bomet rebuild and ping back here.

## Localization keys to seed

Both ship with English fallbacks so this is safe to merge before strings land:
- `CS_FILE_TOO_LARGE` — "File is too large"
- `CS_FILE_UPLOAD_FAILED` — "File upload failed"

## Test plan

- [ ] Citizen create-complaint upload of unsupported format (e.g. PDF on a JPEG-only tenant) → toast appears with backend's error message instead of silent failure
- [ ] Upload >2MB image → "File is too large" toast (or localized translation)
- [ ] Successful image upload still works (no regression to happy path)

### DIFF
--- a/digit-ui-esbuild/packages/react-components/src/atoms/ImageUploadHandler.js
+++ b/digit-ui-esbuild/packages/react-components/src/atoms/ImageUploadHandler.js
@@ -4,8 +4,7 @@ import Toast from "./Toast";
 import UploadImages from "./UploadImages";
 
 export const ImageUploadHandler = (props) => {
-  // const __initImageIds = Digit.SessionStorage.get("PGR_CREATE_IMAGES");
-  // const __initThumbnails = Digit.SessionStorage.get("PGR_CREATE_THUMBNAILS");
+  const { t } = useTranslation();
   const [image, setImage] = useState(null);
   const [uploadedImagesThumbs, setUploadedImagesThumbs] = useState(null);
   const [uploadedImagesIds, setUploadedImagesIds] = useState(props.uploadedImages);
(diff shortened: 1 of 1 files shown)


KDwevedi:

Issue #476, opened 2026-04-28
Nai Pepea: Configurator: Unable to edit the existing employee details


Gurjeet-eGov:
Unable to edit the existing employee details.

`{
    "ResponseInfo": null,
    "Errors": [
        {
            "code": "NullPointerException",
            "message": "An unhandled exception occurred on the server",
            "description": "Cannot invoke \"org.egov.hrms.model.AuditDetails.setLastModifiedBy(String)\" because the return value of \"org.egov.hrms.model.Assignment.getAuditDetails()\" is null",
            "params": null
        }
    ]
}`

https://github.com/user-attachments/assets/406fdf3b-bbf1-4b27-9895-68954b23a392

<img width="1366" height="584" alt="Image" src="https://github.com/user-attachments/assets/62cd7912-7b95-41a1-88a2-0fa61f9d713b" />

ChakshuGautam:
## Fix

**PR**: [ChakshuGautam/digit-configurator#43](https://github.com/ChakshuGautam/digit-configurator/pull/43)
**Commit**: [`1e9c66f`](https://github.com/ChakshuGautam/digit-configurator/commit/1e9c66f)

`toAssignmentRow()` and `toJurisdictionRow()` now pass through `auditDetails` from the HRMS response. Previously they dropped it, causing HRMS to NPE on `assignment.getAuditDetails().setLastModifiedBy()` during employee edit. Also added `auditDetails?` to `EmployeeAssignment` and `EmployeeJurisdiction` types.

**Deployed**: Nairobi (`naipepea.digit.org`) and Bomet (`bometfeedbackhub.digit.org`)
**E2E test**: `employee-edit.spec.ts` → "edit preserves assignments and jurisdictions"

Gurjeet-eGov:
General details are updated but **employment status update fails**
`{
    "ResponseInfo": null,
    "Errors": [
        {
            "code": "ERR_HRMS_INVALID_DEACT_REASON",
            "message": "Employee deactivation reason is invalid.",
            "description": null,
            "params": null
        }
    ]
}`

https://github.com/user-attachments/assets/4989cb52-98e7-4d3c-9244-553de25d10ff

ChakshuGautam:
## Round-2 fix

**PR**: [ChakshuGautam/digit-configurator#45](https://github.com/ChakshuGautam/digit-configurator/pull/45)

Root cause of `ERR_HRMS_INVALID_DEACT_REASON`:

[...]

Issue #471, opened 2026-04-28
Nai Pepea: Configurator: UI does not clear the form once the new employee has been registered


Gurjeet-eGov:
Configurator: UI does not clear the form once the new employee has been registered.

**To Reproduce**
1. Login to configurator
2. Go to Employee UI
3. Click on Create 
4. Enter all details
5. Click Submit

**Expected behavior**
Once the employee is created, the UI shall clear the form.

https://github.com/user-attachments/assets/bdeafc0b-e970-493a-bd8e-74dd92bc69fe

ChakshuGautam:
## Fix

**PR**: [ChakshuGautam/digit-configurator#43](https://github.com/ChakshuGautam/digit-configurator/pull/43)
**Commit**: [`1e9c66f`](https://github.com/ChakshuGautam/digit-configurator/commit/1e9c66f)

Verified working — `DigitCreate` already passes `redirect='list'` to `CreateBase` and has a custom `onSuccess` with success toast. After the rebase to `origin/main` (33 commits), the redirect works correctly. No additional code change was needed.

**Deployed**: Nairobi (`naipepea.digit.org`) and Bomet (`bometfeedbackhub.digit.org`)
**E2E test**: `employee-create.spec.ts` → "form has all expected sections" (verifies form renders correctly; redirect tested manually)

Gurjeet-eGov:
@prtkjakhar @KDwevedi 
unresolved, the UI neither redirects to list or cleared form when employee is registered successfully

https://github.com/user-attachments/assets/2772cbf5-f456-4fbb-ac82-bd55f37924ef

prtkjakhar:
should be fixed by https://github.com/ChakshuGautam/digit-configurator/pull/49

prtkjakhar:
Fix live on https://naipepea.digit.org/configurator/ ([PR #49](https://github.com/ChakshuGautam/digit-configurator/pull/49)).

@Gurjeet-eGov — please verify Create and Edit on Employees (and any other resource) now redirect to the list with a toast.

Gurjeet-eGov:
@prtkjakhar 
Unable to edit or create employee because mobile number is not validating even with correct format 

https://github.com/user-attachments/assets/3715ec8a-5352-4df5-b407-8163df37817e

KDwevedi:
[...]

Issue #459, opened 2026-04-26
Nai Pepea: Configurator: User created for incorrect tenant id and not based on option selected from UI


Gurjeet-eGov:
When user tries to create an employee for a particular tenant, it does not refer to the selected tenant and creates user based on one default tenant

<img width="364" height="170" alt="Image" src="https://github.com/user-attachments/assets/2cd787e2-b579-4d9a-aac6-066f549f2980" />

https://github.com/user-attachments/assets/ea3e55a2-ad77-43e6-a6c3-212b16c8a8f6

ChakshuGautam:
## Round-2 fix

**PR**: [ChakshuGautam/digit-configurator#46](https://github.com/ChakshuGautam/digit-configurator/pull/46)

PR #34 fixed `EmployeeCreate.transform` (it was clobbering the form-picked `tenantId` with the session tenant). With that change in place, raw HRMS `/_create` calls against `ke.nairobi` *do* land at the right tenant — verified.

But the form was still hard to use end-to-end: the AssignmentEditor's department/designation pickers fetched via `useGetList` which closed over the session tenant. ADMIN logged in at root `ke` would see root-only codes, pick one, submit, and HRMS would reject the write at `ke.nairobi` with `ERR_HRMS_INVALID_DEPT`.

This PR closes that loop:

- New `__tenantId` filter key on the data provider lets a single `useGetList` call pin to a specific tenant (consumed by `mdmsGetList`, ignored by `clientFilter`).
- `AssignmentEditor` watches `form.tenantId` via `useFormContext` + `useWatch` and threads it through, so the dept/desig dropdowns repopulate when the operator switches tenant in the form.

**Deployed**: Nairobi (`naipepea.digit.org`) and Bomet (`bometfeedbackhub.digit.org`). Ready to close after re-test.

Gurjeet-eGov:
Unable to verify because the create employee is blocked by mobile number validation. 

https://github.com/user-attachments/assets/facf8246-51ef-485d-8a32-bace3e9511e3

KDwevedi:
Mobile-validation rejection fixed in #674 — fallback now accepts both `712345678` (bare 9) and `0712345678` (everyday Kenyan trunk-zero form). Should unblock you from getting past the create form. Will follow up on the actual 'creates user for wrong tenant' once you can land in the create step.

KDwevedi:
[...]

PR #674, opened 2026-05-30
fix(configurator): accept Kenyan mobile with optional trunk-zero in fallback (#459 #471)

KDwevedi:
## Summary

Unblocks Gurjeet's mobile-number rejection on **#459** and **#471** (both reported \"Unable to edit or create employee because mobile number is not validating even with correct format\"). This is also the surface contributor on #476's edit path.

## Root cause

MDMS \`ValidationConfigs.mobileNumberValidation\` is **not seeded** for tenant \`ke\` on naipepea / bomet. Verified directly:

\`\`\`
curl /mdms-v2/v2/_search?tenantId=ke
  ... ValidationConfigs.mobileNumberValidation → empty
  ... common-masters.UserValidation.mobile → pattern: \"^[6-9][0-9]{9}$\", prefix: \"+91\"  (legacy Indian seed)
\`\`\`

So \`useMobileValidator\` hits its FALLBACK constant, which previously was \`^[17][0-9]{8}$\` — **strict** 9 digits, no leading 0. Operators enter \`0712345678\` (everyday Kenyan trunk-zero form) → rejected.

## What changed

\`configurator/src/admin/hrms/useMobileValidator.ts\` — relax the FALLBACK to allow optional trunk zero, matching \`phoneKE\` in \`src/admin/validation.ts\`:

\`\`\`diff
- pattern: '^[17][0-9]{8}\$',
+ pattern: '^0?[17][0-9]{8}\$',
  minLength: 9,
- maxLength: 9,
+ maxLength: 10,
\`\`\`

Plus updated comment + error-message text to match.

When MDMS DOES provide a rule (\`parseRules\` path), behavior is unchanged — this only affects the fallback that fires today on every Kenya tenant.

## Scope

This PR fixes the **frontend** form-acceptance side. The MDMS seed for \`ValidationConfigs.mobileNumberValidation\` is a separate follow-up (MCP / tenant_bootstrap) so the citizen UI gets the same rule path instead of leaning on the legacy \`common-masters.UserValidation.mobile\` Indian seed.

#476's backend NPE half (\`Assignment.getAuditDetails() returning null\`) is also separate; that lives in egov-hrms.

## Test plan

- [ ] Configurator → Create Employee → enter \`0712345678\` → accepted, can submit
- [ ] Same, enter \`712345678\` → still accepted (backwards-compat with bare-9 form)
- [ ] Same, enter \`9876543210\` (Indian-style) → still rejected
- [ ] Same, enter \`07123\` → rejected (too short)
- [ ] Edit existing employee with bare-9 stored mobile → form pre-fills + edits accepted

### DIFF
--- a/configurator/src/admin/hrms/useMobileValidator.ts
+++ b/configurator/src/admin/hrms/useMobileValidator.ts
@@ -10,15 +10,19 @@ export interface MobileRules {
 }
 
 // Kenya default — used when MDMS `ValidationConfigs.mobileNumberValidation`
-// is missing or unreadable. The canonical naipepea seed is `^[17][0-9]{8}$`
-// with min=max=9, matching the local subscriber number convention.
+// is missing or unreadable. Accept both the bare 9-digit subscriber number
+// (`712345678`) AND the everyday form prefixed with a trunk `0`
+// (`0712345678`) — Gurjeet's regression on #459/#471/#476 was hitting the
+// strict 9-digit fallback after typing the latter (correct local format).
+// Matches `phoneKE` in `src/admin/validation.ts` so all employee forms
+// share one accept-set.
 const FALLBACK: MobileRules = {
-  pattern: '^[17][0-9]{8}$',
+  pattern: '^0?[17][0-9]{8}$',
   minLength: 9,
-  maxLength: 9,
+  maxLength: 10,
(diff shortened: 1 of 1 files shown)


KDwevedi:

Issue #556, opened 2026-05-07
[Bug] Citizen unable to save edit profile changes


Gurjeet-eGov:
Citizen unable save to edit profile changes .

**To Reproduce**
1. Go to citizen ui
2. Go to edit profile 
3. edit details and save (keep the name as it is)

By default the citizen ui is setup with mobile number as the Name of user. When user tries to edit the Name field validation does not allow numbers 

https://github.com/user-attachments/assets/5f763c19-254e-454b-afa6-15cbf8c164de

prtkjakhar:
Fixed, pls review @Gurjeet-eGov.

UI: https://github.com/theflywheel/digit-ui-esbuild/pull/101 (merged + deployed to https://naipepea.digit.org).

The default Name regex in `UserProfile.js` was `/^[a-zA-Z ]+$/i` which rejected the mobile-number-as-name seed value. Relaxed to `/^[a-zA-Z0-9 .'\-]+$/i` so digits + apostrophes / hyphens / periods are accepted (real names with these chars also unblock). MDMS `UserProfileValidationConfig` overrides still win when present.

Verified end-to-end on localhost: typed `777777777` (the 9-digit Naipepea-format mobile number that Denise's seeded `name` field carries) into the citizen Edit Profile name field and Save accepts without the pattern error.

Gurjeet-eGov:
Name validation is successful

- Profile image is uploaded but not reflected in UI upon refresh 
- Gender details are not saved 

https://github.com/user-attachments/assets/86927b6d-53c1-44f6-a544-5b87f336da63

prtkjakhar:
fixed, pls review

Gurjeet-eGov:
Edit works, RESOLVED
Profile image is uploaded but its not displayed in citizen left menu 

https://github.com/user-attachments/assets/e570a3ee-06fd-475e-b032-b47f149943dd

KDwevedi:
[...]

PR #678, opened 2026-05-30
fix(profile): refresh citizen sidebar avatar after Edit Profile save (#556)

KDwevedi:
## Summary

Closes Gurjeet's remaining #556 sub-bug: *"Profile image is uploaded but its not displayed in citizen left menu"* (2026-05-20 retest after the Edit-save half was confirmed RESOLVED).

### Layer 1: `UserProfile.js` setUser patch omitted `photo`

The post-save handler refreshes the session-cached user object, but only with name / mobileNumber / emailId / permanentCity:

```js
Digit.UserService.setUser({
  ...user,
  info: {
    ...user.info,
    name, mobileNumber, emailId, permanentCity,
    // photo NOT here
  },
});
```

The backend got the new photo via the `_update` payload, but the in-memory session stayed stale. Sidebar — which reads `Digit.UserService.getUser().info` — kept showing the old picture until a full re-login.

### Layer 2: `CitizenSideBar` Profile useEffect dep was misshaped

```js
const [profilePic, setProfilePic] = React.useState(null);
React.useEffect(async () => {
  // ... userSearch + setProfilePic ...
}, [profilePic !== null]);
```

The dep `[profilePic !== null]` evaluates to a boolean. It flips `false → true` once on first fetch and then stays put — so after the initial userSearch the effect never runs again, even if the underlying user info changes. The effect also always fired a userSearch on mount, ignoring whatever the session already had.

## Fix

**`UserProfile.js`** — include the photo in the post-save session patch:

```js
Digit.UserService.setUser({
  ...user,
  info: {
    ...user.info, name, mobileNumber, emailId, permanentCity,
    ...(profilePic ? { photo: profilePic } : {}),
  },
});
```

**`CitizenSideBar.js` Profile** — rewrite so:
- State seeds from `info.photo` if present (session is source of truth).
- Effect uses session value first; falls back to userSearch only when `info.photo` is missing (e.g. first login).
- Deps are `[info?.uuid, info?.photo]` so the next save propagates immediately.
- Guard against the fire-and-forget async overwriting state after unmount (cancellation flag).

## Out of scope

- Employee-side avatar refresh — that path doesn't use `CitizenSideBar`. If Gurjeet sees the same staleness on /employee, that's a separate TopBar component and a separate ticket.
- The legacy non-Individual update branch (`Digit.UserService.updateUser`) also routes through the same setUser patch, so the photo fix covers it too.

## Test plan (for QA on a deployable env)

- [ ] Citizen login → Edit Profile → upload a new image → Save → left-menu avatar updates **without** logging out and back in
- [ ] Same flow but with name/mobile changes only (no new image) → no regression to existing fields
- [ ] First-time login of a user who has a photo from server-side seeding → avatar appears via the userSearch fallback path
- [ ] First-time login of a user with no photo → defaultImage shown, no broken userSearch call left in flight when navigating away

### DIFF
--- a/digit-ui-esbuild/packages/modules/core/src/components/TopBarSideBar/SideBar/CitizenSideBar.js
+++ b/digit-ui-esbuild/packages/modules/core/src/components/TopBarSideBar/SideBar/CitizenSideBar.js
@@ -11,19 +11,37 @@ import { LogoutIcon } from "@egovernments/digit-ui-react-components";
 import ImageComponent from "../../ImageComponent";
 
 const Profile = ({ info, stateName, t }) => {
-  const [profilePic, setProfilePic] = React.useState(null);
-  React.useEffect(async () => {
-    const tenant = Digit.ULBService.getCurrentTenantId();
+  // Prefer the session-cached photo so an in-place Edit-Profile save
+  // shows up immediately — UserProfile.js writes the new value into
+  // `Digit.UserService.getUser().info.photo` after a successful
+  // update (CCRS#556 sub-bug pair fix). Fall back to the per-uuid
+  // userSearch lookup only if the session doesn't already carry one
+  // (e.g. on first login).
+  const [profilePic, setProfilePic] = React.useState(
+    info?.photo ? info.photo.split(",").at(0) : null,
+  );
+  React.useEffect(() => {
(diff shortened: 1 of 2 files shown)


KDwevedi:

PR #675, opened 2026-05-30
fix(configurator): postal-code validator on complaint forms (#478 — postal half)

KDwevedi:
## Summary

Half of the original #478 fix bundle, split per one-concern-per-PR. The other half (LocalityPicker leaf-only filter) is now in **#680**.

Closes Gurjeet's *"user is able to create complaint with random pincode"* point on his 2026-05-19 retest of #478.

## What changed

`configurator/src/admin/validation.ts` — new exported validator:
```ts
export const postalCodeKE = raRegex(
  /^[0-9]{5}$/,
  'Enter a valid 5-digit postal code (e.g. 00100)',
);
```

`configurator/src/resources/complaints/ComplaintCreate.tsx` + `ComplaintEdit.tsx` — wire `validate={v.postalCodeKE}` to `address.pincode`. Stays optional (empty passes); only shape-checks when filled.

## Scope / out-of-scope

- Kenya-specific pattern by name — matches what `useMobileValidator` already does for mobile in the same codebase. A future tenant-pluggable form (sourcing from MDMS `ValidationConfigs.postalCodeValidation` or similar) is the right next step; explicitly out of scope here.
- LocalityPicker leaf-only enforcement — split out to **#680** so it can land/iterate on its own (it's a UX change that affects every tenant, not just Nairobi).

## Test plan

- [ ] Configurator → Complaints → Create → enter pincode `00100` → accepted
- [ ] Same form, enter pincode `12abc` → validation error shown
- [ ] Same form, leave pincode blank → form submits (optional field)
- [ ] ComplaintEdit also rejects malformed pincode on save

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #680, opened 2026-05-30
fix(configurator): restrict LocalityPicker to LEAF boundary types only (#478 — leaf half)

KDwevedi:
## Summary

Companion half to **#675**. Original #478 retest had two distinct items; splitting per the one-concern-per-PR principle so each can land/iterate independently.

Closes Gurjeet's *"user is able to create complaint at Nairobi City County level"* point on his 2026-05-19 retest of #478.

## What changed

`configurator/src/resources/complaints/LocalityPicker.tsx` — filter the boundary-type select to **leaf** types per hierarchy (types that aren't anyone's `parentBoundaryType` within the same hierarchy).

On Nairobi (ADMIN: County → Sub-County → Ward → Locality) only Locality remains in the type select; the operator is structurally prevented from selecting a non-leaf rather than seeing a late toast. Mirrors the digit-ui-side `isBoundaryLeaf` rule already enforced for citizens.

## UX impact + why this is split

This change affects **every** hierarchy in **every** tenant — anywhere `LocalityPicker` is used, only leaf boundaries become selectable. That's the intended behavior (PGR complaints have to route to a ward to assign), but it's a meaningful UX change worth landing on its own merits rather than bundled with the postal-code fix.

If a tenant has a hierarchy where higher-level filing makes sense, this PR is the wrong shape and we'd need a configurable "allow non-leaf" prop on the picker. Open to that direction.

## Out of scope

- Backend enforcement. PGR-services itself currently accepts a complaint with a non-leaf `locality.code`, so an API-direct caller can still file at County level. Separate service-layer guard worth opening if that ever surfaces.

## Test plan

- [ ] Configurator → Complaints → Create → Locality cascade: pick ADMIN hierarchy → Boundary Type select only shows Locality (not County / Sub-County / Ward)
- [ ] Existing complaint loaded for Edit at a non-leaf boundary (legacy data) → form still mounts cleanly, but operator can't switch the type to a non-leaf
- [ ] Tenant with a flat hierarchy (single boundaryType, no parents) → that one type still appears as a leaf

### DIFF
--- a/configurator/src/resources/complaints/LocalityPicker.tsx
+++ b/configurator/src/resources/complaints/LocalityPicker.tsx
@@ -75,19 +75,36 @@ export function LocalityPicker({
     return hierarchies.map((h) => ({ value: h.hierarchyType, label: h.hierarchyType }));
   }, [hierarchies]);
 
+  // Only expose the LEAF boundary type(s) per hierarchy — types that
+  // aren't anyone's `parentBoundaryType` within the same hierarchy.
+  // Gurjeet's #478 retest showed operators filing complaints at
+  // "Nairobi City County" (root), which the PGR backend will route to
+  // no ward and the resolver UI then can't show a sane assignment
+  // surface. The original digit-ui side enforced this with
+  // `isBoundaryLeaf` at submit; mirror that constraint here at the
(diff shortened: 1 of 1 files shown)


KDwevedi:

Issue #622, opened 2026-05-28
[Bug] egov-user encryption drift breaks ADMIN login when STATE_LEVEL_TENANT_ID flips


KDwevedi:
## Symptom

After a successful `./deploy.sh maputo` (`state_root: mz`) followed by a working compose-file rewrite that actually flips `egov-user`'s `STATE_LEVEL_TENANT_ID` from `pg` → `mz` (via `up -d --force-recreate`, not `restart`), **every login fails with `Invalid login credentials`** — ADMIN at `pg`, ADMIN at `mz`, ADMIN at `mz.maputo`, citizen mobile, employees. None work.

The same deploy with `restart` instead of `--force-recreate` produces working logins, because `restart` silently doesn't re-read the compose file and the container stays pinned to `STATE_LEVEL_TENANT_ID=pg`. So the bug is invisible today.

## Root cause

`egov-user`'s search path encrypts the username before SELECT'ing:
```
egov-user → egov-enc-service: _encrypt(value='ADMIN', tenantId=<egov-user's STATE_LEVEL_TENANT_ID>)
```

egov-enc-service returns ciphertext prefixed with the key version that's active for the tenant — e.g. `489366|...` for `pg`, `34085|...` for `mz`. Then egov-user does:
```sql
SELECT * FROM eg_user WHERE username = <ciphertext>
```

The `eg_user` rows from `db/full-dump.sql` were ALL encrypted at pg's key version `489366`. When egov-user is on mz, it generates ciphertext starting with `34085|` — no match. `UserNotFoundException` → `Invalid login credentials`.

Live log excerpt from ovh-cloud-dev:
```
UserNotFoundException: User not found for given criteria: UserSearchCriteria(
  ...,
  userName=34085|QXykjBCZ3I27v5n3GSNF7fBcCF2rv6SByg==,
  ...,
  tenantId=mz, ...)
```

The DB has the same row at key version `489366`.

## Repro on ovh-cloud-dev

1. `./deploy.sh maputo` (state_root: mz, db_fast_path: true, build_mcp: true)
2. Manually `docker compose ... up -d --force-recreate --no-deps egov-user`
3. `docker inspect digit-egov-user-1 | grep STATE_LEVEL_TENANT_ID` → `mz` ✅
4. `curl POST /user/oauth/token` with any user/tenant combination → `Invalid login credentials` ❌

[...]

PR #706, opened 2026-06-01
ansible(bomet): commit per-tenant overlay with egov-user 1044 image pin

ChakshuGautam:
## Summary

Commits `local-setup/docker-compose.bomet.yml` — bomet's per-tenant compose overlay — to source. Two of the three overrides were already running on the live bomet server but had never been committed; every clean `ansible-playbook` run silently dropped them and any local hand-edits got wiped on the next deploy.

This PR folds the live state into source and adds one more pin:

| Override | Source |
|---|---|
| `inbox.STATE_LEVEL_TENANT_ID=ke` + `PARENT_LEVEL_TENANT_ID=ke` | Existing on `/opt/digit/` — inbox defaults to `mz` and crashes at boot on bomet's DataSecurity lookup |
| `egov-url-shortening.SPRING_FLYWAY_ENABLED=false` | Existing on `/opt/digit/` — restored Flyway state has no V1 row, V1 re-run collides with the pre-existing `eg_url_shortener` table |
| **`egov-user.image = 10.0.0.4:5000/egovio/egov-user:dynamic-tenant-bomet`** (new) | Picks up the call-site migration from egovernments/Digit-Core#1044 stacked on top of the ccc0e13 NPE fix |

## Why the image lives at the VPC registry

The image is the c6f5438 tenant-aware encryption overload plus the call-site migration commit (`884c5e6`) from PR #1044, built against the ccc0e13 NPE-fix base so `/user/profile/_update` keeps working. Built on egov-ci, pushed to `10.0.0.4:5000/egovio/egov-user:dynamic-tenant-bomet`.

Bomet's docker daemon already has `10.0.0.4:5000` configured as an `insecure-registries` entry (same Hetzner VPC, plain HTTP). Once #1044 merges upstream and the next preview-registry tag of egov-user includes the call-site migration, we can drop this overlay's `egov-user` block and let the shared compose pin take over.

## Test plan

- [x] Live bomet smoke: container recreated on the new image, healthcheck `healthy`, `OTP send 200`, `/user/citizen/_create 200`, `/user/profile/_update 200`, `/user/_search 200` against pre-existing encrypted rows
- [x] Single-state safety: `getStateLevelTenant("ke")` returns `ke` (matches the `STATE_LEVEL_TENANT_ID=ke` env), so encryption ciphertext prefix is identical to the previous image — no DB-side compat impact
- [ ] After merge: re-run `./deploy.sh bomet` from a clean controller and confirm the overlay survives the run

## Refs

- egovernments/Digit-Core#1044 — upstream egov-user fix (open)
- Fixes #622 (encryption drift on `STATE_LEVEL_TENANT_ID` flip) once that fix is merged

### DIFF
--- a/local-setup/docker-compose.bomet.yml
+++ b/local-setup/docker-compose.bomet.yml
@@ -1,0 +1,37 @@
+# Per-tenant compose overlay for bomet (state tenant `ke`).
+#
+# This file is copied to the target server as
+# /opt/digit/docker-compose.<inventory_hostname>.yml by the per-tenant overlay
+# step in playbook-deploy.yml, then included in every `docker compose` call.
+# Use it for bomet-only env tweaks and image pins that don't belong in the
+# shared docker-compose.egov-digit.yaml.
+services:
+  # Inbox defaults STATE_LEVEL/PARENT_LEVEL to mz; bomet has no DataSecurity at
+  # mz → encryptionPolicyConfiguration crash on boot. Pin both to ke.
+  inbox:
+    environment:
+      STATE_LEVEL_TENANT_ID: "ke"
+      PARENT_LEVEL_TENANT_ID: "ke"
+
+  # bomet's restored Flyway state has no V1 record → re-running V1 collides
+  # with the pre-existing eg_url_shortener table. Skip Flyway entirely.
+  egov-url-shortening:
+    environment:
+      SPRING_FLYWAY_ENABLED: "false"
+
+  # CCRS #622 / Digit-Core #1044: tenant-aware encryption.
+  #
+  # Custom image built on egov-ci (89.167.55.190) by cherry-picking c6f5438
+  # ("add tenant-aware encryption overloads") onto ccc0e13 ("prevent NPE in
+  # _updatenovalidate and profile _update"), then migrating the 6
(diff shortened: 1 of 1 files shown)


ChakshuGautam:
Updated to use the public preview-registry URL instead of the VPC IP.

`registry.preview.egov.theflywheel.in` is a read-only nginx TLS proxy on the egov dev box that fronts the egov-ci VPC registry at `10.0.0.4:5000`. The proxy is already running, the certificate is the wildcard `*.preview.egov.theflywheel.in`, and the image is now pullable from anywhere — confirmed with a clean `docker pull` from this dev box (OVH, outside the Hetzner VPC) and a `docker pull` + container restart on bomet just now.

So there was no blocker — the public tag was already there because the proxy serves the same blob storage as the VPC registry. The earlier overlay just hadn't been pointed at it.

Diff in this push: `10.0.0.4:5000/egovio/egov-user:dynamic-tenant-bomet` → `registry.preview.egov.theflywheel.in/egovio/egov-user:dynamic-tenant-bomet`. Bomet was restarted on the new pin and is `healthy`.

subhashini-egov (approved):

PR #775, opened 2026-06-09
refactor(pgr): MDMS-ify scheduler state-name mapping (drops last hardcoded PGR coupling)

ChakshuGautam:
> ⚠️ **STACKED PR** — built on top of #770 (`feat/escalation-otel-configurator-designer`, last commit `989941bed`). GitHub shows the diff against `develop` (the parent branches live on the fork, so the base cannot be re-pointed); **review only this PR's own commits** via the Commits tab.

## Summary

Stacked on PR #770. The escalation scheduler in PR #770 still hardcodes a switch from PGR workflow state names (PENDINGFORASSIGNMENT, PENDINGATLME, …) onto the canonical SLA-column keys (new, triage, forwarded, …). That switch is the last tenant-specific coupling in the scheduler.

This PR replaces it with a tenant-defined MDMS record so operators can wire any workflow's state names onto the SLA columns without code changes.

- **New schema**: `CRS.WorkflowStateMapping` — singleton per tenant, shape `{ singletonKey: "default", mappings: { <workflowStateName>: "new|triage|forwarded|investigation|awaiting|resolved" } }`
- **Scheduler**: `mapWorkflowStateToKey()` now takes the mapping as an argument; no per-tenant state names referenced in code
- **New skip reason**: `STATE_MAPPING_MISSING` so unmapped workflow states surface explicitly in trace-back + OTEL spans
- **Tests**: 3 unit tests covering mapping hit, mapping absent, mapping typo paths

## Stacked on

PR #770 — egov/ccrs escalation foundation. This PR's diff is just the refactor on top.

## Test plan

- [x] `mvn test -Dtest=EscalationSchedulerStateMappingTest` → 3/3 green
- [x] `mvn compile` → green
- [ ] Live test on Bomet after PR #770 + #A both merge: seed CRS.WorkflowStateMapping with `{PENDINGFORASSIGNMENT: "new", PENDINGATLME: "forwarded"}`; trigger /escalation/_trigger; assert `escalation.slaSource` resolves via the new mapping in Tempo

## Linked

- Discussion #773 (overall escalation design)
- PR #770 (escalation foundation)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #797, opened 2026-06-09
test(pgr): edge-case coverage for STATE_MAPPING_MISSING + CSV cell parsing

ChakshuGautam:
> ⚠️ **STACKED PR** — built on top of #775 (`refactor/scheduler-state-name-mdms`). GitHub shows the diff against `develop` (the parent branches live on the fork, so the base cannot be re-pointed); **review only this PR's own commits** via the Commits tab.

Adds unit-level edge-case coverage around the new escalation-SLA cascade introduced in PR #775 (CRS.WorkflowStateMapping → CategorySLA → StateSLA → v0) and the SLA-matrix CSV parser in the configurator.

**Stacked on PR #775** (`refactor/scheduler-state-name-mdms`) — depends on the `CRS.WorkflowStateMapping` schema + `STATE_MAPPING_MISSING` enum landed there.

### Files added
- `backend/pgr-services/src/test/java/org/egov/pgr/service/EscalationStateMappingEdgeCaseTest.java` — 7 JUnit cases covering empty mapping value, empty dictionary, typo'd canonical key, range-cell MAX collapse, partial category tuple, null CategorySLA cell fallthrough, triple-null fallback.
- `configurator/src/resources/crs/sla-matrix/csvParser.test.ts` — 13 vitest cases (10 pass, 3 `.skip`-d pending TASK #66) over header tolerance + cell encoding (empty/number/range/garbage/out-of-bounds/reversed-range).

### Test plan
```
cd backend/pgr-services && mvn test -Dtest="EscalationStateMapping*"
cd configurator && npx vitest run src/resources/crs/sla-matrix/csvParser.test.ts
```
Both suites pass locally on the parent branch tip.

### Caveats
The csvParser case-sensitivity bug (header lowercased then compared to camelCase `subcategoryL1`) surfaced again — it's tracked as a separate followup, not fixed here. One test pins the current broken behaviour so the regression stays visible; three are `.skip`-d with TODOs that will flip once the case-fix lands.

Cross-link: Discussion #773 (escalation design doc — the open questions section that motivated these tests).


KDwevedi:
Closing as stale and superseded by #2048. The unique tests cover WorkflowStateMapping and the CategorySLA CSV surface, neither of which is part of the selected one-config implementation. The test techniques can be recreated against the final self-loop flow without carrying the unmerged foundation. Full archaeology and reusable-parts assessment: https://github.com/KDwevedi/Citizen-Complaint-Resolution-System/pull/35.

PR #805, opened 2026-06-09
Phase 5: parameterize Playwright suite + unblock citizen flow on KC-overlay deployments

### LINKED ISSUE #685: [Bug] E2E Tests failing against Ansible deploy
**Describe the bug**
Pulled latest code from develop. Deployed a maputo tenant -- mz.maputo -- using the Ansible deployer. Deployment succeeded. However, E2E tests are failing. Unable to log into configurator using newly created tenant. Suspect that latest builds & fixes are not wired up to the ansible deployer. Opening this as a placeholder to resolve those. 

**Expected behavior**
E2E tests should all pass.

**Screenshots**

Image
Image

**Additional context**

subhashini-egov:
## Summary

Phase 5 of the test-tier remediation tracked in #685: makes the 234-test Playwright suite **runnable end-to-end against any DIGIT deployment** (naipepea, subhadev, bomet, …) by removing hard-coded `ke.nairobi` / `+254` / pinned SRIDs from spec files, adding a self-seeding lifecycle setup project, and fixing the digit-ui + helm gaps that surfaced when we actually ran the suite against bomet for the first time.

Builds on the earlier verification work done in this session (the 234-test classification matrix and the per-file misclassification audit).

## What's in this PR

**14 spec refactors** — replace hardcoded tenant / SRID / mobile-rule / theme values with env-driven helpers from `utils/env.ts` + `utils/mdms-mobile.ts`. New env vars: `PINNED_COMPLAINT_SRID`, `KNOWN_RESOLVED_SRID`, `THEME_RECORD_ID`, `TERMINAL_COMPLAINT_SRID`, `NONTERMINAL_COMPLAINT_SRID`, `MOBILE_PREFIX`. All default to the naipepea historicals so naipepea runs stay green; any other deployment overrides via env.

**Lifecycle setup project** (`tests/fixtures/lifecycle.setup.ts`) — seeds two complaints (PENDINGFORASSIGNMENT + CLOSEDAFTERRESOLUTION with rating=4) via the PGR API, writes them to `lifecycle-fixtures.json`, three consumer specs fall back to those SRIDs when no env override is set. Degrades gracefully on deployments where the seed can't be built (e.g. broken user-service) — writes `{ status: 'skipped', skipped_reason }` instead of cascading every chromium test.

**Citizen flow runs on KC-overlay deployments** — bomet has the Keycloak SSO overlay enabled; the OTP login UI flow hangs because KC's `/realms/ke/.../token` endpoint currently 500s. Two test-side fixes:
1. New `citizenLoginViaApi` helper — mirror of the existing employee `loginViaApi`. Drives `/user-otp/_send` + `/user/oauth/token` directly with `userType=CITIZEN`, registers via `/user/citizen/_create` if needed, injects the token into `localStorage`. Works on every deployment (KC or not). `citizenOtpLogin` is now a thin wrapper calling this; the original UI flow is kept as `citizenOtpLoginViaUI` for tests that explicitly want the form.
2. New `forceDigitAuthProvider` — uses `page.route()` to rewrite `/digit-ui/globalConfigs.js` on the fly, flipping `authProvider` from `"keycloak"` to `"digit"` and dropping the `/kc/*` `tokenExchangeUrl` default. Without this, the SPA still routes all subsequent API calls through `/kc/<service>/...` which 502s when there's no KC session.

**Wizard test — modern digit-ui selectors** — bomet's `digit-ui` renders dropdowns as `<button role="combobox">` (shadcn-style) and options as `[role="option"]`, not the legacy `input.digit-dropdown-employee-select-wrap--elipses` + `.digit-dropdown-item`. Selectors now match both. The modern wizard also merges the old Step 4 (location cascade) into Step 3 (location details) — replaced the two separate sections with a single 3-iteration County → SubCounty → Ward loop.

**Helm fix — pgr-services boundary lookup** — pgr-services source reads `${egov.boundary.host}` (env `EGOV_BOUNDARY_HOST`). The pgr-services helm chart already binds it from the configmap key `boundary-service` — but the `egov-service-host` configmap **never defined that key**. So at runtime the env var resolved to empty, Spring fell back to the source default `http://localhost:8081`, the boundary HTTP call failed, response parsed as null, `ServiceRequestValidator.validateBoundary` NPE'd on `.getBoundary()`, and pgr-services returned `BOUNDARY_SERVICE_SEARCH_ERROR` for every complaint create. Adds a single line to the configmap exposing `boundary-service: "http://boundary-service.egov:8080/"`. Unblocks citizen complaint submission (UI was showing "Complaint couldn't be submitted") and the lifecycle.setup.ts seed.

## Verified against `bometfeedbackhub.digit.org`

- `citizen/login.spec.ts` — 2/2 pass (login page + OTP→home via OAuth bypass)
- `citizen/wizard.spec.ts` — walks all 6 wizard steps to Submit. Submit currently rejected on the live deployment because pgr-services hasn't been rolled with this helm fix yet. Once it ships, the wizard goes green.
- Live MCP smoke confirms the boundary-service v2 endpoint is healthy at `https://bometfeedbackhub.digit.org/boundary-service/boundary/_search`.

## Refs

- Closes / partly closes #685 (Phase 5 of the remediation)
- Related upstream tracking: egovernments/Digit-Core PR #1240 (JDK-21 base image — same image rebuild path that would also clear the user-service `SafeHtmlValidator` issue we briefly hit early in the run)

## Test plan

- [ ] Reviewer: run `npx playwright test tests/citizen/login.spec.ts` against naipepea — should stay 2/2 green (refactor preserves backward compat).
- [ ] Reviewer: with the helm fix applied to a tenant where pgr-services has been re-rolled, `npx playwright test tests/citizen/wizard.spec.ts` should walk all 6 steps and end on `/citizen/pgr/response` with an `NCCG-PGR-...` ticket ID.
- [ ] Reviewer: check the new `EGOV_BOUNDARY_HOST` resolution by `kubectl exec` into the pgr-services pod and `env | grep EGOV_BOUNDARY_HOST` — should print the boundary-service ClusterIP URL, not empty.

### DIFF
--- a/devops/deploy-as-code/charts/core-services/configmaps/values.yaml
+++ b/devops/deploy-as-code/charts/core-services/configmaps/values.yaml
@@ -56,6 +56,7 @@ configmaps:
       egov-user: "http://egov-user.egov:8080/"
       egov-user-egov: "http://egov-user.egov:8080/"
       egov-location: "http://egov-location.egov:8080/"
+      boundary-service: "http://boundary-service.egov:8080/"
       egov-filestore: "http://egov-filestore.egov:8080/"
       egov-localization: "http://egov-localization.egov:8080/"
       egov-idgen: "http://egov-idgen.egov:8080/"
--- a/tests/integration-tests/.gitignore
+++ b/tests/integration-tests/.gitignore
@@ -16,3 +16,4 @@ catalog.json
 history.json
 .git-sha
 lifecycle-fixtures.json
+test-results/
--- a/tests/integration-tests/.gitignore
+++ b/tests/integration-tests/.gitignore
@@ -15,3 +15,4 @@ report.json
 catalog.json
 history.json
 .git-sha
+lifecycle-fixtures.json
--- a/tests/integration-tests/package-lock.json
+++ b/tests/integration-tests/package-lock.json
@@ -1459,6 +1459,7 @@
       "integrity": "sha512-QP88BAKvMam/3NxH6vj2o21R6MjxZUAd6nlwAS/pnGvN9IVLocLHxGYIzFhg6fUQ+5th6P4dv4eW9jX3DSIj7A==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=12"
       },
--- a/tests/integration-tests/playwright.config.ts
+++ b/tests/integration-tests/playwright.config.ts
@@ -15,6 +15,7 @@ export default defineConfig({
   testMatch: [
     '**/*.spec.ts',
     'fixtures/auth.setup.ts',
+    'fixtures/lifecycle.setup.ts',
   ],
   timeout: 120_000,
   expect: { timeout: 15_000 },
@@ -40,14 +41,23 @@ export default defineConfig({
       testMatch: /tests\/fixtures\/auth\.setup\.ts$/,
     },
     {
+      // Runs after `setup`. Drives the PGR API end-to-end to seed two
+      // complaints (one PENDINGFORASSIGNMENT, one CLOSEDAFTERRESOLUTION
+      // with rating=4) and writes lifecycle-fixtures.json. Downstream
+      // specs that need a "pinned" SRID read from that file.
+      name: 'lifecycle-setup',
+      testMatch: /tests\/fixtures\/lifecycle\.setup\.ts$/,
+      dependencies: ['setup'],
+    },
+    {
       name: 'chromium',
       use: {
         browserName: 'chromium',
         storageState: 'auth.json',
       },
-      dependencies: ['setup'],
+      dependencies: ['setup', 'lifecycle-setup'],
       // Don't try to run setup itself as part of the chromium project.
-      testIgnore: /tests\/fixtures\/auth\.setup\.ts$/,
+      testIgnore: /tests\/fixtures\/(auth|lifecycle)\.setup\.ts$/,
     },
   ],
 });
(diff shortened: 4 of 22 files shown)


KDwevedi (approved):
Approving. The env-driven parameterization is clean and backward-compatible (naipepea defaults preserved), the lifecycle.setup graceful-skip is the right resilience pattern, and the MDMS-rule-based mobile assertions are the correct tenant-agnostic move.

Two non-blocking notes for follow-up (not gating this merge):

1. **`citizenOtpLogin` is now API-only — a UI-coverage gap.** The default helper routes through `citizenLoginViaApi` (token injection); `citizenOtpLoginViaUI` is preserved but nothing calls it, so no active test exercises the real OTP login form anymore. Worth repointing one citizen-login test at `citizenOtpLoginViaUI` so the form itself still has coverage — exactly where KC/OTP login tends to break.

2. **The helm configmap change won't affect compose deployments.** `boundary-service: http://boundary-service.egov:8080/` is the k8s ClusterIP topology; bomet (and the other compose stacks) already wire `EGOV_BOUNDARY_HOST: http://boundary-service:8081/` directly in docker-compose and never read this configmap. So it's a valid fix *for helm/k8s*, but it can't be what unblocks the bomet citizen-submit failure — that root cause is elsewhere on the compose stack. Suggest correcting the PR description's causal claim (and ideally splitting infra one-liners from test PRs in future).

(Tenant-hardcode finding from the review is already addressed in 88704e3.)

PR #828, opened 2026-06-11
feat(deploy): nightly-develop image pipeline for high-churn container services

KDwevedi:
## Why
A fresh `./deploy.sh` deviates from bomet partly because high-churn services are **built on the box at deploy time** (slow, non-deterministic). This builds them nightly from `develop`, pushes a rolling `nightly-develop` tag (+ immutable `develop-YYYYMMDD`) to the registry, so deploys **pull pre-built images** instead of building.

## What (this PR — container services)
- **`build/build-config.yml`**: add pipelines for `digit-mcp`, `otp-publisher`, `digit-ui-esbuild`.
- **`local-setup/ansible/files/nightly-build-push.sh`**: builds `pgr-services-dev`, `default-data-handler`, `digit-mcp`, `otp-publisher` (amd64) and pushes `nightly-develop` + `develop-YYYYMMDD`. Push registry comes from `$NIGHTLY_PUSH_REGISTRY` (set on the build host) so no internal address is committed. Runs from a fresh develop checkout (the bomet redeploy wrapper invokes it after the develop sync, before the converge — the nightly self-builds what it deploys).
- **compose**: parameterize `pgr-services` image → `${PGR_SERVICES_IMAGE:-<prev default>}`.
- **`digit.env.j2`**: plumb `PGR_SERVICES_IMAGE` + `OTP_PUBLISHER_IMAGE` (`mcp`/`ddh` already plumbed).

**Opt-in / no default change:** every image var defaults to the prior behaviour. Pinning a nightly image is opt-in via host_vars, so this PR changes nothing until a deployment sets the image vars + turns its `build_*` flags off.

## Validated
On bomet: all four services build from current develop and push `nightly-develop` + `develop-20260611` to the registry. (digit-mcp needs current develop — an older checkout has a since-fixed TS error.)

## Follow-up (separate PR)
UIs (`digit-ui-esbuild` / `digit-ui-v2` / `configurator`): their bundles bake tenant build-env (`VITE_KEYCLOAK_REALM` etc.), so they need build-arg-parameterized Dockerfiles (v2 + configurator have none yet) and a pull-and-extract deploy path into the host-nginx dirs. The build-config `digit-ui-esbuild` entry is included here to stage that.

### DIFF
--- a/build/NIGHTLY-BUILDS.md
+++ b/build/NIGHTLY-BUILDS.md
@@ -46,18 +46,32 @@ So PGR is `…/pgr-services:nightly-develop` and `…/pgr-services:develop-20260
 **In scope — everything in `build-config.yml`** (CCRS-owned): `pgr-services`,
 `novu-bridge`, `digit-config-service`, `digit-user-preferences-service`,
 `xstate-chatbot`, `default-data-handler`, `digit-mcp`, `otp-publisher`,
-`digit-ui` (legacy micro-ui), `digit-ui-esbuild`, and the `*-db` flyway images.
+`digit-ui` (legacy micro-ui), `digit-ui-esbuild`, `configurator`, `digit-ui-v2`,
+and the `*-db` flyway images.
 
 **Out of scope — DIGIT core platform services** (`egov-*`, `kong`,
 `boundary-service`, mdms-v2, etc.). These do **not** live in this repo, so this
 nightly does not build them; they are pulled from the registry as today. This
 pipeline owns only CCRS-repo services.
 
-**Deferred — `digit-ui-v2` / `configurator`.** Their bundles bake tenant
-build-env (`VITE_KEYCLOAK_REALM`, etc.) and neither has a build-arg-parameterized
-Dockerfile yet, so a single nightly image can't serve every tenant. Tracked as a
-follow-up; they are intentionally absent from `build-config.yml`'s buildable set
-until that's solved.
+### Vite SPAs — `configurator` and `digit-ui-v2`
+
+Both now have build-arg-parameterized Dockerfiles and are built nightly. They
+differ in how much they bake:
+
+- **`configurator`** reads no `VITE_*` build-env (only `import.meta.env.MODE`),
+  so **one image serves every tenant** — configuration is runtime (login + tenant
+  pick). Its `nightly-develop` image is fully deployable as-is.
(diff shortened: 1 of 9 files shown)


subhashini-egov:
@KDwevedi shouldn't all PGR services in the repo (backend + frontend) be built nightly and pushed? Because people are fixing and pushing across - not just mcp and esbuild.

subhashini-egov (approved):
@KDwevedi Do we need to build the -dev images as well. Also there's a bit of naming drift pgr-services vs pgr-services-dev. Might want to look into it.

KDwevedi:
Addressed the review — nightly now builds **every** service in `build/build-config.yml` (no hand-maintained list; add a service there and it's enrolled), with canonical image names that resolve the `pgr-services` vs `pgr-services-dev` drift (channel lives in the tag: `:nightly-develop` / `:develop-YYYYMMDD`). Conventions written down in `build/NIGHTLY-BUILDS.md`.

Also closed the last pull-side gap: `digit-ui` was the only deployed image still hardcoded (`digit-ui:pgr-fixes`) — it's now `DIGIT_UI_IMAGE`-driven like the rest. Documented the cutover gotcha that bites in practice: pinning `*_image` to `:nightly-develop` only takes effect if the matching `build_*` flag is `false`, since `build_*: true` builds `:local` on-box and overrides the pin.

KDwevedi:
Frontends are now fully covered too — added build-arg-parameterized Dockerfiles for **`configurator`** and **`digit-ui-v2`** and enrolled both in `build-config.yml`, so every CCRS frontend (micro-ui, esbuild, configurator, v2) builds nightly:

- **configurator** reads no `VITE_*` build-env (only `import.meta.env.MODE`) → one image serves every tenant, fully deployable as-is.
- **digit-ui-v2** bakes a tenant env contract (`VITE_KEYCLOAK_REALM` / `VITE_CITIZEN_*`); the Dockerfile exposes them as build-args (defaults = the playbook contract, with `/auth` + `/token-exchange` already neutral), so the nightly builds a tenant-neutral reference image. A single image truly serving every tenant is the runtime-config follow-up.

Both mirror the proven on-box recipe (file: sub-packages built first; configurator uses `vite build` directly since its root `tsc -b` has upstream type errors) and I validated both with a real `docker build` — they produce `dist/index.html` and serve at the correct base path (`/configurator/`, `/citizen/`) with prefixed assets.

PR #841, opened 2026-06-11
Non-India tenant hardening: pincode allowlist, HRMS bootstrap, prebuilt UI bundle + citizen e2e suite

subhashini-egov:
## Summary

Hardens tenant onboarding and the citizen complaint flow for **non-India tenants** (driven by a live `ethiopia` deployment), and brings the citizen e2e suite up to the V2 UI so every fix here is regression-tested. All changes are live-verified on https://subhadev.digitlab.in (tenant `ethiopia`).

### Citizen UI (digit-ui-esbuild)
- **Postal-code field disabled itself after one keystroke** (`d3caecb1`): `pincodeKnown` in `CreatePGRFlowV2.tsx` gated the field on the user's own input; it now gates only on a map-provided pincode.
- **"Sorry we are not providing service in this city" on Submit** (`df395d8b`): the pincode-allowlist gate only honored the *map-resolved* ward. Tenants without boundary polygons can only produce manually selected wards, so any typed postal code outside a (stale) allowlist hard-blocked submission. A manually selected `SelectedBoundary` now supersedes the allowlist, in both the V2 flow and the legacy `FormExplorer`.

### Tenant onboarding
- **default-data-handler seeded India's pincodes onto every new tenant** (`771bd271`): the baked `tenant.tenants` template hardcoded `pincode: [143001–143005]` onto `{tenantid}.citya` — the root cause of the submit veto above. The key is now absent (the schema doesn't require it; the UI treats absence as "all codes serviceable"; mdms-v2 rejects `pincode: []` on update).
- **HRMS crash-loop on non-pg tenants** (`f1811e55`): HRMS looks up its `INTERNAL_USER` system user on `state_root`, which only existed on `pg`. The playbook now seeds it post-bootstrap (mobile generated from the tenant's `core_mobile_configs` — egov-user validates the tenant rule even on `_createnovalidate`) and rewrites the hardcoded `pg` in the HRMS prereq gate to `state_root`.

## Features

- **Config-driven pincode allowlist** (`2e5bfa9f`, `f02df6f1`): new optional `pincode_allowlist` arg on the MCP `tenant_bootstrap` tool — RMWs every active `tenant.tenants` record at the target root with the supplied codes (numeric strings coerced to Number per the schema; idempotent re-runs skip unchanged records). Ansible forwards a new `pgr_pincode_allowlist` host_var on both bootstrap passes (`default(omit)`); documented in `_example.yml`. Leave unset for no allowlist.
- **Prebuilt UI bundle serving** (`4b61d3c7`): `digit_ui_bundle_image` host_var lets static mode serve a registry image (extracted via `docker create`/`cp`) instead of force-resetting `/opt/digit-ui-esbuild` from the flywheel git repo — which is older than this monorepo and silently reverted UI fixes on every deploy. Also fixes the stale `docker/Dockerfile` (missing `products/`, `scripts/`, lockfile, CSS prebuild).
- Compose fallback tags bumped: `digit-mcp:2026-06-11`, `default-data-handler:2.12-draft-20260611-111521` (both published to ghcr).

## Tests

- **Citizen specs bucketed** under `e2e/specs/citizen/` (`b1dc679e`, `57537b2b`).
- **New** `complaint-type-localization.spec.ts` (`2119417f`): fails if the complaint-type combobox renders raw `SERVICEDEFS.*` keys; captures every localization fetch for diagnosis.
- **New** `complaint-submit.spec.ts` (`c3899f43`): full wizard — OTP login, type, map confirm, manual Region→Ward cascade, **typed postal code** (the exact veto input) — through SUBMIT; asserts a complaint reference and that the serviceability veto never appears. Nominatim reverse-geocode is stubbed (a live postcode answer makes the field read-only and would bypass the gate under test). Shared OTP login extracted to `e2e/utils/citizen-auth.ts`.
- **Tenant-aware mobile numbers** (`5d61e4f2`): new `utils/mobile.ts` — every minted mobile in e2e + smoke goes through `uniqueMobile()`/`fixedMobile()`, shaped by `CITIZEN_MOBILE_LENGTH`/`CITIZEN_MOBILE_PREFIX`; the Playwright global setup auto-derives those from the target deployment's `globalConfigs.js` (`coreMobileConfigs`); pg defaults preserve existing behavior. Also modernizes the lifecycle spec to the V2 UI (login selectors, `complaint-details/:id` route, current status labels) and rewrites `flow.spec.ts` (the first-visit language gate no longer exists).
- **Unit + static contract coverage** (`6604d019`): see mapping below.

## Change → test mapping

| Change | Test |
|---|---|
| Postal-code self-disable fix | `citizen/complaint-submit.spec.ts` (types postal key-by-key; `fill()` would mask the regression) |
| Allowlist gate honors manual ward | `citizen/complaint-submit.spec.ts` (manual cascade + typed postal through SUBMIT) |
| SERVICEDEFS localization | `citizen/complaint-type-localization.spec.ts` |
| DDH template strips pincodes | `static/deployment-contracts.test.ts` (template JSON has no `pincode` key) |
| MCP `pincode_allowlist` normalization | `digit-mcp/test-pincode-allowlist.ts` (13 cases: coercion, pass-through, off states) |
| Ansible forwards `pgr_pincode_allowlist` ×2 | `static/deployment-contracts.test.ts` |
| INTERNAL_USER seed + prereq-gate rewrite | `static/deployment-contracts.test.ts` |
| `digit_ui_bundle_image` plumbing | `static/deployment-contracts.test.ts` |
| Compose image fallbacks | `static/deployment-contracts.test.ts` |
| Tenant-aware mobile generators | `static/mobile-rules.test.ts` (env overrides, both shapes, padding/truncation, legacy parity) |

## Verification (live, subhadev / ethiopia)

- `npx playwright test e2e/specs/citizen/` (non-Keycloak specs): **15/15 passed** — includes the full citizen→employee lifecycle (file → inbox → assign → resolve), creating real complaints (e.g. `PG-PGR-2026-06-11-001474`)
- `npx tsx digit-mcp/test-pincode-allowlist.ts`: **13/13 passed**
- `npx jest static/`: **17/17 passed**
- Ethiopia root bootstrap re-run on the new MCP image: success, 0 failures (idempotent); `pincode_allowlist: ["9999"]` spot-test wrote all records, then reverted
- `ansible-playbook playbook-deploy.yml --syntax-check`: clean
- Published images smoke-tested: `digit-ui-esbuild:20260611-101103`/`nightly-develop`, `digit-mcp:2026-06-11`/`latest`, `default-data-handler:2.12-draft-20260611-111521`

## Notes

- `citizen/proxy.spec.ts` and `pgr-complaint.spec.ts` are Keycloak-based and self-skip when KC is down (not running on this box); they received the mobile-generator treatment but weren't exercised.
- The smoke (Jest) suite needs per-service ports (e.g. egov-user `:18107`) that this deployment doesn't publish — its failures predate this branch and are connectivity-only.
- This monorepo is the sole source of truth for digit-ui-esbuild; deployments serve its bundle via `digit_ui_bundle_image` (no upstream sync needed).


ChakshuGautam:
I think we should call the MDMS API to setup the fixtures dynamically and do some of these tests. I feel that would reduce the hardcodes - phone number regexes, etc.

subhashini-egov:
> I think we should call the MDMS API to setup the fixtures dynamically and do some of these tests. I feel that would reduce the hardcodes - phone number regexes, etc.

Yes, this is in place.

ChakshuGautam:
Looked specifically at the mobile-number derivation (`local-setup/tests/utils/mobile.ts` + `global-setup.ts`). One gap worth fixing before merge:

## The tests don't read the rule egov-user actually enforces

The enforcement point is **MDMS `common-masters.UserValidation`** (egov-user's `MobileNumberValidator` reads it via MDMS-v2 and caches it in Redis under `validationRules`). The global setup instead parses **`globalConfigs.js` → `coreMobileConfigs`** — the UI-side copy.

Today the two agree, because both are deploy-time renderings of the same host_var (`core_mobile_configs` → playbook `mobile_regex`/`mobile_length` → MCP-synthesized UserValidation; same var → rendered `globalConfigs.js`). Verified on subhadev: MDMS rule `^[17][0-9]{8}$`/min9/max9 matches `coreMobileConfigs` exactly.

But they drift the moment anyone edits UserValidation at runtime (configurator/MCP + `DEL validationRules` — the documented flow for tuning validation without a redeploy). `globalConfigs.js` only changes on redeploy, so the suite would keep minting numbers against the stale copy and fail (or worse, green-light numbers the server now rejects).

## The projection is also lossy

`global-setup.ts` keeps only `mobileNumberLength` + `mobileNumberAllowedStartingCharacters[0]` and **drops `mobileNumberPattern` entirely**. That works for the current tenants (pg `9XXXXXXXXX`, ethiopia/ke `[17]XXXXXXXX`) because their patterns have no internal structure — but any rule like `^0?[17][0-9]{8}$` (optional leading 0, min≠max) or a pattern constraining middle digits silently produces invalid numbers with no useful failure message.

## Suggestion (small, contained to global-setup + mobile.ts)

1. Resolution order: env > **MDMS UserValidation** (one `_search` for `common-masters.UserValidation` at the state tenant — the setup already has creds/kong) > globalConfigs fallback > pg defaults.
2. Export `CITIZEN_MOBILE_PATTERN` alongside length/prefix, and make `uniqueMobile()`/`fixedMobile()` assert their own output against it, failing fast with an actionable message (`generated 712345678 violates tenant rule ^…$ — check CITIZEN_MOBILE_* / MDMS UserValidation`). That single self-check catches both the drift case and the lossy-projection case regardless of which source won.

Everything else in the mobile-generator change looks right — lazy env reads so the setup's exports are seen, offset for same-millisecond mints, the pg defaults preserving the old hardcoded behavior, and `uniqueMobile()` replacing the `9${timestamp}` / `8${timestamp}` literals in the smoke suite.

KDwevedi (approved):

Issue #757, opened 2026-06-05
[Stretch] Auto-populate boundaries from OSM during account-administrator onboarding


KDwevedi:
## Summary

Boundaries should be auto-populated during the account-administrator onboarding flow, rather than requiring a manually prepared XLSX/GeoJSON boundary template.

## Proposed sources
- **OSM (OpenStreetMap)** — pull administrative boundary polygons by region during onboarding.
- **DIGIT-Health** boundary data — alternative source for health-context tenants.
- **Autopoll from a collated repo** — periodically sync from a maintained boundary repository.

## Notes
- Relates to the onboarding & configuration epic #548 and the existing manual boundary-template path (#686).

@kdwevedi to elaborate.

KDwevedi:
Implemented in PR #844 (OSM path of the dual-path Phase 2): during onboarding, the operator searches a city/region (with Turbopass autocomplete), OSM admin levels are fetched country-scoped from Overpass, mapped to local level names, and created as hierarchy + boundaries **with real polygons** — alongside the existing XLSX/GeoJSON path (#686), which stays intact.

E2E-validated against the live bomet backend on a scratch tenant (Cidade de Maputo → 66 boundaries, 3 levels, real geometry, localization, and the citizen map rendering all 59 bairros with pin→bairro resolution) — evidence in the PR comments.

Of the proposed sources, this covers the OSM one; DIGIT-Health and autopoll-from-collated-repo remain open ideas.

Gurjeet-eGov:
@KDwevedi Could you please check this issue 

https://github.com/user-attachments/assets/0baf5c65-766e-42ff-ada9-34d27c0e0b5c

KDwevedi:
Closing as **completed** — the OSM auto-populate feature shipped and merged via PR #886 (dual-path Phase 2: OSM one-click + Excel), E2E-validated against the live backend (evidence in PR comments).

The robustness gap shown in the video above — the target admin-level detection over-populating/mis-populating the level-selection dropdown when the searched region isn't cleanly matched — is a distinct follow-up and is tracked separately in #1016 (guards for validating the detected target before populating the dropdown). That is separate from #980 (Overpass name-tag match fix).

DIGIT-Health and autopoll-from-collated-repo remain open source ideas but out of scope for this ticket.

Issue #686, opened 2026-05-30
[Bug] Configurator onboarding flow - boundary template is not generated


subhashini-egov:
**Describe the bug**
Unable to get the boundary template XLSX during the onboarding flow. It leaves the user "stuck" in Phase 2

**To Reproduce**
1. Go to /configurator
2. Click on "Onboarding"
3. Start adding a new tenant with branding (see attachment)
4. Move to Phase 2
5. Click on "Download Boundary Template". Pop-up appears saying the template download should appear here. But nothing shows up after 10 minutes.

**Expected behavior**
The boundary template should be downloaded

**Screenshots**
https://www.loom.com/share/664ead243b4648dfb55e134a6b9c6b05

**Additional context**

subhashini-egov:
@priyanshu-egov Have you raised a PR for this?

priyanshu-egov:
https://github.com/egovernments/Citizen-Complaint-Resolution-System/pull/714/changes/c21a4db00bf56e3721d5e88c97da0e8eb555daa0

Gurjeet-eGov:
Resolved  

https://github.com/user-attachments/assets/35b6e381-5385-46e9-8c11-a9628ba3dc2b
[Boundary_Template_ADMIN_TEST_ONE.xlsx](https://github.com/user-attachments/files/28793586/Boundary_Template_ADMIN_TEST_ONE.xlsx)

PR #844, opened 2026-06-12
feat: one-click OSM boundary onboarding alongside Excel upload

KDwevedi:
Implements #757 (OSM source; the XLSX path #686 is preserved).

## What

Configurator **Phase 2 now offers two paths** for boundary setup:

1. **Fetch from OpenStreetMap (one-click)** — type a city/region name (with optional type-ahead from the vendored Turbopass search), pull its full administrative boundary tree + real polygons from Overpass, name each admin level, and create hierarchy + boundaries + localization in one go.
2. **Upload from Excel** — the existing template/upload/verify flow, restored unchanged (including the optional polygon GeoJSON sidecar from #621).

Both paths share one post-create pipeline: boundary + hierarchy-level localization, localization cache-bust, and the `fix_boundary_paths` cleanup.

Builds on @dhruv-1001's `feat/fetch-osm-geojson` (the OSM fetch, level-mapping UX, and MAP_TENANT-driven maps are his design), rebased onto current develop and hardened.

## Changes by area

**configurator**
- Dual-path Phase 2 (`Phase2Page.tsx`, new `utils/osmBoundaries.ts`); OSM path produces name-based `UPPER_SNAKE` codes (diacritics-stripped, `name:en`/`int_name` romanization fallback, deterministic collision suffixes) instead of raw OSM relation ids
- No silent re-parenting: unnamed/non-romanizable/orphaned features are excluded and listed in a review step before creation
- Spatial parenting honors polygon holes (enclaves) and resolves ambiguity to the smallest containing parent; representative points come from the same largest MultiPolygon member that gets persisted
- Overpass query escapes input, restricts to administrative areas, and scopes by the picked suggestion's country (fixes the "Maputo vs Cidade de Maputo" worldwide-name-union problem when a suggestion is chosen)
- Hierarchy-exists conflicts fail fast with both level lists named instead of a slow doomed create
- Phase 4: blocking banner + retry when reference data fails to load (no more wall of bogus "boundary not found" errors); unmatched/ambiguous jurisdictions fail the row instead of silently defaulting to `Ward`/`ADMIN`

**digit-ui-esbuild (products/pgr)**
- New `useTenantBoundaries` hook: when `MAP_TENANT` is configured, complaint maps render the tenant's real leaf-level polygons (two-step relationships→entities fetch, placeholder-geometry filtering); static Nairobi wards remain the fallback, so existing deployments are unchanged until they opt in
- `findWardPath` targets the hierarchy's deepest level instead of hardcoded `Ward`; dropdown labels stay localization-first

**ansible / templates**
- `globalConfigs.js.j2` emits `MAP_TENANT` from `city_tenant`/`ui_state_tenant_id` (runtime path; build-time `REACT_APP_MAP_TENANT` kept as fallback)
- New **default-off** `enable_turbopass` host_var: deploys the search-api container on `127.0.0.1:13301` behind an nginx `/turbopass/` proxy; with the flag unset, rendered nginx output is byte-identical to today

**turbopass/ (new, vendored)**
- `dhruv-1001/osm-mapped-data` @ `ccbaea14`: Overpass scraper, NestJS Trie search API, scraped hierarchy data (54 African countries + India)
- Added Dockerfile, `DATA_DIR` env, `/health`, honored `limit` param, README with provenance + contract

## Validation

- `configurator` `npm run build` (tsc + vite) green; `digit-ui-esbuild` `npm run build` green (7.3 MB bundle, expected size)
- turbopass search-api boot-checked: `/health` → 20,279 locations loaded; `limit` clamp verified end-to-end
- Three-lens adversarial review (regression-vs-develop, pipeline correctness, conventions); all 5 blockers + 8 majors fixed in the `fix(...)` commits — notable catches: boundary entity `_search` reads query params only (the original body-based fetch could never work), and the locality picker `name` key would have blanked every existing deployment
- **Not yet box-validated end-to-end** — suggested test: bootstrap `mz` + `mz.maputo`, run Phase 2 → "Fetch from OpenStreetMap" → search "Cidade de Maputo" (pick the suggestion), map levels, create; then file a citizen complaint and check the locality dropdown + map polygons

## Known limitations

- Overpass matching is exact-name; without a picked suggestion, homonyms worldwide can merge (scoping requires choosing a suggestion)
- Public `overpass-api.de` is rate-limited (90 s timeout); large cities may need the local Overpass instance from `turbopass/scraper`
- `chore(configurator)` commit fixes 27 pre-existing type errors that surface once the local `@digit-mcp/data-provider` is compiled — mechanical, kept separate for easy review


KDwevedi:
## E2E validation — local build against bomet's live backend (2026-06-12)

Ran the full one-click flow with the branch's builds locally, same-origin-proxied to bomet's Kong over an SSH tunnel (so all reads/writes hit the real boundary-service / localization / MDMS), targeting the existing scratch tenant `ke.pwt474429045` (zero hierarchies before the run).

**Configurator (vite dev, `/turbopass` proxied to the vendored search-api running locally):**
- Dual-path landing renders; OSM card → typed "Cidade de Maputo" → Turbopass suggestion (`countryCode: MZ`) appeared and was picked → country-scoped Overpass query returned **3 clean admin levels** (L4: 1, L5: 6 districts, L8: 63 bairros — no LineString pollution, root relation preserved)
- Mapped levels `Municipio / Distrito / Bairro` → review step listed **4 skipped bairros** ("no parent found" — honest exclusion instead of silent re-parenting) → **66/66 boundaries created**
- Server-side verification through the tunnel: ADMIN hierarchy with the 3 levels; 66-node relationship tree rooted at `CIDADE_DE_MAPUTO` (6 Distrito children, 59 Bairro leaves); **name-based codes** (`CHAMANCULO_A..D`, `3_DE_FEVEREIRO`, …); **real polygons** (root ring 1,544 pts spanning 0.26°; bairros 35–121 pts — no unit-square placeholders)
- Post-create pipeline ran on the OSM path: 2× `localization _upsert` → 200 (**132 messages** in `rainmaker-boundary-ADMIN`, names with diacritics intact), `cache-bust` → 200, `fix_boundary_paths` → 404 (bomet's Kong doesn't route the MCP shim) handled as designed, non-fatal

**digit-ui (branch's 7.3 MB build served locally, `MAP_TENANT=ke.pwt474429045` via globalConfigs):**
- `useTenantBoundaries` issued exactly the intended wire calls: `boundary-relationships/_search?tenantId=…&hierarchyType=ADMIN&includeChildren=true` then chunked `boundary/_search?tenantId=…&codes=…&limit=100` (40 + 19 codes)
- **59 bairro polygons rendered** on the complaint-location map; dropping a pin inside one highlighted it and resolved `POLANA_CANICO_B`, with the nominatim reverse-geocode populating the address

Known cosmetic: the map tooltip shows the boundary code (relationship nodes carry no name field); the locality dropdowns localize via `t(code)` as before. Scratch data left on `ke.pwt474429045` as a reusable fixture.

vinothrallapalli-eGov:
@KDwevedi Please resolve the merge conflict

KDwevedi:
Done

vinothrallapalli-eGov (approved):

PR #845, opened 2026-06-12
feat(micro-ui): supervisor dashboard (filters/tables) + complaint map widget

KDwevedi:
## What

The clean delta of the supervisor dashboard work on top of current `develop` (supersedes the conflicting `feature/demo-dashboard` PR #829, which re-adds the already-merged base), **plus a new complaint map widget**.

### Dashboard (by @soniikajal)
- Global **Time-window** filter (daily/weekly/monthly/WoW/MoM) for volume KPIs.
- **Data tables** + KPI inventory, draggable/resizable widgets, layout persistence/reset fixes.

### Complaint map widget (new)
- `ComplaintMap` (plain Leaflet, `circleMarker` — no icon-asset webpack issues): one pin per complaint location, **centred on Bomet**.
- Fed by a new `cl_map_complaints` analytics query (`grain=facts`, `dimensions=[latitude,longitude,service_code,application_status]`, `count`, `has_geo_pin`).
- Registered as a `map` widget type in `DashboardGrid` + default layout; appears in the KPI inventory.
- Adds `leaflet` dependency.

## Dependency
The map query needs the analytics catalog to expose `latitude`/`longitude` as facts dimensions — shipped as the companion backend PR. Without it the query 400s (map shows "No mapped complaints", everything else works).

## Validation
Built + deployed live: page loads, map renders, the pin query returns ~37 complaint locations centred on Bomet. Queries use the merged `/v2/analytics` grammar; no host refs/secrets; the `start`/`build` decoupling and de-Bomet from prior PRs are preserved.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #847, opened 2026-06-12
feat(pgr-analytics): expose latitude/longitude as facts dimensions (map support)

KDwevedi:
## What

Exposes `latitude`/`longitude` as **dimensions** on the `complaint_facts` grain in the analytics catalog, so the query API can return per-complaint pin locations for a map widget:

```json
{ "grain":"facts", "dimensions":["latitude","longitude","service_code","application_status"],
  "measures":[{"name":"n","agg":"count"}], "filters":{"has_geo_pin":true} }
```

The columns already exist on the `complaint_facts` materialized view (64/66 complaints are pinned) — this 2-line change just adds them to the `groupable` set so `dimensions:[latitude,longitude]` no longer 400s.

## Why
Backs the **complaint map widget** in the dashboard PR (#845). Realizes part of the "map support" the visualization-validator design called out.


KDwevedi:
Folded into #845 so the map widget + its backing catalog change land in one reviewable PR.

Issue #753, opened 2026-06-05
[Bug]  File attachment not visible in Citizen view screen


pradeepkumarcm-egov:
# Description

A citizen logs into the system and uploads an image while creating a CMS application. The file is successfully uploaded, and the corresponding fileStoreId is generated and stored along with the CMS application.

However, the uploaded attachment is not visible in the application view screen.

Observation
The fileStoreId is being generated and saved correctly.

While rendering the attachment URL, the application is generating the context path using localhost instead of the configured File Store service endpoint.

As a result, the attachment URL is invalid, and the image cannot be displayed in the UI.

Expected Behavior
The application should generate the attachment URL using the configured File Store endpoint, allowing users to view the uploaded image from the application details screen.

Suggested Fix

This requires a configuration change at the Ingress level to ensure that the File Store service URL is resolved correctly and exposed through the appropriate host/context path instead of localhost.

#Attachment 
The url is localhost instead of filestore url : 
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/db7ca96f-cdeb-4f01-9232-a993c6ad7dfc" />

Localhost url :
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/98e2e955-f283-4303-954f-a73711ce67e2" />

The file saved but url path is wrongly added.
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/87d246c1-926a-4fc9-b9bc-90e61a8ab22c" />

Lokendra-egov:
Root cause identified and fix verified.

The uploaded files and signatures were valid; the issue was limited to incorrect attachment URLs. The instance was deployed with `domain: localhost`, causing nginx to rewrite file URLs as `http://localhost/file-store/...`, which is inaccessible from end users' browsers.

Verified on the live instance (`digit.mctd.gov.mz`):

* Reproduced the issue with a new file upload.
* Confirmed that the same file is accessible when the hostname is corrected to `digit.mctd.gov.mz`.
* Verified that the presigned URL remains valid after host replacement.

A fix has been raised to use nginx runtime `$host` instead of the render-time domain value when rewriting file URLs. This ensures that attachment URLs always use the hostname from the incoming request and prevents similar issues caused by domain misconfiguration.

The issue is resolved from a code perspective. Deployment of the updated nginx configuration (or a manual nginx hotfix) is required on the affected instance.

[...]

PR #856, opened 2026-06-12
fix(deploy): rewrite filestore presigned URLs with runtime $host

Lokendra-egov:
The /filestore/ sub_filter rewrote the internal minio:9000 host in presigned URLs to the render-time inventory `domain` value. A box deployed with `domain: localhost` but served on a real hostname handed browsers unresolvable http://localhost/file-store/... attachment URLs (egovernments/Citizen-Complaint-Resolution-System#753, observed live on digit.mctd.gov.mz — the object serves fine once the host is corrected, since /file-store/ re-sets Host: minio:9000 and the public hostname never participates in AWS4 signature validation).

Use nginx's runtime $host instead, so attachment URLs follow whatever hostname the client actually used, in both digit_ui_mode branches.


pradeepkumarcm-egov (approved):

vinothrallapalli-eGov:
@KDwevedi Please review

PR #866, opened 2026-06-14
Supervisor dashboard: header & filter redesign, KPI cards, and improved grid drag/swap

soniikajal:
### Header redesign
- New `DashboardHeader` with operations title, live subtitle (geography + date range), search field, export button, **Add KPI** dropdown, and reset layout action
- Header integrated into `DashboardLayout` / `AdminDashboard` flow

### Filter redesign
- Redesigned `DashboardFilters` bar: date range pickers, geography dropdown, complaint type dropdown, and clear-filters control
- Filters wired through `useDashboardFilters` and `useDashboardData`
- Geography options can load dynamically via new `boundaryService` (boundary API integration)

### KPI cards & widgets
- Updated `KpiCard` styling with status coloring, context lines, and embedded ranked lists for list-type KPIs
- New `kpiDisplay.js` for thresholds, display titles, and status styling
- New `AddKpiDropdown` to add KPI/chart widgets from the header
- KPI query and data wiring updates in `kpiQueries.js` / `useDashboardData.js`

### Grid layout, dragging & swapping (improved, still WIP)
- Axis-locked drag: horizontal moves affect **X only**; vertical moves affect **Y only**
- Unified CSS transforms during drag so sibling cards slide predictably (workaround for react-grid-layout ignoring layout updates while dragging)
- Horizontal row repack for KPI and table rows during drag
- KPI gap filling after drag, resize, remove, or add (`packKpiLayout` + vertical compact)
- Layout persistence bumped to **v20**
- Empty-dashboard guard and safer layout merge/persist logic

**Known issues / still buggy:**
- Drag-and-swap behavior is improved but not fully reliable in all cases (mixed KPI/table rows, vertical drags, occasional jitter)
- Gap filling after remove/reorder may still miss edge cases (multi-row KPI bands, mixed KPI heights)

### Other changes
- `DashboardGrid.jsx` — RGL config updates (`compactType={null}`, drag handlers, transform-based preview)
- `useDashboardLayout.js` — layout load/save, drag session, optimize-after-change pipeline
- `Sidebar.jsx`, `dashboard.css`, `input.css` — styling aligned with new header/filter layout
- `mapGeoUtils.js` — map geo helpers
- `constants/layoutConfig.js` — default layout and widget sizing updates


frontend/micro-ui/web/src/dashboard/components/DashboardGrid.jsx
```diff
@@ -1,112 +1,196 @@
...
-      lg: layout.map((item) => ({
-        ...item,
-        resizeHandles: RESIZE_HANDLES,
-      })),
-    }),
-    [layout]
-  );
+  const gridWrapRef = useRef(null);
+  const dragStartRef = useRef(null);
+  const [preventCollision, setPreventCollision] = useState(false);
```
KDwevedi: **Resize can leave a permanent overlap → wipes the whole saved layout on next load.** (re: this `preventCollision` state, used at the `<GridLayout preventCollision={preventCollision}>` prop ~L379)

During resize `preventCollision` is true, so RGL does not push neighbours when a widget grows wider/taller. The repair helpers built for exactly this — `pushAdjacentItems` / `resolveOverlapsForItem` in `useDashboardLayout.js` (~L291/L352) — are **never called**. The resize path (`onLayoutStop(..., "resize", activeId)`, ~L330) only runs `optimizeLayoutAfterChange`, whose `reflowAndPack` packs horizontally within existing rows and compacts *upward* — it never resolves a widget overlapping its right-hand neighbour.

Consequence: the overlap gets persisted, and on next load `parseStoredLayout` rejects any overlapping layout and falls back to `DEFAULT_LAYOUT` — the user silently loses their entire arrangement.

Fix: in the resize branch run `pushAdjacentItems(merged, activeId)` (the pinned-id path is purpose-built for this) before compaction/persist. `activeId` is already threaded through `onLayoutStop`.

frontend/micro-ui/web/src/dashboard/hooks/useDashboardLayout.js
```diff
@@ -32,6 +32,261 @@ function hasOverlaps(layout) {
...
+}
+
+function reflowAndPack(layout) {
+  let result = packKpiLayout(layout);
+  result = packNonKpiRows(result);
+  return result;
+}
+
+/** After drag/resize/remove: fill gaps in KPI rows, pack widget rows, compact upward. */
+export function optimizeLayoutAfterChange(layout) {
```
KDwevedi: **`reflowCharts` (and `normalizeLayout`, ~L433) are exported but never called — this is the gap-fill/interleave bug.** (anchored on `optimizeLayoutAfterChange`, where the fix belongs)

`reflowCharts` holds the "KPI band on top, charts below" invariant. But the live load/optimize path here uses `reflowAndPack`, which only packs horizontally and compacts upward, with no KPI-vs-chart vertical separation. So loading a non-trivial saved layout can interleave KPI cards and charts in the same rows — almost certainly the "gap filling still buggy" symptom in the PR description: the correct function exists, it just isn't wired in.

Fix: call `reflowCharts` inside `optimizeLayoutAfterChange` (after `reflowAndPack`, before final compaction), or fold its "charts below `kpiBottom`" logic into `reflowAndPack`. Then either wire in or delete `normalizeLayout`.

KDwevedi (commented):
Two layout-correctness findings from a review pass (flagging for discussion, not hard-blocking). Both are reachable through normal use and both can corrupt or silently lose the user's saved arrangement, so worth addressing before this lands. Anchored near the relevant code; details inline.

KDwevedi (approved):

PR #873, opened 2026-06-16
feat(configurator): selectable, contiguous OSM admin levels in Phase 2

KDwevedi:
## What

In the OSM path of configurator Phase 2, the level-mapping step listed **every** discovered OSM admin level and included any level you gave a name to — with no guard against gaps. Two problems:

- Big regions forced an all-levels tree (e.g. Delhi returns 1,200+ relations across many admin levels — far more hierarchy than an operator wants).
- Naming level 4 and level 8 while leaving 5 blank **silently reparented** level-8 features onto level-4, collapsing the middle tier into a misleading hierarchy.

## Change

- Each discovered level now has an explicit **checkbox** (all selected by default — a contiguous, valid starting point).
- The hierarchy-name field shows only for selected levels.
- The selection must be a **contiguous run** of the discovered levels: you can trim the top (choose a different root) or the bottom (different leaf), but you can't skip a level in the middle — the hierarchy is a strict parent→child chain. The create button is disabled until the selection is valid (≥2 contiguous levels, all named), with an inline explanation when it isn't.

Parenting was already done against the immediate *included* parent level, so a contiguous subset produces a clean chain with no code change there.

## Validation

- `configurator` build green (sub-packages + vite); `tsc -b` clean.
- Deployed to bomet for live testing against the `ke.boundarytest` tenant + the self-hosted Overpass.

### DIFF
--- a/configurator/src/pages/Phase2Page.tsx
+++ b/configurator/src/pages/Phase2Page.tsx
@@ -46,6 +46,43 @@ type Step =
 
 type BoundaryPath = 'osm' | 'excel' | null;
 
+// The included levels, ascending by OSM admin_level.
+function getSelectedLevels(levels: OsmAdminLevel[]): OsmAdminLevel[] {
+  return [...levels].filter(l => l.selected).sort((a, b) => a.level - b.level);
+}
+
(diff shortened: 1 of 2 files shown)


KDwevedi:

PR #875, opened 2026-06-16
feat(deploy): self-hosted Overpass for configurator boundary fetch (opt-in)

KDwevedi:
## Why

The configurator's Phase 2 OSM step fetches admin-boundary polygons from Overpass. It was hardcoded to the **public** `overpass-api.de`, which is rate-limited and frequently returns `504 Gateway Timeout` under load (hit during live boundary testing). This adds an **opt-in self-hosted Overpass**, so a deployment can serve the boundary fetch entirely on-box.

This is the last piece of the OSM boundary onboarding work that wasn't yet codified — it had been validated live on bomet but existed only as on-box hot-patches.

## What

Mirrors the existing turbopass pattern (gated, default-off, byte-identical when off):

- **configurator** (`Phase2Page.tsx`): the Overpass endpoint is now `import.meta.env.VITE_OVERPASS_URL || 'https://overpass-api.de/api/interpreter'`. Zero-config still uses the public instance.
- **ansible**: `enable_overpass` host_var (default `false`) runs a standalone `bomet-overpass` container (`wiktorn/overpass-api`, loopback `127.0.0.1:12346`, `OVERPASS_RULES_LOAD=10` for the `area[]` index the query needs) against an operator-prepared boundary extract, renders a gated nginx `/overpass/` proxy, and sweeps the container when flipped off. The configurator build task bakes `VITE_OVERPASS_URL=/overpass/api/interpreter` when the flag is on.
- **overpass/**: `prepare-extract.sh` (Geofabrik pbf → `osmium tags-filter boundary=administrative` → merge → `.osm.bz2`) + README. Extracts are operator-prepared (country/size are deployment-specific) and intentionally not committed.

## Validation

- configurator builds green; default bundle → public Overpass; with `VITE_OVERPASS_URL` set → bakes `/overpass/api/interpreter` (confirmed in the built bundle).
- playbook + `_example.yml` YAML parse; nginx jinja `{% if %}/{% endif %}` balanced — vhost renders byte-identical when `enable_overpass` is off.
- Mechanism proven live on bomet: Mozambique + Kenya + India admin boundaries served via `/overpass`; 504s eliminated (Nairobi 157 / Delhi 1229 / Cidade de Maputo 72 relations).

## Cutover note (bomet)

To make bomet's current on-box Overpass durable: set `enable_overpass: true` + `overpass_planet_file` in its host_vars after this merges (the container + data are already on the box from testing).

### DIFF
--- a/configurator/src/pages/Phase2Page.tsx
+++ b/configurator/src/pages/Phase2Page.tsx
@@ -50,6 +50,13 @@ type BoundaryPath = 'osm' | 'excel' | null;
 // proxies it to the search-api container); override via VITE_TURBOPASS_URL.
 const TURBOPASS_BASE: string = import.meta.env.VITE_TURBOPASS_URL || '/turbopass';
 
+// Overpass endpoint for the boundary-polygon fetch. Defaults to the public
+// instance (zero-config, but rate-limited / 504-prone under load). A deploy
+// that self-hosts Overpass sets VITE_OVERPASS_URL (e.g. same-origin
+// '/overpass/api/interpreter', proxied by nginx to the on-box container behind
+// the enable_overpass gate) at configurator build time.
+const OVERPASS_URL: string = import.meta.env.VITE_OVERPASS_URL || 'https://overpass-api.de/api/interpreter';
+
 // The OSM path always writes the ADMIN hierarchy (the Excel path lets the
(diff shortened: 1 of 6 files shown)


KDwevedi:

PR #877, opened 2026-06-16
feat: mobile number validation driven by mobileNumberRegex (remove mobileNumberLength)

### LINKED ISSUE #596: [Bug] Jarring transition when messaging asks user to enter an acceptable phone number
**Describe the bug**
Jarring transition when messaging changes asking user to enter an acceptable phone number

**To Reproduce**
1. Go to https://naipepea.digit.org/digit-ui/citizen/login
2. Stay on sign in
3. Enter first digit. Enter 9th digit.
4. See transition. This issue only impacts the desktop experience and not the phone experience.

**Expected behavior**

https://github.com/user-attachments/assets/b61aa15a-1e51-4d69-a73e-7ea4f6bdb521

Keep container size static. Change messaging if needed. 

**Screenshots**
Attaching the video of the experience.

**Desktop (please complete the following information):**
 - OS: macOS M1 2020, Sequoia 15.7.3
 - Browser Chrome 
 - VVersion 148.0.7778.179 (Official Build) (arm64)

**Smartphone (please complete the following information):**

**Additional context**

### LINKED ISSUE #715: Mobile validation uses two divergent MDMS masters — FE/backend pattern mismatch
## Problem

Mobile number validation is sourced from **two different MDMS masters** that disagree on the accepted pattern. This can cause numbers that pass login to fail at notification dispatch time.

### Master 1: `common-masters.UserValidation`
- **Pattern:** `^[17][0-9]{8}$` (strict 9 digits, no leading 0)
- **Used by:** novu-bridge backend (`DispatchPipelineService.formatRecipientPhone()`), FE `useMobileValidation` hook → `window.__DIGIT_USER_VALIDATION.mobile`
- **File:** `ansible/nairobi-mdms/mdms/common-masters/UserValidation.json`

### Master 2: `ValidationConfigs.mobileNumberValidation`
- **Pattern:** `^0?[17][0-9]{8}$` (9–10 digits, optional leading 0)
- **Used by:** FE citizen login page, user profile page
- **File:** `utilities/default-data-handler/src/main/resources/mdmsData/ValidationConfigs/ValidationConfigs.mobileNumberValidation.json`

### Additionally: 4 total sources of truth

| # | Source | Pattern | Consumers |
|---|--------|---------|-----------|
| 1 | `common-masters.UserValidation` | `^[17][0-9]{8}$` | novu-bridge, `useMobileValidation` hook |
| 2 | `ValidationConfigs.mobileNumberValidation` | `^0?[17][0-9]{8}$` | Login page, User profile |
| 3 | `globalConfigs.CORE_MOBILE_CONFIGS` | deployment JS config | FE fallback |
| 4 | Library defaults (`mobileValidation.js`) | `^[6-9][0-9]{9}$` (India) | last-resort fallback |

### Concrete failure scenario

1. User logs in on citizen portal — login page reads Master 2, allows `0712345678` (with leading 0)
2. User files a complaint with that number
3. novu-bridge tries to send SMS — reads Master 1, `"0712345678".matches("^[17][0-9]{8}$")` → **fails**
4. Notification silently dropped, user never gets SMS updates

## Recommendation

Consolidate to **one master** — `common-masters.UserValidation` — since the backend already depends on it:

1. Make login page (`packages/modules/core/src/pages/citizen/Login/index.js`) and user profile (`packages/modules/core/src/pages/citizen/Home/UserProfile.js`) read from `common-masters.UserValidation` instead of `ValidationConfigs.mobileNumberValidation`
2. Deprecate/remove `ValidationConfigs.mobileNumberValidation` or make it a thin redirect
3. Align the pattern — decide whether leading `0` is valid (if yes, update `UserValidation.json`; if no, keep as-is)
4. `pgr-services` backend (`ServiceRequestValidator`) has **no** mobile validation at all — consider adding it there too for API-direct callers

### Files to touch

- `digit-ui-esbuild/packages/modules/core/src/pages/citizen/Login/index.js` (lines 71–92)
- `digit-ui-esbuild/packages/modules/core/src/pages/citizen/Home/UserProfile.js` (lines 257–278)
- `ansible/nairobi-mdms/mdms/common-masters/UserValidation.json` (if pattern needs updating)
- `utilities/default-data-handler/src/main/resources/mdmsData/ValidationConfigs/` (deprecate)

### Related

- #478 — postal code + boundary validation fix (confirmed fixed on FE, but exposed this master divergence during review)

### LINKED ISSUE #831: [Bug] UserValidation schema: switch x-unique key from fieldType → zone
MCP bootstrap was creating the UserValidation schema itself, causing a race condition when it ran before the data-handler. The schema must exist before any INSERT. Also x-unique: ['fieldType'] meant only one UserValidation record per field type (e.g. mobile) could exist. India's record blocked any country-specific record from being created, causing Mozambique/Kenya mobile validation to fail.

vinothrallapalli-eGov:
## Summary

- Replace all references to `mobileNumberLength`, `DEFAULT_MOBILE_MAX_LENGTH`, and `DEFAULT_MOBILE_MIN_LENGTH` with `computeMobileLengths(pattern)` — a utility that derives min/max digit counts purely from the MDMS `mobileNumberRegex`
- `common-masters.MobileNumberValidation` is now the single source of truth for mobile validation across configurator and digit-ui (pattern, prefix, lengths — all from the regex)
- Fixed React error #31 in configurator's `DigitFormInput.tsx` when ra-core validators return `{message, args}` i18n objects instead of plain strings
- Wired `useMobileValidator` (configurator) and regex-derived `maxLength` into all mobile input fields: create complaint, create employee, edit employee
- Removed separate length checks from form submit handlers and duplicate-check effects — `pattern.test()` is the sole gate

## Files changed

**Configurator:**
- `DigitFormInput.tsx` — fix React #31 (unwrap `@@react-admin@@` i18n objects); add `maxLength` prop
- `useMobileValidator.ts` — remove `minLength` from `MobileRules`; validator uses regex only
- `ComplaintCreate.tsx`, `EmployeeCreate.tsx`, `EmployeeEdit.tsx` — wire `useMobileValidator` + `maxLength`

**digit-ui-esbuild:**
- `libraries/mobileValidation.js` — add `computeMobileLengths()`, remove `DEFAULT_MOBILE_MAX_LENGTH`/`MIN` exports
- `core/Login/index.js`, `SelectMobileNumber.js` — maxLength from regex, no `mobileNumberLength`
- `core/Home/UserProfile.js` — remove `gc?.mobileNumberLength` fallbacks
- `hrms/createEmployee.js`, `EditForm.js` — regex-only validation in select + effects + submit
- `hrms/EmployeePhoneNumber.js`, `Inbox.js`, `useInboxGeneral/searchFields.js` — maxLength from regex
- `pgr/CreateComplaint/index.js`, `CreateComplaintConfig.js`, `useMobileValidation.js` — same

## Test plan

- [ ] Configurator create employee: mobile field capped at regex-derived length, invalid pattern rejected
- [ ] Configurator edit employee: same
- [ ] Configurator create complaint: citizen mobile field uses MDMS regex (Ethiopia `^[79][0-9]{8}$` → 9 digits)
- [ ] digit-ui login: mobile input capped at 9 digits, `+251` prefix shown
- [ ] digit-ui create complaint (employee): mobile field capped correctly
- [ ] digit-ui edit profile: mobile field uses MDMS regex, not hardcoded length


backend/novu-bridge/src/main/java/org/egov/novubridge/service/MdmsServiceClient.java
```diff
@@ -108,13 +97,13 @@ private Map<String, Object>  fetchDefaultRecord(String tenantId,
...
-                throw new CustomException("NB_MDMS_PHONE_PREFIX_UNAVAILABLE",
-                        "MDMS returned non-2xx or empty body while resolving phone prefix for tenantId=" + tenantId);
+                throw new CustomException("NB_MDMS_MOBILE_VALIDATION_UNAVAILABLE",
+                        "MDMS returned non-2xx or empty body while resolving mobile validation config for tenantId=" + tenantId);
             }
 
             List<Map<String, Object>> mdmsList = (List<Map<String, Object>>) response.getBody().get("mdms");
             if (mdmsList == null || mdmsList.isEmpty()) {
-                throw new CustomException("NB_MDMS_PHONE_PREFIX_UNAVAILABLE",
+                throw new CustomException("NB_MDMS_MOBILE_VALIDATION_UNAVAILABLE",
```
subhashini-egov: What does NB mean? Please review and remove if not necessary.

configurator/packages/data-provider/src/providers/resourceRegistry.ts
```diff
@@ -115,8 +115,8 @@ export const REGISTRY: Record<string, ResourceConfig> = {
   // These get the same generic CRUD as the entries above; richer per-field widgets
   // are layered on later via src/admin/schemaDescriptors/ (Stage 1+).
   'theme-config':           { type: 'mdms', label: 'Theme Config',             schema: 'common-masters.ThemeConfig',               idField: 'code',              nameField: 'name' },
-  'user-validation':        { type: 'mdms', label: 'User Field Validation',    schema: 'common-masters.UserValidation',            idField: 'zone',              nameField: 'zone' },
-  'mobile-validation':      { type: 'mdms', label: 'Mobile Number Validation', schema: 'ValidationConfigs.mobileNumberValidation', idField: 'validationName',    nameField: 'validationName' },
+  'user-validation':        { type: 'mdms', label: 'Mobile Number Validation', schema: 'common-masters.MobileNumberValidation',    idField: 'countryCode',       nameField: 'countryCode' },
```
subhashini-egov: Please see if you need two separate entries for validation or can we use a single one?

digit-ui-esbuild/packages/modules/pgr/src/pages/employee/CreateComplaint/index.js
```diff
@@ -4,12 +4,48 @@ import { useDispatch } from "react-redux";
...
+  DEFAULT_MOBILE_PREFIX,
+} from "@egovernments/digit-ui-libraries";
 
 import { FormComposer } from "../../../components/FormComposer";
 import { createComplaint } from "../../../redux/actions/index";
 
 export const CreateComplaint = ({ parentUrl }) => {
   const { data: cities, isLoading }  = Digit.Utils.getMultiRootTenant()? Digit.Hooks.useTenants() :Digit.Hooks.pgr.useTenants();
+
+  const stateLvlTenantId = window?.globalConfigs?.getConfig("STATE_LEVEL_TENANT_ID");
```
subhashini-egov, line 20: The current strategy is to look for masters within a tenant (self-contained) and if not found, fall back to a state tenant.

digit-ui-esbuild/products/pgr/src/hooks/pgr/useMobileValidation.js
```diff
@@ -86,43 +79,40 @@ const useMobileValidation = (tenantId, validationName = "defaultMobileValidation
...
+  const resolvedPattern =
+    mdmsConfig?.mobileNumberRegex ||
+    globalConfig?.mobileNumberPattern ||
+    defaultValidation.rules.pattern;
+
+  const { min: resolvedMin, max: resolvedMax } = computeMobileLengths(resolvedPattern);
+
   const validationRules = {
     allowedStartingDigits:
-      mdmsConfig?.rules?.allowedStartingCharacters ||
```
subhashini-egov, line 100: We should be reading the MDMS regex for the allowedStartingDigits?

digit-ui-esbuild/products/pgr/src/hooks/pgr/useMobileValidation.js
```diff
@@ -86,43 +79,40 @@ const useMobileValidation = (tenantId, validationName = "defaultMobileValidation
...
 
-    maxLength:
-      mdmsConfig?.rules?.maxLength ||
-      globalConfig?.mobileNumberLength ||
-      defaultValidation.rules.maxLength,
+    minLength: resolvedMin,
+    maxLength: resolvedMax > 0 ? resolvedMax : 15,
 
     errorMessage:
-      mdmsConfig?.rules?.errorMessage ||
```
subhashini-egov, line 125: Review this once. Do we not need a default error message when we read pattern from MDMS?

local-setup/nginx/globalConfigs.js
```diff
@@ -19,6 +19,11 @@ var globalConfigs = (function () {
   var pgrBoundaryLowestLevel = "Ward";
   var mapCenter = { lat: -1.0, lng: 37.0 };
   var useInboxV1 = true;
+  var coreMobileConfigs = {
```
subhashini-egov, line 28: We need a comment here. The strategy is to read from MDMS; this should ideally be the same as MDMS - provided as backup in case there's a miss in MDMS. Developers should know that changing this won't help if masters are not changed.

subhashini-egov (dismissed):
Left some comments. Please review and re-submit for verification

digit-ui-esbuild/products/pgr/src/hooks/pgr/useMobileValidation.js
vinothrallapalli-eGov: Verified the message is constructing based on regex, and remove unused code.

local-setup/nginx/globalConfigs.js
vinothrallapalli-eGov: Updated

digit-ui-esbuild/packages/modules/pgr/src/pages/employee/CreateComplaint/index.js
vinothrallapalli-eGov: Updated

backend/novu-bridge/src/main/java/org/egov/novubridge/service/MdmsServiceClient.java
vinothrallapalli-eGov: Refactored

pradeepkumarcm-egov (approved):

backend/novu-bridge/src/main/java/org/egov/novubridge/service/DispatchPipelineService.java
```diff
@@ -268,11 +268,11 @@ private String formatRecipientPhone(String mobile, String tenantId, String chann
...
             }
             MobileValidationConfig validationConfig = mdmsServiceClient.getMobileValidationConfig(tenantId, requestInfo);
-            if (!normalized.matches(validationConfig.getPattern())) {
+            if (!normalized.matches(validationConfig.getMobileNumberRegex())) {
                 throw new CustomException("INVALID_MOBILE_NUMBER",
-                        "Mobile number is not matching with default mobile pattern in MDMS");
+                        "Mobile number does not match the configured pattern for tenantId=" + tenantId);
             }
-            e164 = validationConfig.getPrefix() + normalized;
+            e164 = validationConfig.getCountryCode() + normalized;
```
KDwevedi, line 275: Bug: the default pattern now accepts a leading 0 (`^0?[17][0-9]{8}$`), so `0712345678` passes the check. Then we do `countryCode + number`, which gives `+2540712345678` — that extra 0 makes it an invalid phone number for SMS/WhatsApp. Fix: either drop the `0?` from the default pattern, or strip a leading 0 before adding the country code.

docker-compose.egov-digit.yaml
```diff
@@ -575,7 +575,7 @@ services:
     restart: "no"
 
   egov-user:
-    image: registry.preview.egov.theflywheel.in/egovio/egov-user:updatenovalidate-fix
+    image: registry.preview.egov.theflywheel.in/egovio/egov-user:master-e22c7c5
```
KDwevedi: The image tag here (`master-e22c7c5`) does not match the one in the upgrade doc (`v2.11-user-service-upgrade.md` says `master-d69ce29`). This PR deletes the old `UserValidation` master and expects egov-user to read the new `MobileNumberValidation` one. If the running image still reads the old master, it falls back to a 10-digit India rule and rejects 9-digit numbers when creating users. Please confirm the right tag and that it reads the new master.

digit-ui-esbuild/packages/libraries/src/constants/mobileValidation.js
```diff
@@ -1,51 +1,176 @@
...
- * See `products/pgr/src/hooks/pgr/useMobileValidation.js` for the
- * resolution code.
+ * See `products/pgr/src/hooks/pgr/useMobileValidation.js` for the hook that
+ * applies this priority chain.
  */
 
-/** Full-anchor regex (use with `new RegExp(pattern)`). */
-export const DEFAULT_MOBILE_PATTERN = "^[6-9][0-9]{9}$";
+/** Full-anchor regex — the single source of truth for all derived constraints. */
+export const DEFAULT_MOBILE_PATTERN = "^0?[17][0-9]{8}$";
```
KDwevedi: These defaults are Kenya-specific (`+254`, `^0?[17][0-9]{8}$`). They used to be India so existing tenants would not break. Any deployment that relies only on this fallback (no MDMS, no globalConfigs) now quietly validates as Kenya. Country-specific values should come from MDMS/globalConfigs, not be hardcoded here — please keep the default country-neutral.

digit-ui-esbuild/packages/libraries/src/constants/mobileValidation.js
```diff
@@ -1,51 +1,176 @@
...
+ * the regex is the single source of truth.
+ *
+ * Examples:
+ *   computeMobileLengths("^[79][0-9]{8}$")     → { min: 9,  max: 9  }
+ *   computeMobileLengths("^[6-9][0-9]{9}$")    → { min: 10, max: 10 }
+ *   computeMobileLengths("^0?[17][0-9]{8}$")   → { min: 9,  max: 10 }
  */
-export const DEFAULT_MOBILE_ERROR_MESSAGE =
-  "Please enter a valid 10-digit mobile number starting with 6-9";
+export function computeMobileLengths(pattern) {
```
KDwevedi: This regex parser is easy to break. For example `computeMobileLengths("^(254)[17][0-9]{8}$")` returns length 10, but the real number is 12 digits — it counts the whole `(254)` group as one character. That would cap the input box too short. The same parser is copy-pasted in three other files and the copies have already drifted apart. Please move it to one shared helper and add a few tests for tricky patterns (groups, `\d`, alternation, no anchors).

digit-ui-esbuild/packages/libraries/src/constants/mobileValidation.js
```diff
@@ -1,51 +1,176 @@
...
+/**
+ * Build a human-readable validation error message from a mobile regex pattern.
+ *
+ * Examples:
+ *   buildMobileErrorMessage("^[17][0-9]{8}$")   →
+ *     "Please enter a valid mobile number (9-10 digits, starting with 1 or 7)"
+ *   buildMobileErrorMessage("^[6-9][0-9]{9}$")  →
+ *     "Please enter a valid mobile number (10 digits, starting with 6-9)"
+ */
+export function buildMobileErrorMessage(pattern) {
```
KDwevedi: The old setup stored the error message as a translation key (`CORE_COMMON_MOBILE_ERROR`); this builds a fixed English sentence instead. On Swahili/Portuguese tenants it will always show English, and operators can no longer set their own message. Suggestion: keep an optional `errorMessage` key on the master and fall back to this generated text. (Related to @subhashini-egov at useMobileValidation.js:125.)

backend/novu-bridge/src/main/java/org/egov/novubridge/service/DispatchPipelineService.java
vinothrallapalli-eGov: The default pattern has been updated. If a country's telecom regulations allow mobile numbers with a leading `0`, and the number is later concatenated with the country code, should it still be considered a valid mobile number?

digit-ui-esbuild/packages/libraries/src/constants/mobileValidation.js
vinothrallapalli-eGov: Updated the default validation pattern to the Indian mobile number format, as the backend services now use the Indian mobile number pattern by default.

docker-compose.egov-digit.yaml
vinothrallapalli-eGov: Updated the latest tag

digit-ui-esbuild/packages/libraries/src/constants/mobileValidation.js
vinothrallapalli-eGov: I have localized the mobile number error message. The earlier code, `CORE_COMMON_MOBILE_ERROR`, was also a static error message. There is a possibility that the regex validation and the static error message could become out of sync, resulting in inconsistent behavior. To avoid this, the static message was removed from the master, and the error message is now generated dynamically based on the configured validation pattern.

Issue #874, opened 2026-06-16
digit-ui: replace static ke_nairobi_wards sidecar with boundary-service API for complaint maps


KDwevedi:
## Summary

The PGR complaint maps in `digit-ui-esbuild` still ship a 249 KB **static, Nairobi-specific** GeoJSON (`ke_nairobi_wards.json`) as a fallback ward layer. Now that boundaries are fully fed per-tenant (configurator OSM/Excel onboarding → boundary-service), we should drop the sidecar and rely solely on the boundary APIs.

## How digit-ui reads boundaries today

`products/pgr/src/hooks/pgr/useTenantBoundaries.js` is the single source for the map ward layer:
- When `MAP_TENANT` is set (`globalConfigs.getConfig("MAP_TENANT")`, now derived from `city_tenant`/`ui_state_tenant_id`), it does a **two-step boundary-service fetch**: `boundary-relationships/_search` (hierarchy tree) → chunked `boundary/_search?codes=…` (geometry), joined into a `FeatureCollection`. This is the "fully fed API" path and it works (validated live on bomet: 59 polygons rendered + pin→ward resolution).
- Otherwise — `MAP_TENANT` unset, fetch error, or all geometry placeholder/degenerate — it falls back to the bundled `keNairobiWardsFallback`.

Sidecar usages (3 files):
- `hooks/pgr/useTenantBoundaries.js:2,72,144,148` (import + 3 fallback branches)
- `components/ComplaintLocationMap.js:9,47` (import + synchronous null-case)
- `components/GeoLocations.js:10,181` (import + synchronous null-case)
- asset: `products/pgr/src/assets/boundaries/ke_nairobi_wards.json` (249 KB, IEBC/HDX 2019 Nairobi wards)

## Why replacing it is correct (and feasible)

The sidecar is not a real fallback — it's **Nairobi wards**, which are wrong for every non-Nairobi tenant (e.g. a Mozambique city would silently render Nairobi on the map). The API path already exists and is proven. So:
- For any tenant with fed boundaries → the API already drives the map; the sidecar never helps.
- For a tenant *without* fed boundaries → an empty/"boundaries not configured" map is **more honest** than showing another city's wards.

## Proposed change

1. Remove the `keNairobiWardsFallback` import + all fallback branches in the 3 files; delete the asset.
[...]

PR #870, opened 2026-06-16
fix(micro-ui): handle .jsx in webpack so the digit-ui image builds

KDwevedi:
## Problem

The nightly `digit-ui` (legacy `frontend/micro-ui`) image build fails on develop:

```
ERROR in ./src/App.js
Module not found: Error: Can't resolve './dashboard/AdminDashboard' in '/app/web/src'
error Command failed with exit code 1.   # yarn build:webpack
```

`web/src/dashboard/` (the admin dashboard feature — `AdminDashboard.jsx` + 9 component `.jsx` files) is authored in **`.jsx`**, but `web/webpack.config.js` never handled `.jsx`:

- the app babel rule matched only `/\.js$/`, so the JSX-syntax files were never run through `@babel/preset-react`;
- there was **no** `resolve.extensions`, so webpack's defaults (which don't include `.jsx`) couldn't resolve the extensionless `import AdminDashboard from "./dashboard/AdminDashboard"`.

So the build has been broken since the dashboard feature landed.

## Fix

`web/webpack.config.js`:
- app babel rule `test: /\.(js)$/` → `/\.(jsx?)$/` (transpile `.jsx` with preset-react)
- add `resolve.extensions: [".js", ".jsx", ".json"]` (resolve extensionless `.jsx` imports)

## Validation

Ran the exact nightly build on bomet:

```
docker build -f frontend/micro-ui/web/docker/Dockerfile frontend/micro-ui/
```

Before: failed at `yarn build:webpack` on the `AdminDashboard` import.
After: webpack compiles clean, image exports, `BUILD_EXIT=0`.

This clears the last `FAILED` entry in the nightly image set (alongside #863 for `digit-ui-esbuild`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #886, opened 2026-06-18
Boundary auto-population: level selection, self-hosted Overpass, and Management maps

KDwevedi:
## Boundary auto-population — features & fixes

Completes the one-click OSM boundary auto-population flow and adds boundary visibility in the Management view. Builds on the OSM one-click base already in `develop`.

### Phase 2 — boundary creation
- **Selectable, contiguous OSM admin levels**: operators tick which levels to create and name each. The selection must be a contiguous slice of the hierarchy (you can drop the outer levels, but not skip one in the middle), with inline validation and a gated submit.
- **Self-hosted Overpass + Turbopass (opt-in)**: the configurator uses `VITE_OVERPASS_URL || public`; a gated `enable_overpass` deploy runs the Overpass container + nginx `/overpass/` proxy and bakes `VITE_OVERPASS_URL` into the build. Adds `overpass/prepare-extract.sh` + README for preparing the boundary data. This removes the dependency on (rate-limited / timeout-prone) public `overpass-api.de` and lets place-name autocomplete (`/turbopass`) work same-origin.

### Management view — boundary maps (new)
- **Maps in the boundary Management view**: a Leaflet map on the boundary detail (`BoundaryShow`) and a "Show map" overview on `BoundaryList` (`BoundaryOverviewMap`) that highlights a tenant's boundaries on an OSM basemap.
  - Entity-based (a single `/boundary/_search`, no codes), so it renders real geometry even when the relationship tree is incomplete.
  - Polygons are drawn largest-first so a containing boundary (e.g. a county) doesn't cover the smaller ones inside it — hover resolves to the specific boundary.

### Boundary list — correctness fixes
- **Dedup the relationship-tree flatten**: the relationships endpoint repeats children under each parent, so a small tree was rendering as hundreds of duplicate rows.
- **Scope the relationship search by query params**: `boundary-relationships/_search` binds its criteria from query params only — passing `tenantId` in the body ran the search unscoped and leaked other tenants' boundaries into a tenant's list.

### PGR complaint map
- **Remove the hardcoded ward sidecar** (#874): the complaint map fell back to a bundled `ke_nairobi_wards.json` whenever `MAP_TENANT` was unset or the tenant had no geometry, drawing one fixed region's wards on every deployment regardless of the actual boundary data (and mis-resolving pins). Boundaries now come solely from boundary-service for the configured `MAP_TENANT`; when there's no data, there's simply no overlay. `useTenantBoundaries` / `ComplaintLocationMap` / `GeoLocations` are null-safe, and the LocalityPicker dropdown remains the required field.

### Deploy
- micro-ui webpack `jsx` resolve/transpile fix so the digit-ui image builds.
- Raise the keycloak healthcheck `start_period` (60s → 180s) so a slow realm import doesn't flip the container `unhealthy` mid-startup and abort the converge.

### Notes
- Supersedes #873, #875, #870, #869 (folded in here) and addresses ticket #874.
- To show a complaint-map overlay, set `MAP_TENANT` to a tenant that has real boundary geometry (e.g. one populated via the Phase 2 OSM flow).


configurator/src/api/services/boundary.ts
```diff
@@ -198,6 +198,57 @@ export const boundaryService = {
...
+  },
+
+  // Fetch ALL boundary entities for a tenant in one shot. /boundary/_search
+  // with only tenantId (no codes) returns every entity (capped ~300 by the
+  // service) with its geometry — the same call the MCP's boundary_entity_search
+  // makes. Used by the Management overview map so it renders whatever has real
+  // geometry, independent of the (sometimes inconsistent) relationship tree.
+  async getAllBoundaryEntities(
+    tenantId: string,
+    limit = 300,
```
Lokendra-egov: @KDwevedi  A tenant with >300 boundaries will get a silently incomplete map with no indication.

Lokendra-egov (approved):

Lokendra-egov (approved):

PR #899, opened 2026-06-19
feat(configurator): Complaint Types two-level accordion + search (increment 1)

singhalkarun:
## Summary

Reworks the configurator **Complaint Types** screen from a flat list into a two-level **Type → Sub-Type** management UI, and fixes how type display names are sourced/edited.

## What's included
- **Two-level accordion** — Complaint **Types** are derived by grouping the flat `RAINMAKER-PGR.ServiceDefs` records by `menuPath` (case-insensitive); expand a Type to see its **Sub-Types** (name, service code, department, SLA, status). Records with no `menuPath` fall into an **Uncategorized** group (last). Types ordered by min `order`; sub-types by `order` then `serviceCode`.
- **Search** — matches a Type label OR any sub-type name/serviceName/serviceCode; matching Types auto-expand and show only matching sub-types, with a distinct "no matches" state.
- **Sub-type CRUD** — Add (from inside a Type, menuPath pre-filled + locked), Edit (existing form), Delete (soft-delete; warns when deleting the last sub-type, which removes the whole Type).
- **Type CRUD** — Add Complaint Type; Edit = rename the display name (localization).
- **Type label sourcing** — labels resolved from the tenant's `rainmaker-pgr` `SERVICEDEFS.*` localization; unnamed types show their raw menuPath code in monospace; renaming promotes them to a real label.
- **Long codes** — service codes / unnamed-type codes are shortened with an ellipsis, full value in a tooltip + one-click copy.

## Decisions
- **Type rename updates only the ACTIVE locale.** Localization is stored per `(tenant, module, locale, code)`, so writing every locale on rename would overwrite other languages' existing translations with the single typed string. We write only the current locale; other locales are translated separately (Localization screen / bulk import).
- **"Complaint Type" is not a stored entity** — it's the `menuPath` shared across sub-types. So "create type" is a combined Type + first-Sub-Type form, and deleting a type's last sub-type removes the type. No standalone delete-type or menuPath-code rename.
- **Type labels can't use react-admin `translate`** in the configurator — its i18n only loads the `configurator-ui` module, not `rainmaker-pgr`. Labels are fetched separately (cached, refetched after rename).
- **Unnamed types show the raw menuPath code (monospace), not a humanized guess** — a generated pretty name is indistinguishable from a real one; a code reads as "needs naming" and renaming promotes it.
- **Long codes are shortened in JS, not via CSS** — CSS `truncate` doesn't work in auto-layout table cells, so the cell never shortened and the tooltip duplicated the visible text.
- **Frontend-only** — no backend, data-provider, registry, or routing changes.

## Test Plan
- [x] Unit tests (Vitest) across the complaint-types resource + `CopyableCode`; full suite green
- [x] `eslint` + `tsc` clean on changed files
- [ ] Manual against a live tenant: grouping/expand, search, add/edit/delete sub-type, create type, rename type (verify only the active locale changes), long-code tooltip + copy

## Notes
- Draft: pending manual/visual verification against a live tenant.
- Unrelated `leaflet` dependency was missing from local `node_modules` (declared upstream via boundary-maps); resolved with `npm ci`. Not part of this diff.


singhalkarun:
Pushed a batch of sub-type CRUD fixes (verified against a live tenant):

- **Edit re-seeds localization** — editing a sub-type now upserts its `SERVICEDEFS.*` name labels (active locale only, parent type label untouched) so the new name shows in the citizen/employee PGR UI, not just MDMS. Added an `afterUpdate` hook to `DigitEdit`.
- **Delete no longer navigates to Show** — the Delete trigger (wrapped in Radix `AlertDialogTrigger`) was leaking the click to the row's navigate handler; now stopped at the actions-cell boundary so the confirm dialog opens and you stay on the list.
- **Delete/rename toasts now appear** — switched from ra-core `useNotify` (no `<Notification>` is mounted) to the shadcn `toast` used elsewhere, with the item name as the description.

Full unit suite green (75 tests).

KDwevedi (commented):
Reviewed the full source diff plus the surrounding code it touches. Strong work overall — the flat datagrid → Type/Sub-Type accordion rework is clean, the pure logic (`groupComplaintTypes`, `filterComplaintTypeGroups`, `menuPathCode`, `buildServiceDefLabelMap`) is well extracted and unit-tested, the `SERVICEDEFS.<MENUPATH_UPPER>` casing is consistent between the seeder (`localization.ts:191`) and the group lookup, and the edit/rename flows handle the localization-clobber problem thoughtfully (active-locale-only, omit `menuPath`).

A few findings below; #1 is the one I'd block on.

### 1. (High) "Add Sub-Type" silently overwrites the parent Type's display name — in every language

The new "Add Sub-Type" action reuses `ComplaintTypeCreate` with the parent's `menuPath` pre-filled and locked. But `afterCreate` (unchanged by this PR, `ComplaintTypeCreate.tsx:30-53`) still passes `menuPath: data.menuPath` into `uploadComplaintTypeLocalizations`, looping over **all** locales. That reaches `buildComplaintTypeLocalizations`, which at `localization.ts:191` does:

```js
push(`SERVICEDEFS.${opts.menuPath.toUpperCase()}`, opts.menuName || opts.menuPath);
```

No `menuName` is supplied, so it upserts the parent label = the **raw menuPath code**, overwriting the operator's existing display name ("Sanitation & Waste" → "SANITATION") across every configured language.

This is exactly the clobber the PR carefully guards against in `ComplaintTypeEdit.afterUpdate` ("deliberately omit `menuPath` ... active locale only") and in the rename flow — the create-via-Add-Sub-Type path reintroduces it. Suggested fix: when `presetMenuPath` is set, omit `menuPath` from the `afterCreate` seed (or pass the existing label as `menuName`), mirroring the edit-flow guard.

### 2. (Medium) "Delete last sub-type → removes the whole type" warning is wrong during an active search

`SubTypeTable` computes `isLastSubType = subTypes.length === 1`, but under search `filterComplaintTypeGroups` replaces `subTypes` with only the matching rows. A type with 3 sub-types showing 1 match will warn that deleting it removes the entire complaint type — false; the two filtered-out sub-types remain. The "last sub-type" determination should be based on the unfiltered group count, not the rendered slice.

### 3. (Low) "Add Complaint Type" conflates display-name and identifier

The create form labels the `menuPath` field "Complaint Type Name", so an operator types e.g. "Water & Sanitation" and that string becomes both the grouping key and the localization code (`SERVICEDEFS.WATER & SANITATION` — spaces/ampersand in a key). Per the increment-4 plan this is intentional, but it's fragile and the later rename only updates the label, never the code. Worth a follow-up.

**Nits:** `groupComplaintTypes`/`filterComplaintTypeGroups` run unmemoized per render (cheap, fine); the uncategorized bucket's expand-state is keyed on `''` (works for a single bucket); clipboard fail-silent in `CopyableCode` is reasonable.

singhalkarun:
Thanks @KDwevedi — great catches. Addressed #1 and #2 in c92ed5be; #3 noted as a follow-up.

**#1 (High) — Add-Sub-Type clobbering the parent type label:** fixed. `ComplaintTypeCreate.afterCreate` now omits `menuPath` from the localization seed when `presetMenuPath` is set (i.e. adding a sub-type to an existing type), so it no longer upserts `SERVICEDEFS.<menuPath>` and can't overwrite the parent's display label. A brand-new type (no preset) still seeds its own parent label. This mirrors the edit/rename guards you referenced. Tests: asserts the seed entry has no `menuPath` in the Add-Sub-Type path, and does include it when creating a new type.

**#2 (Medium) — last-sub-type warning during search:** fixed. `ComplaintTypeList` now passes the unfiltered group total (`totalSubTypes`) to `SubTypeTable`, and the "removes the entire complaint type" warning is computed from that instead of the rendered (filtered) slice. Test: a type with 3 sub-types showing 1 filtered row no longer warns.

**#3 (Low) — name/identifier conflation in Add Complaint Type:** agreed it's fragile (typed name becomes both grouping key and localization code, and rename only fixes the label). Leaving as-is for this PR per the increment-4 plan; I'll track it as a follow-up (e.g. derive a slug code separate from the display name).

Full unit suite green (78 tests). Nits acknowledged — no changes (all reasonable as-is).

configurator/src/resources/complaint-types/filterComplaintTypeGroups.ts
```diff
@@ -0,0 +1,40 @@
...
+  for (const g of groups) {
+    if (g.label.toLowerCase().includes(q)) {
+      result.push(g);
+      continue;
+    }
+    const matchingSubs = g.subTypes.filter(
+      (s) =>
+        (s.name ?? '').toLowerCase().includes(q) ||
+        (s.serviceName ?? '').toLowerCase().includes(q) ||
+        s.serviceCode.toLowerCase().includes(q),
```
vinothrallapalli-eGov, line 28: **[BUG — Crash] `s.serviceCode.toLowerCase()` throws when `serviceCode` is null/undefined**

Any MDMS record missing `serviceCode` will throw `TypeError: Cannot read properties of undefined (reading 'toLowerCase')` the moment the operator types into the search box. The error propagates to the list render, showing a blank screen until page refresh.

Same unguarded call at `groupComplaintTypes.ts:66`.

```ts
// fix
s.serviceCode?.toLowerCase().includes(q)
```

configurator/src/resources/complaint-types/groupComplaintTypes.ts
```diff
@@ -0,0 +1,88 @@
...
+    const label = isUncategorized
+      ? translate('app.complaint_types.uncategorized', { _: 'Uncategorized' })
+      : translate(`SERVICEDEFS.${key}`, { _: original });
+    groups.push({
+      menuPath: key,
+      label,
+      count: subs.length,
+      activeCount: subs.filter((s) => s.active === true).length,
+      isUncategorized,
+      minOrder: Math.min(...subs.map(orderOf)),
```
vinothrallapalli-eGov, line 77: **[BUG — Crash] `Math.min(...subs.map(orderOf))` risks RangeError on large datasets**

Spreading all sub-type order values as variadic arguments hits V8's ~65k argument limit on tenants with a large bulk MDMS import sharing one `menuPath`. Throws `RangeError: Maximum call stack size exceeded`, crashing the entire list render.

Same spread pattern at line 64.

```ts
// fix — use reduce instead of spread
const minOrder = subs.reduce((m, s) => Math.min(m, orderOf(s)), Infinity);
```

configurator/src/resources/complaint-types/ComplaintTypeList.tsx
```diff
@@ -1,35 +1,308 @@
...
+  // yet, fall back to its raw menuPath code (last segment), which the UI shows
+  // in monospace so it reads as an unnamed code rather than a real name.
+  const labelTranslate = (key: string, opts?: { _?: string }) =>
+    key.startsWith('SERVICEDEFS.')
+      ? serviceDefLabels[key] ?? menuPathCode(opts?._ ?? '')
+      : translate(key, opts);
+
+  const handleRenameType = async (menuPath: string, newName: string) => {
+    const tenantId = digitClient.stateTenantId;
+    if (!tenantId) return;
```
vinothrallapalli-eGov, line 52: **[BUG — Silent failure] Rename silently no-ops when `tenantId` is falsy**

`handleRenameType` early-returns without throwing when `digitClient.stateTenantId` is unset (e.g. timing issue on initial load). The dialog closes normally (`setOpen(false)` fires), no toast is shown, but nothing was written to localization. The operator believes the rename succeeded.

Fix: throw (or surface an error toast) instead of silently returning:
```ts
if (!tenantId) {
  notify('Tenant ID not available — please reload and try again.', { type: 'error' });
  return;
}
```

configurator/src/resources/complaint-types/ComplaintTypeList.tsx
```diff
@@ -1,35 +1,308 @@
...
 export function ComplaintTypeList() {
+  const translate = useTranslate();
+  const navigate = useNavigate();
+  const [expanded, setExpanded] = useState<Set<string>>(new Set());
+  const [query, setQuery] = useState('');
+
+  const { data, isPending, isFetching, error, refetch } = useGetList(
+    'complaint-types',
+    {
+      pagination: { page: 1, perPage: 1000 },
```
vinothrallapalli-eGov, line 31: **[BUG — Silent truncation] `perPage: 1000` silently hides records beyond the cap**

Deployments with >1000 PGR sub-types will have records entirely absent from the view with no error, no "showing X of Y" indicator, and no way for operators to discover or manage the hidden records.

Consider fetching all pages (loop until `data.length < perPage`) or at minimum surfacing a warning when `data.length === 1000`.

configurator/src/resources/complaint-types/ComplaintTypeList.tsx
```diff
@@ -1,35 +1,308 @@
...
+                        totalSubTypes={subTypeTotals.get(g.menuPath) ?? g.subTypes.length}
+                        onDelete={handleDeleteSubType}
+                      />
+                      <button
+                        type="button"
+                        onClick={(e) => {
+                          e.stopPropagation();
+                          const raw = g.isUncategorized
+                            ? ''
+                            : g.subTypes[0]?.menuPath ?? '';
```
vinothrallapalli-eGov, line 286: **[BUG — Data corruption] Add Sub-Type passes raw-cased `menuPath` instead of canonical group key**

`g.subTypes[0]?.menuPath` carries whatever casing the first MDMS record happens to have (e.g. `"Sanitation"`) after sort, while the group key is normalized to uppercase (`"SANITATION"`). The new sub-type is saved to MDMS with the raw casing, creating a casing inconsistency among siblings.

Fix: use the group key directly:
```ts
// line 286
params.set('menuPath', g.key); // canonical upper-cased key, not subTypes[0].menuPath
```

Same issue at line 319.

configurator/src/resources/complaint-types/ComplaintTypeEdit.tsx
```diff
@@ -1,10 +1,44 @@
...
+  // translations (same policy as type rename).
+  const afterUpdate = async (record: RaRecord) => {
+    const data = record as unknown as {
+      serviceCode?: string;
+      name?: string;
+      department?: string;
+    };
+    const serviceCode = data.serviceCode?.trim();
+    const name = data.name?.trim();
+    if (!serviceCode || !name) return;
```
vinothrallapalli-eGov, line 27: **[BUG — Silent data loss] Localization re-seed silently skipped for legacy records missing `serviceCode`/`name`**

When an MDMS ServiceDef record has no `serviceCode` or `name` (e.g. partially-migrated legacy data), the early return fires after the MDMS write succeeds. The operator sees the "updated" toast and is redirected normally, but `SERVICEDEFS.*` localization keys are never updated — the old display name persists in all UIs with no indication.

Consider logging a warning or showing a secondary notification when the re-seed is skipped, rather than silently bailing.

configurator/src/resources/complaint-types/ComplaintTypeCreate.tsx
```diff
@@ -44,10 +57,19 @@ export function ComplaintTypeCreate() {
     // Dedupe via Set in case StateInfo declares `en_IN` explicitly.
     const targetLocales = new Set<string>([...locales.map((l) => l.value), 'en_IN']);
 
+    // When adding a sub-type to an EXISTING type (presetMenuPath), omit
+    // `menuPath` from the seed: buildComplaintTypeLocalizations would otherwise
+    // upsert SERVICEDEFS.<menuPath> = the raw code, overwriting the parent
+    // type's display label in every locale. Only a brand-new type (no preset)
+    // seeds its own parent label. Mirrors the edit/rename clobber guards.
+    const seedEntry = presetMenuPath
```
vinothrallapalli-eGov, line 65: **[BUG — Silent data loss] Add Sub-Type flow never seeds parent's display label**

When `presetMenuPath` is set, `afterCreate` omits `menuPath` from `seedEntry`, skipping the `SERVICEDEFS.<MENUPATH>` localization key. If that parent type has no existing localization row yet, it will render as a raw code in both the configurator accordion and the citizen/employee PGR menu indefinitely — the only workaround is using the Rename button separately.

Fix: seed the parent label unconditionally (or only when the key doesn't already exist):
```ts
const seedEntry = {
  serviceCode: data.serviceCode,
  name: data.name,
  menuPath: presetMenuPath ?? data.menuPath, // always include
};
```

vinothrallapalli-eGov (commented):
Automated review — 8 finder angles, 41 candidates, 12 refuted, **10 confirmed**. Findings ranked most-severe first.

**Crash bugs (2):** null `serviceCode` crash on search; `Math.min(...spread)` RangeError on large datasets.
**Silent data loss (4):** rename silently no-ops when tenantId unset; Add Sub-Type never seeds parent label; Edit skips localization re-seed for legacy records; Add Sub-Type writes wrong-cased menuPath.
**UX (2):** dialog description overstates scope (multi-locale); >1000 records silently hidden.
**Performance/UX (2):** redirect deferred behind network calls (hung-page appearance); grouping/filter recomputed on every render with no useMemo.

configurator/src/resources/complaint-types/RenameTypeDialog.tsx
```diff
@@ -0,0 +1,84 @@
...
+  };
+
+  return (
+    <Dialog open={open} onOpenChange={handleOpenChange}>
+      <DialogTrigger asChild>{trigger}</DialogTrigger>
+      <DialogContent>
+        <DialogHeader>
+          <DialogTitle>Rename Complaint Type</DialogTitle>
+          <DialogDescription>
+            Updates the display name across all configured languages. The internal
```
vinothrallapalli-eGov, line 55: **[UX — Misleading] Dialog description overstates rename scope**

The text says *"Updates the display name across all configured languages"* but `handleRenameType` only writes the currently active locale. Admins in multi-language deployments (e.g. `en_IN` + `sw` + `fr`) will assume all locales were updated when only one was written.

Either update the description to reflect the actual scope, or extend `handleRenameType` to write to all configured locales (matching what `uploadComplaintTypeLocalizations` does in Create/Edit).

configurator/src/admin/DigitEdit.tsx
```diff
@@ -175,13 +180,27 @@ export function DigitEdit({ title, children, resource, id, redirect = 'list', tr
...
+              await afterUpdate(data as RaRecord);
+            } catch (e) {
+              toast({
+                title: 'Post-update step failed',
+                description: e instanceof Error ? e.message : String(e),
+                variant: 'destructive',
+              });
+            }
+          }
           if (redirect && effectiveResource) {
```
vinothrallapalli-eGov, line 204: **[UX — Hung-page appearance] Redirect deferred until after localization network calls**

`afterUpdate` in `ComplaintTypeEdit` makes two sequential network calls (`uploadComplaintTypeLocalizations` + `cacheBust`). Because the redirect fires inside `onSuccess` after `await afterUpdate(...)`, a slow or rate-limited localization API leaves the operator staring at the edit form for 1–3 s after the "updated" toast fires, with no loading indicator — appearing as a hung page.

Consider firing the redirect immediately and running the localization/cache-bust calls in the background, or show a loading state on the Save button until the redirect fires.

configurator/src/resources/complaint-types/ComplaintTypeList.tsx
```diff
@@ -1,35 +1,308 @@
...
+    // propagates to DeleteConfirmDialog, which shows the message in-dialog.
+    await dataProvider.delete('complaint-types', {
+      id: record.id,
+      previousData: record as unknown as RaRecord,
+    });
+    toast({ title: 'Sub-type deleted', description: record.name ?? record.serviceName });
+    await refetch();
+  };
+
+  const allGroups = groupComplaintTypes(
```
vinothrallapalli-eGov, line 82: **[Performance] `groupComplaintTypes`, `filterComplaintTypeGroups`, `subTypeTotals` recomputed on every render**

All three are computed inline without `useMemo`. Every keystroke in the search input and every accordion toggle triggers a full O(N·log N) re-grouping + sort + filter pass. With 200+ ServiceDef records this causes visible UI stutter on each keypress.

```ts
const allGroups = useMemo(
  () => groupComplaintTypes(data ?? [], labelTranslate),
  [data, serviceDefLabels] // labelTranslate recreated each render — dep on serviceDefLabels instead
);
const groups = useMemo(
  () => filterComplaintTypeGroups(allGroups, query),
  [allGroups, query]
);
const subTypeTotals = useMemo(
  () => Object.fromEntries(allGroups.map(g => [g.key, g.subTypes.length])),
  [allGroups]
);
```

KDwevedi:
Closing as superseded. This PR reworks Complaint Types into a **two-level** Type → Sub-Type UI, but the underlying model moved to an **N-level complaint hierarchy** in #917 (merged to `develop` 2026-06-24, after this branch's last commit on 06-22). The two-level premise no longer holds, so this increment is stale. If a configurator management UI is still wanted for the N-level structure, it should be rebuilt against the new hierarchy (see the `ComplaintHierarchyComponent` work) rather than rebased. Reopen if I've misjudged.

Issue #490, opened 2026-04-29
Nai Pepea: CMS: Application timeline does not display correct information for assignee/creator


Gurjeet-eGov:
Application timeline does not display correct information for assignee/creator.

**EMPLOYEE**

- When complaint is **created** the timeline displays name of person who registered complaint 
- When **GRO assigns to LME** the timeline displays the name of **Assignee**, which is correct
- When **LME re-assigns to LME** the timeline displays the name of **Person who re-assigned (LME)** instead of GRO name, **this one needs to be updated**
- When **no assignee is selected** the timeline displays the name as **undefined**, currently product has same, please update as required by nairobi. 

**CITIZEN**

- **Pending at LME** displayed twice in citizen timeline and in same line assignee name also displayed. either remove it from same line or display it below comments

<img width="1365" height="650" alt="Image" src="https://github.com/user-attachments/assets/158d404c-0041-486c-a28c-6c5406b904f0" />

<img width="1365" height="650" alt="Image" src="https://github.com/user-attachments/assets/a0ec5561-11fb-4df6-b254-74d3338f6ab7" />

KDwevedi:
## Fix shipped — [theflywheel/digit-ui-esbuild#116](https://github.com/theflywheel/digit-ui-esbuild/pull/116)

Verified live first via the workflow API on `NCCG-PGR-2026-05-08-025980` (E2E escalation test complaint):

| Action | `assigner.name` (actor — current render) | `assignes[0].name` (new render) |
|---|---|---|
| APPLY | E2E Escalation Citizen (citizen who filed) | — |
| ASSIGN | Super Admin | **Hibrahim Otieno** |
| ESCALATE (level 0) | Super Admin | **Mr. Kinoti** |
| ESCALATE (level 1) | Super Admin | **Sarah Gichamba** |

So today the timeline reads "Super Admin" on all three assign/escalate rows; after this lands those rows correctly read the LME the work moved to.

### What the PR fixes against your 4 sub-bugs

1. **Created → registrant name.** Already correct; new rule preserves it (APPLY falls through to `assigner.name`).
2. **GRO assigns to LME → assignee.** Now actually pulls from `assignes[0]` on ASSIGN rows.
3. **LME reassigns to LME → new assignee.** Same rule (REASSIGN routes to `assignes[0]`).
4. **No assignee → "undefined".** Replaced with `t("CS_NA")` (verified seeded in en_IN + sw_KE; resolves to "NA").
[...]

PR #910, opened 2026-06-23
fix(pgr): show all role-filtered employees in assignee dropdown when …

Lokendra-egov:
…department is undefined (#490)


Hari-egov (approved):
hey @Lokendra-egov  and @vinothrallapalli-eGov ,
if department  check is removed from pgr-service at ASSIGN  workflow this will work else the flow will stuck at Invalid Department

service-code and Department validate is happening at This ASSIGN Flow

Issue #547, opened 2026-05-06
[Enhancement] SaaS: Comms Infra Setup


ChakshuGautam:
Communications infra - WhatsApp, E-mail, SMS configuration should be exposed via the onboarding flow with the option of allowing someone to skip.

pradeepkumarcm-egov:
@ChakshuGautam  The Novu configuration documents are as follows,

https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/master/backend/docs/COMPREHENSIVE_WHATSAPP_SETUP_GUIDE.md

https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/master/backend/novu-bridge/README.md

KDwevedi:
Review configuration scripts for whatsapp configuration

KDwevedi:
# Novu in the DIGIT/CCRS notification stack — full reference

The notification stack is **three custom DIGIT services** wrapped around **self-hosted Novu**. The split is **decide → render → deliver**:

| Layer | Component | Lang | Role |
|---|---|---|---|
| Decide | `novu-bridge` | Java 17 | DIGIT-side orchestrator: consume events, gate on consent, resolve tenant config, trigger Novu, log/retry |
| Render | `novu-bridge-endpoint` | Node | Novu Framework "Bridge" — Novu calls back into it to render the message body |
| Deliver | **Novu** (self-hosted) | — | Subscriber mgmt + dispatch to Twilio (→ WhatsApp/SMS) |
| Consent | `digit-user-preferences-service` | Go | Per-user channel consent + language preference (system-of-record for opt-in) |

---

## 1. What Novu exposes

Self-hosted Novu is deployed as a backbone stack (helm chart, images `ghcr.io/novuhq/novu/* :2.3.0`):

| Sub-service | Port | Purpose |
|---|---|---|
| `novu-api` | 3000 | The API DIGIT calls — `POST /v1/events/trigger`, integrations, environments, subscribers |
| `novu-worker` | — | Processes triggered jobs, calls the Bridge to render, dispatches to provider |
| `novu-ws` | 3002 | WebSocket (in-app feed; unused for our WhatsApp/SMS path) |
| `novu-dashboard` | 4000 | Admin UI, ingress at `host/novu` |
| `novu-mongodb` | 27017 | Novu's datastore (db `novu-db`) |
| `novu-redis` | 6379 | Queues/cache (DB index 2) |

[...]

PR #918, opened 2026-06-23
feat(novu-bridge): per-tenant channel enable/disable, multi-channel dispatch + retry (#547)

singhalkarun:
Backend for **SaaS Comms Infra Setup** (#547). Makes the per-tenant `NotificationChannel` enable/disable real in `novu-bridge`, end to end. Pairs with the configurator UI in **#919** (which is inert until this ships).

## Problem
`NotificationChannel` config existed in `digit-config-service` but **nothing read it** — the bridge dispatched on a single deploy-time channel (`NOVU_BRIDGE_CHANNEL`) regardless. There was no per-tenant way to turn channels on/off.

## What this does
- **Per-tenant gate.** Dispatch resolves a tenant's enabled `NotificationChannel` codes from config-service and only sends on those.
- **Legacy fallback — no backfill needed.** Three distinct states:
  - *no records* → fall back to the deploy allow-list (`NOVU_BRIDGE_CHANNEL`) = pre-change behaviour, so **existing tenants keep working untouched**;
  - *records, all disabled* → dispatch nothing;
  - *config-service hard error* → throw → retry/DLQ (never a silent wrong-send or drop).
- **Multi-channel dispatch.** Fans out over `enabled ∩ allow-list`; a per-channel failure is isolated (recorded `FAILED`) so it never blocks sibling channels or re-sends ones that already succeeded.
- **Bounded retry + backoff before DLQ.** Transient failures re-queue on a **separate** retry `@KafkaListener` (`max.poll.records=1`, fixed backoff) up to `maxRetries`, then DLQ — so a brief config-service blip self-recovers without a manual replay, and the backoff never head-of-line-blocks live notifications.
- **Schema + default-off for new tenants.** `default-data-handler` registers the `NotificationChannel` schema per tenant and seeds a **disabled** default; `notif-mdms-seed` registers the schema for the local path (records not seeded → legacy fallback).

## Notable design decisions
- `null` (unconfigured) vs empty (all-disabled) vs throw (error) are kept strictly distinct so an outage can't be mistaken for "off".
- Diagnostic endpoints (`_validate`/`_dry-run`) are side-effect-free and don't 5xx on a config-service blip.
- Channel codes normalised lowercase end to end (schema enum is uppercase; provider/template channels are lowercase).

## Tests & build
- novu-bridge suite **23 green**: unit (gate states, multi-channel, per-channel isolation incl. unexpected-error), a **real-HTTP contract test** against config-service's `_resolve`/`_search`, and a Kafka retry **round-trip** test.
- `mvn package` produces deployable jars for `novu-bridge` and `default-data-handler` (NotificationChannel config + schema bundled).
- An independent adversarial review was applied — see PR comments (diagnostic side-effects, per-channel isolation, retry-listener safety, seed comment).

## Deploy / merge notes
- **#919 is inert until this is deployed** — the configurator toggles only take effect once the bridge honours them.
- The `NotificationChannel` schema registration ships here (via `default-data-handler`).
- Still environment-gated (post-merge): a live end-to-end (UI → config-service → event → `nb_dispatch_log`) and confirming the configurator user has config-service write permission.

### DIFF
--- a/backend/novu-bridge/src/main/java/org/egov/novubridge/config/NovuBridgeConfiguration.java
+++ b/backend/novu-bridge/src/main/java/org/egov/novubridge/config/NovuBridgeConfiguration.java
@@ -63,6 +63,12 @@ public class NovuBridgeConfiguration {
     @Value("${novu.bridge.max.retries:3}")
     private Integer maxRetries;
 
+    // Fixed backoff applied before reprocessing a message consumed from the retry topic, so a brief
+    // downstream outage (e.g. config-service) gets spacing between attempts instead of burning all
+    // retries instantly. Bounded by maxRetries. 0 disables the wait (used in tests).
+    @Value("${novu.bridge.retry.delay.ms:5000}")
+    private Integer retryDelayMs;
+
     @Value("${novu.bridge.preference.enabled:true}")
     private Boolean preferenceEnabled;
 
--- a/backend/novu-bridge/src/main/java/org/egov/novubridge/config/NovuBridgeConfiguration.java
+++ b/backend/novu-bridge/src/main/java/org/egov/novubridge/config/NovuBridgeConfiguration.java
@@ -7,7 +7,11 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Import;
 import org.springframework.stereotype.Component;
 
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 import java.util.TimeZone;
+import java.util.stream.Collectors;
 
 @Component
 @Data
@@ -26,12 +30,32 @@ public class NovuBridgeConfiguration {
     @Value("${novu.bridge.kafka.dlq.topic:novu-bridge.dlq}")
     private String dlqTopic;
 
-    // Default channel for dispatch. SMS is the safer default — works
-    // with any Twilio SMS-capable sender out of the box. WhatsApp
-    // requires a pre-approved Twilio Programmable WhatsApp sender
-    // (sandbox or production). Override via NOVU_BRIDGE_CHANNEL env.
-    @Value("${novu.bridge.channel:SMS}")
+    // Global channel allow-list for dispatch, as a comma-separated value
+    // (e.g. "whatsapp" or "whatsapp,sms"). This is the deployment-wide guard:
+    // a tenant can only dispatch on a channel that is BOTH enabled in its
+    // NotificationChannel config AND present here. Lets ops keep a channel
+    // globally paused (e.g. SMS pending Twilio A2P) regardless of tenant
+    // toggles. Override via NOVU_BRIDGE_CHANNEL env.
+    @Value("${novu.bridge.channel:whatsapp}")
     private String channel;
+
+    /**
+     * The {@link #channel} CSV parsed into a normalised (lowercase, de-duped) allow-list.
+     * Channel codes are compared lowercase everywhere so they match the lowercase
+     * channel values used in ProviderDetail/TemplateBinding seed data and the
+     * lowercased NotificationChannel.code values.
+     */
+    public List<String> getAllowedChannels() {
+        if (channel == null) {
+            return Collections.emptyList();
+        }
+        return Arrays.stream(channel.split(","))
+                .map(String::trim)
+                .filter(s -> !s.isEmpty())
+                .map(s -> s.toLowerCase())
+                .distinct()
+                .collect(Collectors.toList());
+    }
 
     @Value("${novu.bridge.default.locale:en_IN}")
     private String defaultLocale;
--- a/backend/novu-bridge/src/main/java/org/egov/novubridge/consumer/DomainEventConsumer.java
+++ b/backend/novu-bridge/src/main/java/org/egov/novubridge/consumer/DomainEventConsumer.java
@@ -8,8 +8,6 @@ import org.egov.novubridge.service.DispatchPipelineService;
 import org.egov.novubridge.web.models.ComplaintsDomainEvent;
 import org.egov.tracer.model.CustomException;
 import org.springframework.kafka.annotation.KafkaListener;
-import org.springframework.kafka.support.KafkaHeaders;
-import org.springframework.messaging.handler.annotation.Header;
 import org.springframework.stereotype.Component;
 
 import java.util.HashMap;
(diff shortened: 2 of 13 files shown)


singhalkarun:
**Added: NotificationChannel schema wiring + default-OFF reconciliation** (commit 8613d688)

The gate needs the `NotificationChannel` schema registered in MDMS to resolve. Findings + changes:
- `default-data-handler` **already** registers `schema/NotificationChannel.json` per tenant (via the `classpath:schema/*.json` glob) and seeds `configData/NotificationChannel.json`. So schema registration was already wired on the canonical tenant-onboarding path.
- **But** that default configData had `WHATSAPP enabled: true` — every new tenant would send WhatsApp before anyone opted in, contradicting the agreed **default OFF**. Flipped it to `enabled: false`.
- Added the schema to the local-setup seed path too (`notif-mdms-seed`): `schemas/NotificationChannel.json` + registration loop — schema only, no channel data seeded, so local/dev tenants also default OFF.

Net: schema resolves on both onboarding paths; no channel is ever on until an operator enables it via the configurator (#919).

singhalkarun:
**Contract test for the config-service calls** (commit 81782d3b)

Running the full `novu-bridge` standalone needs Kafka + Postgres + config-service + user-preferences + MDMS + Novu live, which isn't practical to stand up ad hoc. Instead, `ConfigServiceClientContractTest` runs the **real `ConfigServiceClient`** (real `RestTemplate` + JSON + HTTP) against a local `HttpServer` stub that mimics config-service's `_resolve`/`_search` contract — verifying the actual on-the-wire request shapes and response parsing the mocked unit tests don't exercise.

It checks:
- `isChannelEnabled` emits `{ resolveRequest:{ schemaCode:NotificationChannel, tenantId, criteria.code } }` (lowercase channel → uppercase code) and parses `configData.data.enabled` (+ disabled / no-record default-off).
- `getEnabledChannels` emits `{ criteria:{ schemaCode, tenantId, criteria.enabled } }` and parses + lowercases `configData[].data.code` (+ empty).

Worth noting it caught a real content-negotiation detail: the client only parses a JSON body when the response carries `Content-Type: application/json` (else it fails closed). Real config-service sends it; flagging because any stub/gateway in front must preserve it.

Full novu-bridge suite: **18 green**. The behavioral end-to-end (Kafka event → `nb_dispatch_log`) still needs a live stack.

singhalkarun:
## ⚠️ Behavioral change requiring a data backfill before deploy

This is the one regression no test catches, so calling it out explicitly.

**Before:** `NotificationChannel` was inert — the bridge sent on `NOVU_BRIDGE_CHANNEL` regardless. Any tenant with a `ProviderDetail` + consent was receiving WhatsApp.

**After:** dispatch is gated on an **enabled `NotificationChannel`** record, default OFF. Existing live tenants have **no such record** (it was never read before), so on deploy they will **stop receiving notifications** until a channel is explicitly enabled.

**Required before/with deploy to any env with live tenants:** backfill an enabled WhatsApp `NotificationChannel` for each tenant currently expected to notify, e.g.
```
POST /config-service/config/v1/_create/NotificationChannel
{ "RequestInfo": {...},
  "configData": { "tenantId": "<tenant>", "uniqueIdentifier": "<tenant>.WHATSAPP",
    "data": { "code": "WHATSAPP", "name": "WhatsApp", "enabled": true, "providerName": "twilio", "priority": 1 } } }
```
(New tenants via `default-data-handler` get a **disabled** WHATSAPP record by default — operators opt in via the onboarding UI #919.)

**Verification before sign-off:** for each existing notifying tenant, confirm `_search {schemaCode: NotificationChannel}` returns `enabled: true` for the channel; then fire one complaint event and confirm `nb_dispatch_log` shows `SENT` (not `SKIPPED / NB_CHANNEL_DISABLED`).

Suggest this lands behind a feature window or with the backfill run as part of the same change.

singhalkarun:
## ✅ Update: backfill no longer required (legacy fallback added, 1789a2cd)

The silent-tenant migration risk above is resolved in code, so the ⚠️ backfill is **no longer needed**.

The gate now distinguishes **unconfigured** from **configured-but-disabled**:
- A tenant with **no `NotificationChannel` records** → falls back to legacy behaviour (dispatch on the `NOVU_BRIDGE_CHANNEL` allow-list). **Existing tenants keep working untouched.**
- A tenant with **any record** → respected exactly (enabled ∩ allow-list).

This preserves everything: existing tenants need no migration; new tenants (seeded a **disabled** record by `default-data-handler`) stay default-OFF and opt in via the UI; explicit disable still works. On a transient config-service error we fall back to legacy (favour delivery over dropping).

Net for reviewers: **deploy is safe with no data migration step.** The only behavioural change is that a tenant which has *explicitly configured* channels and disabled them all will send nothing — which is the intended control.

singhalkarun:
## Update: config-lookup errors now throw → DLQ (cce3c87b)

Refining the error path from my earlier "favour delivery on transient error" note. The legacy fallback now applies **only** to a genuine "no records" result. A **hard** `NotificationChannel` lookup failure (non-2xx / unreachable / unparseable) **throws** instead of falling back, so the event goes down the consumer's existing retry/DLQ path and is reprocessed when config-service recovers.

Why: a transient outage was being read as "unconfigured," which could ignore an explicit per-tenant disable (fail-open). Throwing avoids both that and the silent-drop alternative — it just defers during the outage.

- `null` = genuinely unconfigured → legacy allow-list (unchanged).
- empty list = configured, all disabled → dispatch nothing (unchanged).
- error = throw `NB_CHANNEL_CONFIG_SEARCH_FAILED` → DLQ/retry (new).

Tests added: throws on RuntimeException + non-2xx (unit), real 5xx propagates (contract), throw bubbles out of `processEnabledChannels` (pipeline). Suite 17 green.

singhalkarun:
## Deploy note — config-service write permission (for the paired UI #919)

The bridge reads config-service **in-cluster** (`digit-config-service.egov:8080`), bypassing the gateway/access-control — so the bridge's NotificationChannel reads need no role. ✅

But the configurator (#919) **writes** channel config through the gateway, which gates config-service by role (repo `ACCESSCONTROL-ACTIONS`/`ROLEACTIONS`): `_create`/`_update` require **`SUPERUSER` or `CONFIG_ADMIN`** (`_search` is open to `EMPLOYEE`). So when provisioning a tenant/operator for this feature, ensure the operator who sets channels has `SUPERUSER`/`CONFIG_ADMIN`, and that the env's access-control registers the config-service actions (repo defines them under `ACCESSCONTROL-ACTIONS-TEST` — confirm the prod set matches). No impact on the bridge itself.

singhalkarun:
## ✅ Live-verified on a real DIGIT env (bomet dev server)

Ran the gate against a full DIGIT stack (docker-compose: config-service, redpanda/Kafka, MDMS, Kong, Postgres, Novu) — **not mocks**. Approach: built this branch's image and ran it as an **isolated** container alongside the untouched live `novu-bridge`, on **idle Kafka topics + a distinct consumer group** (so it consumed zero live traffic), and drove the `_dry-run` diagnostic endpoint. The live bridge and live notifications were never touched.

**Gate proven** (real config-service):
| `ke` WHATSAPP config | dry-run | gate |
|---|---|---|
| no record | `Validation only mode` | ✅ passes (legacy fallback) |
| active, `enabled=false` | `Channel not enabled for tenant` (`enabledChannels=[]`) | ✅ blocks |
| active, `enabled=true` | passes | ✅ |

**Bonus validations:** the **split input/retry `@KafkaListener`s** registered correctly against real Kafka, and the **legacy-fallback-for-unconfigured-tenants** path confirmed live (the env had the schema for 18 tenants but zero records → every tenant legacy-falls-back → no disruption on deploy).

### 🐛 Found + fixed a real bug only a live env surfaces (commit 58803277c)
config-service `_search` returns **inactive (soft-deleted) rows** unless filtered — but `_resolve` (the runtime path) ignores them. `getEnabledChannels` used `_search` without an `isActive` filter, so a deactivated record still counted as "configured" and **wrongly suppressed the legacy fallback**. Fix: add `isActive:true` to the search (matches `_resolve`). Re-verified live: a soft-deleted record now correctly falls back to legacy instead of blocking. Unit + contract tests updated; suite 21 green.

### Notes for whoever deploys/tests further (env issues, not this PR's code)
- **Auth on bomet is broken**: legacy `/user/oauth/token` throws a SQL error in egov-user (this box uses Keycloak/token-exchange). We bypassed it by calling config-service directly (it has no authz of its own) + the test bridge's HTTP directly.
- **`TemplateBinding` `_resolve` intermittently fails** on bomet (`NB_CONFIG_RESOLVE_FAILED`) — code path unchanged by this PR (identical to the deployed bridge), so it's a config-service/enc-service env issue. Doesn't affect the gate; does mean a full real-send test on this box isn't currently reliable.

All test data cleaned up (ke back to 0 records), test container/image torn down, live bridge healthy throughout.

KDwevedi:
Closing as stale. The per-tenant channel enable/disable + multi-channel dispatch this targets is consolidated in the merged notifications foundation #1059 (config-driven, MDMS-routed, multi-channel, self-service SMS/Email/WhatsApp) plus the in-flight #1284 (WhatsApp Content-SID delivery + Twilio sync + enablement runbook). Both edit the same bridge surfaces (`DispatchPipelineService`, `NovuBridgeConfiguration`), so we're keeping one line of work to avoid divergent approaches. Thanks @singhalkarun — if anything here (e.g. the retry path) isn't covered by #1059/#1284, flag it and we'll fold it in.

PR #919, opened 2026-06-23
feat(configurator): communications channel enable/disable — onboarding + management (#547)

singhalkarun:
Frontend for **SaaS Comms Infra Setup** (#547). Adds a per-tenant notification-channel enable/disable UI in the configurator, in two places that share one component + hook. Pairs with the backend in **#918** (the toggles only take effect once #918 is deployed).

## What this does
- **Onboarding step (Phase 5 — "Communications").** A new, **skippable** step with WhatsApp / SMS / Email toggles, **off by default**. Skip writes nothing; Save persists and continues.
- **Management settings page** (Management → **System → Communications**, `/manage/communications`). The "configure it later" home — same toggles for the logged-in tenant, with Save.
- **Shared internals.** `ChannelToggleList` + `channels.ts` + `useNotificationChannels` hook are reused by both screens so they behave identically.
- **config-service integration.** New `configService` writes `NotificationChannel` records via `/config-service/config/v1/_create|_update|_search`. Since config-service has no upsert (schema x-unique on `code`), it **creates then reconciles via `_update` on a duplicate** (case-insensitive match), mirroring `notif-mdms-seed/seed.sh`.

## Wiring
- New `phase/5` route; Phase 4 "Continue" → Phase 5; `completePhase` cap 5→6; completion screen stepper/table + `onboarding_complete` telemetry updated for the new final phase.
- Sidebar entry under the **System** group; `/manage/communications` custom route; i18n label added.

## Robustness (from independent review)
- **Best-effort save** (`Promise.allSettled`): every channel is attempted; a combined error names which failed instead of aborting mid-loop and masking partial success.
- **Read-failure guard:** if the initial load fails, a warning is shown and **Save is blocked** so a transient read can't overwrite all channels to OFF.
- Clear error (not a misleading `INVALID_ID`) when a duplicate can't be located; friendlier 403 message; state resets on tenant change; the disabled toggle is properly `disabled`/`aria-disabled`.

## Scope (per team, 2026-06-23)
Enable/disable **only**. Credentials and message templates are **out of scope** — assumed pre-seeded on the Twilio/SendGrid side; message text lives in Twilio as approved templates.

## Tests
`vitest` **40 green** — `configService` (create / duplicate→update / not-found / case-insensitive / partial-failure / search), the onboarding page, and the management page (render/default-off, pre-populate, save+toast, skip, error). `tsc -b` + lint clean. Browser-level QA is manual (checklist in comments).

## Depends on
**#918** — without it these toggles write config that nothing consumes yet.


singhalkarun:
**Added QA tests** (commit af2ea689)

- **`CommunicationsPage.test.tsx`** (RTL, runs in CI now): default-off render, pre-populate from existing config, save payload + toast + navigate, skip-writes-nothing, and save-error shows an alert without navigating. 5 cases.
- **`e2e/tests/communications.test.mjs`** (Puppeteer): onboarding login → enable WhatsApp → Save → **reload `/phase/5` and assert the toggle persisted** (proves the config-service write round-tripped, not just local UI state) → restore default-OFF for idempotency. Runs against a deployed origin via `E2E_BASE_URL`, so it goes green once this branch has a preview/staging deploy.

Full vitest suite **32 green**, `tsc -b` clean, e2e syntax-checked. The behavioral end-to-end (enable → fire a `complaints.domain.events` event → assert `nb_dispatch_log` SENT/SKIPPED) still needs backend #918 deployed — out of scope for this PR.

singhalkarun:
## Manual QA checklist — Communications step

**Prerequisites**
- This branch deployed/served somewhere reachable, with a DIGIT backend.
- The `NotificationChannel` schema registered in MDMS for the test tenant (auto-registered by `default-data-handler`; also covered by the seed wiring in backend #918).
- Log in via **Onboarding** mode (the step lives at `/phase/5`).

**Navigation & gating**
- [ ] Complete Phase 4 → "continue" lands on the **Communications** step (phase 5).
- [ ] The stepper shows a 5th step ("Comms"); it's reachable after Phase 4.
- [ ] Direct-navigating to `/phase/5` while authenticated renders the step.

**Default-off**
- [ ] On a tenant with no channel config, **all three toggles (WhatsApp / SMS / Email) are OFF**.
- [ ] SMS and Email show their "may not be wired / globally paused" notes.

**Enable → save → persistence (the key check)**
- [ ] Toggle **WhatsApp ON** → "Save & Continue" → success toast lists "WhatsApp" → lands on the completion screen.
- [ ] Re-enter `/phase/5` → **WhatsApp is still ON** (proves the write persisted, not just local UI state).
- [ ] Independent read-back: `POST /config-service/config/v1/_search` with `{criteria:{tenantId,schemaCode:"NotificationChannel"}}` returns the WHATSAPP record with `enabled: true`.

**Re-toggle (update path)**
- [ ] Re-enter, toggle **WhatsApp OFF**, Save → re-enter → it's OFF (the create-then-_update reconcile works; no duplicate-record error surfaces).

**Skip**
- [ ] From a fresh entry, click **"Skip for now"** → lands on completion, and **no** `NotificationChannel` records were written/changed for that tenant.

**Error handling**
- [ ] (If you can force a config-service failure, e.g. schema not registered) Save shows a **red error alert** and stays on the page — no navigation.

**Behavioral end-to-end** (needs backend #918 deployed)
- [ ] With WhatsApp enabled, file/transition a complaint for the tenant → `nb_dispatch_log` row `SENT` for `whatsapp`.
- [ ] With WhatsApp disabled, same event → `SKIPPED` / `NB_CHANNEL_DISABLED`, no message sent.

Automated CI covers the page logic (`CommunicationsPage.test.tsx`: default-off, pre-populate, save payload, skip, error). This checklist covers the live browser + real config-service persistence that the unit tests mock out.

singhalkarun:
**Review fixes applied** (commit 5ca3c838) — from an independent adversarial review:

- **Reconcile robustness:** `upsertNotificationChannel` matches the existing row **case-insensitively** and throws a clear error instead of sending `_update` with an `undefined` id (which produced a misleading `INVALID_ID`) when a duplicate can't be located. [F#1]
- **Partial-failure handling:** `saveNotificationChannels` is now best-effort (`Promise.allSettled`) — every channel is attempted and a combined error names which failed, instead of aborting mid-loop and masking partial success. [F#2]
- **Silent-overwrite footgun closed:** if the initial read fails, the hook surfaces a `loadError`, the pages show a warning, and **Save is blocked** — so a transient read can't overwrite all channels to OFF. [F#5]
- **403 UX + tenant-change reset + a11y:** friendlier permission-denied message; state resets on tenant change; the disabled toggle is now actually `disabled`/`aria-disabled` with a visual cue. [F#3/F#4/F#6]

Tests added for the not-found, case-insensitive, and partial-failure paths. tsc clean; vitest **40 green**.

(One review note was a false alarm — "the bridge doesn't gate on `enabled`" — because it reviewed this PR against develop, where #918 isn't merged. With #918 the bridge does gate. Confirms the cross-PR dependency: #919 is inert until #918 ships.)

singhalkarun:
## ⚠️ Deploy prerequisite — operator role for saving channels

Saving channel toggles writes to **config-service** (`_create`/`_update`), which DIGIT access-control gates by role (verified from the repo's `ACCESSCONTROL-ACTIONS`/`ROLEACTIONS`):

- **Read / pre-populate** (`config/v1/_search`): `EMPLOYEE`, `CITIZEN`, `SUPERUSER`, `CONFIG_ADMIN` — broad, so the screen loads fine for any operator.
- **Save** (`config/v1/_create` + `_update`): **only `SUPERUSER` or `CONFIG_ADMIN`.**

**Implication:** the onboarding/management operator must have **`SUPERUSER` or `CONFIG_ADMIN`** to save. A plain `EMPLOYEE` can view/toggle but **Save will 403** (the page now shows a clear permission-denied message for this case). This is a provisioning step, not a code change.

To confirm per environment: (a) the operator account carries one of those roles, and (b) the deployment's access-control actions/role-actions match the repo defaults (these are currently defined under `ACCESSCONTROL-ACTIONS-TEST`, so verify the prod set registers config-service `_create`/`_update`/`_search` and maps them to the operator's role). Derived statically — no live env needed to know the requirement; only the per-user role assignment needs checking in the target env.

singhalkarun:
## Live-test follow-up + a consistency fix (commit 8170b874d)

The backend gate (#918) was live-verified against a real DIGIT env, which surfaced a config-service behaviour worth mirroring here: `_search` returns **inactive (soft-deleted)** records unless filtered. To keep this UI consistent with the gate, `getNotificationChannels` (used to pre-populate the toggles) now searches with `isActive:true`, so a soft-deleted record doesn't pre-populate a toggle with stale state.

The reconcile search in `upsertNotificationChannel` is intentionally left **broad** (no `isActive` filter) so it can still locate **and revive** a soft-deleted duplicate on save. Test added; vitest green.

Note: a real end-to-end UI test on the bomet dev box is currently blocked by that env's auth (its legacy `/user/oauth/token` is erroring; the box uses Keycloak) — not a code issue here. The gate this UI drives is live-verified on #918.

KDwevedi:
Closing as stale. The configurator communications config is consolidated in #1059 (config-driven, self-service SMS/Email/WhatsApp foundation) + #1284 (Notification Providers management + Sync-from-Twilio map-and-confirm UI). Keeping one line of work to avoid overlapping configurator/i18n edits. Thanks @singhalkarun — if the onboarding-flow channel-toggle surface specifically is still needed on top of the consolidated stack, let's scope it as a follow-up rather than a parallel PR.

PR #924, opened 2026-06-24
feat(novu-bridge): bounded auto-retry with backoff before DLQ (#547 follow-up)

singhalkarun:
**Follow-up to #918** (stacked on it — base is `feature/comms-channel-enable-gate`; rebase to `develop` once #918 lands).

## Why
#918 made transient failures (e.g. a config-service blip) **throw → DLQ**, which is safe but the DLQ is a terminal holding pen — recovery needed a **manual replay**. The service already had retry scaffolding (a `novu-bridge.retry` topic the consumer listens to, a `maxRetries` setting) but **nothing fed it**. This wires it up so short outages self-recover.

## What
- On any processing failure, the consumer **re-queues the event on the retry topic** with an incremented `attempt`; once `attempt` exceeds `maxRetries` it routes to the **DLQ**. Bounded, nothing silently dropped.
- Messages consumed from the retry topic **wait `novu.bridge.retry.delay.ms` (default 5s, 0 disables)** before reprocessing — so a brief outage gets spacing between attempts instead of burning all retries in milliseconds.
- Retry-topic messages are the wrapper `{ event, attempt, ... }`; input-topic messages stay the raw event (distinguished by the topic header).

## Behavior
- Transient single failure → retried, usually succeeds → no DLQ, no manual step.
- Sustained outage → `maxRetries` spaced attempts, then DLQ (same safety as before, now after trying).
- Success → nothing produced.

## Tests
`DomainEventConsumerTest`: success → no produce; input failure → retry topic `attempt=1`; retry below max → incremented attempt; at max → DLQ. Backoff set to 0 in tests. Full novu-bridge suite **21 green**.

## Known limitation
Backoff is **blocking** (`Thread.sleep` on the retry consumer). Fine for this low-volume, event-driven path. If throughput ever demands it, a non-blocking delayed-retry (Spring `@RetryableTopic` or a scheduler) is the next step — noted, not done here.

### DIFF
--- a/backend/novu-bridge/src/main/java/org/egov/novubridge/config/NovuBridgeConfiguration.java
+++ b/backend/novu-bridge/src/main/java/org/egov/novubridge/config/NovuBridgeConfiguration.java
@@ -63,6 +63,12 @@ public class NovuBridgeConfiguration {
     @Value("${novu.bridge.max.retries:3}")
     private Integer maxRetries;
 
+    // Fixed backoff applied before reprocessing a message consumed from the retry topic, so a brief
+    // downstream outage (e.g. config-service) gets spacing between attempts instead of burning all
+    // retries instantly. Bounded by maxRetries. 0 disables the wait (used in tests).
+    @Value("${novu.bridge.retry.delay.ms:5000}")
+    private Integer retryDelayMs;
+
     @Value("${novu.bridge.preference.enabled:true}")
     private Boolean preferenceEnabled;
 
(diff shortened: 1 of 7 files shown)


singhalkarun:
**Review fixes applied** (commit 9d43ef9e) — from an independent adversarial review of the backend stack:

- **`_validate`/`_dry-run` regressions fixed:** the diagnostic `process()` no longer 500s when config-service is unavailable (returns a diagnostic result) and no longer writes a `SKIPPED` dispatch-log row on `send=false`. [B#1]
- **Per-channel isolation widened:** `dispatchForChannel` now catches `Exception` (not just `CustomException`), so an unexpected error on one channel is recorded `FAILED` instead of bubbling and re-queuing the whole event — which would re-send channels that already succeeded (duplicates). New test locks it. [B#3]
- **Retry listener split + capped:** input and retry topics now have **separate** `@KafkaListener`s, and the retry listener uses `max.poll.records=1`, so the blocking backoff can't head-of-line-block live notifications or breach `max.poll.interval.ms` under a backlog. [B#2]
- **`seed.sh` comment corrected:** it registers the schema only; seeded tenants fall back to the legacy allow-list (not "default OFF"). [B#8]

novu-bridge suite **23 green**, full `mvn package` produces deployable jars (both novu-bridge and default-data-handler, with the NotificationChannel config/schema bundled).

Note: these fixes touch #918's gate code as well as #924's retry code; since this PR is stacked on #918, they live on this tip branch — merge the backend stack together (#918 → #924).

Issue #903, opened 2026-06-20
[Enhancement] Cross-departmental complaint search page for supervisors and central roles


subhashini-egov:
## Context
Today employees can only see complaints through their Inbox, which is scoped to their own assignment queue. Supervisors and central roles (e.g. Screening Officer, Reception Officer) need to find complaints assigned to other people — across departments — to triage, monitor reportees, and re-route.

## Change
- New employee search page, distinct from Inbox.
- Visibility scope driven by an MDMS-managed role master:
  - Department-scoped roles (e.g. Case Manager) → restricted to their department.
  - Supervisor roles → scoped to their reportees.
  - Central roles (Screening Officer, Reception Officer) → unrestricted across departments.
- Filters: complaint number, mobile (PII-masked per existing rules), date range, status, department, assignee.
- Adding/removing a role from the cross-dept master must not require a code change.

## Acceptance criteria
- New `/search` page reachable from employee nav for configured roles.
- RBAC verified across 4 personas (Case Manager dept-scoped; Supervisor reportee-scoped; Screening Officer central; Reception Officer central).
- Visibility behaviour matches the MDMS role master without code redeploy.
- Search responses respect the existing PII masking rules.

## Why this matters
- Mozambique IGE/IGSAE roll-out needs supervisor and central screening visibility from day one.
- Foundation for the broader role-scoped search work in 2.20.

## Notes
- Break down into UI and backend slices during planning.
- Confirm whether the necessary backend search APIs already exist or need extension.

KDwevedi:
superceded by inbox visibility feature

PR #965, opened 2026-06-29
feat(pgr-inbox): My/All visibility tabs (inbox v2)

KDwevedi:
## What

Inbox **v2** visibility: replaces the "Assigned to me / Assigned to all" filter-card radio with two **tabs** above the PGR employee inbox, per the CMS visibility PRD (visibility v1 tab) and a first slice toward #903.

| | Before | After |
|---|---|---|
| Scope control | radio inside the filter card | **My Complaints / All Complaints** tabs above the list |
| Default | Assigned to all | **My Complaints** (matches PRD reference image) |
| "My" backing | radio → `assignee` (no-op before #942) | `assignee = <my uuid>`, resolved server-side via #942 |
| Switch behaviour | n/a | form **resets** on tab switch (composer remount) |

## How

- **`PGRInbox.js`** — renders the two tabs with count badges; the active tab is passed to `PGRInboxConfig.preProcess` through the existing 2nd arg (`configs.additionalDetails.assigneeScope`). `key={activeTab}` remounts the `InboxSearchComposer` so the search/filter form resets on switch (the PRD's "if a filter was applied and the tab switched, the filter should reset").
- **`UICustomizations.js`** — `preProcess` reads `assigneeScope`: `MY` → `params.assignee = <my uuid>` (single string, matching the backend `String` field), `ALL` → no assignee filter. Old `filterForm.assignedToMe` block removed.
- **`PGRSearchInboxConfig.js`** — removes the `assignedToMe` radio field + default.
- **`PGRService.count()`** — applies the same assignee resolution as `search()` (extracted into `resolveAssigneeFilter`). Without this the "My" badge counted **all** complaints, since an unset `serviceRequestIds` adds no SQL clause. This is a parity fix for #942.
- **`overrides.css`** — `.v2-pgr-inbox-tabs` styling (active accent + count pill).

## Scope of this increment

✅ Tabs, default-to-My, scope wiring, reset-on-switch, per-tab count badges, count() parity fix.

⏭️ **Deferred** (separate increments):
- PRD badge semantics "unopened" (My) / "new since last opened" (All) + red dot — needs per-user read-state tracking.
- Supervisor **reportee / role N+1** scoping for All Complaints — needs HRMS reportee→assignee resolution.

## Notes
- Tab labels use i18n keys `PGR_INBOX_TAB_MY_COMPLAINTS` / `PGR_INBOX_TAB_ALL_COMPLAINTS` with English fallbacks; seed for other locales.
- **Behaviour change:** employees now land on *My Complaints* by default (was *all*); an employee with nothing assigned sees an empty initial list.
- Validated: full read of the search/count/workflow paths, controller query-param binding for both `_search` and `_count`, and esbuild syntax-check of all three JS files. Not yet run against a live stack — pending e2e on a running tenant.

Relates to #903.

### DIFF
--- a/digit-ui-esbuild/products/pgr/src/pages/employee/PGRInbox.js
+++ b/digit-ui-esbuild/products/pgr/src/pages/employee/PGRInbox.js
@@ -1,9 +1,27 @@
 import React, { useMemo, useState, useEffect } from "react";
 import { InboxSearchComposer, HeaderComponent, Toast, Loader } from "@egovernments/digit-ui-components";
+import { Request } from "@egovernments/digit-ui-libraries";
 import { useTranslation } from "react-i18next";
 import _ from "lodash";
 import PGRSearchInboxConfig from "../../configs/PGRSearchInboxConfig";
 import { useLocation } from "react-router-dom";
+
+/**
+ * Inbox v2 visibility tabs. The active tab drives the assignee scope that
+ * PGRInboxConfig.preProcess reads off `configs.additionalDetails.assigneeScope`:
+ *   MY  -> params.assignee = <my uuid> (complaints currently assigned to me)
+ *   ALL -> no assignee filter (every complaint in the tenant)
+ * "My Complaints" is the default, matching the visibility PRD reference image.
+ */
+const INBOX_TABS = [
+  { key: "MY", labelKey: "PGR_INBOX_TAB_MY_COMPLAINTS", fallback: "My Complaints" },
+  { key: "ALL", labelKey: "PGR_INBOX_TAB_ALL_COMPLAINTS", fallback: "All Complaints" },
+];
+
+// Default open / actionable states the inbox shows when no status filter is
+// applied. Kept in sync with PGRInboxConfig.preProcess (UICustomizations.js) so
+// the tab count badges reflect the same rows the list shows.
+const INBOX_OPEN_STATES = ["PENDINGFORASSIGNMENT", "PENDINGFORREASSIGNMENT", "PENDINGATLME", "PENDINGATSUPERVISOR"];
 
 /**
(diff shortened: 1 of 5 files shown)


digit-ui-esbuild/products/pgr/src/pages/employee/PGRInbox.js
```diff
@@ -1,10 +1,28 @@
...
+ */
+const INBOX_TABS = [
+  { key: "MY", labelKey: "PGR_INBOX_TAB_MY_COMPLAINTS", fallback: "My Complaints" },
+  { key: "ALL", labelKey: "PGR_INBOX_TAB_ALL_COMPLAINTS", fallback: "All Complaints" },
+];
+
+// Default open / actionable states the inbox shows when no status filter is
+// applied. Kept in sync with PGRInboxConfig.preProcess (UICustomizations.js) so
+// the tab count badges reflect the same rows the list shows.
+const INBOX_OPEN_STATES = ["PENDINGFORASSIGNMENT", "PENDINGFORREASSIGNMENT", "PENDINGATLME", "PENDINGATSUPERVISOR"];
```
vinothrallapalli-eGov, line 24: @KDwevedi This we can't make it dynamic or configurable, because where more levels are introduced to default workflow with new states then those case may break.

KDwevedi:
Superseded by #1052 (Visibility V1 step 1, per the #1020 design). Same My/All tabs UX on the same inbox surface, but the V1 design derives "My" from role-actionable workflow states and defers the assignee axis (this PR's approach, built on #942) to Visibility V2 — see VISIBILITY-DESIGN.md in #1052. Closing to keep one lineage; the assignee-resolution parity fix for PGRService.count() from this branch will be folded into V2 when that lands.

Issue #923, opened 2026-06-23
[Bug] CMS: Citizen/Employee: Invalid file attached message does not get cleared automatically or manually


Gurjeet-eGov:
# Description

Citizen/Employee: Invalid file attached message does not get cleared automatically or manually

---

# Steps to Reproduce

1. Login to citizen
2. Go to create complaint 
3. Enter details
4. Upload an image exceeding 5mb limit 

---

# Expected Behavior

Toast message shall clear automatically and also display close icon 

---

# Actual Behavior

<img width="1366" height="656" alt="Image" src="https://github.com/user-attachments/assets/81a4cfa4-b3a5-4cd0-88ba-f73b80028413" />

singhalkarun:
Fixed in #976 (verified live on the citizen create flow with a 25 MB image).

Root cause: the upload widget (`ImageUploadHandler`, used by citizen create + reopen) renders the shared `Toast` for the "file too large" error, but (a) the `Toast` error variant only shows its close button when `isDleteBtn` is passed — it wasn't — so there was no × to dismiss it, and (b) nothing cleared the error on a timer, so the banner stayed forever.

Fix: pass `isDleteBtn` (manual close ×) and auto-dismiss the error after 5s.

Note: the employee create form has no image-upload field, so this case doesn't arise there; its toasts already auto-dismiss and its document uploads use inline (self-clearing) validation errors. So the fix targets the citizen path the repro describes.

Gurjeet-eGov:
@singhalkarun fix the css, close icon is just a plain white box 

<img width="1366" height="659" alt="Image" src="https://github.com/user-attachments/assets/48d7478e-d795-4558-8809-76c19c23b16a" />

https://github.com/user-attachments/assets/8b5c8d0b-a050-4b22-a86b-498a8975b6d3

PR #976, opened 2026-06-29
fix(pgr): auto-dismiss + close button on file-upload error toast (#923)

singhalkarun:
Fixes #923.

## Problem
On **citizen → Create Complaint → upload an image that's too large / unsupported**, the error toast ("file too large") stayed on screen permanently — it didn't auto-dismiss and had no close icon, so the user couldn't get rid of it.

## Root cause
The upload widget is `ImageUploadHandler` (citizen create reuses it via the registered `SelectImages`; Reopen-complaint uses it too). It renders the shared `Toast` atom for the error, but:
- The `Toast` **error variant only renders the close button when `isDleteBtn` is set** — `ImageUploadHandler` didn't pass it ⇒ **no close icon**.
- The atom has **no auto-dismiss timer**, and the caller never cleared the error on a timeout ⇒ **the banner persisted forever**.

## Fix (`ImageUploadHandler.js`)
- Pass `isDleteBtn` so the error toast shows a **manual close button**.
- **Auto-dismiss** the error after 5s.

## Scope note
The **employee** Create-Complaint form has no image-upload field (only text/dropdown/location components), so the persistent-toast case doesn't arise there; its other toasts already auto-dismiss (3s) and its document uploads use inline validation errors that clear on re-upload/delete. The reported repro is citizen-only.

## Verification
- `npm run build` passes.
- Manual QA: rejected-file toast now shows a close (×) and disappears on its own after a few seconds.

### DIFF
--- a/digit-ui-esbuild/packages/react-components/src/atoms/ImageUploadHandler.js
+++ b/digit-ui-esbuild/packages/react-components/src/atoms/ImageUploadHandler.js
@@ -43,6 +43,15 @@ export const ImageUploadHandler = (props) => {
     }
   }, [imageFile]);
 
+  // Auto-dismiss the upload error toast so a rejected file (too large /
+  // unsupported) doesn't leave a banner stuck on screen forever (CCRS#923).
+  // The toast also gets a manual close button (isDleteBtn below).
+  useEffect(() => {
+    if (!error) return undefined;
+    const timer = setTimeout(() => setError(null), 5000);
+    return () => clearTimeout(timer);
+  }, [error]);
+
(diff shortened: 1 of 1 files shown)


singhalkarun:
Auto hides after 5s now

https://github.com/user-attachments/assets/1cbd837c-a7a6-4c4d-9ae7-4cd0ca5c87eb

KDwevedi (approved):
Approved. Minimal, correct fix for #923 — adds the close button and 5s auto-dismiss; timer cleanup is sound. Optional follow-up: reset the file input value so re-selecting the identical rejected file re-triggers the toast.

Issue #722, opened 2026-06-02
[Bug]  Postal Code validation issue


pradeepkumarcm-egov:
# Description
 
In the host variable file, the postalCodeLength is configured as 4, and the backend validation is also using this flag and  validating a 4-digit postal code.

However, in the UI, the system is validating the postal code as a 5-digit number and does not allow the application to be submitted when a valid 4-digit postal code is entered. The UI validation should be aligned with the configured postalCodeLength value and backend validation rules.

# Environment

<img width="1470" height="956" alt="Image" src="https://github.com/user-attachments/assets/d1cfa563-5141-47fb-8d85-2f90edd94b89" />

<img width="1470" height="956" alt="Image" src="https://github.com/user-attachments/assets/c6849a21-ea34-41fb-9118-f4eba33a08ae" />

KDwevedi:
I'll replicate in Bomet and fix

Gurjeet-eGov:
**Employee UI** create complaint issue is resolved, it works with 4 digit pincode only
**Citizen UI** create complaint issue is still there, even when user does not select any pin on map, the pincode is autofilled to 6 digits and complaint is registered successfully. So, **even the service level validation for 4 digit pincode does not work** 

https://github.com/user-attachments/assets/187ea28d-2f25-4034-9eb1-1269a1a7de7b
https://github.com/user-attachments/assets/d493950e-4712-49dc-90ab-0122842b1668

singhalkarun:
Citizen side fixed in #978 (employee was already resolved). Notes on what was found:

The active citizen flow (`CreatePGRFlowV2`) had **no real postal-code validation** — the pincode was auto-filled from the map pin, locked, and submitted as-is. Fixes:
- **Config-driven validation** against `CORE_POSTAL_CONFIGS.postalCodePattern` (same source the backend length uses); the location step is blocked while the pincode doesn't match. Mirrors how the employee form was fixed.
- **Error message now reflects the configured length** ("valid 4-digit…") — it was hardcoded to "5 digit" via the stock localization. Derived from the pattern, with a tenant override (`postalCodeErrorMessage`) honored.
- **Field is now editable** (was disabled whenever the map produced a value) so a wrong Nominatim pincode can be corrected.
- **Fixed a revert loop**: the map→postalCode mirror `useEffect` had `formData.postalCode` in its deps and forced the value back on every keystroke, making the field feel locked. It now seeds from the map only when the pin's pincode actually changes.
[...]

PR #978, opened 2026-06-29
fix(pgr): align citizen postal-code validation with configured length (#722)

singhalkarun:
Fixes #722 (citizen side; the employee side was already resolved).

## Problem
On the **citizen** create-complaint flow, the postal code is auto-filled from the map pin (Nominatim), **locked**, and submitted as-is. Per the issue + latest comment:
- A valid **4-digit** code (matching the backend's `postalCodeLength`) had no way through; a wrong **6-digit** Nominatim code got locked in and submitted.
- The error message said **"5 digit"** regardless of the configured length.
- Moving the pin didn't reset a stale pincode, and the field couldn't be edited.

## Fix (`CreatePGRFlowV2.tsx`)
- **Config-driven validation** — validates `postalCode` against `CORE_POSTAL_CONFIGS.postalCodePattern` (the same config the employee fix uses, derived from the tenant's `postalCodeLength`). The location step can't continue while the pincode doesn't match; empty stays allowed (optional).
- **Length-accurate error message** — derived from the configured pattern (e.g. "Please enter a valid **4-digit** postal code") instead of a hardcoded "5 digit". Uses a length-parameterized key (`CS_COMPLAINT_POSTALCODE_INVALID_ERROR_LEN`) with an English fallback; a tenant can still override via `CORE_POSTAL_CONFIGS.postalCodeErrorMessage`.
- **Editable field** — the postal code is no longer disabled when the map auto-fills it, so a wrong reverse-geocoded value can be corrected.
- **Fixed the revert loop** — the mirror `useEffect` listed `formData.postalCode` as a dependency and forced it back to the map value on every keystroke (field felt locked). It now seeds from the map **only when the pin's pincode actually changes** (ref-tracked).
- **Pin-change reset** — moving the pin resets the pincode to the new pin's value, or clears it when the new location has none, rather than keeping a stale code.

## Verification
- `npm run build` passes.
- Verified live on the citizen flow (dev build, real-backend bridge, a simulated 4-digit `CORE_POSTAL_CONFIGS`): 4-digit accepted, non-4-digit shows the "valid 4-digit" error and blocks Continue, the field is editable, and the pincode resets on pin move.

## Note
The latest issue comment also mentions the **backend** accepting a 6-digit pincode ("service level validation … does not work") — that's a separate service-side concern (different repo), outside this UI PR. This change stops the **UI** from submitting a non-conforming pincode.

### DIFF
--- a/digit-ui-esbuild/products/pgr/src/pages/citizen/Create/CreatePGRFlowV2.tsx
+++ b/digit-ui-esbuild/products/pgr/src/pages/citizen/Create/CreatePGRFlowV2.tsx
@@ -54,6 +54,51 @@ function tr(t: (k: string) => string, key: string, fallback: string): string {
 
 declare const Digit: any;
 
+// Postal-code validation is config-driven so the UI honours the same length the
+// backend does, per tenant (CCRS#722). The employee create form and the legacy
+// FormExplorer already read `CORE_POSTAL_CONFIGS.postalCodePattern`; this citizen
+// v2 flow previously had no postal validation at all, so a wrong (e.g. 6-digit
+// Nominatim) pincode auto-filled from the map could be submitted. Optional field
+// — only the format is enforced, and only when a value is present.
+function getPostalConfig(): { pattern: string; errorMessage?: string } {
+  const cfg = (window as any)?.globalConfigs?.getConfig?.("CORE_POSTAL_CONFIGS") || {};
+  return {
+    pattern: cfg.postalCodePattern || "^[0-9]{5}$",
+    errorMessage: cfg.postalCodeErrorMessage, // optional explicit tenant override
+  };
+}
+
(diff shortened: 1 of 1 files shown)


singhalkarun:
https://github.com/user-attachments/assets/e4a9720f-2ae4-4803-a93f-80cc6c1e7b6e

KDwevedi (approved):
Approved. Config-driven citizen postal-code validation at parity with the employee side; field is editable and Continue is correctly gated. Follow-ups (non-blocking): confirm CORE_POSTAL_CONFIGS.postalCodePattern=^[0-9]{4}$ is seeded for the affected tenant, and seed the new CS_COMPLAINT_POSTALCODE_INVALID_ERROR_LEN localization key (currently falls back to hardcoded English).

PR #983, opened 2026-06-30
feat(pgr): add encryption, extended attributes, and complaint schema …

priyanshu-egov:
…support

- Add EncryptionDecryptionService for field-level encryption/decryption via egov-enc-service
- Add ExtendedAttributesValidationService for dynamic field validation against MDMS schema
- Add ComplaintTemplateTypeConfig and ExtendedAttributes models for schema-driven complaint metadata
- Add database migration to create extended_attributes column
- Integrate extended attributes parsing in PGRRowMapper with JSON deserialization
- Add encryption configuration properties (host, encrypt/decrypt endpoints) to PGRConfiguration
- Add MDMS master data files for complaint schema, template types, and related-to mappings
- Update PGRService, EnrichmentService, and UserService to support extended attributes workflow
- Update persister configuration to handle extended_attributes column in database
- Supports field-level encryption based on x-security schema declarations and role-based masking for confidential complaints

### DIFF
--- a/backend/pgr-services/src/main/java/org/egov/pgr/service/EncryptionDecryptionService.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/service/EncryptionDecryptionService.java
@@ -1,0 +1,173 @@
+package org.egov.pgr.service;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import lombok.extern.slf4j.Slf4j;
+import org.egov.common.utils.MultiStateInstanceUtil;
+import org.egov.pgr.config.PGRConfiguration;
+import org.egov.pgr.web.models.ComplaintTemplateTypeConfig;
+import org.egov.pgr.web.models.ExtendedAttributes;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+import org.springframework.web.client.RestTemplate;
+
+import java.util.*;
+
+/**
+ * Encrypt/decrypt/mask extendedAttributes fields via egov-enc-service.
+ *
+ * Which fields to encrypt is declared in the ComplaintSchema MDMS master's x-security array
+ * (merged into ComplaintTemplateTypeConfig by MDMSUtils). This mirrors the x-security convention
+ * from JSON Schema used by digit-config-service.
+ *
+ * Masking is all-or-nothing: when isConfidential=true and the caller lacks
+ * CONFIDENTIAL_COMPLAINT_VIEWER, ALL dynamic fields are replaced with "****" and
+ * decryption is skipped entirely.
+ */
+@Slf4j
+@Service
+public class EncryptionDecryptionService {
+
+    private final RestTemplate restTemplate;
+    private final PGRConfiguration config;
+    private final ObjectMapper objectMapper;
+    private final MultiStateInstanceUtil multiStateInstanceUtil;
+
+    @Autowired
+    public EncryptionDecryptionService(RestTemplate restTemplate,
+                                       PGRConfiguration config,
+                                       ObjectMapper objectMapper,
+                                       MultiStateInstanceUtil multiStateInstanceUtil) {
+        this.restTemplate = restTemplate;
+        this.config = config;
+        this.objectMapper = objectMapper;
+        this.multiStateInstanceUtil = multiStateInstanceUtil;
+    }
+
+    /**
+     * Encrypts fields listed in cfg.xSecurity. No encryptedFields list is stored —
+     * the schema is the source of truth, read again at decrypt time.
+     */
+    public ExtendedAttributes encrypt(ExtendedAttributes ext, ComplaintTemplateTypeConfig cfg,
+                                      String tenantId) {
+        if (ext == null || cfg == null) return ext;
+
+        List<String> secureFields = cfg.getXSecurity();
+        if (secureFields == null || secureFields.isEmpty()) return ext;
+
+        List<String> encKeys     = new ArrayList<>();
+        List<String> plainValues = new ArrayList<>();
+
+        for (String fieldKey : secureFields) {
+            Object raw = ext.getField(fieldKey);
+            if (raw == null) continue;
+            encKeys.add(fieldKey);
+            plainValues.add(raw.toString());
+        }
+
+        if (plainValues.isEmpty()) return ext;
+
+        List<String> ciphers = encryptWithFallback(plainValues, tenantId);
+        if (ciphers != null) {
+            for (int i = 0; i < encKeys.size(); i++)
+                ext.putField(encKeys.get(i), i < ciphers.size() ? ciphers.get(i) : plainValues.get(i));
+        }
+
+        return ext;
+    }
+
+    /**
+     * Decrypts fields listed in cfg.xSecurity. Uses the schema as the source of truth
+     * instead of a stored encryptedFields list. Skip when masking applies — use maskAll() instead.
+     */
+    public ExtendedAttributes decrypt(ExtendedAttributes ext, ComplaintTemplateTypeConfig cfg) {
+        if (ext == null || cfg == null) return ext;
+
+        List<String> secureFields = cfg.getXSecurity();
+        if (secureFields == null || secureFields.isEmpty()) return ext;
+
+        List<String> keys    = new ArrayList<>();
+        List<String> ciphers = new ArrayList<>();
+
+        for (String key : secureFields) {
+            Object c = ext.getField(key);
+            if (c == null) continue;
+            keys.add(key);
+            ciphers.add(c.toString());
+        }
+
+        if (ciphers.isEmpty()) return ext;
+
+        try {
+            List<String> plains = callDecryptBatch(ciphers);
+            for (int i = 0; i < keys.size(); i++)
+                ext.putField(keys.get(i), i < plains.size() ? plains.get(i) : "****");
+        } catch (Exception e) {
+            log.error("Batch decryption failed for fields {}; masking values", keys, e);
+            keys.forEach(k -> ext.putField(k, "****"));
+        }
+
+        return ext;
+    }
+
+    /**
+     * All-or-nothing masking: replaces every dynamic field with "****".
+     * Call instead of decrypt() when the caller lacks CONFIDENTIAL_COMPLAINT_VIEWER.
+     */
+    public ExtendedAttributes maskAll(ExtendedAttributes ext) {
+        if (ext == null || ext.getDynamicFields() == null || ext.getDynamicFields().isEmpty())
+            return ext;
+        new ArrayList<>(ext.getDynamicFields().keySet())
+                .forEach(k -> ext.putField(k, "****"));
+        return ext;
+    }
+
+    // ── private helpers ──────────────────────────────────────────────────────
+
+    private List<String> encryptWithFallback(List<String> plainValues, String tenantId) {
+        String stateTenant = multiStateInstanceUtil.getStateLevelTenant(tenantId);
+        try {
+            return callEncryptBatch(plainValues, tenantId);
+        } catch (Exception cityEx) {
+            log.warn("Encryption failed for tenant '{}', retrying with state tenant '{}'",
+                    tenantId, stateTenant, cityEx);
+        }
+        try {
+            return callEncryptBatch(plainValues, stateTenant);
+        } catch (Exception stateEx) {
+            log.error("Encryption failed for both tenant '{}' and state tenant '{}'; storing plain text",
+                    tenantId, stateTenant, stateEx);
+            return null;
+        }
+    }
+
+    private List<String> callEncryptBatch(List<String> values, String tenantId) {
+        String url = config.getEncHost() + config.getEncEncryptEndpoint();
+        List<Map<String, Object>> requests = new ArrayList<>();
+        for (String v : values) {
+            Map<String, Object> r = new HashMap<>();
+            r.put("tenantId", tenantId);
+            r.put("type", "Normal");
+            r.put("value", v);
+            requests.add(r);
+        }
+        Map<String, Object> payload = new HashMap<>();
+        payload.put("encryptionRequests", requests);
+        JsonNode response = restTemplate.postForObject(url, payload, JsonNode.class);
+        return extractStringList(response);
+    }
+
+    private List<String> callDecryptBatch(List<String> ciphers) {
+        String url = config.getEncHost() + config.getEncDecryptEndpoint();
+        JsonNode response = restTemplate.postForObject(url, ciphers, JsonNode.class);
+        return extractStringList(response);
+    }
+
+    private List<String> extractStringList(JsonNode node) {
+        List<String> result = new ArrayList<>();
+        if (node == null) return result;
+        if (node.isArray()) node.forEach(n -> result.add(n.asText()));
+        else result.add(node.asText());
+        return result;
+    }
+}
--- a/backend/pgr-services/src/main/java/org/egov/pgr/service/EncryptionDecryptionService.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/service/EncryptionDecryptionService.java
@@ -7,6 +7,7 @@ import org.egov.common.utils.MultiStateInstanceUtil;
 import org.egov.pgr.config.PGRConfiguration;
 import org.egov.pgr.web.models.ComplaintTemplateTypeConfig;
 import org.egov.pgr.web.models.ExtendedAttributes;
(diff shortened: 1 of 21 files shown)


backend/pgr-services/src/main/java/org/egov/pgr/service/EncryptionDecryptionService.java
```diff
@@ -0,0 +1,173 @@
...
+            Object raw = ext.getField(fieldKey);
+            if (raw == null) continue;
+            encKeys.add(fieldKey);
+            plainValues.add(raw.toString());
+        }
+
+        if (plainValues.isEmpty()) return ext;
+
+        List<String> ciphers = encryptWithFallback(plainValues, tenantId);
+        if (ciphers != null) {
```
vinothrallapalli-eGov: **[CRITICAL #1] Silent plaintext-fallback permanently corrupts PII**

`encryptWithFallback` returns `null` when `egov-enc-service` is unavailable. The guard `if (ciphers != null)` at this line then skips the write-back loop entirely, so the `ExtendedAttributes` object still holds plaintext values which are persisted to JSONB.

On the next `search()` call, `decrypt()` reads those plaintext values from `xSecurity` and posts them to the enc-service decrypt endpoint as if they were ciphertext. The enc-service returns an error; the catch block masks every secure field with `"****"` — **permanently**. The complaint owner and operators can never read the data again.

**Fix:** Either (a) throw and surface the enc-service failure rather than falling back to plaintext, or (b) persist an `encryptionFailed=true` marker so `decrypt()` can skip those fields instead of sending them as ciphertext.

vinothrallapalli-eGov (commented):
## Code Review — PR #983 (high effort, automated)\n\n10 confirmed findings across the new `extendedAttributes` encryption/decryption feature. **3 findings should block merge** (#1 data corruption, #2 PII never encrypted, #3 plaintext leak in plainSearch). Inline comments are on the nearest reachable diff line; see each comment for the exact root-cause line.\n\n| # | Severity | File | Finding |\n|---|----------|------|---------|\n| 1 | CRITICAL | EncryptionDecryptionService | Silent plaintext-fallback permanently masks PII as "****" |\n| 2 | CRITICAL | EnrichmentService / ExtendedAttributes | email & complainantAddress never encrypted |\n| 3 | CRITICAL | PGRService | plainSearch returns raw ciphertext |\n| 4 | HIGH | PGRService | null-cfg leaks ciphertext to creator |\n| 5 | HIGH | PGRService | Employee access blocked for confidential complaints |\n| 6 | HIGH | MDMSUtils | JSONPath injection via caseRelatedTo/code |\n| 7 | MEDIUM | ExtendedAttributesValidationService | Double MDMS round-trip can reject valid complaints |\n| 8 | MEDIUM | PGRService | Redundant encrypt→decrypt roundtrip on every create/update |\n| 9 | MEDIUM | PGRService | N sequential MDMS calls in search hot path |\n| 10 | LOW | EncryptionDecryptionService | Bypasses ServiceRequestRepository |\n\nGenerated by Claude Code · high-effort workflow review

backend/pgr-services/src/main/java/org/egov/pgr/service/EnrichmentService.java
```diff
@@ -149,6 +150,28 @@ public void enrichSearchRequest(RequestInfo requestInfo, RequestSearchCriteria c
     }
 
 
+    public void enrichUserContactDetails(ServiceRequest request) {
+        ExtendedAttributes ext = request.getService().getExtendedAttributes();
+        if (ext == null) return;
+
+        String email   = ext.getEmail();
```
vinothrallapalli-eGov, line 157: **[CRITICAL #2] `email` and `complainantAddress` are never encrypted**

`encrypt()` calls `ext.getField(fieldName)` which reads only from `dynamicFields`. But `email` and `complainantAddress` are declared as explicit `@JsonProperty`-bound fields on `ExtendedAttributes` — Jackson deserializes them directly into their named Java fields, never routing them through `@JsonAnySetter` into `dynamicFields`.

Result: even when an operator lists `"email"` in the `xSecurity` config, `encrypt()` finds `null` for it, skips it, and the plaintext value is forwarded here to User Service unencrypted.

`validate()` in `ExtendedAttributesValidationService` also uses `getField()` for checking required fields, so it rejects `email` as an unknown field even when it is supplied.

**Fix:** `getField()` / `putField()` need to be aware of named first-class fields, or `email`/`complainantAddress` must be moved into `dynamicFields` and accessed uniformly.

backend/pgr-services/src/main/java/org/egov/pgr/service/PGRService.java
```diff
@@ -214,7 +307,7 @@ public List<ServiceWrapper> plainSearch(RequestInfo requestInfo, RequestSearchCr
             return new ArrayList<>();
         }
 
-        userService.enrichUsers(serviceWrappers);
+        userService.enrichUsers(serviceWrappers, requestInfo);
```
vinothrallapalli-eGov, line 280: **[CRITICAL #3] `plainSearch()` returns raw ciphertext in `extendedAttributes`**

`search()` (lines ~175–210) has an explicit decrypt/mask loop for each complaint based on confidentiality and viewer roles. `plainSearch()` calls `enrichUsers` + `enrichWorkflow` and then returns — **no decrypt or mask is ever applied**.

Any caller of the plain-search endpoint (e.g. admin bulk-export, migration jobs) receives garbled ciphertext strings in all encrypted `extendedAttributes` fields.

**Fix:** Extract the decrypt/mask loop from `search()` into a shared helper and call it from `plainSearch()` as well.

backend/pgr-services/src/main/java/org/egov/pgr/service/PGRService.java
```diff
@@ -134,8 +169,44 @@ public List<ServiceWrapper> search(RequestInfo requestInfo, RequestSearchCriteri
...
+        for (ServiceWrapper wrapper : enrichedServiceWrappers) {
+            Service svc = wrapper.getService();
+            if (svc.getExtendedAttributes() == null) continue;
+            ComplaintTemplateTypeConfig cfg = configCache.get(svc.getExtendedAttributes().getCaseRelatedTo());
+            List<String> viewerRoles = (cfg != null && !CollectionUtils.isEmpty(cfg.getAllowedViewerRoles()))
+                    ? cfg.getAllowedViewerRoles()
+                    : List.of(ROLE_CONFIDENTIAL_VIEWER);
+            boolean isCreator = callerUuid != null && callerUuid.equals(svc.getAccountId());
+            if (svc.getExtendedAttributes().getIsConfidentialSafe() && !isCreator && !hasAnyRole(requestInfo, viewerRoles)) {
+                encryptionDecryptionService.maskAll(svc.getExtendedAttributes());
```
vinothrallapalli-eGov: **[HIGH #4] null-cfg silently leaks ciphertext to complaint creator**

When MDMS can no longer resolve a complaint's `caseRelatedTo` (e.g. the record was deactivated), `configCache.get()` returns `null`. The `isCreator` check at line ~202 passes for the complaint owner, so `decrypt(ext, null)` is called.

`decrypt()` has a null-cfg early-return guard and returns `ext` unchanged — meaning the creator receives **raw ciphertext strings** for all encrypted PII fields with no error or masked placeholder.

**Fix:** When `cfg == null` inside the creator branch, either throw a meaningful error or apply `maskAll()` rather than silently returning plaintext ciphertext.

backend/pgr-services/src/main/java/org/egov/pgr/service/PGRService.java
```diff
@@ -134,8 +169,44 @@ public List<ServiceWrapper> search(RequestInfo requestInfo, RequestSearchCriteri
...
+
+        // Decrypt and conditionally mask extendedAttributes for each result.
+        // Viewer roles come from MDMS allowedViewerRoles; fallback to hardcoded constant if not configured.
+        // All-or-nothing rule: confidential + no matching role → mask ALL fields + citizen PII, skip decryption.
+        // Creator self-access exception: complaint owner always sees plain text regardless of confidentiality.
+        String callerUuid = requestInfo.getUserInfo() != null ? requestInfo.getUserInfo().getUuid() : null;
+        for (ServiceWrapper wrapper : enrichedServiceWrappers) {
+            Service svc = wrapper.getService();
+            if (svc.getExtendedAttributes() == null) continue;
+            ComplaintTemplateTypeConfig cfg = configCache.get(svc.getExtendedAttributes().getCaseRelatedTo());
```
vinothrallapalli-eGov: **[HIGH #5] Employee access blocked for confidential complaints**

The `isCreator` check compares the caller's UUID against `accountId`. `accountId` is always the **citizen's UUID** — it is set in `EnrichmentService.enrichCreateRequest` only when the caller type is `CITIZEN`. An employee acting on a confidential complaint has a different UUID and is not `CONFIDENTIAL_COMPLAINT_VIEWER`, so the code falls through to `maskAll()`, hiding all extended attributes including fields the employee needs to resolve the complaint (e.g. location details).

**Fix:** Introduce an explicit `EMPLOYEE_CAN_VIEW_ASSIGNED` role check, or pass the `cfg.getAllowedViewerRoles()` list to employees assigned to the complaint.

backend/pgr-services/src/main/java/org/egov/pgr/util/MDMSUtils.java
```diff
@@ -170,4 +179,121 @@ public StringBuilder getMdmsSearchUrl() {
...
+        try {
+            MdmsCriteriaReq req = MdmsCriteriaReq.builder()
+                    .requestInfo(requestInfo)
+                    .mdmsCriteria(MdmsCriteria.builder()
+                            .tenantId(tenantId)
+                            .moduleDetails(List.of(ModuleDetail.builder()
+                                    .moduleName(MDMS_MODULE_NAME)
+                                    .masterDetails(List.of(MasterDetail.builder()
+                                            .name(MDMS_COMPLAINT_RELATED_TO_MAP)
+                                            .filter("$.[?(@.active==true && @.code=='" + code + "')]")
```
vinothrallapalli-eGov, line 223: **[HIGH #6] JSONPath injection via unsanitised `caseRelatedTo` / `code`**

User-supplied values are string-concatenated directly into JSONPath filter expressions, e.g.:

```java
"$.[?(@.active==true && @.caseRelatedTo=='" + caseRelatedTo + "')]"
```

A crafted input (e.g. containing `']]`) breaks the expression. The exception is swallowed, causing `isValidCaseRelatedTo` to return `false` (rejecting a valid complaint type) or `fetchComplaintTemplateTypeConfig` to load an unintended schema.

**Fix:** Validate `caseRelatedTo` / `code` against `[A-Za-z0-9_-]+` before use in JSONPath, or use the MDMS filter API's structured query parameters instead of free-form JSONPath strings.

backend/pgr-services/src/main/java/org/egov/pgr/service/ExtendedAttributesValidationService.java
```diff
@@ -0,0 +1,89 @@
...
+    @Autowired
+    public ExtendedAttributesValidationService(MDMSUtils mdmsUtils) {
+        this.mdmsUtils = mdmsUtils;
+    }
+
+    public void validate(ExtendedAttributes ext, ComplaintTemplateTypeConfig config,
+                         Service service, RequestInfo requestInfo, String tenantId) {
+        if (ext == null) return;
+
+        if (!mdmsUtils.isValidCaseRelatedTo(requestInfo, tenantId, ext.getCaseRelatedTo()))
```
vinothrallapalli-eGov: **[MEDIUM #7] Double MDMS round-trip can reject valid complaints**

`PGRService.create()` already called `fetchComplaintTemplateTypeConfig` (which queries MDMS) and confirmed `cfg != null` before calling `validate()`. But `validate()` unconditionally calls `isValidCaseRelatedTo` here, which fires its **own separate MDMS call**.

If that second call times out or returns an empty result due to a transient network hiccup, it returns `false` and throws `INVALID_CASE_RELATED_TO` — rejecting a complaint that had already passed config lookup 20ms earlier.

**Fix:** Pass the already-fetched `cfg` into `validate()` and skip the redundant `isValidCaseRelatedTo` check (the non-null `cfg` already proves `caseRelatedTo` is valid).

backend/pgr-services/src/main/java/org/egov/pgr/service/PGRService.java
```diff
@@ -93,10 +105,33 @@ public ServiceRequest create(ServiceRequest request) {
...
+		ComplaintTemplateTypeConfig cfg = null;
+		if (ext != null) {
+			if (ext.getIsConfidential() == null) ext.setIsConfidential(false);
+			cfg = mdmsUtils.fetchComplaintTemplateTypeConfig(
+					request.getRequestInfo(), tenantId, ext.getCaseRelatedTo());
+			if (cfg == null)
+				throw new CustomException("INVALID_CASE_RELATED_TO",
+						"No MDMS config found for caseRelatedTo: " + ext.getCaseRelatedTo());
+			extendedAttributesValidationService.validate(ext, cfg, service, request.getRequestInfo(), tenantId);
+			service.setExtendedAttributes(
```
vinothrallapalli-eGov, line 122: **[MEDIUM #8] Redundant `encrypt → decrypt` roundtrip on every create/update**

After `encrypt()` + `producer.push()`, the code immediately calls `decrypt()` on the now-encrypted attributes purely to return plaintext to the caller. This doubles enc-service round-trips on every write.

Worse: if enc-service has a transient failure on this second call, the caller receives an error **even though the complaint was already successfully persisted**.

**Fix:** Hold a reference to the original plain `ExtendedAttributes` before calling `encrypt()`, then restore it for the response instead of calling `decrypt()`:

```java
ExtendedAttributes plainExt = deepCopy(ext);  // or clone before encrypt
service.setExtendedAttributes(encrypt(ext, cfg, tenantId));
producer.push(...);
service.setExtendedAttributes(plainExt);       // no decrypt round-trip
```

backend/pgr-services/src/main/java/org/egov/pgr/service/PGRService.java
```diff
@@ -134,8 +169,44 @@ public List<ServiceWrapper> search(RequestInfo requestInfo, RequestSearchCriteri
...
+        String tenantIdForMdms = criteria.getTenantId() != null
+                ? criteria.getTenantId() : requestInfo.getUserInfo().getTenantId();
+        Set<String> categoryTypes = enrichedServiceWrappers.stream()
+                .map(w -> w.getService().getExtendedAttributes())
+                .filter(Objects::nonNull)
+                .map(ExtendedAttributes::getCaseRelatedTo)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toSet());
+        Map<String, ComplaintTemplateTypeConfig> configCache = new HashMap<>();
+        for (String cat : categoryTypes) {
```
vinothrallapalli-eGov: **[MEDIUM #9] N sequential synchronous MDMS calls in search hot path**

`search()` loops over distinct `caseRelatedTo` values and fires a separate blocking `fetchComplaintTemplateTypeConfig` MDMS call for each. A page with 10 complaint types blocks on 10 sequential round-trips before any result is returned.

Since each call fetches an independent type, they are fully parallelisable.

**Fix:** Issue the MDMS lookups in parallel (e.g. `CompletableFuture` + `allOf`), or batch them into a single MDMS request if the MDMS API supports multi-filter queries.

backend/pgr-services/src/main/java/org/egov/pgr/service/EncryptionDecryptionService.java
```diff
@@ -0,0 +1,173 @@
...
+        for (String v : values) {
+            Map<String, Object> r = new HashMap<>();
+            r.put("tenantId", tenantId);
+            r.put("type", "Normal");
+            r.put("value", v);
+            requests.add(r);
+        }
+        Map<String, Object> payload = new HashMap<>();
+        payload.put("encryptionRequests", requests);
+        JsonNode response = restTemplate.postForObject(url, payload, JsonNode.class);
```
vinothrallapalli-eGov, line 160: **[LOW #10] Bypasses `ServiceRequestRepository`, inconsistent error handling**

`callEncryptBatch` and `callDecryptBatch` call `restTemplate.postForObject(...)` directly. Every other external call in this service uses `ServiceRequestRepository.fetchResult`, which catches `HttpClientErrorException` and translates it to `ServiceCallException`.

Bypassing it means enc-service 4xx errors surface as raw `HttpClientErrorException` instead of `ServiceCallException`, producing a different exception type than all other external calls. This complicates uniform error handling in callers and makes monitoring inconsistent.

**Fix:** Route through `serviceRequestRepository.fetchResult(url, payload, JsonNode.class)` the same way other services make external calls.

pradeepkumarcm-egov (approved):

PR #985, opened 2026-06-30
fix(ui): normalize JPEG file extensions and MIME types on upload (#445)

### LINKED ISSUE #445: Nai Pepea:  In Edit Profile Screen, Upload Profile Image feature not working

dhruv-1001:
## Issue (#445)
**Edit Profile → "Upload Profile Image" does not work** for JPEG images — the user picks a JPEG, but the profile picture never updates.

## Root cause
Depending on the device/browser, the file picker hands over a JPEG with an **uppercase extension** (`.JPG`) or an **inconsistent MIME type** (`image/jpg`, or empty). The downstream filestore upload rejects/ignores those, so the profile image silently fails to save.

## Fix
Normalize every JPEG to a canonical form **before** upload — lowercase `.jpeg` filename + `image/jpeg` MIME — in both upload entry points:
- `packages/digit-ui-components/src/atoms/UploadImage.js`
- `packages/modules/core/.../ImageUpload/UploadDrawer.js`

Also reset the file input after the attempt so the same file can be re-selected if a retry is needed.

## Verification
On a live stack, uploading `photo.JPG` (uppercase) now sends `filename="photo.jpeg"` + `Content-Type: image/jpeg` to filestore and the profile image saves (verified via Playwright network interception).

Closes #445.


subhashini-egov (approved):
Please resolve the merge conflicts.

Issue #812, opened 2026-06-10
[Bug] UI/UX: CMS: Employee edit profile change password button is not clearly visible


Gurjeet-eGov:
# Description

UI/UX: CMS: Employee edit profile change password button is not clearly visible 

---

# Steps to Reproduce

1. Login to employee
2. Go to edit profile 
3. Check for change password button

---

# Expected Behavior

Change Password button shall be clearly visible

---

# Actual Behavior

Currently its just a plain text with white background. Its not clear if its clickable until hovered

<img width="1366" height="650" alt="Image" src="https://github.com/user-attachments/assets/ecb1370d-0d0c-42e9-883c-225df2325a5d" />
---

KDwevedi:
Fix raised in #970 — moves `@import tokens.css` to the top of the v2 `tailwind.css` so design-token CSS vars load and the Change Password button styling resolves (CSS drops an `@import` placed after `@tailwind`).

Gurjeet-eGov:
Resolved 

https://github.com/user-attachments/assets/b51d9f9d-f3a4-450a-9972-dde8448beb1e

PR #986, opened 2026-06-30
fix(ui): move @import tokens.css to top in tailwind.css (#812)

dhruv-1001:
## Issue (#812)
**Employee → Edit Profile: the "Change Password" button is not clearly visible.** It renders as plain text on a white background and doesn't look clickable until hovered.

## Root cause
That button is a `digit-ui-components-v2` component; its background/border/text colors come from CSS custom properties defined in `tokens.css`, pulled into `tailwind.css` via `@import "./tokens.css"`. But the `@import` was **not the first statement** in the file. Per the CSS spec, any `@import` that isn't at the very top is dropped — so the design tokens never loaded and the button fell back to unstyled (transparent background / plain text).

## Fix
Move `@import "./tokens.css"` to the **top** of `packages/digit-ui-components-v2/src/theme/tailwind.css` (and the built `public/vendor/tailwind.css`) so the tokens load and the Change Password button gets its proper button styling.

## Verification
Rebuilt digit-ui: the design-token custom properties load on the page and the Change Password button renders as a styled, clearly-clickable button instead of plain text. Adds `change-password-visibility-812.spec.ts`.

Closes #812.


KDwevedi (approved):
Approved. Verified the fix is strictly additive: rule-level diff of the rebuilt `public/vendor/tailwind.css` shows the only new rules are the `:root{--v2-*}` token block and the `.v2-scope` brand-bridge — nothing existing is altered or removed. On develop the token block was entirely absent (late `@import` dropped by postcss-import per CSS spec), so every v2 component using `var(--v2-*)` was rendering unstyled; this restores them library-wide. Blast radius is broken→styled, not a regression from any working state. Minor: the .gitignore additions are unrelated scope creep (and lack a trailing newline) — fine to leave.

PR #989, opened 2026-06-30
fix(pgr): enforce 1-hour complaint reopen deadline (UI + backend) (#925)

### LINKED ISSUE #925: [Bug] CMS: Employee: CSR user is able to reopen complaint even after the 1 hr reopen limit has lapsed on citizen ui
# Description

If a complaint has been rejected or resolved, there's a 1 hr time window within which a citizen or csr employee can reopen the complaint. Post that the reopen functionality gets hidden from citizen ui but the CSR is still able to reopen. Ideally it used to show a toast message

---

# Steps to Reproduce

1. Login with CSR
2. Open a rejected or resolved complaint 
3. Click on reopen (if it was resolved/rejected more than 1 hr ago) and submit 

---

# Expected Behavior

While reopening a complaint past deadline, the ui shall display a validation toast message in employee. For citizen the reopen button is already hidden

---

# Actual Behavior

https://github.com/user-attachments/assets/35a57775-0096-4035-885e-91bfa13c259a

dhruv-1001:
## Issue (#925)
A resolved/rejected complaint is meant to be reopenable only within a **1-hour** window. The citizen UI hides the reopen button after 1h, **but a CSR/employee can still reopen the complaint after the 1-hour limit has lapsed**, and no validation message is shown.

## Root cause
Two gaps:
1. The **employee UI** didn't block reopen past the deadline or surface a toast.
2. The **backend deadline was effectively disabled.** `ServiceRequestValidator.validateReOpen()` rejects REOPEN once `now - lastModifiedTime > pgr.complain.idle.time`, but `pgr.complain.idle.time` defaulted to `864000000` (**10 days**) — so reopen succeeded via the API long past 1 hour even when the UI tried to block it.

## Fix
- **Frontend** (employee): block REOPEN past the deadline and show a validation toast (`products/pgr/.../ComplaintDetails.js`, `PGRDetails.js`).
- **Backend**: `pgr.complain.idle.time` `864000000` (10 days) → `3600000` (1 hour) so `validateReOpen()` enforces the same window server-side; the UI guard is no longer bypassable via the API. Overridable per-deployment via `PGR_COMPLAIN_IDLE_TIME`.

## Verification
On a live stack, as an employee (skips the citizen-ownership branch), ageing a complaint's `lastModifiedTime`:
- resolved **120 min ago** → `400 INVALID_ACTION: "Complaint is closed"` (blocked) ✅
- resolved **30 min ago** → deadline check passes ✅

Closes #925.

## Summary by CodeRabbit

* **New Features**
  * Added configurable complaint reopening deadlines through tenant settings.
  * Reopening is now consistently allowed or blocked based on the configured deadline.
  * Added a clear error message when a complaint can no longer be reopened.

* **Bug Fixes**
  * Improved reopening validation using the complaint’s persisted details and last update time.
  * Added fallback handling when the configured deadline is unavailable.

* **Tests**
  * Added coverage for reopening complaints both within and beyond the configured deadline.


backend/pgr-services/src/main/resources/application.properties
```diff
@@ -85,7 +85,12 @@ egov.workflow.businessservice.search.path=/egov-workflow-v2/egov-wf/businessserv
 egov.workflow.processinstance.search.path=/egov-workflow-v2/egov-wf/process/_search
 
 #PGR Varaibles
-pgr.complain.idle.time=864000000
+# Reopen deadline (#925): a resolved/rejected complaint can be reopened only within this window.
+# validateReOpen() rejects REOPEN with "Complaint is closed" once now - lastModifiedTime exceeds it.
+# Was 864000000 (10 days), which let CSR/employees reopen long past the 1-hour window the citizen UI
+# (and issue #925) intend — the UI block alone was bypassable via API. 3600000 ms = 1 hour.
+# Overridable per-deployment via env PGR_COMPLAIN_IDLE_TIME.
+pgr.complain.idle.time=3600000
```
subhashini-egov: @dhruv-1001 I hope we're picking this from MDMS.

digit-ui-esbuild/packages/modules/pgr/src/pages/employee/ComplaintDetails.js
```diff
@@ -351,8 +359,17 @@ export const ComplaintDetails = (props) => {
         setDisplayMenu(false);
         break;
       case "REOPEN":
-        setPopup(true);
-        setDisplayMenu(false);
+        const lastModifiedTime = complaintDetails?.service?.auditDetails?.lastModifiedTime;
+        const ComplainMaxIdleTime = 3600000; // 1 hour in ms
```
subhashini-egov: @dhruv-1001 Ensure you are picking this up from MDMS or introduce a new master for this (if needed). Check with @Hari-egov or @vinothrallapalli-eGov . Let's not hard code this.

digit-ui-esbuild/products/pgr/src/pages/employee/PGRDetails.js
```diff
@@ -727,6 +727,20 @@ const PGRDetails = () => {
               label={t("ES_COMMON_TAKE_ACTION")}
               onOptionSelect={(selected) => {
                 console.log("*** Log ===> selected", selected);
+                if (selected.action === "REOPEN") {
+                  const lastModifiedTime = pgrData?.ServiceWrappers?.[0]?.service?.auditDetails?.lastModifiedTime;
+                  const ComplainMaxIdleTime = 3600000; // 1 hour in ms
```
subhashini-egov: Same comment as before. We should not hard code this.

tests/integration-tests/tests/employee/reopen-limit-925.spec.ts
```diff
@@ -0,0 +1,179 @@
+import { test, expect } from "@playwright/test";
+import { loginViaApi } from "../utils/auth";
+import { BASE_URL } from "../utils/env";
```
subhashini-egov: Make sure the test is also reading configuration, not hard coding.

subhashini-egov (changes requested):
We should make the re-open time limit configurable and not hard code to an hour.

subhashini-egov:
@dhruv-1001 - can you please close this?

Hari-egov:
hey @dhruv-1001 can you check the suggested changes by @subhashini-egov.

Hari-egov (changes requested):
hey @dhruv-1001 can you check the comments and make the required changes

backend/pgr-services/src/main/resources/application.properties
dhruv-1001: Good catch, and it turned out this property was a red herring — it is now reverted, and this PR no longer touches the backend at all.

The 1-hour window users actually saw never came from `pgr.complain.idle.time` (which was 10 days). It came from a hardcoded `ComplainMaxIdleTime = 3600000` default parameter in the citizen timeline components (`resolved.js` / `rejected.js`), which took effect because the MDMS lookup meant to supply that prop had been commented out in citizen `ComplaintDetails.js`. So `RAINMAKER-PGR.UIConstants.REOPENSLA` — already seeded for every tenant, schema-defined and editable in the configurator — was being ignored entirely, which is also why changing it had no effect (see @Gurjeet-eGov on #925).

The fix now reads REOPENSLA and makes it the single source of truth for the citizen and employee flows alike. No hardcoded window anywhere, and no backend change needed.

digit-ui-esbuild/packages/modules/pgr/src/pages/employee/ComplaintDetails.js
dhruv-1001: Done — and no new master turned out to be needed. `RAINMAKER-PGR.UIConstants.REOPENSLA` already exists for exactly this: seeded for every tenant, schema-defined, and exposed in the configurator as "PGR UI Constants". Nothing was reading it.

This page now reads it through a new `useReopenWindow(tenantId)` hook (mirroring the existing `useMapConfig` pattern in this module) rather than any hardcoded duration. The same hook feeds the citizen timeline, so the employee guard and the citizen button can no longer disagree about the deadline.

Note the window is now whatever REOPENSLA says — currently seeded at `432000000` (5 days), not 1 hour. If 1 hour is the intended rule, that is a seed-value change rather than a code change; flagging it so QA is not surprised.

digit-ui-esbuild/products/pgr/src/pages/employee/PGRDetails.js
dhruv-1001: Same fix here — this page now reads the window from `RAINMAKER-PGR.UIConstants.REOPENSLA` via the shared `useReopenWindow(tenantId)` hook. No hardcoded duration remains in either employee page or in the citizen timeline components.

tests/integration-tests/tests/employee/reopen-limit-925.spec.ts
dhruv-1001: The spec stubs `RAINMAKER-PGR.UIConstants.REOPENSLA` and derives every timestamp and expectation from the stubbed value, so it asserts the contract ("the UI honours the configured window") rather than any particular duration.

It deliberately stubs a 6h window rather than 1h, which is what pins the regression: the old hardcoded `3600000` build fails the in-window case.

backend/pgr-services/src/main/java/org/egov/pgr/util/MDMSUtils.java
```diff
@@ -150,6 +168,83 @@ public List<Object> getNotificationTemplates(String tenantId) {
...
+            if (millis <= 0) {
+                log.warn("Ignoring non-positive REOPENSLA ({}) for tenant {}", millis, tenantId);
+                return null;
+            }
+            return millis;
+        } catch (Exception e) {
+            log.error("Failed to read REOPENSLA from RAINMAKER-PGR.UIConstants for tenant {}", tenantId, e);
+            return null;
+        }
+    }
```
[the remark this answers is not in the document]
dhruv-1001: Moot now — `MDMSUtils.java` is no longer touched by this PR. The backend changes have been reverted and the fix is frontend-only, so this code path does not exist on the branch anymore.

dhruv-1001:
@claude review.

digit-ui-esbuild/packages/modules/pgr/src/hooks/useReopenWindow.js
```diff
@@ -0,0 +1,36 @@
...
+//
+// Until #925 the window was a hardcoded `ComplainMaxIdleTime = 3600000` default parameter in the
+// reopen timeline instances, which silently won over REOPENSLA because the MDMS lookup that was
+// meant to supply it had been commented out — so every tenant got 1 hour regardless of config.
+const useReopenWindow = (tenantId) => {
+  const { data } = Digit.Hooks.useCustomMDMS(
+    tenantId,
+    "RAINMAKER-PGR",
+    [{ name: "UIConstants" }],
+    {
```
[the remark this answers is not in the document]
dhruv-1001: Acknowledged, and agreed it should not block — closing as known/intentional.

This is a deliberate scoping decision rather than an oversight. #925 reports a UI bug ("the ui shall display a validation toast message in employee"), and REOPENSLA is stricter than the server’s existing `pgr.complain.idle.time` backstop, so the server rejects nothing the UI allows and no backend change is needed to fix the reported issue. As the comment notes, this PR narrows the gap rather than widening it (previously 1h in the UI vs 10d server-side; now ~5d vs 10d).

Server-side enforcement of REOPENSLA — and anchoring `validateReOpen()` to the persisted record rather than the client-supplied request body, which is the more interesting hole, since the current check reads `lastModifiedTime` straight from the payload — was prototyped on this branch and deliberately reverted to keep this PR reviewable. It will be raised as its own issue against pgr-services.

utilities/default-data-handler/src/main/resources/localisations/default/rainmaker-pgr.json
```diff
@@ -773,6 +773,12 @@
         "module": "rainmaker-pgr",
         "locale": "default"
     },
+    {
+        "code": "CS_CANNOT_REOPEN_COMPLAINT_PAST_DEADLINE",
+        "message": "The window for reopening this complaint has closed",
+        "module": "rainmaker-pgr",
+        "locale": "default"
+    },
```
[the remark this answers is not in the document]
dhruv-1001: Good catch — this was real, fixed in ff5006d7.

Verified both halves before acting rather than taking it on trust:
- The key was indeed only in `default/rainmaker-pgr.json`, absent from `en_IN/`.
- `translations/index.js` sets `fallbackLng: Digit.Utils.getDefaultLanguage()`, which resolves to `<locale>_<region>` (`en_IN`) and never the literal `"default"` — so `default` is a separately-seeded locale record, not a fallback, exactly as described.
- The `CS_COMPLAINT_DETAILS_PIN_LOCATION` precedent confirms the convention of carrying a key in both files.

So on a standard en_IN deployment the toast would have rendered the raw key. That is worth fixing rather than waving through, since the toast *is* the deliverable of #925 ("the ui shall display a validation toast message in employee"). Seeded the same entry into `en_IN/rainmaker-pgr.json`; JSON validated.

One correction: there is no `hi_IN` file in this repo — only `default` and `en_IN` — so that part of the suggestion is moot.

Leaving the spec’s `REOPEN_BLOCKED_MSG` regex permissive (raw key *or* message) on purpose: it runs against envs where default-data-handler may not have seeded localisation, so tightening it would make the test fail for reasons unrelated to the behaviour under test.

subhashini-egov (approved):

Hari-egov (approved):
Approved
cc:  @dhruv-1001 and @subhashini-egov

Issue #920, opened 2026-06-23
[Bug] CMS: Citizen: Map pin selection keeps changing automatically and does not auto populate pincode


Gurjeet-eGov:
# Description

Map pin selection keeps changing automatically (check screen recording) and does not auto populate pincode

---

# Steps to Reproduce

1. Login to citizen ui
2. Go to create complaint 
3. Select a pin on map 
4. Click next and check locality and pincode fields to be auto-filled

---

# Expected Behavior

- Pincode shall be auto filled based on pin selected on map
- Map pin shall not change automatically

---

# Actual Behavior

https://github.com/user-attachments/assets/791cd608-a124-4028-9000-5f9efc621741

KDwevedi:
Verified live on naipepea (esbuild build, the deployed CMS frontend). Summary of the three reported symptoms:

- **Pincode auto-fill — working ✅.** Dropping a pin populates Postal Code (e.g. `00507`) on the Location Details step. The map-pin → `postalCode` bridge is live in `CreatePGRFlowV2`. No change needed.
- **Locality/County auto-fill — root-caused to CONFIG, not frontend code.** The auto-fill code (`BoundaryComponent.findWardPath`, CCRS#491) is present and correct; it silently no-ops because `ke.nairobi` has **no ward geometry seeded** and **`MAP_TENANT` is unset**, so the map falls back to static wards whose codes don't match the seeded `NAIROBI_CITY_*` hierarchy. Split out as **#1140** (config + surfacing).
- **Pin "keeps changing" — minor stale-geocode race** (a superseded reverse-geocode can overwrite a newer pin). Tracked for a small hardening fix in `GeoLocations.fetchAddress`; not blocking.

Note: an earlier attempt (PR #1000) targeted the legacy `frontend/micro-ui` tree and was closed — the deployed CMS serves `digit-ui-esbuild`, where the pincode behavior is already correct.

Moving to **To Review**: pincode auto-fill is live; locality depends on the config/data work in **#1140**.

Gurjeet-eGov:
Resolved 

https://github.com/user-attachments/assets/5045f17a-0d2d-4ccc-8be2-9b6bae5912fb

PR #1000, opened 2026-07-01
fix(pgr): stop citizen map pin jump + auto-fill pincode (partial #920)

KDwevedi:
Partially addresses #920 (CMS citizen create-complaint map). This PR fixes **two of the three** reported defects; the third (locality auto-fill) is a separate regression — see below.

### 1. Pin keeps changing on its own
`GeoLocations` had two competing sources of truth for the marker: local state (set instantly on click) and a `useEffect` that re-syncs from the `formData` prop — which is `FormComposerV2`'s react-hook-form `watch()` snapshot, a **new object identity on every form change**. Two races resulted:
- `updateLocation` committed coordinates to the form only **after** the async Nominatim reverse-geocode, so any re-render in that window re-applied the *stale* form value and snapped the marker back → **commit lat/lng synchronously** on selection + make the sync effect **idempotent** (only move when coords actually differ).
- Reverse-geocode responses are **unordered**: the mount-time lookup for `MAP_CENTER`, or an earlier click, could resolve *after* a newer pin and overwrite it → added a **monotonic request id**; a superseded response is dropped.

### 2. Pincode did not auto-fill
The pincode was bridged into `postalCode` by **mutating `formData` at render time** and passing it as RHF `defaultValues`, which RHF ignores after mount (single form instance, never reset between steps) → **bridge via `setValue()`** in `onFormValueChange`, applied **once per new map pincode** (via a ref) so a value the user typed by hand isn't clobbered. Also removed the hardcoded `value: "560001"` default that masked it, and relaxed the India-only 6-digit pincode extraction + field validation to 4–6 digits so non-India tenants (e.g. Kenya) aren't blocked.

## What this does NOT fix — locality auto-fill (separate regression)
The issue also asks for **locality** to auto-fill. That broke in commit `1b2b7d224` ("Citizen Side Boundary changes", 2026-01-22), which replaced the custom `SelectAddress` component — which **filtered the city/locality dropdowns by the entered pincode** — with a generic manual `type: "boundary"` admin-tree picker (`boundaryComponent`) that ignores pincode entirely. Restoring auto-fill is a design + MDMS-data decision (pincode↔boundary mapping, which many tenants lack), not a mechanical fix, so it is intentionally left for a follow-up. **Do not auto-close #920 on merge.**

## Files
- `components/GeoLocations.js` — sync commit + idempotent marker + request-ordering guard + locale-agnostic pincode extraction
- `pages/citizen/Create/FormExplorer.js` — `setValue` pincode bridge (once per pin); remove dead render-time mutation
- `pages/citizen/Create/steps-config/locationDetails.js` — drop hardcoded `560001`; relax postalCode pattern to 4–6 digits

## Testing
All changed files parse (JSX-aware). Diagnosis independently reviewed with a second tool (codex), which caught the clobber + async-ordering gaps now fixed, and confirmed the locality regression. Manual UI verification (pin select → Next → postalCode populated, pin stable under rapid clicks) still needed by QA.

### DIFF
--- a/frontend/micro-ui/web/micro-ui-internals/packages/modules/pgr/src/components/GeoLocations.js
+++ b/frontend/micro-ui/web/micro-ui-internals/packages/modules/pgr/src/components/GeoLocations.js
@@ -55,6 +55,11 @@ const GeoLocations = ({ t, config, onSelect, formData }) => {
   const mapRef = useRef(null);
   const searchInputRef = useRef(null);
   const hasInitialized = useRef(false);
+  // Monotonic id for reverse-geocode requests. Reverse geocoding is async and
+  // unordered: the mount-time lookup for MAP_CENTER, or an earlier click, can
+  // resolve *after* a newer selection and overwrite it (the pin "jumps back").
+  // Each fetch captures an id and only commits its result if it is still latest.
+  const geocodeSeqRef = useRef(0);
 
   // Initialize with India center location on first load or from Session Storage
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


KDwevedi:
Closing — this PR targets the wrong tree and is redundant.

**Wrong tree:** it edits `frontend/micro-ui` (the legacy `digit-ui` webpack container per `build/NIGHTLY-BUILDS.md:147-150`). The CMS deployment #920 was filed against serves the **`digit-ui-esbuild`** static bundle (`digit-ui-esbuild/products/pgr`), not this tree. So these edits never run.

**Redundant:** verified against the live naipepea (esbuild) build — pincode **already auto-fills** from the map pin (Postal Code populated `00507`), so the pincode half of this PR is already solved on the deployed code.

**What #920 actually is (verified live):**
- Pincode autofill — already works ✅
- Locality autofill — fails because the pin-resolved ward doesn't match the tenant's seeded `CMS-BOUNDARY` hierarchy (the CCRS#491 auto-fill code is deployed and correct; it silently no-ops on boundary-data drift). Fix is **boundary seeding**, not frontend code.
- Pin UX — clicks on a ward `GeoJSON` polygon are swallowed by Leaflet's interactive vector layer, so the pin only drops on gaps. Real code bug, to be fixed in `digit-ui-esbuild/products/pgr/GeoLocations.js`.

Follow-up work will land on the esbuild tree + boundary data. Do not auto-close #920.

PR #1044, opened 2026-07-03
fix(local-setup): fix egov-user image pin and mobile-regex derivation for new tenant deploys

vinothrallapalli-eGov:
## Summary

Four related issues found and fixed while onboarding a new tenant (`tamilnadu`) end-to-end via `./deploy.sh tamilnadu`, and via the Configurator UI's onboarding wizard.

### 1. `egov-user` image pin was silently wrong (`docker-compose.egov-digit.yaml`)
The `egov-user` image line was missing its registry host, so it silently resolved to an unpatched Docker Hub tag lacking the tenant-aware encryption fix (Digit-Core#1044). New tenants' ADMIN users could be created but never log in again ("Invalid login credentials" despite the user row existing). Pinned to `registry.preview.egov.theflywheel.in/egovio/egov-user:1044-preview` instead — the only tag on that registry carrying the fix. This is a shared/base-compose change, affecting the one `egov-user` container serving all tenants, not just `tamilnadu`.

### 2. Mobile-regex derivation missing (`local-setup/ansible/playbook-deploy.yml`, new `filter_plugins/mobile.py`)
`playbook-deploy.yml` expected separate `mobileNumberLength`/`mobileNumberAllowedStartingCharacters`/`mobileNumberErrorMessage` host_vars fields, but the current tenant config schema only provides `countryCode`/`mobileNumberRegex`. Added `filter_plugins/mobile.py` (`derive_valid_mobile`) to derive length + allowed leading digits directly from the regex, mirroring digit-mcp's own `deriveValidMobile()`. Wired into the HRMS `INTERNAL_USER` seed task and the two MCP `/v1/tenant/bootstrap` calls.

### 3. New tenants never registered with `egov-enc-service` (`digit-mcp`, `configurator`)
`egov-enc-service` discovers tenants via an MDMS search scoped to its own `STATE_LEVEL_TENANT_ID` env var — a brand-new tenant root is invisible to it until explicitly registered, so any encrypt/decrypt call for that tenant (e.g. creating its ADMIN user) fails with `"Tenant Id not found"`, surfaced to callers as a generic `"Unknown error occurred in encryption process"`.

Three call sites hit this because none of them called `egov-enc-service`'s `POST /crypto/v1/_generatekey` (the same call `playbook-deploy.yml` already makes before its own MCP-driven bootstrap, for the same reason — see lines ~1930-1973):

- **digit-mcp's `tenant_bootstrap` tool** (`src/tools/mdms-tenant.ts`) — added `digitApi.generateEncKey()` (mirroring the existing `encryptData`/`decryptData` raw-fetch pattern in `digit-api.ts`) and call it first, before schema/data copy or ADMIN provisioning.
- **digit-mcp's `city_setup` tool** (`src/tools/mdms-tenant.ts`) — same fix. `city_setup` provisions its own dual-scoped ADMIN user for a *distinct* city tenantId, which needs its own enc-service key independent of the root's.
- **configurator's onboarding wizard** (`src/api/services/tenantBootstrap.ts`) — this is a separate, duplicated client-side reimplementation of the same bootstrap flow (does not call digit-mcp), so it needed the identical fix independently. Added `registerEncKey()` as Step 0 of `bootstrapStateRoot()`.

### 4. Fixed a schema regression merged in from #1042 (`local-setup/ansible/playbook-deploy.yml`)
While rebasing on `develop`, picked up #1042's new "re-provision ADMIN with enc key (`user_only`)" tasks, which referenced `core_mobile_configs.mobileNumberPattern`/`.mobilePrefix` with no `default()` fallback — the old schema this branch already migrated away from (see #2 above). Any host_vars using the current `countryCode`/`mobileNumberRegex` schema (e.g. `tamilnadu.yml`) would hit the same undefined-variable crash fixed in #2. Aligned to `mobileNumberRegex`/`countryCode` with defaults, matching the other two calls in the same file.

### 5. `user_only` re-key silently no-op'd on the exact case it exists to fix (`digit-mcp`)
While verifying #4, found that the merged-in `user_only` re-provisioning path (from #1042) always hit its "unhandled exception" branch: it detected duplicates by string-matching `err.message` for "duplicate", but egov-user wraps `DuplicateUserNameException` behind a generic message ("An unhandled exception occurred on the server") — the real exception type only survives in the error's `code` field, which was never checked. Since the user *always* already exists at this point (that's the whole premise of "post-bootstrap re-provision"), this branch fired on every call and surfaced as `admin_user_provision_error`.

Deeper problem: even when duplicate detection worked, the old code just declared success without updating anything — the stale, wrongly-encrypted user row this endpoint exists to fix was left untouched. Replaced create-then-catch-duplicate with a direct search-then-update: fetch the existing user and update it in place (re-encrypting `password`/`mobileNumber` under whichever key is currently active), falling back to `userCreate` only if no user is found.

## Test plan
- [x] Ran `./deploy.sh tamilnadu` to completion (`PLAY RECAP ... failed=0`)
- [x] Verified `ADMIN` login on the newly bootstrapped `tamilnadu` tenant returns `HTTP 200` with a valid token (previously failed with "Invalid login credentials")
- [x] Confirmed no unhealthy/restarting containers after the `egov-user` image switch
- [x] Ran digit-mcp's `tenant_bootstrap` against a fresh test tenant end-to-end (rebuilt + hot-swapped into the running container) and confirmed its ADMIN could log in immediately, no manual `_generatekey` call needed
- [x] Ran digit-mcp's `city_setup` (via `/v1/tenant/city`) against a fresh city tenant end-to-end, confirmed dual-scoped ADMIN provisioned and login succeeded
- [x] Rebuilt and redeployed the configurator SPA, confirmed the same `_generatekey` call succeeds through Kong with a real session token
- [x] After merging `develop` (#1042), rebuilt digit-mcp and re-verified full bootstrap → `user_only` re-key → login sequence still returns `HTTP 200`
- [x] Fixed and re-verified the `user_only` unhandled-exception bug (#5): `admin_user_provisioned: true` now, instead of the prior error; also verified the no-existing-user fallback path still works
- [ ] Re-verify login still works for pre-existing tenants (pg, ke, statea, ...) after they pick up the new shared `egov-user` image on next redeploy — this is a base-compose change, not tenant-scoped
- [ ] Full click-through of the Configurator onboarding wizard for a brand-new tenant end-to-end (verified the previously-failing step directly; didn't re-run the full multi-phase wizard UI)

## Summary by CodeRabbit

* **New Features**
  * Added tenant encryption key generation as part of tenant bootstrap.
  * Updated mobile-number validation to use `countryCode` and `mobileNumberRegex` (regex-first) across the UI and setup configs.
  * Added local setup support to derive valid sample mobile numbers from the configured rules.
* **Bug Fixes**
  * Improved admin user provisioning to update existing users or create them when absent.
  * Bootstrap now continues even if encryption-key setup fails.
* **Documentation**
  * Updated Ansible/compose setup examples and inventory defaults to the new mobile-number configuration format.


KDwevedi (commented):
## Review: coordinate the mobile-schema migration before this lands

The individual fixes here are sound in mechanism — the enc-key registration, the search-then-update re-key, and the image pin all do what they say. My concern is **one change**: switching the playbook to read the new mobile fields.

### The gap

This PR makes `playbook-deploy.yml` read the **new** field names — `mobileNumberRegex` and `countryCode` — but nothing else in the repo uses those names yet. Every checked-in config still uses the **old** names (`mobileNumberPattern`, `mobilePrefix`, `mobileNumberLength`, `mobileNumberAllowedStartingCharacters`): `_example.yml`, `bomet.yml.example`, `maputo.yml.example`, the README, and the real (gitignored) host_vars on each box.

Because the new lookups have `| default(...)`, the mismatch doesn't error — it **silently falls back to the India defaults**. On any existing tenant (all of them today), that means:

- **INTERNAL_USER seed** derives a 10-digit India number (`6000000000`) instead of a valid one. On a 9-digit tenant (Kenya `^[17][0-9]{8}$`, Mozambique `^8[0-9]{8}$`) the `1044-preview` image rejects it → **HRMS crash-loops** — the exact failure this seed task exists to prevent.
- **SPA globalConfigs** loses the dial prefix (`countryCode` undefined) → citizen login falls back to +91.
- **Fresh non-10-digit tenants** get an unsatisfiable rule: `mobile_length` was dropped, so the synthesized UserValidation ends up `minLength=maxLength=10` while the pattern wants 9.

bomet's nightly cron redeploys from develop, so this reaches the fleet automatically ~24h after merge — with no human action. Our two live reference installs (bomet, maputo) are both 9-digit, i.e. exactly the failing case.

### This is bigger than a rename — the layers aren't migrated together

Moving to the new schema touches five layers that all have to agree:

1. Real host_vars (per box) + the examples
2. `playbook-deploy.yml` (this PR) — but it also **dropped** `mobile_length` / `allowedStartingCharacters` / `errorMessage`
3. `globalConfigs.js.j2` → the **SPA code** that reads `CORE_MOBILE_CONFIGS.mobilePrefix`
4. Seeded MDMS data — and note the synth is **create-if-absent**, so existing tenants' rules are **frozen**; a redeploy won't fix them, they need a manual/forced update
5. The **egov-user image** and which master it reads

That last layer is the real coupling: **ChakshuGautam/Digit-Core#2** is the backend PR that moves egov-user onto the flat `countryCode`/`mobileNumberRegex` schema (upstream #967) and to a **new master** — `common-masters.MobileNumberValidation`. That PR is still **open / not merged**, ships a DB column migration, and its own notes say existing tenants must be backfilled. Meanwhile CCRS still seeds `common-masters.UserValidation` — `MobileNumberValidation` appears nowhere in this repo. So the frontend/config half is being renamed here while the backend half hasn't shipped.

### Two ways forward

**Option A — land it as a coordinated set.** Merge Digit-Core#2, build and pin the egov-user image, update tenant_bootstrap to seed the new master, migrate all host_vars (real + examples) and the SPA consumer, and backfill existing tenants' seeded rules — all together, verified on the ovh bomet repro against a pre-existing DB. Correct end state, but it's a real cross-repo rollout, not a single PR.

**Option B — make this PR backward-compatible so it can land now.** Accept both names in the playbook and keep length/error passthrough:
```jinja
mobile_regex:  "{{ core_mobile_configs.mobileNumberRegex | default(core_mobile_configs.mobileNumberPattern) | default('^[6-9][0-9]{9}$') }}"
mobile_prefix: "{{ core_mobile_configs.countryCode       | default(core_mobile_configs.mobilePrefix)        | default(omit) }}"
```
Then old-name tenants keep working untouched, new-name tenants work too, and the full schema switch (Option A) becomes a separate, deliberate rollout instead of riding in on a nightly cron.

vinothrallapalli-eGov:
Will go Ahead with Option A

vinothrallapalli-eGov:
Thanks for the thorough review, @KDwevedi — went through the actual code before applying this, and the picture turned out better than the review suggests in one respect, but there was a real bug that wasn't in the original write-up either. Went with **Option A**.

**The "MobileNumberValidation appears nowhere in this repo" claim is stale.** PR #1022 (`fix/877-mobile-validation-schema-update`) already migrated `digit-mcp` (`tenant_bootstrap`/`city_setup`), the `common-masters.MobileNumberValidation` schema + seed data in `default-data-handler`, `configurator`, and all three frontends (esbuild, v2, and the PGR product hook) to the flat `countryCode`/`mobileNumberRegex` schema, with length/leading-digit derivation done purely from the regex (`computeMobileLengths`, `deriveValidMobile`). So most of the cross-repo "Option A" surface was already landed before this PR — the real gap was narrower than described.

**What was genuinely still broken, now fixed (pushed in adaa7c0aa):**
- `local-setup/ansible/inventory/host_vars/{_example.yml, bomet.yml.example, maputo.yml.example}` and `README.md` still declared the retired field names (`mobilePrefix`, `mobileNumberPattern`, `mobileNumberLength`, `mobileNumberAllowedStartingCharacters`, `mobileNumberErrorMessage`). Since this PR's `playbook-deploy.yml` reads the new names with `| default(...)`, that's exactly the silent-fallback-to-India bug you flagged — migrated all of them.
- Added a preflight assertion in `playbook-deploy.yml` that fails the deploy with an actionable message if `core_mobile_configs` still carries legacy keys without the new ones. The real (gitignored) `bomet.yml`/`maputo.yml` on the live boxes aren't in this repo, so this is the backstop: it forces whoever holds SSH access to migrate those host_vars before their next deploy succeeds, instead of silently masking the mismatch.
- Found independently (not flagged in the review) that `digit-ui-esbuild`'s citizen Edit Profile page (`UserProfile.js`) read only `mobileNumberPattern` with no fallback to `mobileNumberRegex` — so on Tamil Nadu or any already-migrated tenant, profile-mobile validation was silently degrading to a generic 6-15 digit pattern and maxLength 15.
- Corrected a stale example in `ansible/nairobi-mdms/v2.11-user-service-upgrade.md` that told operators to override `CORE_MOBILE_CONFIGS` with the retired field names — since that override takes priority over MDMS, following it as written would have actively broken validation rather than just no-op'd.

**Not in scope of this push:** the `Digit-Core#2` / egov-user image work is being handled separately, and I didn't touch the live `bomet`/`maputo` boxes — those still need their real host_vars migrated before the next deploy, which the new preflight guard will now enforce rather than let slide.

KDwevedi (approved):
Reviewed. The five fixes are sound and well-scoped:

1. **egov-user image pin** — correct fix for the silent Docker Hub resolution. Note the actual diff pins `registry.preview.egov.theflywheel.in/egovio/egov-user:mobilevalidation-jdk8-4984479` (freshly published, carries MobileNumberValidation + NPE guard + security bumps), not `1044-preview` as the body text says — worth updating the body to match. This image is the right choice given this same PR migrates config to the flat `MobileNumberValidation` schema, which `1044-preview` couldn't read.
2. **mobile-regex derivation** (`filter_plugins/mobile.py`) — mirrors digit-mcp's `deriveValidMobile()`, correctly replaces the removed length/prefix host_vars.
3. **enc-service `_generatekey` registration** in mcp/configurator/playbook — correct root-cause fix for new-tenant 'Tenant Id not found'; matches the existing playbook pattern.
4. **#1042 schema regression** aligned to `mobileNumberRegex`/`countryCode` with defaults — good catch.
5. **`user_only` re-key** search-then-update instead of create-catch-duplicate — fixes the code-field vs message mismatch and the silent no-op.

CI `test`/`tilt-test` are red but pre-existing and repo-wide (same two checks fail on unrelated #1065; failure is a JS harness TypeError, not introduced here). Base-compose change → all develop-deployed tenants pick up the new shared egov-user image on next redeploy; bomet is unaffected (its /opt/ccrs fork pins separately).

Issue #1023, opened 2026-07-02
[Bug] Image ↔ DB-migration version skew: floating tags on egov-user/db-migrations can reintroduce BadSqlGrammar on login


KDwevedi:
## Problem

`egov-user` and the `db-migrations` container both track **floating/mutable image tags** with nothing enforcing that they carry the **same schema version**. When the service image gets ahead of the migration bundle, the service queries `eg_user` columns the bundle never created and every request throws `BadSqlGrammarException`.

This is the structural cause behind #992 ("Bad SQL Error at the point of login") — it took down **login + citizen registration on all tenants** (ke, ke.etoebeta, ke.bomet) on Bomet, because the error is on both `INSERT into eg_user (...)` and `SELECT userdata.title, ...`.

## Root cause

- Unified migrations run in one container `registry.preview.egov.theflywheel.in/tilt-demo-db-migrations:**latest**` (introduced in #959), and per-service `SPRING_FLYWAY_ENABLED=false` — this part is by design and works.
- BUT both the migration bundle (`:latest`) and the service images float. There is **no lockstep** guaranteeing `egov-user`'s expected schema == the schema the `:latest` bundle applies.
- Result: an image bump on `egov-user` (or a lagging `db-migrations:latest`) silently drifts the two apart → `BadSqlGrammar` at runtime. Migrations *are* running; they're just running a **different schema version** than the floating service image expects.

## Current mitigation (not a real fix)

Images were pinned to `master-e22c7c5` — `egov-user`, `user-otp`, `egov-notification-sms` — in `be79ce809` (PR #877, reached develop bundled in #963). This is a **manual alignment** that holds today but will drift again the moment either side moves, because `db-migrations` is still `:latest`.

## Evidence (live Bomet, verified)

- `digit-db-migrations-1` exited 0; egov-user block: *"Successfully validated 28 migrations, current version `20211029175430`, schema up to date."*
- `eg_user` has all 34 columns (`title`, `alternatemobilenumber`, `signature`, `dob`, `password` present).
- `digit-egov-user-1` (`egovio/egov-user:master-e22c7c5`) healthy, **0 `BadSqlGrammar` in 24h** since the pin.

## Proposed fix

1. Pin `db-migrations` to a **fixed digest/tag** that matches the pinned service image set (retire `:latest` for it).
2. Document the **image ↔ migration lockstep** requirement in the deploy notes (any service image bump must ship with / verify against a matching migration bundle version).
[...]

Issue #1020, opened 2026-07-02
PGR Inbox Visibility V1 — My/All complaint tabs


KDwevedi:
Part of the PGR Inbox Visibility epic (V1 → V3). Design: `CCRS/VISIBILITY-DESIGN.md`. PRD: CMS_Escalation_PRD → "visibility v1" tab.

## V1 scope (from the PRD)

Replace the current `assignedToMe` filter-card radio with a **two-tab inbox**. Visibility is **role/state + jurisdiction** based (no per-user assignee yet — that's V2).

### UI (per the reference image)
- Two tabs above the inbox card:
  - **"My Complaints (##)"** — static label; `(##)` = count of **unopened** complaints; **red dot** next to the number, vertically centred.
  - **"All Complaints (##)"** — static label; `(##)` = count of complaints **newly added since the tab was last opened**; red dot next to the number.
- The table stays the same (no column changes).
- **Filter card**: remove the assigned-to control. The filter card applies to **both** tabs. If a filter is applied and the tab is switched, **the filter resets**.

### Tab logic (V1)
`hasReportees` = user has ≥1 direct reportee via the `reportingTo` chain. Every tab is additionally scoped to the acting user's **jurisdiction** (`locality ∈ myLocalities`).

| Role | My Complaints | All Complaints |
|---|---|---|
| no reportees | complaints in the queue states of my role(s) | same (= my role) |
| has reportees (N+1, configurable; default N=1) | complaints in the queue states of my role(s) | complaints in the queue states of the roles **reporting to me** |

Where "queue states of role R" = states `S` in the PGR workflow BusinessService for which some `action ∈ S.actions` lists role R (`statesFor(R)`), matched against `applicationStatus`.

---

### Step 1 — FE experience first; BE minimal & unoptimised
Goal: ship the complete tab **experience** end-to-end, accepting a rough backend.

- **Frontend (full):**
  - Render the `Tab` atom (`@egovernments/digit-ui-components`) in the `PGRInbox.js` header slot, above `<InboxSearchComposer/>`; `My`/`All` tabs with count + red-dot indicator.
  - Remove the `assignedToMe` radio from `PGRSearchInboxConfig.js`.
  - Filter-reset-on-switch via `key={activeTab}` remount of the composer.
[...]

Issue #1019, opened 2026-07-02
PGR Inbox Visibility (My/All tabs): V1 → V3


KDwevedi:
## Summary

Give PGR employees a **two-tab inbox** — **My Complaints** and **All Complaints** — where each tab's contents are driven by the user's role, their position in the `reportingTo` hierarchy, and their jurisdiction. This epic tracks the full rollout across **V1 → V2 → V3**.

- PRD: CMS_Escalation_PRD → "visibility v1" tab.
- Full design: `CCRS/VISIBILITY-DESIGN.md` (to be committed with V1).

## The core model: two orthogonal HRMS axes, resolved server-side

Today the inbox has no tabs (visibility is an `assignedToMe` radio in the filter card) and **the backend reads none of the reporting hierarchy at inbox time** — it only reads one hop of `reportingTo` inside the escalation path. Both decisions are reversed here.

Every tab in every version is a predicate over two independent axes, **AND**-ed together:

| Axis | Question | HRMS source | PGR filter |
|---|---|---|---|
| **Queue (who)** | whose work is this? | `reportingTo` chain + `user.roles` + workflow `Action.roles` | `applicationStatus IN statesFor(roles)`; V2+ `assignee` |
| **Jurisdiction (where)** | is it in my area? | employee `jurisdictions` → boundary subtree → leaf localities | `locality IN (...)` |

**Hierarchy resolution lives in the backend.** `pgr-services` reads a materialized HRMS projection and composes the search criteria; the frontend just asks for a tab.

Notation: `statesFor(roles)` = states whose queue those roles own (`{ S in PGR BusinessService : ∃ a ∈ S.actions, a.roles ∩ roles ≠ ∅ }`); `reporteeRoles(N)` = roles held by employees transitively under me in the `reportingTo` tree to depth N (default 1); `me` = my user UUID; `∅` = unassigned. Every predicate below is additionally `AND locality ∈ myLocalities`.

## Version matrix (from the PRD)

`hasReportees` = user has ≥1 direct reportee.

### V1 — queue axis = role/state only
| | My Complaints | All Complaints |
|---|---|---|
| no reportees | `status ∈ statesFor(myRoles)` | `status ∈ statesFor(myRoles)` |
| has reportees | `status ∈ statesFor(myRoles)` | `status ∈ statesFor(reporteeRoles(N))` |

### V2 — queue axis adds assignee
| | My Complaints | All Complaints |
|---|---|---|
[...]

PR #1052, opened 2026-07-06
feat(pgr): My/All complaint inbox tabs — Visibility V1 (FE, step 1)

KDwevedi:
Draft — **Step 1** of #1020 (PGR Inbox Visibility V1). FE-first; the optimised server-side backend is Step 2.

## What this does
Adds the two **My Complaints / All Complaints** tabs from the CMS_Escalation_PRD "visibility v1" tab — table and search/filter cards unchanged (per the PRD: "retain what we have").

- **Tabs** (`PGRInboxTabs`): rendered **directly above the complaint list** (new `resultsHeader` slot on `InboxSearchComposer`), per the PRD placement. Literal `My Complaints (N)` / `All Complaints (N)` text tabs, keyboard-accessible, brand-palette theming via overrides.css.
- **Badges are alert counts, not totals** (PRD): `(N)` = complaints new in the tab since the user last opened it — high-water-mark cursor per (user, tenant, tab), localStorage in Step 1, keyed into the react-query key so cursor advances refetch deterministically. Red dot accompanies `N > 0`. Opening a tab zeroes its badge (channel-unread semantics).
- **Per-tab visibility — My = assigned to me** (PO decision after the bomet demo; the tabs replace the assigned-to-me radio, which carried person semantics): "My" = complaints whose workflow assignee is me (`assignee` param, the radio's exact mechanism); "All" = every open complaint. Both tabs share the same status scope (all open/actionable states from the PGR BusinessService). Includes a **pgr-services `_count` parity fix**: `count()` now resolves `assignee` to serviceRequestIds like `search()` does — previously the param was silently ignored, so the My badge would have overcounted. Accepted tradeoff: a GRO's unassigned assignment queue lives in All, not My, until V2's server-side union (`assignee = me ∨ unassigned-in-my-queue`).
- **Filter card**: applies within the active tab and **resets on tab switch** (composer remount). An explicit status filter wins on both tabs (terminal states like Resolved stay reachable) and composes with My's assignee scope — e.g. My + Resolved = my resolved complaints. The now-redundant `assignedToMe` radio is removed.
- **Localization**: `PGR_INBOX_TAB_MY`/`PGR_INBOX_TAB_ALL` seeded in DDH localisations (default + en_IN).
- **Feature-flagged (off = the legacy inbox)**: one state-level MDMS record, `RAINMAKER-PGR.InboxVisibilityConfig` `{ code: "INBOX_VISIBILITY", enabled, version, reporteeDepth, jurisdictionScoped }`, gates the whole feature per tenant. **Default is OFF** (absent record / fetch error / `enabled:false` are identical): no tabs, the `assignedToMe` radio is restored, default search is the legacy open-states list, and none of the workflow/`_count` requests fire — the feature merges dark and rollout/rollback is a pure MDMS flip, no redeploy. Schema seeded via DDH; Step 2's BE gets `PGR_VISIBILITY_ENABLED` env + the same MDMS check (design §6.1). Verified both states live on bomet (seeded `enabled:true` at `ke`). Note: MDMS responses are localStorage-cached per `ApiCachingSettings` TTL, so flips are eventually-consistent per browser session.
- **`VISIBILITY-DESIGN.md`**: the full V1→V3 design (two-axis server-side resolver, `eg_pgr_hrms_projection`, cursor counts) this step builds toward.

## Step 2 (reportee core) — now included, shipped dark

The server-side reportee resolver from the design doc (§4.2/§4.3) is implemented in pgr-services and ships **dark behind two switches**: `PGR_VISIBILITY_ENABLED` (deploy env, surfaced in the compose env block) and `InboxVisibilityConfig.serverSide` (per-tenant MDMS cutover). Until both are flipped, runtime behavior is exactly the Step-1 client composition above.

- **`eg_pgr_hrms_projection`** — local projection of the HRMS `reportingTo` hierarchy, fed by a Kafka consumer on the HRMS employee save/update topics, with a nightly full-rebuild backstop and a boot-time backfill when empty. Inbox resolution makes zero live HRMS calls.
- **`VisibilityService`** — MY → `assignee = me`; ALL → for users with projected reportees: complaints assigned to them or their subtree (`reporteeDepth` from MDMS, default 1) **plus the unassigned queues**; **empty-subtree fallback** → tenant-wide open view (unmaintained `reportingTo` data can never produce an empty All tab).
- **`POST /request/inbox/_search` + `/inbox/_count`** — same criteria surface as the plain endpoints plus `tab`; the team-or-unassigned predicate is one SQL clause so pagination/count stay correct.
- **FE cutover** — `serverSide: true` points the composer and badge counts at the new endpoints and drops client-side assignee composition.

**Validation status:** backend compiles (JDK17), FE builds, and the client mode was regression-verified against bomet after the change. The server path itself has **not** been e2e-validated yet — that runs on the local stack before any tenant flips `serverSide` (and bomet's live MDMS schema needs the documented ops tweak first, see design §6).

## Explicitly out of scope (Step 2 / later versions)
- **Unoptimised by design**: visibility is FE-composed against the existing pgr search — no backend changes. Step 2 moves resolution server-side (`VisibilityService` + `eg_pgr_hrms_projection`) and the cursor server-side.
- **"Unopened" is cursor-based** (accepted redefinition, design §5.5): the badge counts *new since the tab was last opened*, not per-complaint read state; and Step 1's cursor keys off `createdTime`, so complaints re-routed into the queue don't count until Step 2 keys off state-entry time.
- **Reportee-scoped "All" and the role-queue axis**: `statesFor(roles)` (with the `state→ownerRole` refinement) is now entirely a Step-2 resolver concern — V1's tabs differ only on the assignee axis. The hybrid "assigned to me ∪ my unassigned queue" for assigners is a V2 candidate.
- **Jurisdiction axis, reportee-scoped "All", assignee (V2), escalation (V3)** — per the design doc.

## Testing
Re-verified after the semantics revision (KE_GRO vs bomet): My lists only complaints owned by County GRO; All unchanged (everyone's). Verified e2e against bomet's backend (local esbuild dev server → SSH-tunnelled Kong), KE_GRO: tabs sit directly above the list; badges show alert counts and zero on tab open; cursor survives reload (tenant-scoped key); filter resets on switch; All+Resolved returns resolved complaints; My+Resolved returns an honest empty result; no new console errors. Production `npm run build` passes.

> Note: bomet's `_search` was independently broken (missing `extended_attributes` column — image↔migration skew, see #1023); fixed on bomet so this could be validated.

Refs #1019, #1020.

## Summary by CodeRabbit

* **New Features**
  * Added “My Complaints” / “All Complaints” tabs to the PGR inbox with visibility-based, server-driven scoping.
  * Introduced visibility-aware inbox search and count operations to power per-tab badges/new-item indicators.
  * Added MDMS-driven configuration (including reportee depth and jurisdiction scoping) plus a deployment kill switch for controlled rollout.
  * Implemented an HRMS reporting-hierarchy projection to enable reportee-based “All” results.
* **Bug Fixes**
  * Corrected inbox count alignment so “My” badge totals match the scoped results.
* **Documentation**
  * Added the full inbox visibility design and rollout specification.


backend/pgr-services/src/main/java/org/egov/pgr/service/PGRService.java
```diff
@@ -260,6 +188,18 @@ public ServiceRequest update(ServiceRequest request){
      * @return
      */
     public Integer count(RequestInfo requestInfo, RequestSearchCriteria criteria){
+        // Mirror search()'s assignee handling: resolve the assignee to
+        // serviceRequestIds via workflow before counting. Without this the
+        // assignee param was silently ignored on _count, so count and search
+        // disagreed for assignee-scoped queries (e.g. the My-tab badge).
+        if (criteria.getAssignee() != null) {
+            String tenantId = criteria.getTenantId() != null ? criteria.getTenantId() : requestInfo.getUserInfo().getTenantId();
+            Set<String> serviceRequestIds = workflowService.getServiceRequestIdsByAssignee(requestInfo, tenantId, criteria.getAssignee());
```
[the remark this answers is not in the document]
KDwevedi: Fixed in dde0bfed0 — null-guarded the `getUserInfo()` fallback deref here, in the `count()` twin, and in the `tenantIdForMdms` resolution, matching the null-safe pattern `plainSearch()` already used.

backend/pgr-services/src/main/java/org/egov/pgr/service/VisibilityService.java
```diff
@@ -0,0 +1,152 @@
...
+        Map<String, Object> mdmsConfig = fetchVisibilityConfig(requestInfo, tenantId);
+        if (mdmsConfig == null || !Boolean.TRUE.equals(mdmsConfig.get("enabled"))) {
+            throw new CustomException("PGR_VISIBILITY_DISABLED",
+                    "Inbox visibility is not enabled for tenant " + tenantId + " (RAINMAKER-PGR.InboxVisibilityConfig)");
+        }
+
+        String me = requestInfo.getUserInfo() != null ? requestInfo.getUserInfo().getUuid() : null;
+        if (me == null) {
+            throw new CustomException("PGR_VISIBILITY_NO_USER", "Inbox visibility requires an authenticated user");
+        }
```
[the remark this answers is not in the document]
KDwevedi: Fixed in dde0bfed0 — the authenticated-user guard now runs first, so both derefs sit behind it (took the shape proposed here).

backend/pgr-services/src/main/java/org/egov/pgr/service/VisibilityService.java
```diff
@@ -0,0 +1,152 @@
...
+            // means ALL stays the tenant-wide open view, never an empty tab.
+            return;
+        }
+
+        Set<String> team = new HashSet<>(reportees);
+        team.add(me);
+        Set<String> assignedIds = new HashSet<>();
+        for (String memberUuid : team) {
+            assignedIds.addAll(workflowService.getServiceRequestIdsByAssignee(requestInfo, tenantId, memberUuid));
+        }
```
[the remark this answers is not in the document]
KDwevedi: Verified against the live workflow-v2: `/process/_search` matches `assignee` as ONE literal string — a comma-separated pair returns 0 rows — so a batched assignee query isn't available to switch to. Mitigated in dde0bfed0 instead: the team fan-out is now bounded (`pgr.visibility.team.fanout.max`, default 25); past the cap the ALL tab degrades to the same tenant-wide open view as the empty-reportees fallback, so the inbox never stalls behind an unbounded sequential chain. The search/count duplication is inherent to them being separate HTTP requests; a shared short-TTL resolution cache is a possible follow-up if it shows up in practice.

backend/pgr-services/src/main/java/org/egov/pgr/web/models/RequestSearchCriteria.java
```diff
@@ -59,6 +59,19 @@ public class RequestSearchCriteria {
...
+     * `(serviceRequestId IN visibilityIds OR applicationStatus IN visibilityUnassignedStates)`
+     * — team-assigned complaints plus the unassigned queues. Both restrict the
+     * result set; they can't widen a search.
+     */
+    @JsonIgnore
+    private Set<String> visibilityIds;
+
+    @JsonIgnore
+    private Set<String> visibilityUnassignedStates;
+
```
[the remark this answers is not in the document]
KDwevedi: Fixed in dde0bfed0 — added an `@InitBinder` disallow-list on the controller covering `visibilityIds`, `visibilityUnassignedStates`, and the other server-side-only criteria fields (`serviceRequestIds`, `userIds`, `isPlainSearch`). Good catch: on the empty-reportees fallback path `resolve()` returns without overwriting these, so a client-supplied `visibilityIds` would have survived into the OR-predicate.

digit-ui-esbuild/products/pgr/src/pages/employee/PGRInbox.js
```diff
@@ -156,17 +158,40 @@ const PGRSearchInbox = () => {
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [location, visLoading, visibilityEnabled]);
 
   /**
    * Show loader until necessary data is available
    */
-  if (isLoading || isValidationLoading || !pageConfig || serviceDefs?.length === 0) {
+  if (isLoading || isValidationLoading || visLoading || !pageConfig || serviceDefs?.length === 0) {
     return <Loader />;
   }
```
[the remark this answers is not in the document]
KDwevedi: Fixed in dde0bfed0 — `useServiceDefs` now starts at `null` (loading) and resolves to the array, so `[]` unambiguously means the tenant has no defs and the Loader gate checks `serviceDefs == null`. Audited the live consumers for the new null: `PGRInbox`/`ComplaintsList` already guarded, `createComplaintForm` got a default on its one unguarded call; the `packages/modules/pgr` copies are dead code behind the esbuild `module-pgr → products/pgr` alias.

backend/pgr-services/src/main/java/org/egov/pgr/config/PGRConfiguration.java
```diff
@@ -246,6 +246,20 @@ public MappingJackson2HttpMessageConverter jacksonConverter(ObjectMapper objectM
...
 
+    // Visibility (V1 Step-2 reportee core) — deploy-level kill switch plus
+    // resolver defaults; the per-tenant switch is MDMS InboxVisibilityConfig.
+    @Value("${pgr.visibility.enabled:false}")
+    private Boolean visibilityEnabled;
+
+    @Value("${pgr.visibility.reportee.depth.default:1}")
+    private Integer visibilityReporteeDepthDefault;
+
+    @Value("#{'${pgr.visibility.unassigned.states:PENDINGFORASSIGNMENT,PENDINGFORREASSIGNMENT}'.split(',')}")
```
subhashini-egov, line 272: @KDwevedi Can you please also document where all end users can make these configurations for the Ansible set up? .env file, docker compose etc..

backend/pgr-services/src/main/java/org/egov/pgr/consumer/HrmsProjectionConsumer.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Keeps eg_pgr_hrms_projection in sync with HRMS by consuming the employee
+ * create/update topics (VISIBILITY-DESIGN.md §4.3). Org changes reflect on
+ * the next inbox load with no complaint re-stamping; the scheduled rebuild in
+ * HrmsProjectionService is the backstop for missed events.
+ */
+@Slf4j
+@Component
+@ConditionalOnProperty(value = "pgr.visibility.enabled", havingValue = "true")
```
subhashini-egov, line 39: @KDwevedi Let's add comments on how to enable this feature flag and where are all the places we can enable this.

backend/pgr-services/src/main/java/org/egov/pgr/service/VisibilityService.java
```diff
@@ -0,0 +1,165 @@
...
+ * Gated twice: the PGR_VISIBILITY_ENABLED env kill switch (service-wide) and
+ * the per-tenant MDMS RAINMAKER-PGR.InboxVisibilityConfig `enabled` flag —
+ * the same record the frontend decides from, so flag-off tenants never reach
+ * this code path end-to-end.
+ */
+@Slf4j
+@Service
+public class VisibilityService {
+
+    public static final String TAB_MY = "MY";
```
subhashini-egov: @KDwevedi should this be externalized to localisation? Also seems like a UI abstraction bleeding into backend code. Should this just be renamed as a filter for "mine" vs. "reportees"?

backend/pgr-services/src/main/java/org/egov/pgr/util/PGRConstants.java
```diff
@@ -189,6 +189,10 @@ public class PGRConstants {
 
     public static final String MDMS_ESCALATION_CONFIG = "EscalationConfig";
```
subhashini-egov, line 195: @KDwevedi Assuming these are masters. Let's add a comment above to indicate that.

subhashini-egov (changes requested):
Minor feedback and changes requested.

backend/pgr-services/src/main/java/org/egov/pgr/config/PGRConfiguration.java
KDwevedi: Done in e4f864cbb — PGRConfiguration's visibility block now documents the full operator surface: `PGR_VISIBILITY_ENABLED` in the pgr-services env block of `local-setup/docker-compose.egov-digit.yaml` (deploy-wide), per-box overrides via the per-tenant overlay `docker-compose.<inventory_hostname>.yml` that the Ansible playbook picks up, the remaining `pgr.visibility.*` knobs via Spring relaxed binding (e.g. `PGR_VISIBILITY_TEAM_FANOUT_MAX`), and the per-tenant runtime flags living in MDMS rather than env.

backend/pgr-services/src/main/java/org/egov/pgr/consumer/HrmsProjectionConsumer.java
KDwevedi: Done in e4f864cbb — the consumer's class doc now spells out the two independent switches (deploy env `PGR_VISIBILITY_ENABLED`, which gates this consumer via `@ConditionalOnProperty`, the rebuild, and the inbox endpoints; and per-tenant MDMS `InboxVisibilityConfig.enabled`/`serverSide`, runtime-editable via the configurator), where each is set, and the resulting behavior for each on/off combination.

backend/pgr-services/src/main/java/org/egov/pgr/service/VisibilityService.java
KDwevedi: Both points taken, done in e4f864cbb. Localisation: no — these are machine enum values on the wire, never rendered; the labels users see are localized frontend-side (`PGR_INBOX_TAB_*`), and the constants' doc now says so. Naming: agreed it was a UI abstraction leaking in — renamed the API to filter semantics: param `tab=MY|ALL` → `scope=MINE|TEAM` (`MINE` = assigned to me; `TEAM` = reportee subtree + unassigned queues, with the tenant-wide fallback). Went with TEAM rather than REPORTEES since the scope also includes the requester's own assignments and the unassigned queues. The FE (in this PR) maps its tab ids to scope at the request boundary; no external consumers exist yet.

backend/pgr-services/src/main/java/org/egov/pgr/util/PGRConstants.java
KDwevedi: Done in e4f864cbb — comment added: it's the MDMS master name under module RAINMAKER-PGR (mdms-v2 schema code `RAINMAKER-PGR.InboxVisibilityConfig`, seeded by default-data-handler), with the JSONPATH constant noted as where that master lands in an MdmsRes payload.

subhashini-egov (approved):

Issue #916, opened 2026-06-23
[Bug] CMS: Employee: Inbox pagination navigation not working


Gurjeet-eGov:
# Description

CMS: Employee: Inbox pagination navigation not working. There are more than 10 complaint but user is unable to navigate to next page of records 

---

# Steps to Reproduce

1. Login with CSR
2. Create complaints 
3. Go to inbox and check pagination options 

---

# Expected Behavior

User shall be able to navigate across pages

---

# Actual Behavior

Navigation is blocked 

https://github.com/user-attachments/assets/8655848b-f53d-4780-9a5b-ba373c4fd698

Lokendra-egov:
https://github.com/egovernments/Citizen-Complaint-Resolution-System/pull/964

Lokendra-egov:
Fix's PR: https://github.com/egovernments/Citizen-Complaint-Resolution-System/pull/1014

(two bugs, not one)
ResultsDataTableWrapper received TotalCount=undefined, collapsed totalPages to 1, and permanently disabled Next.

Even with that fixed, libraries/usePGRInboxSearch was returning totalCount: wrappers.length (current page size = 10) instead of calling _count — so indexOfLastRow (10) >= totalCount  (10) still disabled Next on page 1.

Fix:
- Added totalCountJsonPath: "totalCount" to frontend/.../pgr/src/configs/PGRSearchInboxConfig.js
- Updated digit-ui-esbuild/packages/libraries/src/hooks/pgr/usePGRInboxSearch.js to call _search and _count in parallel (same as the products hook), using the real server-side total.
- Added 9 static regression tests in local-setup/tests/static/pgr-inbox-pagination.test.ts to prevent either bug from regressing

Gurjeet-eGov:
@Lokendra-egov pagination not fixed 
user is unable to navigate across pages. 
https://github.com/user-attachments/assets/dc3b32dd-1cbb-4f6b-b653-c3abbb0227d8

subhashini-egov:
@Lokendra-egov Please add automated tests location for this use case and how someone can run it.

Lokendra-egov:
[...]

PR #1058, opened 2026-07-07
fix(pgr): stop inbox total-count from capping at page size (#916)

Lokendra-egov:
pgr-services' _count wraps the same LIMIT/OFFSET-bound query used for _search in count(*), so it never reports more than the requested limit — passing the UI's page size (10) as the count's limit meant the reported total was always capped at one page, keeping Next disabled/pagination broken for any inbox with more than one page.

Request the count with a limit far above any realistic per-tenant filtered inbox instead of the page size, so it reflects the true total. This is a ceiling, not a fix for the underlying backend query builder — verified against the live deployment: limit=10 capped the count at 10 for a tenant with 396 matching complaints; limit=10000 correctly returned 396.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed inbox search pagination so the total count now reflects the full result set instead of the current page.
  * Improved count handling to avoid including pagination limits in total-count lookups.
  * Search results still load normally even if count retrieval fails, with a safe fallback for totals.

### DIFF
--- a/digit-ui-esbuild/packages/libraries/src/hooks/pgr/usePGRInboxSearch.js
+++ b/digit-ui-esbuild/packages/libraries/src/hooks/pgr/usePGRInboxSearch.js
@@ -10,15 +10,6 @@ import { CustomService } from "../../services/elements/CustomService";
  * Returns data in the same shape as the inbox-v2 API so existing column
  * configs and additionalCustomizations work unchanged.
  */
-// pgr-services' _count wraps the same LIMIT/OFFSET-bound query used for
-// _search in `count(*)`, so it never reports more than the requested limit
-// (#916) — it isn't a true unpaginated count. Fixing that is a backend
-// query-builder change; until that ships, request a limit far above any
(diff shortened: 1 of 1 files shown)


digit-ui-esbuild/packages/libraries/src/hooks/pgr/usePGRInboxSearch.js
```diff
@@ -19,9 +28,10 @@ const usePGRInboxSearch = (reqCriteria) => {
   const fetchData = async () => {
     // 1. Call PGR search + count in parallel
     const countUrl = url.replace("_search", "_count");
+    const countParams = { ...params, limit: COUNT_QUERY_LIMIT_CEILING, offset: 0 };
```
vinothrallapalli-eGov: @Lokendra-egov We have a separate Count API that returns the total number of records matching the specified search criteria. Unlike the Search API, it is not paginated and is not limited by the page size or pagination parameters.

vinothrallapalli-eGov (approved):

PR #1062, opened 2026-07-07
feat(digit-ui-esbuild): embed PGR supervisor dashboard as ACS-gated employee module

KDwevedi:
## Summary

Migrates the PGR supervisor dashboard frontend (from the #963 lineage, `soniikajal:feature/demo-dashboard` head) into **digit-ui-esbuild** as `products/dashboard/`, mounted as a normal module **inside the DigitUI employee chrome** (standard topbar + sidebar, native navigation) and gated through the platform's standard ACS mechanisms. Data-side RBAC/ABAC is untouched — the dashboard keeps calling `pgr-services /v2/analytics`, where scope injection, catalog `visibleTo`, and the PII gate continue to run server-side.

This retires the standalone deployment model: no more separate `/var/www/dashboard-ui` webroot, no more hand-run webpack build with a manually added DefinePlugin. The dashboard ships inside the ansible-managed `/digit-ui/` static build.

## What changed

**Ported content (byte-identical):** `digit-ui-esbuild/products/dashboard/src/` — AdminDashboard + all tile/chart/filter/map components, copied verbatim from the source tree (17 dead files pruned). The subtree is fully self-contained: zero imports reach outside it.

**Integration layer (new code, outside the verbatim zone):**
- `products/dashboard/Module.js` — `DashboardModule` route component: role guard via tenant-agnostic `Digit.UserService.hasAccess(DASHBOARD_ROLES)` (roles live at the state root, so the tenant-scoped helper would wrongly deny at city tenants), renders `<AdminDashboard embedded />`, else redirects to `/employee`.
- `products/dashboard/DashboardCard.js` — home card, same role check, SPA link.
- `products/dashboard/roles.js` — shared `DASHBOARD_ROLES = [SUPERVISOR, PGR_SUPERVISOR, GRO, DGRO, PGR_LME, PGR_ADMIN, SUPERUSER]`.
- `packages/modules/core/.../AppModules.js` — always-on fallback route for `/employee/dashboard` (registry-driven, sits after MDMS-driven routes so the citymodule route wins when the row exists; mirrors the pattern of `/login` etc.). Without it, the route only mounts when `tenant.citymodule` contains `Dashboard`.
- **`embedded` mode** (minimal edits to `AdminDashboard.jsx` / `DashboardLayout.jsx` + an appended `.dashboard-embedded` CSS block): suppresses the dashboard's standalone shell — its internal dark sidebar, its `DashboardLogin` gate, and `100dvh` viewport sizing — so it lives inside the host chrome. Default (standalone) behavior is unchanged.

### 1. Build-time (esbuild)
- `REACT_APP_ANALYTICS_BASE` define added to `esbuild.build.js` + `esbuild.dev.js`; defaults to `/pgr-services/v2/analytics`, overridable via env at build. **Mandatory** — `services/analyticsService.js` reads `process.env` at import time and throws without the define.
- Dependency pins (exact, no carets): `apexcharts@3.45.2`, `react-apexcharts@1.4.1`, `react-grid-layout@1.3.4`, plus `react-draggable@4.4.5` via overrides/resolutions. **Do not upgrade apexcharts** — `config/lineChartPresentation.js` manipulates Apex DOM internals; 4.x would be a project.
- `public/index.html`: Inter font `<link>` (the CSS `@import` can be dropped mid-bundle).
- New npm script `build:dashboard-css` regenerates `products/dashboard/src/styles/dashboard.css` from `input.css` (committed compiled CSS is used as-is otherwise; Tailwind is `tw-`-prefixed, `preflight:false`, all rules scoped under `.dashboard-root` — no collision with the app's unprefixed build).

### 2. MDMS / ACS seeds (`ansible/nairobi-mdms/mdms/`)
- `tenant/citymodule.json`: `Dashboard` module row → gates the **home card** slot (+ the MDMS-driven route).
- `ACCESSCONTROL-ACTIONS-TEST/actions-test.json`: action **id 4557**, `url:"url"`, `navigationURL:/digit-ui/employee/dashboard` → the **left-nav entry**, filtered server-side by egov-accesscontrol.
- `ACCESSCONTROL-ROLEACTIONS/roleactions.json`: bindings for SUPERVISOR / GRO / DGRO / SUPERUSER. (`PGR_SUPERVISOR`, `PGR_ADMIN` intentionally not bound — they don't exist in the ke roles master.)
- **Card and sidebar are independent**: disable the action row → sidebar-only removal; deactivate the citymodule row → card-only removal. The route stays reachable either way via the fallback (with its own role guard).
- **Upserting into a live env**: mirror the live conventions, not this repo's file verbatim — e.g. live bomet's max action id is 2568 (4557 is free) and its roleactions carry inner `tenantId:"ke"` where this file uses `"pg"`. mdms-v2 has no `_delete`; removal = `is_active:false`, and roleactions must be deactivated **before** their action (the `actionid` x-ref is validated on update).

### 3. Localization
Upsert to `rainmaker-common` / `en_IN` at the state root: `ACTION_TEST_DASHBOARD` (sidebar), `DASHBOARD_CARD_HEADER` (card) → "Dashboard". Must ship with the full cache-bust: redis `messages` hash server-side, and clients hold bundles in sessionStorage `Digit.initData` **and** localStorage `Digit.Locale.<locale>.<module>` (survives hard refresh).

### 4. Runtime (globalConfigs) — all optional, with fallbacks
`STATE_LEVEL_TENANT_ID`, `DASHBOARD_STATE_LABEL`, `STATE_NAME`, `DASHBOARD_PRODUCT_LABEL`, `DASHBOARD_SYSTEM_TITLE`, `DASHBOARD_BRAND_PRIMARY/DARK/SLATE`, `MAP_CENTER_LAT_LNG`, `DASHBOARD_MAP_CITY_LABEL`.

### 5. Deployment
No new deploy surface: the dashboard is part of the digit-ui esbuild build in all three `digit_ui_mode`s. Environments running the old standalone dashboard can retire `/var/www/dashboard-ui` and its nginx block after cutover.

## Validation

Run locally (esbuild dev server) against the **live bomet backend** over an SSH Kong tunnel:
- Full render with real data: KPI tiles + sparklines + deltas, all ApexCharts kinds incl. the dashed multi-series line with custom markers, Leaflet choropleth with complaint pins, Complaints-at-risk table. Zero console errors.
- Guards: signed-out → employee login; role mismatch → redirect to `/employee`; GRO user → server-side pack filtering returns the correct empty state.
- Navigation: home ↔ dashboard, complaint-ID deep links into the PGR module, browser back/forward.
- ACS end-to-end: MDMS rows were temporarily seeded on bomet — home card + left-nav "Dashboard" entry appeared for SUPERVISOR and clicked through to the dashboard — then deactivated (bomet is back to pre-demo state).

## Known gaps / follow-ups
- Sequencing with the analytics catalog PR #963 (this FE was ported from its head; it runs fine against the currently deployed bomet backend).
- Localization upsert is not automated in this PR (see §3).
- Sidebar/card icon is a placeholder borrowed from an existing entry.
- Card role list is a code constant (`roles.js`); could be made MDMS-driven if per-deployment control is wanted.

## Summary by CodeRabbit

* **New Features**
  * Added a new Dashboard experience with sign-in, role-based access, KPI cards, charts, tables, filters, search, export, and an “add KPI” panel.
  * Introduced geography map views and complaint-risk dashboards with drill-down, tooltips, and richer visual summaries.
  * Enabled dashboard modules for new tenant setup and added improved chart/table interactions.

* **Bug Fixes**
  * Improved access handling so users without the right roles are redirected appropriately.
  * Added fallback behavior for missing dashboard configuration and empty data states.

### DIFF
--- a/ansible/nairobi-mdms/mdms/ACCESSCONTROL-ACTIONS-TEST/actions-test.json
+++ b/ansible/nairobi-mdms/mdms/ACCESSCONTROL-ACTIONS-TEST/actions-test.json
@@ -26211,5 +26211,23 @@
       "serviceCode": "MDMS Search",
       "parentModule": "306"
     }
+  },
+  {
+    "tenantId": "ke",
+    "data": {
+      "id": 4557,
+      "url": "url",
+      "name": "Dashboard",
+      "path": "Dashboard",
+      "enabled": true,
+      "leftIcon": "action:dashboard",
+      "rightIcon": "",
+      "displayName": "Dashboard",
+      "orderNumber": 2,
+      "queryParams": "",
+      "serviceCode": "DASHBOARD",
+      "parentModule": "",
+      "navigationURL": "/digit-ui/employee/dashboard"
+    }
   }
 ]
--- a/ansible/nairobi-mdms/mdms/ACCESSCONTROL-ROLEACTIONS/roleactions.json
+++ b/ansible/nairobi-mdms/mdms/ACCESSCONTROL-ROLEACTIONS/roleactions.json
@@ -26398,5 +26398,45 @@
       "tenantId": "statea",
       "actioncode": ""
     }
+  },
+  {
+    "tenantId": "ke",
+    "data": {
+      "id": 1740,
+      "actionid": 4557,
+      "rolecode": "SUPERVISOR",
+      "tenantId": "pg",
+      "actioncode": ""
+    }
+  },
+  {
+    "tenantId": "ke",
+    "data": {
+      "id": 1741,
+      "actionid": 4557,
+      "rolecode": "GRO",
+      "tenantId": "pg",
+      "actioncode": ""
+    }
+  },
+  {
+    "tenantId": "ke",
+    "data": {
+      "id": 1742,
+      "actionid": 4557,
+      "rolecode": "DGRO",
+      "tenantId": "pg",
+      "actioncode": ""
+    }
+  },
+  {
+    "tenantId": "ke",
+    "data": {
+      "id": 1743,
+      "actionid": 4557,
+      "rolecode": "SUPERUSER",
+      "tenantId": "pg",
+      "actioncode": ""
+    }
   }
 ]
--- a/ansible/nairobi-mdms/mdms/tenant/citymodule.json
+++ b/ansible/nairobi-mdms/mdms/tenant/citymodule.json
@@ -46,5 +46,22 @@
         }
       ]
     }
+  },
+  {
+    "tenantId": "ke",
+    "data": {
+      "code": "Dashboard",
+      "order": 14,
+      "active": true,
+      "module": "Dashboard",
+      "tenants": [
+        {
+          "code": "ke"
+        },
+        {
+          "code": "ke.nairobi"
+        }
+      ]
+    }
   }
 ]
--- a/digit-mcp/src/tools/mdms-tenant.ts
+++ b/digit-mcp/src/tools/mdms-tenant.ts
@@ -1326,6 +1326,14 @@ export function registerMdmsTenantTools(registry: ToolRegistry): void {
         'Workflow.AutoEscalationStatesToIgnore',
         // ── inbox ──
         'INBOX.InboxQueryConfiguration',
+        // ── dashboard analytics catalog ──
+        // KPI defs and packs are platform-level definitions (queries + viz +
+        // role visibility) with no tenant identity inside — without them a new
+        // root gets a working dashboard shell but an empty catalog ("no tiles
+        // in the catalog pack for this role"). Complaint-type–specific data
+        // (ServiceDefs/ComplaintHierarchy) stays operator-owned; these are not.
+        'dss.KpiDefinition',
+        'dss.DashboardPack',
         'ACCESSCONTROL-ROLEACTIONS.roleactions',
       ];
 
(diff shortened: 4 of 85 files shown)


KDwevedi:
Marked ready for review — requesting this one first, today if possible.

Since the draft: rebased onto current develop tip and added a parity commit syncing the ported dashboard with develop's #1013 sanitizeFilters null-guard (the port source predates it; without it any date-filter change crashes the embedded view).

Sequencing with the dashboard bug fixes: the four bug/schema PRs are being consolidated into a single follow-up PR that lands **after** this one; it will carry a mirror commit applying the frontend fixes to the ported `products/dashboard` copy so the two copies can't diverge.

KDwevedi:
Added `dss.KpiDefinition` + `dss.DashboardPack` to tenant_bootstrap's `essentialSchemas` allowlist (digit-mcp). Previously a freshly bootstrapped state root received the full view-access surface (ACS rows, citymodule, localization) but an empty KPI catalog — the dashboard shell rendered with "no tiles in the catalog pack for this role". KPI defs/packs are platform-level definitions with no tenant identity, so they copy cleanly; city tenants were already covered via state-root inheritance. Docs in #1081 updated to match.

KDwevedi:
Since #1081 merged ahead of this PR, the declared mirror flipped direction: this branch now carries the mirror commit instead. Rebased onto current develop (includes #1081 and #1072) and synced the ported `products/dashboard` copy with develop's dashboard fixes — map empty-state overlay, ABAC-scoped filter options with ComplaintHierarchy labels/optgroups, and the globalConfigs-derived MDMS context path (the same pattern @vinothrallapalli-eGov requested on #1081). Build green; the two dashboard copies are in lockstep as of this commit. No other changes — still ready for review.

vinothrallapalli-eGov (approved):

PR #1067, opened 2026-07-07
fix(pgr): citizen boundary cascade queries the complaint's resolved tenant

Hari-egov:
## Problem
On digit.mctd.gov.mz the citizen Location step renders no boundary cascade (only Landmark). Network shows the component's lookup going to `tenantId=mz` → **400 HIERARCHY_DEFINITION_DOES_NOT_EXIST**, while the same search at `mz.ige`/`mz.igsae` returns the full 235-node tree — the data exists, the render fails.

## Root cause
`BoundaryService.fetchBoundaries` **unconditionally overrides** the caller's `tenantId` with the citizen's `CITIZEN.COMMON.HOME.CITY` for citizen users. On multi-authority envs the citizen's home tenant is the state root (`mz`), which has no tree — so the wizard's carefully resolved tenant (from the *Complaint related to* selection) was discarded.

## Fix
Prefer the explicitly passed tenant (the authority resolved by the dropdown); fall back to the home city only when none is passed. Employee flows unaffected (override only applied to CITIZEN users).

## Testing
- Local (mz.ige/mz.igsae): citizen Location step renders the full Província→Distrito→Município cascade for both authorities; selection persists to submit.
- Live diagnosis on digit.mctd.gov.mz: state-tenant 400 vs city-tenant 235 nodes, matching this exact code path.

### DIFF
--- a/digit-ui-esbuild/products/pgr/src/components/BoundaryComponent.js
+++ b/digit-ui-esbuild/products/pgr/src/components/BoundaryComponent.js
@@ -279,7 +279,17 @@ useEffect(() => {
               <BoundaryDropdown
                 key={key}
                 fieldKey={key}
-                label={`${t(`${hierarchyType}_${key?.toUpperCase()}`)}`}
+                // Level-heading key normalized the way seeding does it:
+                // ASCII (accents stripped), uppercase, non-alnum → "_"
+                // (divisao_administrativa + "Província" → DIVISAO_ADMINISTRATIVA_PROVINCIA).
+                // Unseeded → show the human boundaryType ("Província"), never the raw key.
+                label={(() => {
+                  const norm = `${hierarchyType}_${key}`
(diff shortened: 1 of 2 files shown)


Hari-egov:
Second commit: **boundary labels now load from the tree's tenant** — the cascade's display names (`rainmaker-boundary-<hierarchy>` module) are seeded at the city tenant with the boundaries, but the app loads localization at the state tenant, so raw codes rendered. `fetchBoundaries` fetches the label module for the same tenant as the tree (parallel, best-effort) and feeds i18next directly. No copying data to the state tenant.

Hari-egov:
Third commit: **level-heading labels** — key normalization (accents stripped, uppercase: `DIVISAO_ADMINISTRATIVA_PROVINCIA`), targeted fetch of those codes at the tree's tenant, and a human fallback (the boundaryType itself, e.g. "Província") when unseeded — no more raw composite keys on either citizen or employee forms.

PR #1070, opened 2026-07-07
fix(pgr): lazy boundary loading — no pre-selection calls, level order from the cascade's own fetch

Hari-egov:
## Problem (digit.mctd.gov.mz, citizen create)
Opening ANY PGR page fires `boundary-relationships/_search?tenantId=mz` repeatedly (400 × react-query retries) **before the citizen has selected anything** — and the location cascade can render blank even though the correct city-tenant fetch succeeds.

## Root cause
`usePGRInitialization` prefetched the boundary tree at **module mount** to compute `boundaryHierarchyOrder`:
- runs before the authority selection exists → citizen home tenant (state root) → 400
- throws → react-query default retries → the repeated calls in the network tab
- sole writer of `boundaryHierarchyOrder` → on failure the cascade renders nothing

## Fix
- `fetchBoundaries` (the cascade's own data call — already tenant-correct after #1067) now derives `boundaryHierarchyOrder` from the same response and stores it under the key `BoundaryComponent` reads.
- The module-mount prefetch is removed entirely.

**Result:** zero boundary calls until the cascade actually loads; exactly one call, at the resolved tenant; order always consistent with the rendered tree. Employee flow identical (fetch fires with the logged-in tenant when the form renders).

## Testing
- Local citizen flow: no boundary requests on step 1; single `mz.ige`/`mz.igsae` request on the location step; cascade + labels render; employee create unchanged.

### DIFF
--- a/digit-ui-esbuild/products/pgr/src/Module.js
+++ b/digit-ui-esbuild/products/pgr/src/Module.js
@@ -50,20 +50,16 @@ export const PGRModule = ({ stateCode, userType, tenants }) => {
   });
   let user = Digit?.SessionStorage.get("User");
 
-  // Initialize boundary hierarchy for both employee AND citizen users.
-  // Citizens reach the create-complaint flow on naipepea, where the
-  // location step is now driven by `<PGRBoundaryComponent>` (closes
-  // egovernments/CCRS#428 + #433). The component reads
-  // `boundaryHierarchyOrder` from SessionStorage, which this hook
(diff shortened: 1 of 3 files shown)


Hari-egov:
Second commit: fixes the **works-only-after-refresh** symptom — the level-order memo was keyed on tenantId only, so with lazy loading it cached an empty order on first visit; now re-keyed on the fetched tree, the cascade renders on the first pass.

PR #1073, opened 2026-07-07
fix(pgr): citizen details — complaint hierarchy read at the complaint's tenant

Hari-egov:
## Problem
On digit.mctd.gov.mz a citizen's complaint summary shows raw keys for Complaint Type / Sub-Type (`COMPLAINT_HIERARCHY.TRANSPORTATION_LEGALITY`, …).

## Root cause
The details page fetched `RAINMAKER-PGR.ComplaintHierarchy` at the **citizen's home tenant** (state root on multi-authority envs). The hierarchy nodes AND their `COMPLAINT_HIERARCHY.*` localization are onboarded **at the complaint's tenant** (`mz.igsae`) — so the localized lookup missed *and* the node-name fallback had no node to fall back to. Same read-at-state vs seeded-at-city family as #1067/#1070.

## Fix
Fetch the hierarchy at `service.tenantId` (the complaint's own tenant), falling back to the previous behavior until the complaint loads. Labels then resolve via the localized key or the node's seeded name.

## Testing
- Local: citizen details for `mz.igsae`/`mz.ige` complaints show readable Type/Sub-Type; single-tenant (legacy) envs unchanged (`service.tenantId` == home tenant there).

### DIFF
--- a/digit-ui-esbuild/products/pgr/src/pages/citizen/ComplaintDetails.js
+++ b/digit-ui-esbuild/products/pgr/src/pages/citizen/ComplaintDetails.js
@@ -253,18 +253,34 @@ const ComplaintDetailsPage = () => {
       select: (raw) => {
         const defs = (raw?.["RAINMAKER-PGR"]?.ComplaintHierarchyDefinition || []).filter((d) => d?.active !== false);
         const allRows = raw?.["RAINMAKER-PGR"]?.ComplaintHierarchy || [];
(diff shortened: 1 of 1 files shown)


Hari-egov:
Second commit: **leaf-owned definition selection** — on tenants holding multiple hierarchies (e.g. the state root aggregating both authorities), the first-def-with-rows picker mis-selected for the other authority's complaints and collapsed the view to legacy flat rows. The def whose hierarchyType matches the complaint's leaf now wins.

Issue #1026, opened 2026-07-02
[Bug] Configurable Dashbord: Mismtach in Subtypes across charts


aarushibaser-lab:
# Description

As a health supervisor on the dashbord, the open complaints by workflow stage shows 8+ subtypes but the complaint type details shows only 3. 

SImilar behaviour is observed in the water supervisor dashboard also. There I can see 8 in the workflow statge bar chart, but 0 in the complaint type details chart. 

<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/a61cc8e2-b439-4b50-99d7-f0461cd4b7c8" />

<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/2fe58b3c-d06f-4fbc-9fe4-38bb33731cdd" />

---

# Steps to Reproduce

1. Log in as health or water dept supervisor 
2. View the "open complaints by workflow stage" chart and "complaint type details" chart
3. View the discrepency in the complaint subtypes count

---

# Expected Behavior

The same number of subtypes should show across all charts.

---

# Actual Behavior

Difference in number of subtypes shown across charts.

---

# Environment

- **Deployment**:Bomet 
- **Tenant**: 
- **Browser**: 
- **User role**: 

---

# Screenshots / Logs

Attach screenshots, browser console errors, or service logs if available.

---

# Classification

> Add these labels before submitting. See [CONTRIBUTING.md](../CONTRIBUTING.md) for the full list.

- **Feature**: 
- **Area**: 
- **Priority**: 
- **Source**:

Gurjeet-eGov:
Resolved 

<img width="1366" height="662" alt="Image" src="https://github.com/user-attachments/assets/69809275-5f8f-44f9-8d28-04719142a5af" />

<img width="1366" height="662" alt="Image" src="https://github.com/user-attachments/assets/189ac50a-6355-4d3e-b9d7-1f2c19157bb1" />

PR #1074, opened 2026-07-07
fix(dashboard): Complaint Type Details covers all complaints, not week-to-date

KDwevedi:
## Summary
Fixes #1026. The details table had a baked `window: {wtd, filed_at}` while "Open Complaints by Workflow Stage" is a windowless open-snapshot — the two tiles covered different complaint sets, so their subtype lists diverged (worst on Mondays: 1 vs 12 subtypes for a dept supervisor; Arushi saw 3-vs-8 and 0-vs-8 mid-week). Department ABAC was verified NOT to be the cause — both queries scope identically.

MDMS-only change to `dss.KpiDefinition` (`cl_table_complaint_type_details`): drop the baked window (table becomes an all-time per-subtype reference that still narrows when a global date range is set), raise `limit` 30→100 (supervisors sat exactly at the cap), state the scope in the subtitle. No deploy needed — MDMS upsert only.

Verified via live `_query` probes on bomet: without the window the table returns the same subtype set as the stage chart for DEMO_HEALTH / DEMO_WATER / DEMO_SUPERVISOR.

Follow-up (separate concern, not in this PR): the backend never applies `params[].default` for KPI-by-reference calls, so declared defaults are dead config.

### DIFF
--- a/ansible/nairobi-mdms/mdms/dss/KpiDefinition.json
+++ b/ansible/nairobi-mdms/mdms/dss/KpiDefinition.json
@@ -1409,10 +1409,6 @@
       "status": "published",
       "query": {
         "grain": "facts",
-        "window": {
-          "name": "wtd",
-          "timeRole": "filed_at"
-        },
         "dimensions": [
           "service_code",
           "service_group"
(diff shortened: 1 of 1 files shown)


KDwevedi:

Issue #1031, opened 2026-07-02
[Bug]  Configurable Dashboards: Map doesnt work w filter


aarushibaser-lab:
# Description

map component doest work when filters for ward or complaint type are changed. and after selecting any ward/complaint type, the map becomes blank. even after resturning all filters to default, it doesnt go back to normla

---

# Steps to Reproduce

1. sign in as health/water dept supervisor
2. try to change any of the filters
3.

---

# Expected Behavior

Maps should accordingly adjust

---

# Actual Behavior

Map disappears, blank screen is shown

---

# Environment

- **Deployment**: bomet
- **Tenant**: 
- **Browser**: 
- **User role**: 

---

# Screenshots / Logs

Attach screenshots, browser console errors, or service logs if available.

---

# Classification

> Add these labels before submitting. See [CONTRIBUTING.md](../CONTRIBUTING.md) for the full list.

- **Feature**: 
- **Area**: 
- **Priority**: 
- **Source**: 
try to f

Gurjeet-eGov:
Resolved 

https://github.com/user-attachments/assets/158de1f8-8160-4db7-8433-196d31ec0feb

PR #1076, opened 2026-07-07
fix(dashboard): keep Leaflet map mounted through empty filter results

KDwevedi:
## Summary
Fixes #1031. When a filter combination returned zero ward rows (clean 200), the choropleth tile's `showEmpty` branch **unmounted the Leaflet container div**. The map instance — created by a one-shot init effect (`[]` deps + `mapRef.current` guard) — stayed bound to the detached node, so when data returned the fresh div was never initialized: the tile went permanently blank, surviving filter resets and even page reloads (filters persist in localStorage and re-latch on first render). Reproduced end-to-end on the live bomet dashboard; two independent static analyses converged on the same lines, and empty-bounds/double-init/pins-shape hypotheses were checked and disproven (all bounds paths are guarded — nothing throws).

Fix: render the map canvas div unconditionally; show "No geographic data" as an absolutely-positioned overlay mirroring the existing loading overlay. The draw effect already `clearLayers()` + early-returns on empty data, so the map shows bare tiles under the overlay and recovers automatically.

Validated against the bomet backend: empty date window → overlay over live tiles, no crash; clearing filters fully restores polygons/pins — the exact sequence that permanently blanked the deployed dashboard.

### DIFF
--- a/frontend/micro-ui/web/src/dashboard/components/GeographyChoroplethMap.jsx
+++ b/frontend/micro-ui/web/src/dashboard/components/GeographyChoroplethMap.jsx
@@ -847,43 +847,40 @@ const GeographyChoroplethMap = ({
           </div>
         ) : null}
         {showEmpty ? (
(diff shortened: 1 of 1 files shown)


KDwevedi:

Issue #1028, opened 2026-07-02
[Bug] Dashboard: Complaint type detail table does not display data for every type 


aarushibaser-lab:
# Description

The Complaints type details chart does not have data coming in. <img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/aed3b5c5-6b66-4706-9c24-99cc0b49fc9e" />

---

# Steps to Reproduce

1. Login as health supervisor.
2. View COmplaints Type Details chart
3. You will see blank columns

---

# Expected Behavior

Columns should be filled w data. If there is no data, it should say 0 or 0%. SLA column should always be filled since that comes from MDMS and not dependent on complaints filed. 

---

# Actual Behavior

Some data is there, not all.

---

# Environment

- **Deployment**: Bomet

- **Tenant**: 
- **Browser**: 
- **User role**: 

---

# Screenshots / Logs

Attach screenshots, browser console errors, or service logs if available.

---

# Classification

> Add these labels before submitting. See [CONTRIBUTING.md](../CONTRIBUTING.md) for the full list.

- **Feature**: 
- **Area**: 
- **Priority**: 
- **Source**:

Gurjeet-eGov:
Unresolved 

<img width="1366" height="662" alt="Image" src="https://github.com/user-attachments/assets/914c23c2-2d1a-4a3b-b931-a2831d175e03" />

Gurjeet-eGov:
Unresolved 

<img width="1728" height="1117" alt="Image" src="https://github.com/user-attachments/assets/7b5b4954-a8f9-4d71-8148-326970eb818b" />

Issue #1079, opened 2026-07-08
Analytics grains: support arbitrary jurisdiction & complaint hierarchies (materialized path + level registry)


KDwevedi:
## Summary

The V2 analytics grains (`complaint_events` / `complaint_facts` / `complaint_open_state_daily`) do **not** support arbitrary-depth **jurisdiction** or **complaint** hierarchies. Both axes are effectively hardcoded to a fixed shape, which breaks for tenants whose trees aren't exactly the assumed depth (e.g. Maputo's 4-level boundary tree, or any N>2 complaint taxonomy).

The good news: **both axes already have the two ingredients needed to fix this generically** — a prebuilt materialized path *and* a level-definition registry. This issue proposes making the grains registry-driven instead of position-hardcoded.

---

### Jurisdiction — arbitrary path, but position-hardcoded named levels
The full tree is preserved losslessly:
```sql
boundary_path = boundary_relationship.ancestralmaterializedpath || '|' || code   -- root→leaf, arbitrary depth
```
…but the *named* level columns are extracted by fixed position:
```sql
zone_code = split_part(boundary_path, '|', 2)   -- always 2nd segment
ward_code = split_part(boundary_path, '|', 3)   -- always 3rd segment
```
So every "by ward"/"by zone" KPI assumes the tree is exactly `root(1) > zone(2) > ward(3)`. On a 4-level tree, or where "ward" isn't at depth 3, these mislabel or go null. The migration itself flags it: *"position-based; refine per hierarchy."* `boundary_depth` (segment count) is available, but the human-named levels are not depth-agnostic.

### Complaint — only 2 levels, no path
The grain joins the taxonomy **one level only**:
```sql
service_code   -- leaf (the complaint's serviceCode)
service_group  -- the IMMEDIATE parent only (ServiceDefs.menuPath / ComplaintHierarchy.parentCode)
department_code
```
There is **no `complaint_node_path`** column — the only `*path*` column in the grains is `boundary_path`. An N-level complaint taxonomy collapses to leaf + one parent, which breaks for N>2 tenants.

Two related sub-issues:
- **Source-master inconsistency:** the deployed grain on bomet still reads `RAINMAKER-PGR.ServiceDefs` (`menuPath`), while the newer migration reads `RAINMAKER-PGR.ComplaintHierarchy` (`parentCode`). Either way it captures only leaf + immediate parent.
[...]

PR #1075, opened 2026-07-07
fix(dashboard): populate filter options from ABAC-scoped analytics distincts

KDwevedi:
## Summary
Fixes #1030. The ward/complaint-type dropdowns had no repo data source (`filterOptions={null}`); the deployed dashboard relied on an options object **hot-patched into the minified bundle on the host** (unscoped, stale, wiped by any redeploy). A water-dept supervisor saw all departments' types — 22 of 32 options could never return data — plus drifted ward codes and test-tenant wards.

New `useFilterOptions` hook: one batch `/v2/analytics/_query` with two inline distinct queries (`service_code`, `ward_code`, facts grain, window "all"). PrincipalScopeResolver injects the caller's scope into every planned query, so options are department-scoped server-side with zero client scope logic; tenant-wide roles (SUPERVISOR) are unaffected. Options flow through the existing `applyFilterOptions` → `reconcileFiltersWithOptions` plumbing so persisted dead selections are dropped. On fetch failure the selects keep their placeholder fallback.

Labels use the shared `formatDimensionLabel` humanizer (`ward_name` is not a groupable column on the facts grain).

Validated live against the bomet backend: DEMO_HEALTH sees 12 medical-only types and only wards with visible data; DEMO_WATER 12 water types; supervisor sees all 35.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Dashboard filters now load available ward and complaint type options dynamically.
  * Complaint type choices can now appear in grouped sections for easier browsing.

* **Bug Fixes**
  * Improved filter loading so the dashboard keeps working even if option data is unavailable.
  * Filter dropdowns now show more complete, human-friendly labels and an “All” option by default.

PR #1080, opened 2026-07-08
fix(analytics): SLA target from MDMS + registry-driven hierarchy grains

KDwevedi:
## Summary
One grain rebuild, three coordinated changes (per maintainer decision), plus the dashboard-configuration docs.

Fixes #1028. Implements #1079.

### 1. SLA sourcing (#1028)
`sla_target_ms` was the workflow `businessservicesla` joined on **exact tenant** — NULL for every city-tenant complaint (blank SLA column in Complaint Type Details; `sla_breached`/compliance KPIs silently undercounting — 118 open complaints could never breach). New definition, per the decided semantics:
```
sla_target_ms = COALESCE(
  ServiceDefs.slaHours → ms,                     -- per-subtype resolution SLA (MDMS)
  EscalationConfig ladder total,                 -- per-type overrides, else summed defaultSlaByLevel
  workflow SLA (join falls back to state root))  -- legacy last resort
```
Escalation timers are per-assignment-level clocks, so only the **ladder sum** is used, and only as a fallback.

### 2. Registry-driven hierarchy grains (#1079)
- `ward_code`/`zone_code` resolved from `boundary_hierarchy` types instead of `split_part` positions (leaf/parent-of-leaf fallback for tenants without a Ward level); new `boundary_leaf_code`/`boundary_leaf_type`.
- Complaint axis: `complaint_node_path` (from `ComplaintHierarchy.path`, recursive `parentCode` fallback), `complaint_depth`, `service_parent_code`; `service_group` = **root category** (byte-compatible on 2-level taxonomies); leaf detection from `ComplaintHierarchyDefinition.isLeafServiceCode`.
- Planner: new `starts_with` prefix operator (parameterized, LIKE-metachar-escaped) allowed **only** on `boundary_path`/`complaint_node_path`; `_schema` exposes the new columns and operator.

### 3. `params[].default` applied server-side
Declared KPI param defaults were parsed but never applied (dead config). Precedence: explicit caller > declared default > baked query. The now-obsolete window default on `cl_table_complaint_type_details` is stripped in the same file so #1074's all-complaints intent survives (different hunks than #1074; trivial rebase for whichever lands second).

### Docs
`docs/dashboard-configuration/` — 7-part operator/developer guide: what lives in MDMS vs code, KPI catalog + query grammar cookbook, packs & the four RBAC layers, view access via ACS (with the mdms-v2 operational gotchas), filter options, SLA & hierarchy semantics (this PR's end-state), and operations (refresh scheduler, bootstrap coverage).

## Validation (read-only probes on the live bomet DB; both full MV bodies executed)
| Check | Result |
|---|---|
| `sla_target_ms` coverage | 299 → **436/436** (tiers: 406 ServiceDefs, 30 ladder, 0 workflow; all 137 city-tenant NULLs resolved) |
| `sla_breached` | 217 → **350** (undercount fixed); never-breach opens 118 → **0** |
| ward/zone registry vs positional | **436/436 identical** |
| `service_group` root==parent (2-level compat) | **403/403** |
| ComplaintHierarchy path vs recursive rebuild | **447/447** |
| `mvn compile` | clean |

Byte-diffs vs the old `service_group` exist only where bomet's master data is stale/polluted (e.g. 163 rows get the *correct* post-#917 `StaffMisconductGroup`); side-finding logged: 308/446 `ke` ComplaintHierarchy records carry junk `complaints.categories.*` paths — data cleanup, deliberately not coded around.

## Deploy notes
- New flyway migration `V20260708000000__sla_and_hierarchy_grains.sql` recreates both grain MVs; `DashboardRefreshScheduler` picks them up on its 5-minute cycle — no manual refresh step.
- Historical breach numbers move up (that's the fix); `complaint_open_state_daily` history is untouched.

Related: #1074 #1075 #1076 (same bug-family, separate one-concern PRs).

### DIFF
--- a/ansible/nairobi-mdms/mdms/dss/KpiDefinition.json
+++ b/ansible/nairobi-mdms/mdms/dss/KpiDefinition.json
@@ -1585,7 +1585,6 @@
       "params": [
         {
           "name": "window",
-          "default": "last_7d",
           "allowed": [
             "last_1d",
             "last_7d",
--- a/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsCatalog.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsCatalog.java
@@ -29,6 +29,7 @@ public class AnalyticsCatalog {
         public final Set<String> epochMsColumns;        // columns stored as epoch-ms (vs sql date)
         public final Set<String> groupable;
         public final Set<String> filterable;
+        public final Set<String> prefixFilterable;      // #1079: starts_with (path rollup/scope) allowlist — ONLY these
         public final Set<String> measurable;            // numeric -> sum/avg/min/max/percentile
         public final Set<String> distinctable;          // -> count_distinct
         public final String tenantColumn;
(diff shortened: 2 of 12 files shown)


KDwevedi:

Issue #1030, opened 2026-07-02
[Bug] Complaints should be scoped to department and visible only to department users


aarushibaser-lab:
# Description

When logged in as a water dept supervisor, the complaint type filter shows all the complaint types, even ones not belonging to water dept.

---

# Steps to Reproduce

1. log in as water dept supervisor
2. view the complaint type filter

---

# Expected Behavior

only complaint types belonigng to water dept should be shown

---

# Actual Behavior

all complaint types are shown

---

# Environment

- **Deployment**: bomet
- **Tenant**: 
- **Browser**: 
- **User role**: 

---

# Screenshots / Logs

Attach screenshots, browser console errors, or service logs if available.

---

# Classification

> Add these labels before submitting. See [CONTRIBUTING.md](../CONTRIBUTING.md) for the full list.

- **Feature**: 
- **Area**: 
- **Priority**: 
- **Source**:

Gurjeet-eGov:
Unresolved,

Health department user is able to view garbage related and other options under complaint type filter 

https://github.com/user-attachments/assets/fbc3397a-4292-49d4-ba95-6d2202ba5cd5

Gurjeet-eGov:
Unresolved 

- HS_CSR and HS_GRO users are assigned to Street Lights department. A complaint has been created for 'No Streelights', yet GRO is not able to view any complaint in inbox **PENDING**
- dashboard UI complaint type/subtype dropdown **does not** display any category from streetlights  **PENDING**
- Some charts are filtered based on department of user **RESOLVED**
@KDwevedi please confirm if first two points are ABAC related, i can log those separately. Currently the employees are not editable from configurator ui hence the department can't be updated to verify other use cases

<img width="1366" height="660" alt="Image" src="https://github.com/user-attachments/assets/fd4d6cd3-1e63-4ec6-bd17-5d11ddd56179" />

<img width="895" height="498" alt="Image" src="https://github.com/user-attachments/assets/7aaec528-d36d-4910-928b-208f934a7124" />

KDwevedi:
same as #1821

PR #1081, opened 2026-07-08
fix(dashboard): consolidated bug fixes, SLA/hierarchy grains, and configuration docs

KDwevedi:
## Summary
Consolidates the four dashboard PRs (#1074, #1075, #1076, #1080 — now closed in favor of this) into one review, per maintainer decision. Sequenced **after** the dashboard migration PR #1062.

Fixes #1026, #1028, #1030, #1031. Closes #1079.

Five commits, each independently validated against the live bomet deployment:

1. **#1026 — Complaint Type Details window** (MDMS-only): drop the baked `wtd` window so the table covers the same complaint population as the workflow-stage chart; limit 30→100; scope stated in subtitle.
2. **#1031 — map empty-state latch**: keep the Leaflet canvas mounted; empty state becomes an overlay. Fixes the permanently-blank map after any zero-row filter (reproduced on production; recovery verified).
3. **#1030 — filter options**: ABAC-scoped distincts via `/_query` replace the options object that had been hot-patched into the production bundle; dept supervisors see only their own types/wards.
4. **#1030 enhancement**: complaint-type labels + native optgroup grouping from `RAINMAKER-PGR.ComplaintHierarchy` (real display names; humanized fallback; graceful degradation).
5. **#1028 + #1079 — grains** (one MV rebuild): `sla_target_ms = COALESCE(ServiceDefs.slaHours→ms, EscalationConfig ladder total, workflow SLA w/ state-root fallback)`; registry-resolved `ward_code`/`zone_code` (+`boundary_leaf_*`); `complaint_node_path`/`complaint_depth`/`service_parent_code`, `service_group`=root category; `starts_with` prefix operator restricted to the two path columns; KPI `params[].default` applied server-side. Plus **`docs/dashboard-configuration/`** — 7-part operator/developer guide.

## Validation (read-only probes on bomet's live DB + local UI harness against the live backend)
| Check | Result |
|---|---|
| SLA target coverage | 299 → **436/436** (406 ServiceDefs, 30 ladder; all 137 city-tenant NULLs resolved) |
| `sla_breached` | 217 → **350**; never-breach opens 118 → **0** |
| ward/zone registry vs positional | **436/436 identical** |
| `service_group` root==parent (2-level compat) | **403/403** |
| ComplaintHierarchy path vs recursive rebuild | **447/447** |
| Map latch repro → fix | blank-and-latched on prod build → overlay + full recovery with fix |
| Filter scoping/grouping | DEMO_HEALTH: 12 medical-only options in 11 real category optgroups |
| `mvn compile` / esbuild parse checks | clean |

## Deploy notes
- Flyway `V20260708000000__sla_and_hierarchy_grains.sql` recreates both grain MVs; the refresh scheduler picks them up within ~5 min — no manual step.
- Historical breach counts rise (that is the fix). `service_group` changes only where master data was stale (e.g. RudeBehavior → the correct post-#917 `StaffMisconductGroup`).
- Side-finding (data cleanup, not code): 308/446 `ke` ComplaintHierarchy records carry junk `complaints.categories.*` paths from the #917 cutover.

## Sequencing
After #1062 (dashboard migration) merges, this branch gets one mirror commit applying the frontend fixes (map latch, filter options/grouping) to the ported `digit-ui-esbuild/products/dashboard` copy so the two copies stay in lockstep.

## Summary by CodeRabbit

* **New Features**
  * Dashboard filters now load ward and complaint-type options from the backend, including grouped complaint types and improved fallback behavior.
  * Added support for the new **“starts with”** filter option in analytics queries.
  * Complaint and SLA analytics now support deeper hierarchy groupings and more accurate SLA calculations.

* **Bug Fixes**
  * Increased KPI result limits and clarified date-filtered results.
  * Improved empty-state handling for geography maps so the map stays available.

* **Documentation**
  * Added and expanded dashboard configuration, access, filter, SLA, and operations docs.

### DIFF
--- a/ansible/nairobi-mdms/mdms/dss/KpiDefinition.json
+++ b/ansible/nairobi-mdms/mdms/dss/KpiDefinition.json
@@ -1409,10 +1409,6 @@
       "status": "published",
       "query": {
         "grain": "facts",
-        "window": {
-          "name": "wtd",
-          "timeRole": "filed_at"
-        },
         "dimensions": [
           "service_code",
           "service_group"
@@ -1488,7 +1484,7 @@
             "dir": "desc"
           }
         ],
-        "limit": 30
+        "limit": 100
       },
       "supportsSeries": false,
       "viz": {
@@ -1580,7 +1576,7 @@
             }
           }
         ],
-        "subtitle": "Resolution, SLA and reopen metrics per complaint type"
+        "subtitle": "Resolution, SLA and reopen metrics per complaint type - all complaints, narrowed by the date filter"
       },
       "params": [
         {
--- a/ansible/nairobi-mdms/mdms/dss/KpiDefinition.json
+++ b/ansible/nairobi-mdms/mdms/dss/KpiDefinition.json
@@ -1581,7 +1581,6 @@
       "params": [
         {
           "name": "window",
-          "default": "last_7d",
           "allowed": [
             "last_1d",
             "last_7d",
--- a/frontend/micro-ui/web/src/dashboard/services/complaintHierarchyService.js
+++ b/frontend/micro-ui/web/src/dashboard/services/complaintHierarchyService.js
@@ -1,0 +1,135 @@
+import { getTenantId, hasAuth } from "./analyticsService";
+import { formatDimensionLabel } from "../config/labelFormat";
+
+function parseJson(raw) {
+  try {
+    return JSON.parse(raw);
+  } catch {
+    return raw;
+  }
+}
+
+function getEmployeeToken() {
+  const raw = window.localStorage?.getItem("Employee.token");
+  return raw && raw !== "undefined" ? parseJson(raw) : null;
+}
+
+function getEmployeeInfo() {
+  const raw = window.localStorage?.getItem("Employee.user-info");
+  return raw && raw !== "undefined" ? parseJson(raw) : null;
+}
+
+function buildRequestInfo() {
+  const authToken = getEmployeeToken();
+  const userInfo = getEmployeeInfo();
+  return {
+    apiId: "Rainmaker",
+    ver: ".01",
+    ts: Date.now(),
+    action: "_search",
+    msgId: `dashboard-complaint-hierarchy-${Date.now()}`,
+    ...(authToken && { authToken }),
+    ...(userInfo && { userInfo }),
+  };
+}
+
+/**
+ * Display name for a hierarchy node. Live masters carry real display names on
+ * leaves ("Ambulance breakdown / Crew unresponsive") but category records often
+ * have name === code ("MedicalServices", "complaints.categories.X") — those go
+ * through the shared code humanizer instead.
+ */
+function displayName(record) {
+  const code = String(record?.code ?? "").trim();
+  const name = String(record?.name ?? "").trim();
+  if (name && name !== code) return name;
+  return formatDimensionLabel(code);
+}
+
+/**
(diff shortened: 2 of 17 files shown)


frontend/micro-ui/web/src/dashboard/services/complaintHierarchyService.js
```diff
@@ -0,0 +1,135 @@
...
+ * code → { label, rootCode, rootLabel } index.
+ *
+ * Resolves null on any failure (never rejects) so callers can fall back to
+ * humanized flat labels without blocking the dashboard.
+ */
+export async function fetchComplaintTypeIndex() {
+  if (!hasAuth()) return null;
+
+  try {
+    const response = await fetch("/egov-mdms-service/v1/_search", {
```
vinothrallapalli-eGov: @KDwevedi Please avoid hardcoding the MDMS search endpoint. For MDMS v2, the context path is `mdms-v2`. Instead, use the MDMS search hook so that switching between versions is easier. The hook already reads the context path from the global configuration, making the implementation more flexible and maintainable.

ansible/nairobi-mdms/mdms/dss/KpiDefinition.json
```diff
@@ -1580,12 +1576,11 @@
             }
           }
         ],
-        "subtitle": "Resolution, SLA and reopen metrics per complaint type"
+        "subtitle": "Resolution, SLA and reopen metrics per complaint type - all complaints, narrowed by the date filter"
```
vinothrallapalli-eGov, line 1579: Is subtitle will support multiple locale?

vinothrallapalli-eGov (changes requested):

frontend/micro-ui/web/src/dashboard/services/complaintHierarchyService.js
KDwevedi: Fixed in a7b8a6d34 — the URL is now derived from globalConfigs (`MDMS_V1_CONTEXT_PATH` → `MDMS_CONTEXT_PATH` → legacy fallback), so deployments resolve to `/mdms-v2/v1/_search` per their config. One note on the hook suggestion: the dashboard subtree is deliberately self-contained (no `@egovernments` imports / Digit runtime) so the same files run in the standalone build too — reading the context path from globalConfigs is the subtree's existing runtime-config pattern (same as STATE_LEVEL_TENANT_ID, map center, brand keys), and it picks up the identical value the MDMS hook would read. Happy to switch to the hook if/when the standalone build is retired.

ansible/nairobi-mdms/mdms/dss/KpiDefinition.json
KDwevedi: Honest answer: no — `viz.subtitle` is a plain string across the entire catalog today (pre-existing; all 28+ defs), unlike titles which already localize via `titleKey` + `RAINMAKER-PGR.DASHBOARD_KPI_*` messages. This PR only edits the text of one existing subtitle, so it doesn't change that. The right fix is a `viz.subtitleKey` resolved through the same localization path as `titleKey` (tiny FE change in the tile header + catalog-wide key seeding) — I'd do that as a follow-up covering all defs rather than special-casing this one here. Can fold it into this PR if you'd prefer.

vinothrallapalli-eGov: Ok, please do it followup

vinothrallapalli-eGov (approved):

PR #1092, opened 2026-07-08
fix(pgr): prevent masked placeholders from overwriting real data on u…

priyanshu-egov:
…pdates

- Extract mask sentinel "****" to PGRConstants.MASK_SENTINEL for centralized management
- Add restoreMaskedPlaceholders() method to detect and restore censored fields during complaint updates
- Decrypt existing complaint data before copying back masked fields to avoid double-encryption
- Replace hardcoded mask strings in EncryptionDecryptionService with MASK_SENTINEL constant
- Correct MDMS constant name from ComplaintSchema to ComplaintExtendedAttributeSchema
- Handle clients resubmitting cached complaints with stale masked values (e.g. from failed MDMS lookups or UI caching)

## Summary by CodeRabbit

* **New Features**
  * Complaint forms now support updated extended-attribute schemas with clearer field types, labels, required fields, ordering, and validation rules.

* **Bug Fixes**
  * Masked values in complaint updates are now restored correctly before saving, reducing the risk of corrupted placeholder data.
  * Confidential complaint details are now shown only to authorized viewers, while others receive masked content.
  * Schema loading has been improved to fetch the correct complaint configuration more reliably.


backend/pgr-services/src/main/java/org/egov/pgr/util/PGRConstants.java
```diff
@@ -174,8 +174,11 @@ public class PGRConstants {
     // Extended attributes — complaint category config and field schemas
     public static final String MDMS_COMPLAINT_RELATED_TO_MAP = "ComplaintRelatedToMap";
     public static final String MDMS_COMPLAINT_TEMPLATE_TYPE  = "ComplaintTemplateType";
-    public static final String MDMS_COMPLAINT_SCHEMA         = "ComplaintSchema";
+    public static final String MDMS_COMPLAINT_SCHEMA         = "ComplaintExtendedAttributeSchema";
```
vinothrallapalli-eGov, line 177: **correctness (CONFIRMED):** `MDMS_COMPLAINT_SCHEMA` was renamed from `"ComplaintSchema"` to `"ComplaintExtendedAttributeSchema"` with no corresponding update to the seeded MDMS master data (or anywhere else visible in this repo).

`MDMSUtils.doFetchComplaintTemplateTypeConfig()` looks up `$.MdmsRes.RAINMAKER-PGR.<MDMS_COMPLAINT_SCHEMA>[?(@.name=='<schemaRef>')]` inside a try/catch that only logs a warning on failure. If the real MDMS master is still named `ComplaintSchema`, every lookup silently returns nothing, `cfg.getXSecurity()` stays null, and `EncryptionDecryptionService.encrypt()/decrypt()` become no-ops for every category — confidential field values get persisted as plaintext instead of encrypted. This also breaks the new `restoreMaskedPlaceholders()` path, which depends on this same schema lookup to know which fields to decrypt.

Please confirm the MDMS master data has been renamed to match in a companion change, otherwise this is a silent encryption bypass.

backend/pgr-services/src/main/java/org/egov/pgr/service/PGRService.java
```diff
@@ -344,6 +346,40 @@ private Map<String, ComplaintTemplateTypeConfig> buildConfigCache(
...
+        List<ServiceWrapper> existing = repository.getServiceWrappers(criteria);
+        if (CollectionUtils.isEmpty(existing)) return;
+
+        ExtendedAttributes existingExt = existing.get(0).getService().getExtendedAttributes();
+        if (existingExt == null) return;
+
+        // existingExt's x-security fields are ciphertext at rest — decrypt before copying
+        // back, otherwise validation runs on ciphertext and encrypt() double-encrypts it.
+        encryptionDecryptionService.decrypt(existingExt, cfg);
+
```
vinothrallapalli-eGov, line 377: **correctness (CONFIRMED):** `restoreMaskedPlaceholders()` calls `encryptionDecryptionService.decrypt(existingExt, cfg)` but never checks whether decryption actually succeeded. `decrypt()`'s own failure path swallows exceptions and re-masks fields with `MASK_SENTINEL` ("****") instead of throwing.

If the enc-service is transiently unreachable during an update (the exact scenario this PR targets), every secure field on `existingExt` gets set back to the sentinel. `restoreMaskedPlaceholders` then reads that sentinel as a legitimate `existingValue`, copies it into `updatedExt`, and it gets re-encrypted/persisted as if it were real plaintext — permanently overwriting the real confidential data with an encrypted "****". This reintroduces the exact data-loss bug this PR is meant to fix, just triggered via a decrypt failure in the new restore path instead of a stale client cache.

Suggest checking decrypt success (or catching/propagating the underlying failure) before treating the returned value as authoritative.

backend/pgr-services/src/main/java/org/egov/pgr/service/PGRService.java
```diff
@@ -344,6 +346,40 @@ private Map<String, ComplaintTemplateTypeConfig> buildConfigCache(
...
     }
 
+    /**
+     * Clients that fetched a complaint while it was masked (e.g. a transient MDMS lookup
+     * failure, or the citizen UI caching a stale view) may echo the "****" sentinel back
+     * on a later update — the citizen RATE flow resubmits the whole cached service object.
+     * Restore the currently-stored value for any field the client sends back as the
+     * sentinel, so a masked placeholder never permanently overwrites real data.
+     */
+    private void restoreMaskedPlaceholders(ExtendedAttributes updatedExt, String serviceId, String tenantId,
```
vinothrallapalli-eGov, line 360: **correctness/security (CONFIRMED):** `restoreMaskedPlaceholders()` decrypts and restores the real value for any field the client echoes back as the mask sentinel, with no check that the current caller is authorized to see that (possibly confidential) data. Unlike `applyDecryptOrMask()`, there's no role/creator/`isConfidential` gate here — and the restored plaintext flows straight back to the caller in the update response.

Scenario: an employee lacking `CONFIDENTIAL_COMPLAINT_VIEWER` (and not the creator) fetches a confidential complaint and correctly gets every dynamic field masked to "****". If their client resubmits that same object on an update (e.g. a workflow status change), this method sees the sentinel, unconditionally decrypts the persisted service, and copies the real plaintext into `updatedExt` — which is then returned directly in the `/request/_update` response. That's a new data leak that didn't exist before this PR (previously the sentinel was just saved back literally, never decrypted and returned).

Suggest gating this restore path behind the same confidentiality/role check used by `applyDecryptOrMask()`.

vinothrallapalli-eGov (commented):
Automated high-effort code review found 3 confirmed issues (see inline comments).

pradeepkumarcm-egov (approved):

vinothrallapalli-eGov (approved):

PR #1093, opened 2026-07-08
feat(pgr): persist create-complaint drafts across navigation & refresh (citizen wizard + employee form)

Hari-egov:
## Problem

- **Citizen** (3-screen wizard, single route): every field lives in plain component state — navigating browser-back or refreshing wiped all answers and returned to screen 1; even stepping back within the wizard blanked the type/boundary dropdowns (their visible selection is component-private state).
- **Employee** (1-page form): nothing survived refresh; the `COMPLAINT_CREATE` session container existed but was written only in a dormant branch.

## What this does

**Both flows: session-draft persistence** — restore on mount (survives F5 and re-entry), cleared after successful `_create` (kept on failure so *Try Again* restores). Drafts are **stamped `{tenant, user}` and discarded on mismatch** — a home-city/ULB switch or re-login never restores foreign serviceCode/boundary codes. Photos travel as fileStoreIds, so drafts are tiny (quota-guarded writes regardless).

**Dropdown repaint on restore** — restored values now repaint their pickers:
- complaint-type cascade (citizen + employee variants) rebuilds the per-level chain by walking `parentCode`;
- boundary cascade rebuilds the chain from the saved node, including the next level's options for mid-cascade drafts (previously showed another parent's children).

**Cascade resets (dropdown change ⇒ dependent data reset, properly):**
- authority change (citizen) already reset type/subtype/dynamic fields; now also resets `SelectedBoundary` (tree is per-tenant) and `ComplaintImagesPoint` (filestore ids are tenant-scoped);
- employee Type→Sub-Type reset clears only when the held subtype is invalid for the new list (unconditional clear would wipe the restored value on hydrate);
- ward-hint auto-fill got a processed-key guard: a hint replayed from a restored draft no longer overwrites a manual boundary correction, and `childrenData` recreation (HRMS jurisdiction filter settling) no longer re-applies the same hint mid-session.

**Correctness fixes surfaced while wiring this:**
- `stepIsValid` was missing `templateFields`/`hasDispatcher` deps — with a restored last-step draft it never recomputed after templates settled, allowing submit with empty mandatory dynamic fields;
- FormComposerV2 resets the form whenever the `defaultValues` **reference** changes — the employee form passes a mount-frozen snapshot and persists via `onFormValueChange` (JSON change-guard, no loops, no keystroke wipes).

## Review

The diff was reviewed by a 3-lens adversarial pass (state lifecycle / cascade-reset / React loops); all 12 raw findings were independently re-verified against full sources and every confirmed one is fixed in this PR (tenant-unscoped drafts, mid-cascade child-option leak, ward-hint replay, hint-miss blank cascade, stepIsValid staleness, tenant-scoped filestore ids).

## Testing (local, deployed to digit-ui)

- Citizen: fill screens 1–2 → F5 → same step, all values + pickers repainted; complete create → revisit → empty form.
- Citizen: switch home city with a draft pending → draft discarded (no foreign codes).
- Employee: fill form incl. hierarchy type + boundary → F5 → everything restored incl. both cascades; submit → session cleared.
- Pin-drop → manual boundary correction → F5 → correction preserved; fresh pin still re-derives the cascade.

### DIFF
--- a/digit-ui-esbuild/products/pgr/src/components/BoundaryComponent.js
+++ b/digit-ui-esbuild/products/pgr/src/components/BoundaryComponent.js
@@ -152,9 +152,25 @@ useEffect(() => {
   // pick manually.
   const wardHintCode = formData?.GeoLocationsPoint?.ward?.code;
   const wardHintName = formData?.GeoLocationsPoint?.ward?.name;
+  // The hint key this mount has already handled. Guards two replays:
+  // (1) a session-restored draft replays the OLD pin's hint on remount — if
+  //     the draft also carries a SelectedBoundary (possibly a manual
+  //     correction of that very hint), the saved boundary is the user's final
+  //     intent and wins; only a hint CHANGE (fresh pin drop) re-derives;
+  // (2) childrenData is recreated when the HRMS jurisdiction filter settles —
+  //     without the key guard that re-ran the effect and re-applied the same
+  //     hint over a manual correction mid-session.
+  const processedHintRef = React.useRef(null);
   useEffect(() => {
     if (!wardHintCode && !wardHintName) return;
     if (!childrenData || childrenData.length === 0) return;
+    const hintKey = (wardHintCode || "") + "|" + (wardHintName || "");
+    if (processedHintRef.current === hintKey) return;
+    if (processedHintRef.current === null && formData?.SelectedBoundary?.code) {
+      // First hint of this mount with a saved boundary — the draft wins.
+      processedHintRef.current = hintKey;
+      return;
+    }
     // boundaryHierarchyOrder may not be seeded yet (usePGRInitialization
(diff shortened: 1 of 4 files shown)


Hari-egov:
Added a fix for the reported **NEXT dead-lock on the citizen location step**: clearing the map pin (✕) and filling the boundary manually left NEXT permanently disabled — the step required pin AND boundary. Now the **leaf boundary alone gates NEXT** (pin optional; payload keeps `geoLocation: {}` like the employee flow, which the persister accepts). Also hardened the leaf check to trust the emitter's `isLeaf` tag (trees deeper than the rendered levels left `children` on the deepest pickable node → same dead-lock), and the payload's locality now prefers the confirmed cascade value over the raw GeoJSON ward hint so a manual correction can't be silently overridden at submit.

PR #1098, opened 2026-07-08
fix(boundary): migrate client callers off legacy egov-location onto boundary-service

singhalkarun:
## Problem

The legacy boundary API `/egov-location/location/v11/boundarys/_search` is **not served on the Kubernetes stack** — only the compose stack fakes it, via a Kong Lua adapter that rewrites the call onto `boundary-service` and reshapes the response. On K8s there is no such workload, ingress rewrite, or gateway route, yet the deployed `digit-ui` bundle calls the legacy endpoint. Result: the **PGR ward/locality selector** (and boundary tooling in configurator/MCP) **break on K8s**. The backend is already migrated (`pgr-services` uses `boundary-service`); only client callers remained on the legacy path.

Found during the compose-vs-K8s deployment-parity review (item #1).

## Fix

Migrate the remaining **client** callers to `/boundary-service/boundary-relationships/_search`, using a **client-side adapter that mirrors the compose Kong adapter exactly**, so consumers are unchanged and both stacks share one path:
- wrap `hierarchyType` string → `{ code, name }`
- recursively add `name` (=code), `localname` (=code), `label` (=boundaryType) to each boundary node
- normalize missing/empty `children` to `[]`

Like the Kong adapter, the legacy `boundaryType` filter is dropped — the full hierarchy tree is returned and consumers walk it as they already do.

### Changes by frontend
| Frontend | Change |
|---|---|
| **digit-ui-esbuild** (deployed bundle — the real user-facing break) | Central adapter in `LocationService` (`Location.js`); removed the now-dead legacy URL getters (`urls.js`) and the unused `populateLocalityReqCriteria` helper (pgr `UICustomizations.js`) |
| **digit-mcp** | `locationBoundarySearch` → boundary-relationships + reshape; repointed the `Location` health probe to boundary-service; corrected now-misleading tool wording |
| **digit-ui-v2** / **configurator** | Removed the dead `LOCATION_BOUNDARY_SEARCH` constant (no callers; both already use `boundary-relationships`) |

## Validation

The reshape output was verified **byte-equivalent to the proven Kong adapter**, live against tenant `pg`:

| Check | Result |
|---|---|
| `hierarchyType` shape | ✅ `{code:ADMIN, name:ADMIN}` |
| Node field-sets | ✅ identical (`code/name/localname/label/boundaryType/children`) |
| Node count | ✅ 14 = 14 |
| Sample node | ✅ `{PG_STATE, name, localname, label:State, boundaryType:State}` identical |

## Notes / limitations
- Type-check/build were **not run locally** (`node_modules` not installed in these packages) — relying on CI to confirm `tsc`/build/tests. The TS mirrors existing patterns (`URLSearchParams` usage, private-method style).
- No full in-browser UI test (deployed UI runs from a prebuilt bundle); validated at the API/transform layer instead.
- The digit-mcp OpenAPI spec + reachability tests that document the legacy endpoint were intentionally left as-is (documentation, not runtime).
- **Follow-up:** once shipped, the compose Kong `/egov-location` adapter can be dropped so both stacks are fully on `boundary-service`.

## Summary by CodeRabbit

* **New Features**
  * Added support for updated access-role and access-action search flows.
  * Added support for generating IDs through the platform’s ID generation service.

* **Bug Fixes**
  * Location searches now use the newer boundary service and return the expected full hierarchy.
  * Improved location-related requests and health checks to work with the updated service routing.
  * Updated location help text and error guidance to reflect the current service behavior.

### DIFF
--- a/configurator/packages/data-provider/src/client/endpoints.ts
+++ b/configurator/packages/data-provider/src/client/endpoints.ts
@@ -37,7 +37,6 @@ export const ENDPOINTS = {
   ACCESS_ROLES_SEARCH: '/access/v1/roles/_search',
   ACCESS_ACTIONS_SEARCH: '/access/v1/actions/_search',
   IDGEN_GENERATE: '/egov-idgen/id/_generate',
-  LOCATION_BOUNDARY_SEARCH: '/egov-location/location/v11/boundarys/_search',
 
   ENC_ENCRYPT: '/egov-enc-service/crypto/v1/_encrypt',
   ENC_DECRYPT: '/egov-enc-service/crypto/v1/_decrypt',
--- a/digit-mcp/src/services/digit-api.ts
+++ b/digit-mcp/src/services/digit-api.ts
@@ -894,25 +894,64 @@ class DigitApiClient {
     return data.idResponses || [];
   }
 
-  // Location — search boundaries via egov-location service
+  // Location — boundary search.
+  // The legacy /egov-location boundarys API is not served on the Kubernetes stack
+  // (only the compose stack fakes it via a Kong adapter). We call the modern
+  // /boundary-service/boundary-relationships API and reshape its response back into
+  // the legacy TenantBoundary shape (mirroring the compose Kong adapter) so the
+  // location_search tool returns the same structure on both stacks. Like the Kong
+  // adapter, the legacy boundaryType filter is dropped — the full tree is returned.
   async locationBoundarySearch(
     tenantId: string,
     boundaryType?: string,
     hierarchyType?: string
   ): Promise<Record<string, unknown>[]> {
-    const body: Record<string, unknown> = {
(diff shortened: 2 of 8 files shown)


singhalkarun:
Addressed both CodeRabbit findings (`9e7652a7`):
- **reshapeToTenantBoundary** now always returns a stable envelope with `TenantBoundary` normalized to an array (empty on missing/malformed), so callers indexing `TenantBoundary[0]` can't crash on a bad upstream response.
- **`location_search` `boundary_type`** description now states it's ignored (the full tree is returned) — the schema matches `locationBoundarySearch`.

subhashini-egov (changes requested):
@singhalkarun Can you please address the coderabbit comments and resolve?

subhashini-egov (approved):

Issue #1071, opened 2026-07-07
pgr-services complaint search lacks server-side ownership scoping (client-supplied mobileNumber trusted)


singhalkarun:
### Summary
`POST /pgr-services/v2/request/_search` filters complaints by the **client-supplied `mobileNumber`** and does **not** verify that it belongs to the caller. There is no server-side ownership/record-level check, so an authenticated `CITIZEN` can read complaints beyond their own:
- **omit** `mobileNumber` → returns **all** complaints
- **supply another citizen's** `mobileNumber` → returns **that** citizen's complaints

Complaint records include citizen PII (name, mobile number, description, location).

### Severity
**Low** — triaged with the team; filing for tracking/hardening, not urgent.

### Affected
Both deployments — it's a `pgr-services` service-level behaviour (same image):
- The gateway only validates the token and resolves the caller's **role** (action-level authz). It does **not** do record-level ownership.
- `CITIZEN` is legitimately granted the pgr-search action (`ACCESSCONTROL-ROLEACTIONS`, action id 2008) because the "My Complaints" UI calls this endpoint, so the gateway correctly lets citizens through — and nothing then scopes *which* complaints they receive.
- The digit-ui SPA calls this endpoint directly from the browser with the citizen's token; the UI's own-mobile filter is client-side only.

### Expected behaviour
When the caller is a `CITIZEN`, the search must be constrained to their own identity (derive the filter from `userInfo` — `account_id`/uuid), ignoring any client-supplied `mobileNumber`.

### Suggested fix
Apply record-level scoping in the `pgr-services` search path. This pattern already exists in the repo for the analytics path — `PrincipalScopeResolver` / `AnalyticsScope` ("a pure CITIZEN sees only their own complaints, `account_id = their uuid`") — it just needs applying to the main `/request/_search`. Optionally add gateway RBAC in Kong (compose) as defense-in-depth, though that alone would not address the record-level issue.

### Reproduction (brief)
With a valid `CITIZEN` token (self-registerable; the citizen owns zero complaints):
```
POST /pgr-services/v2/request/_search?tenantId=<t>&mobileNumber=<another-citizens-number>   # returns their complaints
POST /pgr-services/v2/request/_search?tenantId=<t>                                           # returns all complaints
```

### Verification status
[...]

PR #1100, opened 2026-07-08
fix(pgr): scope citizen complaint search to own records (fixes #1071)

singhalkarun:
Fixes #1071.

## Problem
`pgr-services` complaint search did not enforce record-level ownership for citizens. `PGRQueryBuilder` scopes results with `ser.accountId IN (:userIds)`, and `userIds` was populated from a **client-supplied `mobileNumber`** (via `UserService.enrichUserIds`). So a citizen could:
- pass **another** user's `mobileNumber` → read that user's complaints, or
- pass **no** `mobileNumber` while sending any other filter (e.g. `tenantId`, which the UI always does) → `criteria.isEmpty()` is false, the self-scoping branch is skipped, `userIds` stays empty, and the query returns **every** complaint.

The pre-existing `criteria.setAccountId(uuid)` in `enrichSearchRequest` was dead code — `PGRQueryBuilder` never reads `getAccountId()`.

## Fix
In `EnrichmentService.enrichSearchRequest`, for a **pure citizen** (type `CITIZEN`, no employee role) force `criteria.userIds = [caller.uuid]` and drop any client-supplied `mobileNumber`. The existing `accountId IN (userIds)` clause then scopes results to the caller's own complaints. **Employees / internal principals are unchanged** and may still look up complaints by `mobileNumber`.

The pure-citizen test is extracted into `PrincipalScopeResolver.isPureCitizen()` — the single source of truth already used by the analytics self-scope path — so the two security-relevant call sites cannot drift. The dead `setAccountId` line is removed.

## Validation (live, on a running stack)
Built the image from this branch and exercised it end-to-end. Attacker = a citizen owning **0** complaints; victim owns 2.

| Test | Before | After |
|---|---|---|
| Attacker unfiltered search | 2 (victim's) | **0** |
| Attacker + another's `mobileNumber` | leaks | **0** |
| Citizen sees their **own** complaint | — | **1** (own only) |
| Employee (ADMIN) search | 2 | **2** (unaffected) |

Maven build exit 0.

## Scope
Backend only; applies to both the Compose and Kubernetes stacks. Record-level ownership like this must live in the service (a gateway/action-level RBAC layer cannot enforce it) — see parity item #5.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restricted both search and count results to the authenticated citizen’s own records.
  * Prevented client-supplied mobile numbers from widening citizen search results; refined how eligible user IDs are derived.
  * Improved access scoping by determining “pure citizen” status from role codes (with safer handling of missing/unrecognized roles), ensuring correct self-scoped analytics visibility.

* **Tests**
  * Added unit tests covering pure-citizen role combinations, edge cases, and failure-closed behavior.


backend/pgr-services/src/main/java/org/egov/pgr/analytics/PrincipalScopeResolver.java
```diff
@@ -74,22 +74,33 @@ public AnalyticsScope resolve(RequestInfo requestInfo, String tenantId, int stat
...
+     * sites cannot drift.
+     */
+    public boolean isPureCitizen(RequestInfo requestInfo) {
+        User u = requestInfo == null ? null : requestInfo.getUserInfo();
+        if (u == null || !"CITIZEN".equalsIgnoreCase(u.getType()))
+            return false;
+        List<Role> roles = u.getRoles();
+        if (roles != null) for (Role r : roles) {
+            String c = r.getCode() == null ? "" : r.getCode().toUpperCase();
+            if (!c.equals("CITIZEN")) return false; // holds an employee role → not a pure citizen
```
subhashini-egov: @singhalkarun - The assumption is that the citizen would only have ONE role assigned. But this will break if the CITIZEN has additional non-employee roles assigned. Make a TODO note here around this.

subhashini-egov (changes requested):
Kindly make a comment/todo note here.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/PrincipalScopeResolver.java
subhashini-egov: Verified with Pradeep & Vinoth. You should check if CITIZEN is one of the roles this person holds and doesn't have EMPLOYEE or COMMON_EMPLOYEE. If so, then scope it to their mobile number.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/PrincipalScopeResolver.java
```diff
@@ -74,22 +74,39 @@ public AnalyticsScope resolve(RequestInfo requestInfo, String tenantId, int stat
...
+     * sites cannot drift.
+     */
+    public boolean isPureCitizen(RequestInfo requestInfo) {
+        User u = requestInfo == null ? null : requestInfo.getUserInfo();
+        if (u == null || !"CITIZEN".equalsIgnoreCase(u.getType()))
+            return false;
+        List<Role> roles = u.getRoles();
+        if (roles != null) for (Role r : roles) {
+            String c = r.getCode() == null ? "" : r.getCode().toUpperCase();
+            // TODO(#1071): this classifies ANY role other than CITIZEN as an employee role,
```
subhashini-egov: @singhalkarun - just check if CITIZEN is one of the roles amongst the set and then scope it. Only CITIZEN will break right away.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/PrincipalScopeResolver.java
singhalkarun: Image

Added a TODO for this.

singhalkarun: Addressed in f0e13e9.                                                                                                                                                                                   
                                                                                                                                                                                                          
  isPureCitizen() is now role-based and no longer assumes a citizen holds only the CITIZEN role. A pure citizen is classified as holding the CITIZEN role and no employee base role (EMPLOYEE /           
  COMMON_EMPLOYEE). Additional citizen-side roles are therefore permitted, while an employee who also holds CITIZEN remains classified as an employee. The TODO(#1071) has been removed as resolved, and  
  mvn compile passes.                                                                                                                                                                                     
                                                                                                                                                                                                          
  Two points for confirmation:                                                                                                                                                                            
  1. COMMON_EMPLOYEE is treated as an employee marker in addition to EMPLOYEE; I can restrict this to EMPLOYEE only if preferred.                                                                         
  2. The prior type == CITIZEN check was dropped in favour of the positive CITIZEN-role check to keep the logic fully role-based; happy to retain it as an additional guard if you'd like.
  
@subhashini-egov

singhalkarun:
Pushed b8f639e1 addressing the three review findings. Summary of what was verified vs. what was changed.

## Finding triage

All three reproduce against the code, but they are not equally severe — one was a live, exploitable leak and two were hardening:

| Finding | Verdict |
|---|---|
| `/_count` bypasses `enrichSearchRequest` | **Real leak, reproduced live** |
| `userIds` bindable via `@ModelAttribute` | Hardening on `/_search`; **load-bearing on `/_count`** |
| `isPureCitizen()` fail-open on abnormal role state | Hardening — not reachable with a normal citizen token |

## Live reproduction (mz.maputo, attacker citizen owns 0 complaints, victim owns 1)

| Probe | Before | After |
|---|---|---|
| attacker `_count` (tenantId only) | **50** | 0 |
| attacker `_count?userIds=<victim-uuid>` | **1** | 0 |
| attacker `_search` (tenantId only) | 0 | 0 |
| attacker `_search?userIds=<victim-uuid>` | 0 | 0 |

`/_count` returned the tenant-wide total to a citizen owning nothing, and `userIds=<victim-uuid>` turned it into a working existence oracle against a specific citizen. The two `_search` rows were already `0` — confirming the reviewer's own caveat that the `userIds` injection isn't reachable on `_search` for a correctly-classified pure citizen, because the pure-citizen branch overwrites the field regardless of what was bound. So finding 2 is hardening on `_search`, but it is what makes the `_count` oracle precise.

## Changes

**1. `PrincipalScopeResolver.isPureCitizen()` — fail CLOSED.** Roles still decide first (employee marker wins; `CITIZEN` role self-scopes, so extra citizen-side roles remain fine per @subhashini-egov's point). But a principal with no employee role and no recognisable citizen role now falls back to the declared user type. Previously such a principal matched *neither* branch in `enrichSearchRequest`, `userIds` stayed empty, and `PGRQueryBuilder` dropped the `accountId IN (userIds)` clause entirely — reopening #1071 via a different trigger.

This also answers the two questions I raised above: the `type == CITIZEN` check is back, but as a *fallback* rather than a precondition, which satisfies the extra-roles concern and the fail-open concern at once. `COMMON_EMPLOYEE` is still treated as an employee marker alongside `EMPLOYEE`.

**2. `userIds` is never trusted from the client.** Confirmed the mechanism rather than assuming it: `/_search`, `/_plainsearch` and `/_count` all bind `RequestSearchCriteria` via `@ModelAttribute`, and a grep across `pgr-services` finds no `@InitBinder`/`setDisallowedFields` blocklisting the field — so `@JsonIgnore` genuinely has no effect on that path. `userIds` is now cleared unconditionally before scoping, so ownership holds by construction rather than relying on a later branch happening to overwrite it.

**3. `/_count` applies the same scoping.** The scoping half is extracted into `EnrichmentService.scopeSearchCriteria()` and reused by `PGRService.count()`.

Worth calling out why it's a *split* and not just a call to `enrichSearchRequest`: `getCountQuery` wraps the search query **including its `LIMIT`**, so applying the pagination defaults inside `count()` would have silently capped every count at the default limit. The ADMIN regression row below (50, uncapped) is what proves that split was necessary.

## No regression

| Check | Result |
|---|---|
| victim citizen sees their own complaint | `_search` 1, `_count` 1 |
| ADMIN `_count` (tenant) | 50 — uncapped |
| ADMIN `_search?limit=5` | 5 |
| ADMIN `_count?mobileNumber=<victim>` | 1 |

## Tests

Adds `PrincipalScopeResolverTest` (11 cases: role precedence, extra citizen roles, employee-also-citizen, case/whitespace, null role entries, null `RequestInfo`, and the fail-closed cases). The 3 fail-closed cases were checked against the *previous* resolver and fail there, so they are real regression guards rather than tests that merely agree with the current code. Full suite 21/21 green.

singhalkarun:
Merged `develop` into this branch (33af85c7) — conflicts resolved, PR is `MERGEABLE` again.

## The conflict: `PGRService.count()`

Two independent changes landed in the same method:

- **develop @ 439d1ef7** — resolves `assignee` → `serviceRequestIds` so `_count` matches `_search` for assignee-scoped queries (the My-tab badge).
- **this branch** — applies record-level ownership scoping to `_count`.

They touch different criteria fields, so **both are kept** rather than picking a side. Ownership scoping runs **first**, so it is pinned before the assignee block's early `return 0`.

## Worth flagging: develop already fixed the `userIds` binding vector

**dde0bfed** independently closed the `@ModelAttribute` binding gap at the controller layer, with an `@InitBinder` disallowing `userIds*` / `serviceRequestIds*` / `isPlainSearch*` — the same issue raised in review here, fixed one layer up. `RequestsApiController` is the only controller binding `RequestSearchCriteria`, so that fully closes the HTTP vector.

That demotes this branch's service-layer `criteria.setUserIds(null)` from *sole guard* to *backstop*. I kept it deliberately — an entry point added later without that binder config would otherwise go unscoped, which is the "ownership by construction, not by convention" point. But the code comment asserted that nothing blocklists the field, which the merge made false, so it has been reworded to describe the layering. A stale comment claiming a vulnerability that no longer exists is worse than no comment.

No other conflicts: develop did not touch `EnrichmentService` or `PrincipalScopeResolver`. The branch's own diff vs `develop` is still only the 4 intended files.

## Re-validated on the merged code (mz.maputo)

Re-ran everything, since `count()` now interleaves both changes:

| Check | Result |
|---|---|
| attacker `_count` / `_count?userIds=<victim>` / `_search` | 0 / 0 / 0 — leaks stay closed |
| victim citizen own records | `_search` 1, `_count` 1 |
| ADMIN `_count` (tenant) | 50 — uncapped |
| ADMIN `_search?limit=5` | 5 |
| **develop's assignee parity** | `_count?assignee=<real>` **7** == `_search?assignee=<real>` **7** |

The assignee row is worth calling out: my first probe used a citizen uuid and returned `0`, which is ambiguous between "correct" and "I broke 439d1ef7". Re-run against a *real* assignee uuid, both endpoints agree at 7 — i.e. the parity that commit exists to establish still holds under the new ordering.

78/78 tests pass (2 pre-existing skips).

@subhashini-egov this is ready for re-review whenever you have a moment — the `CHANGES_REQUESTED` from 14 Jul is the only thing blocking now. The two questions in my earlier comment (`COMMON_EMPLOYEE` as an employee marker, and retaining the `type == CITIZEN` guard) are resolved in code: the type check is back as a *fallback* rather than a precondition, which addresses your extra-roles point and the fail-open finding together.

subhashini-egov (approved):

subhashini-egov (approved):

PR #1101, opened 2026-07-08
fix(kong): emit InvalidAccessTokenException on expired session so digit-ui re-logs in

singhalkarun:
## Problem
digit-ui keys its "session expired → redirect to login" behaviour off receiving an **`InvalidAccessTokenException`** error envelope from the gateway (its axios interceptor checks `err.response.data.Errors[].message` for that string). The Spring Cloud Gateway (K8s) emits it, so re-login works there. The compose **Kong** auth pre-function did **not**: on an invalid/expired token it set `userInfo=nil` and **forwarded** the request, so the backend NPE'd (HTTP 400) and the UI rendered blank/error screens instead of redirecting to login.

Found during the compose-vs-K8s deployment-parity review (item #4). UX gap, not a security hole (the token is genuinely rejected either way).

## Fix
In the auth-enrichment `pre-function`, when a token is **present** but `/user/_details` rejects it (non-200), return the egov `InvalidAccessTokenException` **401** envelope. Two guards keep it safe:
- The **no-token** branch already returns early → open/anonymous endpoints are unaffected.
- A **transient `/user/_details` outage** (no HTTP response at all) does **not** force logout — only an actual token rejection does.

## Validation (live, on the running compose stack)
| Case | Before | After |
|---|---|---|
| Invalid / expired token | HTTP 400 `NullPointerException`, no signal | **HTTP 401** `InvalidAccessTokenException` ✅ |
| Valid token | 200 | **200** `successful` (unaffected) |
| No token, open endpoint (mdms) | 200 | **200** (no leak) |

The returned envelope: `{"ResponseInfo":null,"Errors":[{"code":"InvalidAccessTokenException","message":"InvalidAccessTokenException","description":"Invalid access token or access token expired for the request"}]}` — matches what the UI interceptor checks.

Compose-only change (K8s already behaves this way). A full gateway-side RBAC layer (parity #5) would also cover this.

### DIFF
--- a/local-setup/kong/kong.yml
+++ b/local-setup/kong/kong.yml
@@ -82,11 +82,35 @@ plugins:
           headers = { ["Content-Type"] = "application/json" },
         })
         local resolved = nil
+        local token_rejected = false
         if res and res.status == 200 then
           local ok2, user = pcall(cjson.decode, res.body)
-          if ok2 and type(user) == "table" and (user["uuid"] or user["userName"]) then resolved = user end
+          if ok2 and type(user) == "table" and (user["uuid"] or user["userName"]) then
+            resolved = user
+          else
+            token_rejected = true  -- 200 but no resolvable user
+          end
+        elseif res then
+          token_rejected = true    -- /user/_details rejected the token (invalid / expired)
         else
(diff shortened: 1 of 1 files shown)


subhashini-egov (approved):
Good catch

Issue #1090, opened 2026-07-08
default-data-handler tenant onboarding is broken in two independent layers (/defaultdata/setup)


singhalkarun:
Surfaced while validating the K8s data-bootstrap path end-to-end (Compose-vs-K8s parity work). A fresh tenant onboarded via `default-data-handler` (DDH) is only **partially** provisioned, and DDH reports success anyway.

### Summary
`POST /default-data-handler/defaultdata/setup` (module=PGR) is broken at **two independent layers**, one hidden behind the other:

**Bug 1 — dead `MDMS.json` reference (crash before anything runs)**
`DataHandlerService.addMdmsData()` loads `classpath:MDMS.json`:
```java
Resource resource = resourceLoader.getResource("classpath:MDMS.json");
```
That file was deleted in commit `da0d65a4` ("move data-handler service to utilities and removed unwanted files from config") when MDMS data was reorganized into `mdmsData/`. The loader was left behind →
```
400 CustomException: Error reading or mapping JSON file:
class path resource [MDMS.json] cannot be opened because it does not exist
```
`addMdmsData()` runs **before** `setupDefaultData()` in `DataHandlerController`, so the whole endpoint aborts and the actual seed logic never executes. (The old `MDMS.json` contained only 4 orphaned `SUPERUSER` roleactions — actionids 2200/2201/2203/2204 — that aren't in the actions master, aren't in the dump, and aren't in the current `mdmsData/` bundle. So the data is obsolete; the call should be removed.)

**Bug 2 — missing mdms-v2 endpoint (revealed once Bug 1 is bypassed)**
With `addMdmsData` no-op'd, `setupDefaultData` runs and `createDefaultData()` → `MdmsV2Util.createDefaultMdmsData()` POSTs to:
```
egov.mdms.default.data.create.endpoint = /mdms-v2/defaultdata/_create
```
…which does **not exist** on the deployed mdms-v2 →
```
400 NoResourceFoundException: No static resource defaultdata/_create
  at MdmsV2Util.createDefaultMdmsData(MdmsV2Util.java:37)
```
[...]

PR #1104, opened 2026-07-08
feat(kong): gateway RBAC phase 1 — path classification + authn enforcement (#5)

singhalkarun:
Phase 1 of gateway RBAC on the compose Kong gateway — see the design doc
[`docs/parity-item5-gateway-rbac-design.md`](../blob/develop/docs/parity-item5-gateway-rbac-design.md).

> **Stacked on #1101 (item #4).** This extends the same auth `pre-function`, so the
> diff currently includes #1101's commit. Rebase onto `develop` once #1101 merges.

## What this does
Adds the K8s Spring-gateway request model to the Kong auth `pre-function`:
1. **Classify** every request as **auth-optional** (open ∪ mixed) or **PROTECTED**.
2. On **protected** paths, **require a valid token** (unauthenticated → 401).

It does **not** call `egov-accesscontrol` yet — action-level RBAC is **Phase 2**.
(In Phase 1, open and mixed collapse into one "auth-optional / RBAC-exempt" set;
Phase 2 keeps them exempt while adding the `_authorize` call for protected paths.)

## Ships in AUDIT mode (safe by default)
`ENFORCE_UNAUTH = false`: unauthenticated hits on protected paths are **logged**
(`RBAC-audit(#5)… would 401`), **not rejected**. So merging is a **no-op** until an
operator flips `ENFORCE_UNAUTH = true` after an observation period — first confirming
no logged warning is actually a legitimate anonymous call (if it is, add its path to
`AUTH_OPTIONAL`). This is the audit-first rollout the design doc recommends.

## Whitelist
`AUTH_OPTIONAL` is synced from the Spring gateway whitelists (`env.yaml`
`egov-open-endpoints-whitelist` / `egov-mixed-mode-endpoints-whitelist`) plus the
`boundary-service` search paths CCRS migrated to in #1. Keeping the two in sync is a
known maintenance concern — Phase 3 formalizes a single source of truth.

## Validation (live)
| Case | Enforce mode | Audit mode (default) |
|---|---|---|
| login / mdms / localization / boundary / otp-send (anonymous) | 200 | 200 |
| pgr `_search` / `_create`, **no token** | **401** | passes to backend + logs would-401 |
| pgr `_search`, **valid token** | 200 | 200 |

## Scope / limitations
- POST-request scope (the token lives in the JSON body); GET passes through, matching
  the existing filter. A GET whitelist can follow.
- Depends on `ACCESSCONTROL-ROLEACTIONS` being seeded for Phase 2 (fine on the dump; see #1090).

## Follow-ups
- **Phase 2**: call `egov-accesscontrol /access/v1/actions/_authorize` for protected paths (pin the exact contract first).
- **Phase 3**: single source of truth for the whitelists + CI equality check.

## Summary by CodeRabbit

* **New Features**
  * Requests are now classified into protected and optional paths, with stricter authentication requirements for protected endpoints.
  * Requests lacking valid authentication information now follow clearer unauthenticated behavior.
* **Bug Fixes**
  * Protected requests with missing/empty/malformed request data now fail more consistently instead of proceeding silently.
  * Token validation now distinguishes invalid-token scenarios from upstream reachability issues, returning a clearer 401 response when access tokens are invalid.

### DIFF
--- a/local-setup/kong/kong.yml
+++ b/local-setup/kong/kong.yml
@@ -66,38 +66,47 @@ plugins:
         -- Mirrors the K8s Spring gateway's open / mixed-mode / protected buckets. OPEN and MIXED
         -- are auth-OPTIONAL (anonymous allowed and NOT action-RBAC'd); everything else is PROTECTED
         -- (a valid token is required). Phase 1 does NOT call egov-accesscontrol yet — action-level
-        -- RBAC is Phase 2. The AUTH_OPTIONAL set = open ∪ mixed, kept in sync with the Spring
-        -- gateway whitelists (env.yaml egov-open/-mixed-mode-endpoints-whitelist). See
-        -- docs/parity-item5-gateway-rbac-design.md.
+        -- RBAC is Phase 2.
+        --
+        -- EXACT match (set membership), mirroring the Spring gateway's
+        -- openEndpointsWhitelist.contains(getPath().value()) — NOT prefix (a prefix would over-open,
+        -- e.g. anonymous /localization/messages/v1/_upsert). AUTH_OPTIONAL = open ∪ mixed, kept
+        -- VERBATIM-in-sync with env.yaml egov-open/-mixed-mode-endpoints-whitelist (+ the
+        -- boundary-service search paths CCRS migrated to in #1); a CI check enforces the two match
+        -- (Phase 3). See docs/parity-item5-gateway-rbac-design.md.
         --
         -- Ships in AUDIT mode: unauthenticated hits on protected paths are LOGGED, not rejected.
         -- Flip ENFORCE_UNAUTH=true after an observation period (first confirm no RBAC-audit warning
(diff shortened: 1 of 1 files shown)


singhalkarun:
Addressed the CodeRabbit findings:
- **Prefix matching over-opening** (`/filestore/v1/file/...`) — already fixed: `AUTH_OPTIONAL` is now an **exact-match set** mirroring the Spring gateway's `List.contains(getPath())` (`b38db561`, after decompiling the gateway). This closed a real over-exposure (prefix `/localization/messages` would have opened `…/_upsert`).
- **Fail-open on `/user/_details` unreachable** — now fails **closed** on protected paths when enforcing (`3cf8ad77`): a transient resolve failure returns `503 AuthServiceUnavailable` for protected requests (never forwarded server-unvouched); optional paths and audit mode fall through.

subhashini-egov (changes requested):
Kindly resolve Coderabbit comments

PR #1114, opened 2026-07-09
fix(pgr): create-form validation sweep — future dates, description 20–1000, (Optional) labels (CCSD-1952/1956/1955)

Hari-egov:
Three small create-form tickets, one coherent sweep (same files):

## CCSD-1952 — Date of fact shall not allow future dates
- **Employee**: `PGRDatePicker` supports `populators.maxDate: "today"` — future cells grey out and are unclickable (`pick()` also hard-blocks). Enabled for every extended-attribute date field.
- **Citizen**: native date input gets `max=today` plus an onChange guard (the `max` attribute constrains only the popup, not typed values).

## CCSD-1956 — Description 20–1000 characters (citizen + employee)
- **Citizen**: NEXT requires trimmed length ≥ 20; the character counter becomes a red "Minimum 20 characters" hint until valid; 1000-char cap unchanged.
- **Employee**: `minLength: 20` + an all-whitespace-rejecting pattern (`minLength` alone would accept 20 spaces); error message key `CS_DESC_MIN_CHARS`.

## CCSD-1955 — "(Optional)" on all optional fields
- **Employee**: every non-mandatory field's label gets the localized `CS_OPTIONAL_SUFFIX` appended (labels are pre-translated in `updatedConfig`, so the composer's `t()` passes them through).
- **Citizen**: non-mandatory dynamic fields get the same suffix; landmark/photos/reporter fields already had it.

## Data note
Seeded at `mz`/`en_IN`: `CS_DESC_MIN_CHARS` = "Description must be 20–1000 characters" (new), `CS_OPTIONAL_SUFFIX` already existed. Other envs need the new key (pt pending with the localization backlog).

## Testing (local, deployed)
- Employee + citizen date-of-fact: future dates unpickable/blocked; past/today fine.
- Description: 19 chars blocks with hint/error; 20 passes; 1000 cap holds; whitespace-only rejected.
- Optional labels: employee address/map/postal/landmark + non-mandatory ext-attrs and citizen optional dynamic fields show "(Optional)"; mandatory fields unchanged (asterisk only).

### DIFF
--- a/digit-ui-esbuild/products/pgr/src/components/PGRDatePicker.js
+++ b/digit-ui-esbuild/products/pgr/src/components/PGRDatePicker.js
@@ -162,11 +162,13 @@ const PGRDatePicker = (props) => {
   };
 
   return (
-    <div className="pgr-datepicker" style={s.wrap} ref={ref}>
+    // Callers can override sizing: the employee form caps inputs at 37.5rem,
+    // while the citizen v2 wizard runs fields full-width (style/fieldStyle).
+    <div className="pgr-datepicker" style={{ ...s.wrap, ...(props?.style || {}) }} ref={ref}>
       <style>{FOCUS_CSS}</style>
       <div
         className="pgr-datepicker-field"
-        style={s.field}
+        style={{ ...s.field, ...(props?.fieldStyle || {}) }}
         onClick={toggle}
         role="button"
         tabIndex={disabled ? -1 : 0}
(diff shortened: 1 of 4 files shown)


Hari-egov:
Added commit 2 per review: the citizen Date-of-Fact now renders the **same PGRDatePicker calendar as the employee form** (future days greyed out and unclickable) instead of the native date input — the native `max` enforcement varies by browser and a silently ignored pick read as broken. Both surfaces now behave identically.

PR #1117, opened 2026-07-09
fix(pgr): citizen wizard re-entry starts at step 1 — F5 still resumes the exact step

Hari-egov:
## Problem (user feedback on #1093 draft persistence)

Fill the wizard partway → sidebar **Home** → **File a Complaint**: the citizen landed on **page 3** (the persisted step position), which reads as broken navigation. Fresh entry should start at page 1.

## Fix

Unmount cleanup rewrites the saved draft's `stepIndex` to `0` — so any **in-app navigation away** (Home, back, links) makes the next entry start at step 1 **with the answers still prefilled**. A **hard refresh** doesn't run React cleanup, so mid-flow F5 still restores the exact step — the original purpose of persisting the position. Successful create still deletes the draft outright (a guard stops the cleanup from resurrecting it), and the failure path ("Try Again") now also re-enters at step 1 with answers intact.

## Testing (local, deployed)

- Fill steps 1–2 → Home → File a Complaint → **step 1**, previously entered values present when stepping forward.
- Fill steps 1–2 → **F5** → resumes on the same step with values.
- Submit success → next visit is a clean, empty wizard.

### DIFF
--- a/digit-ui-esbuild/products/pgr/src/pages/citizen/Create/CreatePGRFlowV2.tsx
+++ b/digit-ui-esbuild/products/pgr/src/pages/citizen/Create/CreatePGRFlowV2.tsx
@@ -305,6 +305,21 @@ const STEPS = [
 // key, which Module.js clears on every citizen-home mount.
 const CREATE_DRAFT_KEY = "PGR_CREATE_CITIZEN_DRAFT";
 
+// The saved step POSITION may only be restored when the document itself was
+// RELOADED (mid-flow F5) — and only by the first wizard mount after it. Any
+// other entry (SPA links, and sidebar items that hard-navigate like "Citizen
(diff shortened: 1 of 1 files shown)


Hari-egov:
Commit 2: entries via **hard navigation** (the sidebar "Citizen Complaint Resolution System" item does a full page load, so unmount cleanup never runs) also start at step 1 now. Step restore is gated on the document's navigation type being `reload` and consumed on first mount — F5 mid-flow still resumes exactly, every other path enters at step 1 with answers kept.

Hari-egov:
JIRA: CCSD-1969 (created retroactively per the new PR↔Jira linkage policy; ticket carries problem/expected/acceptance criteria and validation notes).

PR #1118, opened 2026-07-09
fix(libraries): city localization overlay back-fills bootstrap-cached modules (dept names in assignee dropdown)

Hari-egov:
## Problem

The assignee dropdown's department group headers and option suffixes showed raw codes (`BUSINESS_OPS`, `CULTURE_OPS`) even though the display names ("Business & Commerce Operations", …) are seeded — at the **city tenant** in `rainmaker-common`.

## Cause

The city overlay from #1096 applies only to modules being fetched **fresh**. `rainmaker-common` (and the rest of the bootstrap set) loads on the **login page**, where no city tenant exists yet — so it cached state-only, and every subsequent `getLocale` call was a cache hit that skipped the overlay entirely. (Module-entry loads like `rainmaker-pgr` happen post-login, which is why complaint-type labels worked while department names didn't.)

## Fix

`getLocale` keeps a per-`(locale, module)` **city marker**: any *cached* module whose marker doesn't match the current city tenant gets just its city copy fetched and merged — city wins per code, and the merge is persisted into the module cache so future cache hits include it. Markers are written even when the city has no extra keys (no re-query per call), and a ULB switch re-overlays for the new city automatically.

## Testing (local, deployed)

- Fresh incognito → login (bootstrap caches at state) → open a complaint → Assign: department groups show "Business & Commerce Operations" / "Culture Operations" instead of codes.
- Language switch and repeated navigations don't re-fetch the overlay (marker hit).
- Complaint-type labels, state labels etc. from #1096 unchanged.

### DIFF
--- a/digit-ui-esbuild/packages/libraries/src/services/elements/Localization/service.js
+++ b/digit-ui-esbuild/packages/libraries/src/services/elements/Localization/service.js
@@ -94,20 +94,32 @@ export const LocalizationService = {
     // fell back to en_IN, so picking Swahili in the UI did nothing.
     // Locale codes already carry their region (xx_YY), so the append is
     // never correct. Drop it.
+    // City overlay: modules are loaded at the STATE tenant, but data
+    // onboarding seeds label keys at the LOGGED-IN city tenant
+    // (COMPLAINT_HIERARCHY.*, COMMON_MASTERS_DEPARTMENT_*, boundary level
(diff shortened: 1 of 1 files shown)


Hari-egov:
JIRA: CCSD-1970 (created retroactively per the new PR↔Jira linkage policy; ticket carries problem/expected/acceptance criteria and validation notes).

PR #1125, opened 2026-07-09
fix(ansible): derive INTERNAL_USER/ADMIN seed mobile numbers from mobileNumberRegex

### LINKED ISSUE #1126: derive_valid_mobile / deriveValidMobile still collide on multi-char literal-prefix regexes (e.g. ^77[0-9]{6}$)
## Summary

`derive_valid_mobile()` (`local-setup/ansible/filter_plugins/mobile.py`) and its TS mirror `deriveValidMobile()` (`digit-mcp/src/tools/mdms-tenant.ts`) generate a placeholder mobile number for auto-created system/ADMIN/HRMS users when the hardcoded default (`9999999999`, `8888888888`, ...) doesn't satisfy a tenant's `mobileNumberRegex`. Both functions synthesize candidates as `lead-digit + repeated-fill-digit`, then search across candidate lengths and lead/fill combinations.

PR #1125 fixed two real bugs in this derivation:
- the TS version only tried candidate lengths `[n, n+1, n-1]` (default `n=10`), so regexes needing a length far from 10 (e.g. Djibouti's `^77[0-9]{6}$`, 8 digits) never matched and fell through to an invalid placeholder → `INVALID_MOBILE_NUMBER` from egov-user.
- the exhaustive fallback sweep iterated digits `0..9` in the same fixed order regardless of caller, so two callers with different `preferred` values that both fail the regex could converge on the *same* derived number (addressed per CodeRabbit review by seeding the digit order from each caller's `preferred`).

## Remaining gap

The digit-order fix only helps when a regex admits more than one `lead+fill` combination at a given length. For a regex with a **fixed multi-character literal prefix** — e.g. `^77[0-9]{6}$` (Djibouti-style) — the candidate shape `lead + fill*(n-1)` admits exactly **one** valid combination (`lead='7'`, `fill='7'` → `77777777`) no matter what order the search tries digits in. Two different callers on the same tenant (e.g. `ADMIN` and `INTERNAL_USER`) both converge on that single candidate and get issued the identical mobile number.

Verified via a live deploy against `mumbai` (host_vars configured with `mobileNumberRegex: "^77[0-9]{6}$"`): both `ADMIN` and `INTERNAL_USER` were created with `mobileNumber: "77777777"`.

Not currently a hard failure — `egov-user` doesn't appear to enforce mobile-number uniqueness across these two system accounts — but two distinct users sharing a mobile number is fragile: any future uniqueness constraint, OTP-based recovery flow, or lookup-by-mobile logic would misbehave.

## Suggested fix

The candidate-generation scheme itself needs to widen beyond "one lead digit + one repeated fill digit" to admit multiple valid candidates under literal-prefix regexes — e.g. by also trying candidates that vary a suffix digit or embed more of the `preferred` value's digits, rather than collapsing the whole tail to a single repeated character. Should preserve the existing behavior for regexes that already work today.

## Where to look

- `local-setup/ansible/filter_plugins/mobile.py` — `derive_valid_mobile()`, `_seeded_digit_order()`
- `digit-mcp/src/tools/mdms-tenant.ts` — `deriveValidMobile()`, `seededDigitOrder()`
- Related: PR #1125, original CodeRabbit review comment on `mobile.py`

vinothrallapalli-eGov:
## Summary
- `core_mobile_configs` was migrated to the current schema (`countryCode` + `mobileNumberRegex` only, enforced by the preflight check in `playbook-deploy.yml`), but the post-bootstrap `INTERNAL_USER` and `ADMIN` seed tasks still gated mobile-number construction on the removed legacy fields `mobileNumberAllowedStartingCharacters`/`mobileNumberLength`. Since those are never `defined` under the current schema, both tasks always fell back to a hardcoded `9999999999`/`8888888888`, which fails `_createnovalidate` on any tenant whose `mobileNumberRegex` rejects it (e.g. Kenya, Djibouti, Tamil Nadu).
- Both tasks now derive the mobile number via the existing `derive_valid_mobile` Jinja filter (`filter_plugins/mobile.py`), the same way the MCP tenant-bootstrap calls already do.
- Fixed a latent bug in `derive_valid_mobile`'s fallback search: once a `preferred` value failed the regex, the search ignored `preferred` entirely, so INTERNAL_USER (`9999999999`) and ADMIN (`8888888888`) would derive the *same* fallback number on some tenants — defeating the "distinct mobile to avoid collision" intent already documented in the playbook comments. The fallback now first sweeps only the lead digit while keeping `preferred`'s fill digit, before falling back to the full lead×fill sweep.

## Test plan
- [x] `python3 -c "import yaml; yaml.safe_load(open('local-setup/ansible/playbook-deploy.yml'))"` — YAML still parses
- [x] `ansible-playbook playbook-deploy.yml --syntax-check` — passes
- [x] Verified `derive_valid_mobile` against Kenya (`^[17][0-9]{8}$`), Djibouti (`^8[0-9]{8}$`), Tamil Nadu (`^77[0-9]{6}$`), India default (`^[6-9][0-9]{9}$`), and the optional-prefix case (`^0?[17][0-9]{8}$`) — INTERNAL_USER/ADMIN derive distinct, regex-valid numbers in all but the Tamil Nadu multi-literal-digit-prefix case (inherent limitation of the lead+fill candidate shape, pre-existing, not introduced by this change)
- [ ] Run a real `deploy.sh` against a non-`pg` tenant with a stricter `mobileNumberRegex` (e.g. Kenya/bomet) to confirm INTERNAL_USER/ADMIN seed succeeds end-to-end

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mobile number generation during setup to be deterministic and better aligned with the current mobile validation schema.
  * Reduced the chance of setup account mobile-number collisions by deriving distinct fallback values for default system accounts.
  * Strengthened preflight checks to fail fast with clearer error details when required mobile configuration keys are missing or regex validation could be skipped.

### DIFF
--- a/local-setup/ansible/filter_plugins/mobile.py
+++ b/local-setup/ansible/filter_plugins/mobile.py
@@ -9,6 +9,26 @@ host_vars fields.
 import re
 
 
+def _seeded_digit_order(seed):
+    """Rotate '0123456789' to start at seed's leading digit.
+
+    Two callers deriving a fallback mobile number for the same regex with
+    different `preferred` values (e.g. INTERNAL_USER's '9999999999' vs
+    ADMIN's '8888888888') should diverge instead of both walking the
+    exhaustive lead x fill sweep in the same 0..9 order and returning the
+    same first match. Seeding the order from `preferred` makes them diverge
+    whenever the regex admits more than one lead/fill combination at a given
+    length; regexes with a fixed multi-character literal prefix (e.g.
+    '^77[0-9]{6}$') admit exactly one combination regardless of search
+    order, so those still collide.
+    """
+    digits = "0123456789"
+    if not seed or seed[0] not in digits:
+        return digits
+    i = digits.index(seed[0])
+    return digits[i:] + digits[:i]
+
+
 def derive_valid_mobile(regex, length=10, preferred=None):
     try:
         compiled = re.compile(regex) if regex else None
@@ -37,12 +57,14 @@ def derive_valid_mobile(regex, length=10, preferred=None):
     # vs ADMIN) must not collide on the same derived mobile number just
     # because both preferred values failed the regex and fell through to an
     # identical lead x fill sweep below.
+    digit_order = _seeded_digit_order(preferred)
+
     preferred_fill = preferred[-1] if preferred else None
     if preferred_fill:
         for try_len in lengths:
             if try_len <= 0:
                 continue
-            for lead in "0123456789":
+            for lead in digit_order:
                 candidate = lead + preferred_fill * (try_len - 1)
                 if len(candidate) == try_len and matches(candidate):
                     return candidate
@@ -50,8 +72,8 @@ def derive_valid_mobile(regex, length=10, preferred=None):
     for try_len in lengths:
         if try_len <= 0:
             continue
-        for lead in "0123456789":
-            for fill in "0123456789":
+        for lead in digit_order:
+            for fill in digit_order:
                 candidate = lead + fill * (try_len - 1)
                 if len(candidate) == try_len and matches(candidate):
                     return candidate
--- a/local-setup/ansible/filter_plugins/mobile.py
+++ b/local-setup/ansible/filter_plugins/mobile.py
@@ -50,30 +50,24 @@ def derive_valid_mobile(regex, length=10, preferred=None):
     # falling through to an unmatched placeholder.
     lengths = sorted(set(range(6, 16)) | {n, n + 1, n - 1}, key=lambda x: abs(x - n))
 
-    # If the caller's preferred number didn't match outright, first retry
-    # candidates that keep its fill digit (e.g. '9' for '9999999999', '8' for
-    # '8888888888') and only vary the lead digit. Two callers seeding
-    # different usernames with different `preferred` values (INTERNAL_USER
-    # vs ADMIN) must not collide on the same derived mobile number just
-    # because both preferred values failed the regex and fell through to an
-    # identical lead x fill sweep below.
+    # Within each length, try the caller's own fill digit first (e.g. '9' for
+    # '9999999999', '8' for '8888888888') before the rest of digit_order, so
+    # two callers seeding different usernames with different `preferred`
+    # values (INTERNAL_USER vs ADMIN) diverge instead of both landing on the
+    # same lead x fill candidate. This must stay a single pass over `lengths`
+    # — trying preferred_fill across *every* length before the fallback fill
+    # sweep gets a turn would let a farther length "win" over a closer one
+    # that only works with a different fill, breaking the closest-length-
+    # first guarantee documented above.
     digit_order = _seeded_digit_order(preferred)
-
     preferred_fill = preferred[-1] if preferred else None
-    if preferred_fill:
-        for try_len in lengths:
-            if try_len <= 0:
-                continue
-            for lead in digit_order:
-                candidate = lead + preferred_fill * (try_len - 1)
-                if len(candidate) == try_len and matches(candidate):
-                    return candidate
+    fills = ([preferred_fill] if preferred_fill else []) + [d for d in digit_order if d != preferred_fill]
 
     for try_len in lengths:
         if try_len <= 0:
             continue
-        for lead in digit_order:
-            for fill in digit_order:
+        for fill in fills:
+            for lead in digit_order:
                 candidate = lead + fill * (try_len - 1)
                 if len(candidate) == try_len and matches(candidate):
                     return candidate
--- a/local-setup/ansible/filter_plugins/mobile.py
+++ b/local-setup/ansible/filter_plugins/mobile.py
@@ -28,7 +28,26 @@ def derive_valid_mobile(regex, length=10, preferred=None):
     # mobileNumberRegex requires a length far from the hinted default (e.g. 7
     # or 12 digits) still derives a match instead of exhausting n-1/n/n+1 and
     # falling through to an unmatched placeholder.
-    for try_len in sorted(set(range(6, 16)) | {n, n + 1, n - 1}, key=lambda x: abs(x - n)):
+    lengths = sorted(set(range(6, 16)) | {n, n + 1, n - 1}, key=lambda x: abs(x - n))
+
+    # If the caller's preferred number didn't match outright, first retry
+    # candidates that keep its fill digit (e.g. '9' for '9999999999', '8' for
+    # '8888888888') and only vary the lead digit. Two callers seeding
+    # different usernames with different `preferred` values (INTERNAL_USER
+    # vs ADMIN) must not collide on the same derived mobile number just
+    # because both preferred values failed the regex and fell through to an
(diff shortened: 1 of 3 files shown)


local-setup/ansible/playbook-deploy.yml
```diff
@@ -2292,10 +2297,12 @@
     # 400s "Invalid login credentials". The tenant rule is guaranteed live by now
     # (the INTERNAL_USER seed above blocks until it is), so ensure ADMIN exists
     # with the expected password + SUPERUSER. Idempotent (DuplicateUserName ok);
-    # distinct mobile (8s vs INTERNAL_USER's 9s) to avoid any collision.
+    # distinct mobile (8s vs INTERNAL_USER's 9s, both derived via
+    # derive_valid_mobile — see the INTERNAL_USER seed task above) to avoid any
```
pradeepkumarcm-egov: This comment ("...to avoid any collision") overstates the guarantee. For any `mobileNumberRegex` with a fixed literal digit beyond position 0 — e.g. a Tamil-Nadu-style `^77[0-9]{6}$` (already mentioned in this PR's description) — both the INTERNAL_USER and ADMIN `preferred_fill` sweeps in `derive_valid_mobile` fail identically (only `lead` varies; position 1 can't be forced to the literal `7`), so both calls fall through to the same deterministic fallback sweep and derive the *identical* mobile number. Verified: `derive_valid_mobile('^77[0-9]{6}$', 10, '9999999999')` and `derive_valid_mobile('^77[0-9]{6}$', 10, '8888888888')` both return `'77777777'`. `mobile.py`'s own new `_seeded_digit_order` docstring already acknowledges this ("...so those still collide") — worth mirroring that caveat here so a future debugger of a collision on such a tenant isn't misled by this comment.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -2262,7 +2267,7 @@
...
       ansible.builtin.shell: |
-        body='{"RequestInfo":{"apiId":"digit","ver":"1.0"},"User":{"userName":"INTERNAL_USER","name":"Internal Microservice User","mobileNumber":"{{ ((core_mobile_configs.mobileNumberAllowedStartingCharacters | first) ~ ('9' * ((core_mobile_configs.mobileNumberLength | int) - 1))) if (core_mobile_configs.mobileNumberAllowedStartingCharacters is defined and core_mobile_configs.mobileNumberLength is defined) else '9999999999' }}","gender":"MALE","active":true,"type":"SYSTEM","tenantId":"{{ state_root }}","password":"System@123","roles":[{"code":"INTERNAL_MICROSERVICE_ROLE","name":"Internal Microservice Role","tenantId":"{{ state_root }}"}]}}'
+        body='{"RequestInfo":{"apiId":"digit","ver":"1.0"},"User":{"userName":"INTERNAL_USER","name":"Internal Microservice User","mobileNumber":"{{ core_mobile_configs.mobileNumberRegex | default('^[6-9][0-9]{9}$') | derive_valid_mobile(preferred='9999999999') }}","gender":"MALE","active":true,"type":"SYSTEM","tenantId":"{{ state_root }}","password":"System@123","roles":[{"code":"INTERNAL_MICROSERVICE_ROLE","name":"Internal Microservice Role","tenantId":"{{ state_root }}"}]}}'
```
pradeepkumarcm-egov, line 2279: `default('^[6-9][0-9]{9}$')` only substitutes when `mobileNumberRegex` is Jinja-*Undefined* — not when it's defined-but-empty. The preflight assert earlier in this file only checks `mobileNumberRegex is defined`, not that it's non-empty. If `core_mobile_configs.mobileNumberRegex` is ever set to `''` (passes preflight, `default()` doesn't fire), `derive_valid_mobile` receives `regex=''`, which is falsy so `compiled` stays `None`, and `matches()` then returns `True` unconditionally — the function returns the raw hardcoded `preferred` value with zero regex validation applied, silently reintroducing the original hardcoded-mobile bug this PR is fixing. Same applies at the ADMIN task below (`preferred='8888888888'`). Might be worth tightening the preflight assert to also reject an empty string.

local-setup/ansible/filter_plugins/mobile.py
```diff
@@ -28,11 +48,32 @@ def matches(candidate):
...
+    # candidates that keep its fill digit (e.g. '9' for '9999999999', '8' for
+    # '8888888888') and only vary the lead digit. Two callers seeding
+    # different usernames with different `preferred` values (INTERNAL_USER
+    # vs ADMIN) must not collide on the same derived mobile number just
+    # because both preferred values failed the regex and fell through to an
+    # identical lead x fill sweep below.
+    digit_order = _seeded_digit_order(preferred)
+
+    preferred_fill = preferred[-1] if preferred else None
+    if preferred_fill:
```
pradeepkumarcm-egov: This `preferred_fill` pass scans *every* candidate length (closest-to-`n` first) before the fallback sweep below ever runs, which can make the function return a farther-length match over a closer one that only works with a different fill digit. Verified: for `regex='^0?[17][0-9]{8}$'` (the Kenya rule mentioned elsewhere in this file), `length=10`, `preferred='9999999999'` — the pre-PR code returned the 10-digit `'0111111111'` (distance 0 from `n`); this version returns the 9-digit `'199999999'` (distance 1), because this loop exhausts `try_len=10` with fill fixed at `'9'` and succeeds at `try_len=9` before the fallback sweep gets a chance to try `try_len=10` with a different fill. Still regex-valid, just breaks the "closest length first" guarantee documented a few lines up. Nesting the preferred-fill preference inside the single sweep below (try `fill=preferred_fill` first per length, then the rest) instead of a separate full pre-pass would preserve closest-length-first while still doing the divergence.

local-setup/ansible/filter_plugins/mobile.py
```diff
@@ -28,11 +48,32 @@ def matches(candidate):
...
+        for try_len in lengths:
+            if try_len <= 0:
+                continue
+            for lead in digit_order:
+                candidate = lead + preferred_fill * (try_len - 1)
+                if len(candidate) == try_len and matches(candidate):
+                    return candidate
+
+    for try_len in lengths:
```
pradeepkumarcm-egov, line 66: Minor: this loop and the `preferred_fill` loop above it are near-identical (same `lengths`/`try_len<=0` guard/`lead` iteration), differing only in whether `fill` is fixed to `preferred_fill` or varies over `digit_order`. Could likely collapse into one loop by trying `fill in ([preferred_fill] if preferred_fill else []) + list(digit_order)` (dedup'd) — avoids scanning `lengths` twice and sidesteps the closest-length regression noted above, since the preference would be tried first *within* each length rather than across a whole separate pass over all lengths.

pradeepkumarcm-egov (commented):
Automated review pass (correctness-focused). Confirmed by executing the patched `derive_valid_mobile` against several regex shapes. Nothing here blocks the merge outright, but two items are worth a look before relying on this for a stricter-regex tenant.

pradeepkumarcm-egov (approved):

PR #1105, opened 2026-07-08
feat(kong): gateway RBAC phase 2 — action authorization via egov-accesscontrol (#5)

singhalkarun:
Phase 2 of gateway RBAC on the compose Kong gateway — see
[`docs/parity-item5-gateway-rbac-design.md`](../blob/develop/docs/parity-item5-gateway-rbac-design.md).

> **Stacked on #1104 (phase 1) → #1101 (item #4).** The diff currently includes both;
> rebase onto `develop` as they merge.

## What this does
For a **protected** path with a resolved principal (from phase 1), the Kong
`pre-function` now calls **`egov-accesscontrol /access/v1/actions/_authorize`** and
returns **403 `AccessDeniedException`** when the principal's roles may not perform the
action — the action-level RBAC the K8s Spring gateway does.

## Contract (pinned by decompiling the running `egov-accesscontrol` jar)
```
POST http://egov-accesscontrol:8090/access/v1/actions/_authorize
{ "RequestInfo": {...},
  "AuthorizationRequest": { "roles":[{"code","tenantId"}], "uri":"<path>", "tenantIds":["<tenant>"] } }
→ 200 = allow, non-200 = deny   (ResponseEntity<Void>, no body)
```
(The wrapper key is `AuthorizationRequest` with a capital A; the field is `uri`, not `actionUrl`.)

## Ships in AUDIT mode (safe default)
`ENFORCE_RBAC = false`: would-be denials are **logged** (`RBAC-audit(#5 p2)… would 403`),
not enforced. Flip `ENFORCE_RBAC = true` after an observation period. When enforcing it
**fails closed** (deny unless an explicit allow).

## Validation (live)
| Principal → action | Enforce | Audit (default) |
|---|---|---|
| CITIZEN → pgr `_search` / `_create` | 200 (allowed) | 200 |
| CITIZEN → `egov-hrms/employees/_create` | **403 denied** | 400 (passes) + logs would-403 |
| ADMIN → `egov-hrms/employees/_create` | 400 (allowed) | 400 |
| anonymous → open endpoint | 200 | 200 |

## Notes / follow-ups
- **Perf:** adds one accesscontrol call per protected request (on top of `/user/_details`).
  Response caching keyed by roles+uri+tenant is a documented follow-up.
- **Data dependency:** relies on `ACCESSCONTROL-ROLEACTIONS` being seeded (fine on the dump; see #1090).
- **Phase 3:** single source of truth for the open/mixed whitelists + a CI equality check.

[...]

Issue #1109, opened 2026-07-09
Performance benchmark: Dashboard load time at 3K / 50K / 100K records


subhashini-egov:
## Summary

Create a reproducible browser-performance baseline for the Configurable Dashboard at exactly **3K, 50K and 100K complaints**. The deliverable is the deterministic dataset/snapshot harness, Playwright benchmark suite, raw run artifacts, and a versioned benchmark report — not the optimisation of bottlenecks found.

The benchmark must answer:

> On a pinned CCRS build and reference hardware/network profile, how long does the shipped dashboard take to become usable as the complaint corpus grows from 3K to 100K records?

## Current architecture and benchmark decisions

This ticket predates the current dashboard implementation. As of 2026-08-28:

- The canonical catalog has one employee `supervisor-default` pack (11 tiles) and one anonymous `public-default` pack (8 tiles), not three Employee/Supervisor/Admin layouts.
- There is no in-app persona/dashboard-tab switch. Employee identity changes require re-login.
- Dashboard data comes from the pgr-services analytics API and its `complaint_events` / `complaint_facts` grains.
- Production render/query instrumentation from #1110 already exists and must be reused for correlation.
- Playwright + Chromium CDP is the benchmark runner. k6 may be reused for bounded API/bootstrap work, but k6 is not the source of browser paint or memory measurements.
- The target branch and SHA must be pinned in every result because `master` and `develop` currently differ in dashboard access/batching behavior.

## Execution targets and comparability

Dhruv PR #1848 establishes Bomet as the comparison shape: an unthrottled 16 vCPU / 30.6 GiB KVM guest running the full 59-container DIGIT stack. The fixed-dataset repeat used 2,525 complaints. Its `cpu-8` data is a per-service quota profile on that same 30 GiB host, not a measurement from an 8-vCPU machine.

| Target | Allowed data path | Use | Comparable with the Bomet result? |
|---|---|---|---|
| `bomet-live` | Read-only existing corpus | As-deployed browser baseline | Yes, labelled with the current corpus and runtime manifest |
| `bomet-snapshot` | Full DB backup restored to a disposable DB on the Bomet host; exact fixture with teardown | Primary dashboard scale curve | Yes; same host/stack, temporary PGR datasource and scheduler overrides recorded |
| `bomet-clone` | Exact fixture on an external clone | Fallback when Bomet cannot enter maintenance | Yes, after host/service/image/config parity passes |
[...]

Issue #1108, opened 2026-07-09
[P0] Dashboard: localise to Portuguese and populate language dropdown from DIGIT-configured languages


subhashini-egov:
## Summary

The Configurable Dashboard currently renders in **English only**. It needs to be fully localised, starting with **Portuguese**, and the language selector must dynamically enumerate every language configured on the DIGIT deployment (via MDMS / stateInfo) rather than hardcoding a language list.

Priority: **P0 · Urgent**.

## Current behaviour

- Dashboard UI shipping only `en_IN` (or `en_US`) strings — every widget label, filter label, KPI title, table header, empty-state message, and error surface is English.
- Language dropdown in the Dashboard header either does not exist, or has a hardcoded static list that does not reflect the tenant's actual configured languages.
- Portuguese-speaking operators cannot use the dashboard.

## Expected behaviour

1. **All dashboard-specific strings localised to Portuguese** — widget titles, filter labels, KPI names, chart legends, table headers, empty states, tooltips, error messages, date/number formats where applicable.
2. **Language dropdown reads from DIGIT** — populate the picker from the tenant's languages configured in MDMS (`common-masters.StateInfo.languages` or the equivalent path that citizen-side landing-dispatch already consumes). No hardcoded language arrays.
3. **Language persistence** — selection carries through page reloads via existing `locale` localStorage/session pattern used elsewhere in digit-ui.
4. **Fallback** — if a localisation key is missing in the selected locale, fall back to `en_IN` gracefully (don't render the raw key).

## Acceptance criteria

- [x] Portuguese (`pt_MZ` — Mozambique-Portuguese target — or the exact locale code approved by product) added to the localisation JSON bundles for all dashboard-specific keys.
- [x] Dashboard header language selector queries MDMS languages at mount and renders every enabled language as a dropdown option.
- [x] Switching to Portuguese re-renders the dashboard with translated strings across every widget in the default employee/supervisor persona layout.
- [x] Switching languages does not reload the page or reset filter state.
- [x] No raw `LOCALIZATION_KEYS` visible in any language (regression guard).
- [ ] Works on the Maputo tenant on `bometfeedbackhub.digit.org` (or the equivalent Portuguese-target deployment).
- [x] Localisation namespace aligned: the approved implementation uses `rainmaker-dashboard`, with `rainmaker-common` explicitly loaded for shared dimension labels.

## Notes for implementer

[...]

PR #1128, opened 2026-07-09
feat(ci): gateway RBAC phase 3 — one whitelist + parity CI check

singhalkarun:
Phase 3 (final) of gateway RBAC — see [`docs/parity-item5-gateway-rbac-design.md`](../blob/develop/docs/parity-item5-gateway-rbac-design.md).

> **Stacked on #1105 (P2) → #1104 (P1) → #1101 (item #4).** Rebase onto `develop` as they merge.

## Goal
The auth-optional whitelist lived in two places (Kong `kong.yml` + K8s `env.yaml`) that could **drift** — and drift is security-critical (an endpoint anonymous on one stack, protected on the other). This makes them **one identical list** and adds a CI guard so they can never silently diverge.

## What decompiling the gateway unlocked
The Spring gateway matches by `openEndpointsWhitelist.contains(getPath().value())` — **exact membership**. Once Phase 1 was corrected to exact match too (in #1104), the two lists can be **byte-identical sets**, so the CI check is a trivial set-equality — no prefix-normalizer needed.

## Changes
- **`.github/scripts/check-gateway-whitelist-parity.py`** — parses Kong's `AUTH_OPTIONAL` set and the K8s `egov-open`/`-mixed-mode-endpoints-whitelist`, asserts they're equal sets, and names any offending path.
- **`.github/workflows/gateway-whitelist-parity.yml`** — runs it on PRs touching either file.
- **`env.yaml`** — aligned to Kong exactly: removed the disabled DSS `dashboard-analytics` paths from `open`; added the `boundary-service` search paths CCRS migrated to (item #1) to `mixed`. **Both lists are now 39 identical entries.**
- Deleted **`local-setup/kong/auth-enrichment.lua`** — a dead, unwired duplicate of the inline `kong.yml` pre-function.

## Verification
- Check **passes** when aligned: `OK: gateway auth-optional whitelists match (39 entries).`
- Check **fails** on injected drift, printing the offending path and exit 1.

## This completes item #5
Phase 1 (#1104) authn + classification · Phase 2 (#1105) accesscontrol RBAC · **Phase 3 (this) single-source + CI guard**. The compose gateway now matches the K8s gateway's authn + action-RBAC model, and the whitelists can't drift.

## Summary by CodeRabbit

- **New Features**
  - Added configurable gateway authentication and role-based access enforcement.
  - Added support for audit-only or fail-closed handling of unauthenticated requests, authorization failures, and temporary service issues.
  - Improved protection against client-supplied identity information when authentication is missing or invalid.
- **Bug Fixes**
  - Improved handling of invalid tokens, unavailable authorization services, and missing tenant information.
- **Documentation**
  - Updated local setup documentation to reflect the consolidated gateway configuration.


subhashini-egov (commented):
## Review — feat(ci): gateway RBAC phase 3 — one whitelist + parity CI check

### Overview

Solves a real security concern: the compose Kong `AUTH_OPTIONAL` set and the K8s Spring gateway `open ∪ mixed` whitelists could drift silently — an endpoint anonymous on one stack, protected on the other. This PR (a) aligns them to be byte-identical 39-entry sets, (b) adds a CI check that fails the build on drift, and (c) removes a dead `auth-enrichment.lua` duplicate. Right approach, right shape, right size.

### Strengths

- **Right scope** — one CI script, one workflow, one aligned data file, one dead-code deletion. No feature creep.
- **Audit-mode-by-default** posture (`ENFORCE_UNAUTH = false`, `ENFORCE_RBAC = false` in `kong.yml`) is the correct rollout shape for a security enforcement layer.
- **Clear failure output** — the script names the offending path(s) with directional markers (`+ only in Kong` vs `- only in K8s`), which is exactly what a reviewer of a failed CI needs.
- **Path-filtered workflow trigger** — runs only when either whitelist file changes (plus the script/workflow themselves). Minimal CI overhead.
- **Symmetric fail-fast** — the script exits with a clear error if either source file is missing or malformed, rather than silently returning empty sets.
- **Runs on both PR and push-to-develop** — defence-in-depth against direct-to-develop drift.

### Issues to consider

#### Correctness / robustness

**1. Regex-based YAML parsing is fragile.**

`env_whitelist()` parses the YAML value with `re.search(re.escape(key) + r'\s*:\s*"([^"]*)"', text)`. This assumes the value is a single-line quoted string. If someone reformats `env.yaml` to use YAML block scalars (`|` or `>`), or splits the value across lines with `\n`, the regex silently returns nothing → the check fails with a not-found error, but only on that specific reformat. Prefer `pyyaml` — it's already a stdlib-adjacent dep on any GHA runner. Same concern applies to `kong_whitelist()` if anyone adds a nested Lua table into `AUTH_OPTIONAL` (the `.*?\n\s*\}` will terminate on the first inner `}\n`).

**2. Removed `dashboard-analytics` endpoints from `egov-open-endpoints-whitelist`.**

The diff removes:
- `/dashboard-analytics/dashboard/getChartV2`
- `/dashboard-analytics/dashboard/getDashboardConfig/overview`

The PR body says "removed the *disabled* DSS `dashboard-analytics` paths." Confirming: are these endpoints still callable (as auth-required) on the K8s stack, and does digit-ui-v2 handle the 401 correctly when a not-yet-logged-in user hits a dashboard route that fires these? This is exactly the class of change where a security-hardening cleanup accidentally breaks an unauth'd dashboard load. Worth a smoke check against a live env before merge, especially given #1108 (Portuguese dashboard localization) and #1109 (dashboard perf benchmark) are queued.

**3. `resolved["tenantId"]` wrapped as `tenantIds = { resolved["tenantId"] }` in the Phase 2 RBAC call.**

If the resolved user has no `tenantId` (edge case: newly-created citizen before city selection, admin user without state binding), the resulting body has `tenantIds: [null]` — cjson will encode nil in an array as JSON `null`, which egov-accesscontrol may reject as malformed. Consider `local tid = resolved["tenantId"]; local tids = tid and { tid } or {}` before serializing.

**4. `if not (ares and ares.status == 200)` collapses two failure modes.**

In audit mode, the log message doesn't distinguish "accesscontrol unreachable" (transient infra) from "accesscontrol returned non-200 for this role/uri" (legitimate deny). Consider separating them in the audit log so operators can tell which alerts to act on.

#### Hardening

**5. No self-test on the parity check itself.**

The workflow only runs the check against the current file state. A single injected divergence (e.g. in a scratch fixture) as a self-test would ensure the check actually fails when it should. Optional.

**6. No `permissions:` block on the workflow.**

Best-practice hardening: declare `permissions: contents: read` at the workflow level. The default `GITHUB_TOKEN` scopes are minimal in most repo configs, but explicit is safer.

**7. Deletion of `local-setup/kong/auth-enrichment.lua`.**

Confirmed dead from the diff, but worth a `grep -r auth-enrichment` sweep before merge — nothing in a docker-compose, Kong config include, or README references it as an active file?

#### Follow-ups (worth issue-tracking, not blocking this PR)

- **Accesscontrol response caching** — the Phase 2 code makes a per-request round-trip to `egov-accesscontrol`. On a busy stack this doubles the gateway's request cost. The Lua comment calls out caching as a documented follow-up. File it as an issue linked from this PR so it doesn't slip.
- **Migrate the whitelist source-of-truth to a shared YAML/JSON** that both `kong.yml` and `env.yaml` are generated from. The parity check works, but the manual dual-maintain will still burn reviewers periodically. Not blocking, but the natural Phase 4.

### Verdict

Approve after: confirming the `dashboard-analytics` removal is safe (item #2) and swapping the env-parser regex for `pyyaml` (item #1). The rest are polish.

subhashini-egov (approved):

subhashini-egov:
@singhalkarun Please review my comment. And if those are addressed, this can be merged.

singhalkarun:
Thanks for the thorough review — all points addressed. Summary:

## Blocking items

**① Regex YAML parsing → pyyaml.** Done (`5c8a0224`). Both files are now parsed with `yaml.safe_load`; the Lua set keys are regex-extracted from the pyyaml-loaded pre-function block (so a nested Lua table can no longer confuse the old `.*?\n\}` termination), and the env value is read from the parsed tree.

**② `dashboard-analytics` removal — confirmed safe.** I traced every caller across all frontends:
- **No CCRS frontend calls `/dashboard-analytics/*`.** The only references are dead URL *definitions* in `digit-ui-esbuild/.../urls.js` (the DSS module isn't enabled — Item 12) and the RBAC actions-master JSON (metadata, not a caller). A grep for actual `fetch`/call-sites of `/dashboard-analytics/` returned nothing.
- **The real dashboards use a native endpoint.** `digit-ui-v2`'s dashboards (Pgr/Admin/Supervisor/Executive/Citizen) fetch **`/pgr-services/v2/dashboard`** (`usePgrDashboardData.ts:252`), with zero `dashboard-analytics` references in `digit-ui-v2`. That endpoint is protected (reached post-login), so there's no anonymous dashboard load that would hit the removed DSS paths.

So removing them from the open whitelist changes nothing for any live caller (and it strengthens Item 12). The in-flight dashboard PRs (#1108/#1109) are on the native path, not DSS.

## Polish (all done)

- **③ `tenantIds:[null]`** — guarded (`7617215c`, in #1105): `tid and { tid } or {}`.
- **④ audit-log failure modes** — now split: a legitimate DENY logs `warn` (accesscontrol non-200), a transient UNREACHABLE logs `err` (`7617215c`). Enforce mode still fails **closed** on both.
- **⑤ self-test** — added a `--self-test` mode (asserts drift is caught both directions) + a CI step that runs it before the real check (`5c8a0224`).
- **⑥ workflow `permissions`** — added `contents: read` (`5c8a0224`).
- **⑦ dead `auth-enrichment.lua`** — confirmed no active reference (docker-compose / kong include / configs); the only hits were the `local-setup/README.md` file-tree, which I updated to drop the stale entries (`5c8a0224`).

## Follow-ups (issue-tracked, non-blocking)
- **#1129** — cache egov-accesscontrol authorize decisions (per-request round-trip) before enforcing at scale.
- **#1130** — single source-of-truth for the whitelist (generate both from one file) — the "Phase 4" you noted.

Re-validated live after the changes: citizen `pgr/_search` 200, citizen `hrms/_create` 403, anonymous open endpoints 200, and the parity + self-test both pass.

Issue #1131, opened 2026-07-10
bomet.yml pgr-services block missing EscalationScheduler system-user UUID → overlay collision on deploy


KDwevedi:
## Summary
develop's `local-setup/docker-compose.bomet.yml` gained a `pgr-services` block in #1059 (config-driven notification cutover — `PGR_NOTIFICATION_CONFIG_DRIVEN`, `KAFKA_TOPICS_COMPLAINTS_DOMAIN_EVENTS`), but that block does **not** carry `EGOV_INTERNAL_MICROSERVICE_USER_UUID` — the system user the PGR **EscalationScheduler** runs as. On bomet that env var is currently supplied by an on-box overlay in the nightly redeploy wrapper (`ensure_overlay_fixes`), not by the repo.

## Why this bites
The redeploy wrapper appends its own `pgr-services:` block to inject the escalation UUID. After #1059 merged, `bomet.yml` already defines `pgr-services:`, so the wrapper's append produced a **second `pgr-services` mapping key** →

```
failed to parse .../docker-compose.bomet.yml: yaml: construct errors:
  line NN: mapping key "pgr-services" already defined at line MM
```

→ `docker compose config` fails → the wrapper aborts at `ensure_overlay_fixes` **before the converge runs**. This aborted a real `deploy.sh bomet` run off develop on 2026-07-09.

## Impact
- **Deploy fragility:** any converge that layers a UUID-injecting overlay onto the post-#1059 `bomet.yml` hits a duplicate-key parse failure.
- **Fresh-box gap:** a new bomet-style box deploying straight from develop (without the on-box wrapper) gets a `pgr-services` block with **no** escalation system user → the EscalationScheduler resolves the wrong/absent system user and auto-escalation silently doesn't run.

## Root cause
Two sources both try to own the `pgr-services` overlay for bomet: the repo (`bomet.yml`, post-#1059) and the on-box wrapper. The escalation UUID belongs in the repo overlay so there is a single source of truth.

## Fix
Add `EGOV_INTERNAL_MICROSERVICE_USER_UUID` to the existing `pgr-services` block in develop's `bomet.yml`. Then the wrapper's guard (`grep -q EGOV_INTERNAL_MICROSERVICE_USER_UUID`) is satisfied, it skips the append, no duplicate key — and a fresh box gets a correct escalation user. PR incoming.

## Notes
[...]

PR #1132, opened 2026-07-10
fix(bomet): add EscalationScheduler system-user UUID to pgr-services overlay

KDwevedi:
Closes #1131.

## Problem
develop's `local-setup/docker-compose.bomet.yml` gained a `pgr-services` block in #1059 (config-driven notification cutover) but it omits `EGOV_INTERNAL_MICROSERVICE_USER_UUID` — the system user the PGR **EscalationScheduler** runs as. On bomet that env is injected by the nightly redeploy wrapper's `ensure_overlay_fixes`, not by the repo.

Because the wrapper appended its **own** `pgr-services:` block to add the UUID, after #1059 there were **two `pgr-services` mapping keys** →
```
yaml: mapping key "pgr-services" already defined
```
`docker compose config` failed and the wrapper aborted **before the converge** — this broke a real `deploy.sh bomet` off develop on 2026-07-09.

## Fix
Add `EGOV_INTERNAL_MICROSERVICE_USER_UUID` to the **existing** `pgr-services` block in `bomet.yml` (one line + comment). Single source of truth:
- The wrapper's guard (`grep -q EGOV_INTERNAL_MICROSERVICE_USER_UUID`) is now satisfied → it skips its append → no duplicate key.
- A fresh box deploying straight from develop gets a correct escalation system user (otherwise auto-escalation silently resolves the wrong/absent user).

## Verification
- `docker compose config` / `yaml.safe_load` on the merged `bomet.yml`: parses, exactly **one** `pgr-services` key.
- On bomet the escalation UUID is already present and `pgr system UUID present` passed in the redeploy smoke test.

## Notes
- `bomet.yml` is the committed bomet-specific overlay; the value is an internal-microservice user identifier (not a credential).
- Related: the wrapper was hot-patched to inject-not-append as a stopgap; this closes it durably.

## Summary by CodeRabbit

* **Configuration**
  * Updated the Bomet tenant setup with a fixed internal service user identifier.
  * Maintains configuration-driven notifications and complaint event routing.

### DIFF
--- a/local-setup/docker-compose.bomet.yml
+++ b/local-setup/docker-compose.bomet.yml
@@ -31,4 +31,10 @@ services:
     environment:
       PGR_NOTIFICATION_CONFIG_DRIVEN: "true"
       KAFKA_TOPICS_COMPLAINTS_DOMAIN_EVENTS: complaints.domain.events
+      # System user the PGR EscalationScheduler runs as (not in the shared
+      # egov-digit base compose). Kept in this same pgr-services block so the
+      # bomet overlay is a single source of truth — previously injected by the
+      # on-box redeploy wrapper, which appended a *second* pgr-services key here
+      # and broke `docker compose config` once this block existed (see #1131).
+      EGOV_INTERNAL_MICROSERVICE_USER_UUID: a6966a30-947d-46d9-8226-7d94b67de581
 


KDwevedi:
Closing: the escalation system-user UUID is a bomet-instance-specific value and belongs in the server-side layer (the on-box redeploy wrapper / gitignored host_vars), not committed to the shared develop `bomet.yml`. The runtime overlay is the git-tracked repo file, but the divergence is applied on-box by the wrapper (inject-not-append), which is the correct home for per-instance identifiers. Keeping it server-side.

PR #1138, opened 2026-07-10
fix(pgr): localized "Complaint related to" options + stale seed refresh (CCSD-1971 A1)

Hari-egov:
## What (JIRA CCSD-1971, feedback item A1 follow-up)

1. The citizen **"Complaint related to"** dropdown rendered the raw MDMS `name` — one language for all locales. Options (and the confirmation helper below the field) now resolve **`PGR_RELATEDTO_<CODE>`** for the active locale, **falling back to the MDMS name** — zero behavior change where keys aren't seeded. Seeded locally: `en_IN` English names + `pt_PT` the A1 Portuguese names (other envs need the same; values in the commit).
2. `docs/migration/seed/ComplaintRelatedToMap.json` still carried the **pre-A1 names** ("Inspeccão Geral do Estado" …) — the last stale copy in the repo; refreshed to match the DDH seed and live data so future migrations don't resurrect the old names.

## Testing (local, deployed)
- pt_PT session: options show the A1 Portuguese names; en_IN session: English names; both fall back cleanly when a key is missing.
- Note for testers: the browser caches MDMS in IndexedDB for 24h — fresh incognito to see renamed options.

### DIFF
--- a/digit-ui-esbuild/products/pgr/src/pages/citizen/Create/CreatePGRFlowV2.tsx
+++ b/digit-ui-esbuild/products/pgr/src/pages/citizen/Create/CreatePGRFlowV2.tsx
@@ -751,10 +751,13 @@ function RelatedToStepBody({ data, patch, relatedToOptions, t }: StepBodyProps) 
             });
           }}
           placeholder={tr(t, "CS_COMPLAINT_PICK_ONE", "Select…")}
-          options={options.map((o) => ({ value: o.code, label: o.name }))}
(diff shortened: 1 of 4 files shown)


Hari-egov:
Commit 2 — **superseding approach per prod data change**: prod's `ComplaintRelatedToMap.name` now carries localization KEYS (`PGR_RELATEDTO_NAME_<CODE>`), so the dropdown rendered raw keys. Options + helper now render `t(name)` — keys resolve per locale, legacy display-text names pass through unchanged (both conventions safe). Seeds aligned to the key convention; en/pt values seeded on local **and prod**.

PR #1142, opened 2026-07-10
feat(compose): per-service Flyway init containers — DB-migration parity with K8s (item #10)

singhalkarun:
## Plain-English summary

- Our system runs two ways — a **quick local setup** (Docker Compose) and the **real cloud setup** (Kubernetes) — and they applied database changes **differently**, which caused a real incident. This makes the local setup work **exactly like the cloud**.
- Every service now gets its **own "update the database" helper that runs before it starts**, and the local setup **pulls the *same* helper images the cloud uses** — so the two can't drift apart.
- **Future database changes flow in cleanly:** bump a version tag and the helper applies the change on the next startup; if there's nothing new, it does nothing (no risk).
- We had to **rebuild the "what's already applied" record** inside the pre-loaded database snapshot so the cloud helpers understand it (all real data left untouched) — otherwise they'd try to redo everything and crash.
- **Proven, not theory:** rebuilt the whole stack from scratch → every service came up healthy (27 healthy, 0 failed); faked a future change → the helper picked it up and applied it smoothly. Plus an automatic check + a runbook so the team can operate and extend it.

---

## What & why

**Deployment-parity item #10** — make the Docker Compose stack apply DB migrations the **same way Kubernetes does**: one Flyway **init container per service**, run before the app, using the **same `egovio/<service>-db:<pinned-tag>` image K8s pins**, writing to the same `<service>_schema` history table. This replaces compose's previous mix (embedded Flyway in some apps + one shared `db-migrations` container) and closes the gap behind a recent multi-day migration incident.

The whole point: **future migrations propagate cleanly and predictably** — bump a service's `-db` image tag, and its init container applies the pending migration on boot, exactly like K8s.

## What's in here

**Per-service Flyway init containers** (`local-setup/docker-compose.migrations.yml`) for **13 services**:
- **Repo-local** (built from the in-repo `db/Dockerfile`): `pgr-services`, `novu-bridge`, `digit-config-service`.
- **Core** (pull the exact K8s `egovio/<service>-db:<tag>` images): `boundary-service`, `egov-hrms`, `egov-user`, `egov-idgen`, `egov-localization`, `egov-enc-service`, `egov-filestore`, `egov-workflow-v2`, `mdms-v2`, `egov-url-shortening`, `egov-otp`.

Each: a one-shot migrator connected **directly to `postgres-db`** (not the pgbouncer/transaction-mode alias — which also dissolves the old `egov-otp` Flyway incompatibility), the app gated on it, embedded Flyway off. `pgr-services` is additionally **ordered** after the services its analytics views read.

**Dump re-bake** (`local-setup/db/full-dump.sql`) — the dump stored **baseline-only** Flyway history under compose-flavoured names, incompatible with the real K8s `-db` image lineages (they'd re-run migrations → `42P07`). Re-baked so each history table carries the **real** applied rows under the K8s `<service>_schema` names (data tables untouched).

**CI guard** (`.github/scripts/check-flyway-dump-alignment.py` + workflow) — asserts every service's history-table name lines up with the dump, catching the `42P07` naming class in a PR instead of at boot.

**Docs** — design spec, pilot plan, and an operator **runbook** (`docs/db-migration-flow.md`): how future migrations flow, the conditions for it to work, add/test/troubleshoot procedures, and dump-regeneration maintenance.

**Retired** the consolidated `db-migrations` container (no-op entrypoint) — its 3 dependents now self-migrate.

## Validated end-to-end (live)

- **Full-stack boot from the re-baked dump:** all 13 migrators `Exited (0)` (no-op), `pgr-services-migration` ran *after* its gated deps went healthy, all converted apps came up **healthy** — **27 healthy / 0 unhealthy / 0 failed**.
- **Future-migration smoke test:** simulated a newer `-db` image (real image + one new migration) → init container **applied it cleanly** (`Successfully applied 1 migration`), then **no-ops** on re-run (idempotent).
- **CI alignment check green:** 13 aligned / 3 baseline-fresh / 0 pending.

## Deliberately not changed

- `egov-accesscontrol` — kept on embedded Flyway (its K8s chart declares no `schemaTable`).
- `egov-indexer` — baseline-fresh, still embedded.
- `default-data-handler` (no migrations) and `digit-user-preferences-service` (GORM AutoMigrate) — **already at parity**; adding migrators would *create* divergence.

## Follow-ups (out of scope)

- Regenerate the dump from the `-db` images on a cadence (guards content drift — the CI check guards naming, the smoke test guards the apply path). Documented in the runbook.
- Apply the same per-service model to the K8s side is N/A (K8s already does this) — this PR brings compose *to* K8s, not the reverse.

## Draft

Opened as a draft; ready for a review of the approach before merge.

## Summary by CodeRabbit

* **New Features**
  * Added per-service migration orchestration that runs before each app starts, with embedded app migrations disabled.
  * Added a pre-migration “history normalization” step to safely align Flyway history tables with the expected canonical naming.
  * Added a dedicated CI workflow to validate dump/migration parity.
* **Bug Fixes**
  * Prevents Flyway from replaying or corrupting migrations when dump history table names don’t match expectations.
* **Documentation**
  * Updated migration parity, rollout, verification, and troubleshooting guidance for Compose and Kubernetes workflows, plus history normalization testing details.

### DIFF
--- a/.github/scripts/check-flyway-dump-alignment.py
+++ b/.github/scripts/check-flyway-dump-alignment.py
@@ -44,6 +44,7 @@ ROOT = pathlib.Path(__file__).resolve().parents[2]
 DUMP = ROOT / "local-setup/db/full-dump.sql"
 COMPOSE = ROOT / "local-setup/docker-compose.egov-digit.yaml"
 FAST_PATH = ROOT / "local-setup/docker-compose.fast-path.yml"
+MIGRATIONS = ROOT / "local-setup/docker-compose.migrations.yml"
 
 # ── Transitional allowlists (drive these to empty as item #10 lands) ──────────
 # Enabled services that legitimately create their history fresh: they claim a
@@ -107,24 +108,31 @@ def _env(spec) -> dict:
     return {}
 
 
-def merged_services(compose_text: str, fast_path_text: str) -> dict:
-    """Per-service env from the base compose with the fast-path env overlaid."""
-    base = yaml.safe_load(compose_text) or {}
-    over = yaml.safe_load(fast_path_text) or {}
-    services = {n: _env(s) for n, s in (base.get("services") or {}).items()}
-    for name, spec in (over.get("services") or {}).items():
-        services.setdefault(name, {}).update(_env(spec))
+def merged_services(*compose_texts: str) -> dict:
+    """Per-service env from the base compose with each overlay merged in order."""
+    services = {}
+    for text in compose_texts:
+        data = yaml.safe_load(text) or {}
+        for name, spec in (data.get("services") or {}).items():
+            services.setdefault(name, {}).update(_env(spec))
     return services
 
 
 def claimed_tables(services: dict) -> set:
-    """SPRING_FLYWAY_TABLE of every Flyway-enabled service that pins one.
+    """Tables claimed by (a) a per-service migration init container via SCHEMA_TABLE,
+    or (b) a Flyway-enabled app via SPRING_FLYWAY_TABLE.
 
     A service is Flyway-relevant if it carries any SPRING_FLYWAY_* / FLYWAY_ENABLED
     key; enabled unless that flag is explicitly false (the image default is on).
     """
     claimed = set()
     for env in services.values():
+        # (a) init-container migrator: SCHEMA_TABLE is the authoritative name.
+        schema_table = env.get("SCHEMA_TABLE")
+        if schema_table:
+            claimed.add(schema_table.strip())
+            continue
+        # (b) app with embedded Flyway still enabled.
         relevant = any(k.startswith("SPRING_FLYWAY") for k in env) or "FLYWAY_ENABLED" in env
         if not relevant:
             continue
@@ -221,8 +229,17 @@ def self_test() -> int:
     stale2 = analyze(dump, {"pgr_services_schema", "hrms_schema_version"})
     assert "hrms_schema_version" in stale2["stale_baseline_fresh"], stale2
 
+    # A service migrated to an init container claims its table via SCHEMA_TABLE,
+    # not SPRING_FLYWAY_TABLE. That must still count as claimed.
+    PENDING_ENABLE, BASELINE_FRESH = set(), set()
+    migrated = claimed_tables({
+        "pgr-services": {"SPRING_FLYWAY_ENABLED": "false"},
+        "pgr-services-migration": {"SCHEMA_TABLE": "pgr_services_schema"},
+    })
+    assert migrated == {"pgr_services_schema"}, f"migrator SCHEMA_TABLE not claimed: {migrated}"
+
     PENDING_ENABLE, BASELINE_FRESH = saved_p, saved_b
-    print("self-test OK: alignment, mismatch, orphan, and stale-allowlist all detected.")
+    print("self-test OK: alignment, mismatch, orphan, migrator-claim, and stale-allowlist all detected.")
     return 0
 
 
@@ -230,7 +247,8 @@ def main() -> int:
     if "--self-test" in sys.argv:
         return self_test()
     dump = dump_history_tables(DUMP.read_text())
-    claimed = claimed_tables(merged_services(COMPOSE.read_text(), FAST_PATH.read_text()))
+    claimed = claimed_tables(merged_services(
+        COMPOSE.read_text(), FAST_PATH.read_text(), MIGRATIONS.read_text()))
     return report(dump, claimed)
 
 
--- a/.github/scripts/check-flyway-dump-alignment.py
+++ b/.github/scripts/check-flyway-dump-alignment.py
@@ -60,17 +60,18 @@ BASELINE_FRESH = {
 # History tables the dump ships whose owning service currently has Flyway OFF
 # (item #10 decision #2 backlog). When you enable one, set its SPRING_FLYWAY_TABLE
 # to exactly this name and delete the entry here — the check then enforces it.
-PENDING_ENABLE = {
-    "egov_idgen_schema_version",        # egov-idgen        — SPRING_FLYWAY_ENABLED:false
-    "egov_localization_schema_version", # egov-localization — SPRING_FLYWAY_ENABLED:false
-    "egov_user_schema_version",         # egov-user         — FLYWAY_ENABLED:false
-}
+# Emptied in Phase 3: the dump was re-baked to the K8s <service>_schema names and
+# every core service (incl. idgen/localization/user/otp) now has a per-service
+# migration init container claiming its table, so nothing is pending-enable.
+PENDING_ENABLE = set()
 
 _FALSE = {"false", "0", "no", "off"}
 
 # CREATE TABLE public.<name> ( ... \n);  — non-greedy body up to the closing line.
+# The name may be double-quoted and contain hyphens (e.g. K8s's
+# "egov-url-shortening_schema"), so allow '-' inside the identifier.
 _CREATE = re.compile(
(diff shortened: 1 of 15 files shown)


singhalkarun:
## End-to-end verification against a team-supplied dump

Ran the deploy's migration path over a real dump the team handed us
(`egov_backup_20260709_082159.sql` — 57 tables, 97,250 rows, a live tenant: 534 users,
496 employees, 16 complaints, 70k localization messages).

**Result: the deploy ran all 14 migrators and destroyed nothing. 45 data tables,
97,122 rows — byte-identical before and after.**

### Two problems this PR now fixes

**1. The overlay never reached Ansible.** `docker-compose.migrations.yml` was not copied
to the target and not in `compose_files` — so an Ansible deploy ran *none* of the
per-service migrators and silently fell back to the apps' embedded Flyway. Merging the
PR as it stood would have changed nothing for any Ansible-deployed tenant.

> The only reason the migrators run on our box today is that someone copied
> `docker-compose.migrations.yml` to `docker-compose.<tenant>.yml` — the one filename the
> playbook *does* load. It's byte-identical to the overlay, it's untracked, and it exists
> on that one machine. **Upgrade step: delete that file once this lands**, or the same 14
> services get defined twice.

**2. A team dump destroys data.** These dumps carry the LEGACY history names
(`*_schema_version`); the `-db` images look for the K8s names (`<service>_schema`). Flyway
sees an empty history and replays from V1 against populated tables. Nine migrators crash
with 42P07 — that crash is the only thing that saved their data. The two that *don't*
crash are the ones that do the damage, because their V1 opens with `DROP TABLE IF EXISTS`:

| | unguarded |
|---|---|
| `message` | 70,835 → **0** |
| `eg_enc_symmetric_keys` | 27 → **0** |
| `eg_enc_asymmetric_keys` | 27 → **0** |
| **total** | **97,250 → 26,377 (73% destroyed)** |

Both exit **0**. A green deploy that ate the data — and the encryption keys are
unrecoverable: `eg_user.name` and `mobilenumber` are ciphertext and those keys were the
only thing that could decrypt them.

### How it was verified

Not a hand-rolled harness — the **real deploy path**: the compose files the playbook
produces (base + fast-path + migrations + the new ansible overlay), the real fast-path
`initdb` loading the dump, and compose's own `depends_on` ordering. Run in a separate
project namespace so the live stack was never touched.

`db-history-normalize` ran first, as compose gated it to:

```
db-history-normalize: normalizing Flyway history
  renamed  boundary-service: boundary_schema_version -> boundary_service_schema
  renamed  egov-user: egov_user_schema_version -> egov_user_schema
  renamed  mdms-backend: mdms_schema_version -> mdms_v2_schema
  renamed  egov-idgen: egov_idgen_schema_version -> egov_idgen_schema
  renamed  egov-localization: egov_localization_schema_version -> egov_localization_schema
  renamed  egov-enc-service: enc_schema_version -> egov_enc_service_schema
  renamed  egov-filestore: filestore_schema_version -> egov_filestore_schema
  renamed  egov-workflow-v2: workflow_schema_version -> egov_workflow_v2_schema
  renamed  egov-hrms: hrms_schema_version -> egov_hrms_schema
  renamed  egov-url-shortening: egov_url_shortening_schema_version -> egov-url-shortening_schema
  rebuilt  egov-otp: dropped empty eg_token (migrator will recreate)
ok: renamed 10, rebuilt 1, already-aligned 3, skipped 2
```

Then every migrator, all exit 0:

| migrator | verdict |
|---|---|
| boundary-service, egov-user, mdms-backend, egov-idgen, egov-localization, egov-enc-service, egov-filestore, egov-workflow-v2, egov-hrms, egov-url-shortening | `No migration necessary` |
| egov-otp | `Successfully applied 2 migrations` — rebuild path; its empty orphan `eg_token` was dropped, so it *must* rebuild. Previously this hit 42P07 and blocked boot. |
| pgr-services | `Successfully applied 1 migration` — a genuinely **new** migration landing on top of the dump. This is the whole point: incremental migrations still apply. |

`pgr-services-migration` also correctly waited for its four cross-service gates
(`boundary-service`, `egov-user`, `mdms-backend`, `egov-workflow-v2`) to go healthy before
running — the ordering added in this PR, observed working rather than assumed.

### The data

```
DATA tables before: 45  rows=97,122
DATA tables after : 45  rows=97,122
DATA tables that changed at all: NONE — byte-identical

message                  70,835 -> 70,835  ✅
eg_enc_symmetric_keys        27 -> 27      ✅
eg_enc_asymmetric_keys       27 -> 27      ✅
```

Total rows moved 97,250 → 97,254. All +4 are Flyway bookkeeping: `pgr_services_schema`
+1 (its new migration) and `egov_otp_schema` +3 (rebuilt history). **Zero user-data rows
changed.**

### Regression tests

- **Integration proof** — `./local-setup/db/normalize/test-integration.sh <dump.sql>`: any
  dump, real pinned images, asserts byte-identical.
- **Negative test** — same dump, normalizer skipped, asserts the data **IS** destroyed. If
  that ever stops failing, the guard has been silently disabled and the positive test alone
  would not tell us.
- Decision-table unit tests (12), idempotency, fresh-install, and CI now fails if a
  migrator's `SCHEMA_TABLE` is missing from `flyway-history-map.yml`.

### Known limitation (deliberately out of scope)

The rename **assumes** the dump's schema matches what its history claims. True for this
dump — all 11 migrators go clean — but not independently verified. A dump whose schema had
drifted from its history would still be accepted. Closing that properly is the schema-drift
check (build schema from migrations, build from dump, diff); it's its own piece of work and
half-doing it here would give false confidence.

Spec: `docs/superpowers/specs/2026-07-13-ansible-flyway-history-normalization-design.md`

subhashini-egov (changes requested):
Resolve coderabbit comments

subhashini-egov:
@singhalkarun One more qn - does the DB dump include the notifications and maps infra set up etc..?

subhashini-egov (approved):

KDwevedi:
Field report from deploying develop (`7030a905f`) on the Bomet box — two `db-history-normalize` gaps surfaced, both worked around on-box but worth follow-ups:

**1. Baseline-only history + pre-existing data tables → migrator hard-fails the dependency chain.** Bomet historically ran egov-url-shortening with `SPRING_FLYWAY_ENABLED=false`, so its history table held a single `BASELINE` row at version `1` while `eg_url_shortener` already existed. Normalize renamed the table to `"egov-url-shortening_schema"` correctly, but `egov-url-shortening-migration` then re-applied `V20190624185601__eg__ddl.sql` (`relation "eg_url_shortener" already exists`, exit 1), and via `depends_on` the failure left pgr-services, hrms and egov-user stuck in `Created`. On-box fix: `UPDATE "egov-url-shortening_schema" SET version='20190624185601' WHERE type='BASELINE'`. Suggestion: when a mapped service's history is baseline-only but its `data_tables` exist, normalize (or the migrator wrapper) should re-baseline to the image's latest migration version instead of letting Flyway re-run DDL.

**2. Table renames leave the old PK index names behind.** Normalize renames the history tables but not their indexes, so `egov_user_schema_version_pk`, `mdms_schema_version_pk`, etc. survive under the old names. Any Flyway that later baseline-creates a table under a legacy name (a stale compose set that misses the new overlays, or the neutered `db-migrations` image run standalone) dies on the leftover index — `relation "egov_user_schema_version_pk" already exists` — which is a confusing symptom pointing at the wrong culprit. Renaming the PK index alongside the table would make the failure mode honest.

Related operational note for anyone running the Bomet-style self-healing wrapper: its hardcoded compose-file list must now include `docker-compose.migrations.yml` + `docker-compose.migrations.ansible.yml`, otherwise a heal-loop `compose up` recreates services with in-service Flyway pointed at the pre-normalize table names and they crash-loop exactly as in (2).

PR #1143, opened 2026-07-10
fix(ui/configurator): deployment-agnostic fixes — tenant de-hardcoding, onboarding logoId, boundary leaves, citizen pages, data-provider

dhruv-1001:
## What

Deployment-agnostic application fixes surfaced by running the full integration suite against a fresh non-Kenya (Mozambique/Maputo) deployment, plus review hardening. No behavior change for correctly-seeded Kenya deployments.

### configurator
- **Login/pre-login tenant de-hardcoded**: placeholder + initial `tenantCode` + pre-login default now derive from build-time `VITE_STATE_TENANT_ID` (neutral fallback), replacing the hardcoded `'ke'`/`DEFAULT_TENANT_CODE`. Ansible threads the env into the configurator build; `configurator-build.sh` now warns loudly when it's unset (a manual rebuild used to silently revert this).
- **UserCreate/UserEdit mobile validation → MDMS-driven** (`useMobileValidator`), replacing the hardcoded Kenyan regex that made citizen-user create/edit impossible on any non-Kenya tenant. Hard fallback is now permissive `^[0-9]{7,15}$` (no country assumption) instead of India's rule.
- **Tenant-create sends no `logoId` when absent** (was `logoId: null`, which the `tenant.tenants` schema rejects with `expected type: String, found: Null`) — this blocked the entire onboarding wizard at Phase 1 "Upload to DIGIT".
- **data-provider**: (a) complaint-type leaf create maps `serviceCode→code` when the form carries an empty string (fixes MDMS `UNIQUE_IDENTIFIER_EMPTY_ERR`); (b) PGR create resolves `address.tenantId` to the boundary's city — with `boundarySearch` verification in both single- and multi-city cases, session-tenant fallback (root-seeded trees stay correct); (c) added the missing `user` update branch (UserEdit's Save previously threw "Update not supported").
- Misc de-hardcoding: Kenya examples/help-text → generic; OSM hierarchy type + XLSX sample coords read config.

### digit-ui-esbuild
- **Citizen FAQ / How-it-works** no longer throw into the error boundary when MDMS data is absent (graceful empty state; happy path unchanged).
- **Profile dial-code chip** read `mdmsValidationData?.countryCode`, which never exists (the hook maps it into `.prefix`) — chip was permanently `+91`. Now reads `.prefix` (matches the validation path).
[...]

Issue #412, opened 2026-04-21
Nai Pepea: Configurator: Credentials are auto filled on login page with password


Gurjeet-eGov:
**Describe the bug**
Credentials are auto filled on login page with password, any user would be able to login

**To Reproduce**
Open Login URL: https://naipepea.digit.org/configurator/login

**Expected behavior**
Username, password, tenant field shall be blank.

**Actual behavior**
Details are auto filled
https://github.com/user-attachments/assets/e4ba0eb2-e02a-482f-a601-5812e08579f6

KDwevedi:
Fixed in [ChakshuGautam/digit-configurator#28](https://github.com/ChakshuGautam/digit-configurator/pull/28) (merge commit `d501bbf`), deployed to https://naipepea.digit.org/configurator/.

**Root cause.** `LoginPage.tsx` seeded `username: 'ADMIN'`, `password: 'eGov@123'`, `tenantCode: 'ke'` into form state on mount, so the very first submit taught browsers to autosave those credentials.

**Fix.** Start with empty fields, add `autoComplete="off"` on the `<form>`, `new-password` on the password input, `off` on the other two.

**Validate.** Sign out, reload `/configurator/login` — fields empty, no "Save password?" banner after a real login. Safe to close.

Gurjeet-eGov:
Resolved, 

https://github.com/user-attachments/assets/7fba83f0-97a8-4a6c-a2d6-5efeb59e6f9a

PR #1145, opened 2026-07-10
test(integration): audit stale specs + make the suite deployment-agnostic (#1047)

### LINKED ISSUE #1047: Enhancement: Improving testing suite
# Integration tests: audit stale specs + make the suite deployment-agnostic

## Context

The integration test suite has drifted from the current code. A number of specs fail, and several fail only after a long per-test timeout — which also causes later tests to not run at all within the suite's overall time budget. Separately, the suite is currently wired to run against a specific deployment, which makes it hard to point at others.

This issue covers two related pieces of work.

---

## 1. Audit stale specs — update or remove

Go through the suite and decide, per spec, whether it should be **updated** to match the current app or **removed**.

- Fix specs that assert UI/behaviour that has since changed.
- Unmask specs that guard fixes which have already landed.
- Remove specs that are obsolete or purely diagnostic (no real assertions).
- Ensure no single spec can consume a large share of the run budget (tighten per-test timeouts / split slow specs) so the whole suite finishes and nothing is silently skipped.

**Done when:** the suite completes within its time budget on a healthy deployment, every remaining spec either passes or is explicitly and justifiably skipped, and no test is masking an already-fixed issue.

---

## 2. Run the suite against any deployment

**Status: to be planned — no tasks yet.**

Today the suite is tied to particular deployments via per-deployment config, and its defaults assume specific tenants/users that don't exist everywhere. We want to be able to run the same suite against any online deployment without code changes.

dhruv-1001:
## What

Integration-test audit + deployment portability (#1047), verified against both shapes: the Kenya defaults and a fresh Mozambique (Maputo) deployment.

- **Audit (#1047):** stale specs updated/removed/renamed per the outdated-tests report; crash-guards on fixture reads; ~15 specs renamed to intent-revealing names; 5 dead debug specs removed.
- **De-hardcoded location specifics** across citizen/employee/admin/api/onboarding: `ke`/`ke.nairobi`/`naipepea` hosts, Kenya mobiles, `00100` postal, hardcoded SRIDs → env/fixtures/live discovery (`generateValidMobile`, live serviceCode/boundary/FK resolution, `pgr-idgen` prefix discovery). Kenya-shaped defaults remain (suite runs on ke with zero env).
- **Kenya-compat hardening:** SRID matching is prefix-agnostic (`[A-Z]+-PGR-…`, was `PG-PGR-…` in two places); lifecycle seeding drops `verificationDocuments: []` (400s on CRS); `PGR_ID_PREFIX` fallback aligned to `NCCG`; exact-English asserts carry raw-key alternates.
- **New coverage:** employee inbox filters/search actually *narrow* (XHR param + every visible row asserted), full ASSIGN→RESOLVE and REJECT driven through the real Take-Action UI with server-side state verification, plus a shared portable `employee-ui` helper.
- **login.spec:** tenant-field assertion updated — the LoginPage intentionally pre-fills the configured root tenant (upstream introduced the same pre-fill hardcoded to `'ke'`; asserting empty was stale against upstream's own page).
- **Fixtures:** employee logins probe CITY→ROOT tenant; onboarding fresh-tenant codes are letters-only (egov-user's `^[a-zA-Z. ]*$` tenantId rule); boundary fixture uses the wizard's actual default hierarchy.

## Testing
`tsc --noEmit` clean; 313 tests in 97 files collect. Full sequenced suite run against a local Maputo deployment on current develop + the companion app-fix PRs (citizen/employee/admin/onboarding/api/keycloak/lifecycle/specs) — headline: onboarding, api, smoke, misc all green; remaining reds are deployment-data items documented per-spec.

Closes #1047


dhruv-1001:
## Remaining-failures report — full-suite run on the merged state

Run context: local Mozambique (`mz.maputo`) deployment on current `develop` + this PR + the companion app PRs (#1143, #1144), sequenced per-area, `workers=1`. Headline: **166 ✓ / 15 ✘ / 88 ⊘** in the batch run → **168 ✓ / 13 ✘** effective after re-verifying two suspects in isolation (below). `onboarding`, `api`, `smoke`, `keycloak/lifecycle/specs` and the whole **employee** area are at **0 failures**.

### Re-verified green after the run (no longer failing)
| Spec | What happened |
|---|---|
| `admin/login.spec.ts` | The test asserted the tenant input renders **empty**, but the LoginPage intentionally **pre-fills** the configured root tenant (upstream introduced the same pre-fill hardcoded to `ke` — the assertion was stale against upstream's own page). Fixed in this PR (`047a9531`); username/password emptiness still guards the #412 autofill regression. Re-run: pass. |
| `citizen/complaint-detail-page.spec.ts` | One-off `page.goto` 30 s timeout on the citizen home; 33 other citizen tests (incl. the home-page spec) passed in the same run. Re-run in isolation: pass. Transient. |

### The 13 remaining failures, by root cause

**A. Stale walk logic in two citizen specs (test-side follow-ups)**
| Spec | Evidence | Why it fails |
|---|---|---|
| `citizen/full-complaint-lifecycle-ui.spec.ts` (t2) | `waiting for locator('input.digit-dropdown-employee-select-wrap--elipses')` | Its boundary-cascade walk still targets the **legacy input-style dropdown**; the modern wizard renders `button[role="combobox"]`. The primary wizard spec (`file-complaint-wizard`) got the modern N-level walk and passes — this spec needs the same port. |
| `citizen/complaint-attachment.spec.ts` | `#555 — Attachments section … must surface an ` | Clears the (fixed) cascade but its own walk-to-upload/detail-image step loop dead-ends before the detail-page image renders. Needs the same locator/walk modernization. |

**B. Async-create race in configurator create flows (test-side candidates, same class already fixed for `users.spec.ts`)**
| Spec | Evidence |
|---|---|
| `admin/complaints.spec.ts` (t1) | `page.waitForURL` 30 s timeout after Create — the URL predicate is satisfied by the create page itself / the async react-admin POST hadn't fired when navigation was awaited. The proven fix (applied to `users.spec.ts` in this PR's lineage) is to await the `_create` **response** before asserting the redirect. |
| `admin/complaint-types.spec.ts` (t2) | Same `waitForURL` 30 s pattern after leaf create. |
| `admin/complaint-form-validators-2026-05-31.spec.ts` | 120 s test timeout inside the users+create+validators chain — same create-then-navigate family. |
| `admin/departments.spec.ts` (t2) | 120 s timeout in the create → edit → deactivate round-trip — same family. |

**C. Deployment-data / seed gaps on this (fresh Mozambique) stack — expected red here, expected green on a seeded Kenya deployment**
| Spec | Why |
|---|---|
| `admin/localization.spec.ts` (t1, then 2–7 skip serially) | en_IN row-count parity `ke` vs `ke.nairobi` — the Kenya localization seed doesn't exist on the mz stack (delta ≫ 2). Runs (not skips) on ke by design. |
| `admin/recently-shipped-fixes.spec.ts` | `localization row missing for ESCALATE` — the Kenya `rainmaker-pgr` seed (sentence-cased ESCALATE label) is absent on mz. |
| `admin/tenants.spec.ts` (t1) | List asserts `> 1` tenant row; the fresh mz stack has a single onboarded city. Data, not code. |
| `admin/workflow-action-select-521.spec.ts` | Needs a complaint parked at `PENDINGATLME` to expose the Action select; none existed at run time (the Edit view correctly shows terminal-no-actions otherwise). Seed-dependent. |
| `admin/boundaries-ui.spec.ts` | 180 s timeout in the create-two-boundaries-then-search chain (depends on the Phase-1 tenant/branding path on a fresh tenant). Needs an isolated re-triage; grouped here because the failure moves with tenant data, not with the code under test. |

**D. Genuine app-side flag, intentionally left red (1)**
| Spec | Why |
|---|---|
| `admin/employees.spec.ts` (t2) | `getByLabel(/^Date of Birth/i)` `toHaveAttribute('required')` fails — the DOB input enforces required-ness via the form validator + visual `*` but never emits the HTML `required` attribute. Kept as a red rather than weakening the assertion; it's a one-line app fix (flagged for a follow-up to #1143). |

### Bottom line
None of the 13 are regressions from this PR or the companion PRs — 6 are test-side follow-ups (A+B, all with a known fix pattern), 5 are seed/data conditions of the fresh test deployment (C), 1 is a deliberate app-bug canary (D), and the 13th (`boundaries-ui`) needs one isolated re-triage. Happy to fold the A/B fixes into this PR or a follow-up — maintainer's preference.

vinothrallapalli-eGov:
## Review: `test(integration): audit stale specs + make the suite deployment-agnostic (#1047)`

**Scope:** 109 files, +7583/-3478, entirely under `tests/integration-tests/` plus one stray doc file.

### Overview

This PR does three things to the Playwright integration suite: (1) removes/renames stale specs, (2) replaces Kenya-tenant-specific literals (hosts, mobile numbers, postal codes, SRIDs, PGR-id prefixes) with env-driven or live-discovered values so the suite also runs against a fresh Mozambique/Maputo tenant, and (3) adds genuinely new coverage — inbox filter/search narrowing checks and full ASSIGN→RESOLVE/REJECT flows with server-side state verification via a new shared `employee-ui.ts` helper. The stated goal is well-executed in most of the diff; the issues below are concentrated in a handful of files.

### Blocking / high-severity issues

1. **`playwright.config.ts` default `BASE_URL` is stale and now inconsistent with `env.ts`.** `playwright.config.ts:3` still hardcodes `https://naipepea.digit.org`, while `tests/utils/env.ts` was rewritten in this same PR (with a comment explaining exactly why) to default to `http://localhost`. `use.baseURL` drives every `page.goto()` call, so a bare `npx playwright test` will now navigate the browser to a dead demo host while every fetch-based helper (built on `env.ts`) hits `localhost` — undermining the PR's own stated fix. Same stale literal is duplicated in `tests/utils/citizen-provision.spec.ts:42`. Fix: import `BASE_URL` from `./env` in both places.

2. **`pgr-idgen.ts` reads the wrong admin-credential env vars.** It uses `process.env.ADMIN_USER` / `ADMIN_PASS` / `ADMIN_PASSWORD` directly instead of importing `ADMIN_USER`/`ADMIN_PASS` from `./env` (which derive from `DIGIT_USERNAME`/`DIGIT_PASSWORD` — the convention every other helper in this PR follows). A deployment that only sets `DIGIT_USERNAME`/`DIGIT_PASSWORD` will have this login silently fail and fall back to `DEFAULT_FALLBACK = 'NCCG'` (a Nairobi prefix) with no surfaced error — exactly the kind of silent Kenya-default leak this PR is meant to eliminate.

3. **`docs/superpowers/plans/2026-06-26-playwright-consolidation.md` (1037 lines) looks like an accidental inclusion, out of scope.** It's a separate implementation plan targeting a different branch (`feat/fix_citizen_tests_automation`), describes work this PR doesn't do (adding `smoke`/`api` projects — though coincidentally some of that appears to have landed anyway; porting specs from `local-setup/tests/e2e/`; later deleting `tests/playwright/`), and embeds internal hostnames (`subhadev.digitlab.in`, `bometfeedbackhub.digit.org`) plus `ADMIN`/`eGov@123` credential pairs in shell examples. It doesn't belong in a "test audit" PR — recommend dropping it.

### Medium-severity issues

- **`login-mobile-input-validation.spec.ts`**: the new error-locator regex (`/\d+-digit|valid.*mobile|.../i`) can match the same idle-state helper-hint paragraph asserted a few lines earlier, so the test could pass even if the UI never renders a distinct validation error — weakening the CCRS#429 regression guard it exists for.
- **`tenants.spec.ts`**: test 6 uses an unconditional `test.skip(true, …)`, permanently dropping tenant-create coverage on every deployment (not gated on a live-data check like its siblings).
- **`employees.spec.ts`**: `resolveSeedFks()` falls back to Kenya-only literals (`NAIROBI_CITY`, `DEPT_7`, `DESIG_58`) with no skip-guard if no live employee resolves them — will fail confusingly rather than skip cleanly on a data-sparse fresh tenant.
- **`hrms-proxy.spec.ts`**: assertions only check `status < 500` / shallow shape, never the actual outbound request params the test claims to verify (offset/limit/tenantId injection) — a regression in that injection logic could go undetected.
- **`escalate-action-521.spec.ts`**: still uses a single fixed `waitForTimeout(3000)` + one-shot assertion, unlike sibling workflow specs in this same PR that correctly moved to `expect.poll(...)` for server-state verification — inconsistent and more flake-prone now that it seeds a fresh complaint per run.
- Onboarding: 3 of 16 phase specs (`phase1.spec.ts`, `phase1-branding.spec.ts`, `phase1-validation.spec.ts`) weren't migrated to the new shared `utils/onboarding.ts` helpers and still hand-roll tenant-code minting — reproducing the exact "same fix needed in seven places" drift the new helper's own doc-comment cites as the reason it was created.
- `README.md`'s spec-listing table appears to mis-map `citizen-otp-login.spec.ts` under the `tests/employee/` section (that file actually lives under `tests/citizen/`; `tests/employee/login.spec.ts` wasn't renamed).

### Low-severity / nits

- `boundary-jurisdiction-496.spec.ts`: converted from a hard-fail to `test.skip` on tenants lacking the Kenya persona — reasonable for portability, but the regression it guards effectively goes untested everywhere except that one deployment, with no tracked follow-up.
- `citizen-help-and-faq.spec.ts`: a regression test silently switched to `test.fixme` with no rationale in the diff — out of scope for a de-hardcoding pass.
- `pgr-idgen.ts` prefix regex (`[A-Z0-9]+`) vs. `employee-ui.ts` inbox-row regex (`[A-Z]+`) could diverge if a deployment's prefix contains a digit.
- Keycloak specs (`sso-config.spec.ts`) carry over a pre-existing tautological assertion (`clientSecret || clientSecret === '**********'`) — not introduced by this PR, but ported without cleanup.
- Minor stray literal: a comment in `wizard-pin-and-boundary-cascade.spec.ts` bakes in one test run's generated mobile number.

### What's solid

- The employee inbox filter/search and workflow ASSIGN/RESOLVE/REJECT specs are the strongest part of the PR: they genuinely assert result-set narrowing (not just the XHR param) and use `expect.poll` against real server state rather than trusting a UI toast.
- The Keycloak SSO/provisioning port is a faithful, careful de-hardcoding with no weakened assertions.
- `hardcoding.spec.ts`, `boundary-hierarchies.spec.ts`, `designations.spec.ts`, most citizen specs, and the onboarding tenant-code letters-only generation (`0→a…9→j` bijection, verified against `^[a-zA-Z. ]*$`) are correctly implemented.
- Good judgment calls: preferring `test.skip` with a clear reason over fabricating a passing assertion when tenant data is genuinely absent (seen in several files).

### Recommendation

Before merge: fix the `BASE_URL`/`pgr-idgen.ts` credential inconsistencies (#1, #2 — both are real, easily-hit correctness bugs for anyone running the suite fresh), and remove the unrelated planning doc (#3). The medium/low items are worth a follow-up pass but aren't blocking given the PR's stated scope and the thorough manual verification described in the PR body.

---
🤖 Review generated with [Claude Code](https://claude.com/claude-code)

vinothrallapalli-eGov (approved):

PR #1155, opened 2026-07-10
feat(pgr): implement employee department-scoped complaint search

priyanshu-egov:
- Add EmployeeDepartmentScopeService to restrict employee searches to their HRMS department(s)
- Introduce configurable department scope bypass roles (PGR_ADMIN, SUPERUSER, etc.) via pgr.department.scope.bypass.roles
- Add departmentCodes filter to PGRQueryBuilder to scope complaints by stored department code
- Extend RequestSearchCriteria with departmentCodes field for server-resolved scoping
- Update PGRService to invoke department scope resolution during employee search
- Update HRMSUtil to fetch current employee department assignments from HRMS
- Add department scope configuration to PGRConfiguration with sensible defaults
- Align employee search scoping with analytics row-scope rules for consistency
- Enable fail-closed behavior: employees with unresolved departments see no results

## Summary by CodeRabbit

* **New Features**
  * Added optional department-based access controls for employee complaint searches.
  * Employees with configured roles now see, count, and search complaints from their current HRMS departments only.
  * Searches return no results when an employee’s department information cannot be retrieved.
  * Existing unrestricted behavior remains available when no roles are configured.


vinothrallapalli-eGov:
## Review: employee department-scoped complaint search

This adds department-based access control to `_search`/`_count` so employees only see complaints in their own HRMS department. The intent is solid, but as implemented it has a correctness bug that defeats the feature plus a backward-compatibility gap.

### Blocking issues

1. **The department filter compares a NAME to a CODE, so it essentially never matches** (`PGRQueryBuilder.java:31`, `:146`). `PGRService#getDepartmentFromMDMS` stores the MDMS-resolved *display name* on `additionalDetails.department` (e.g. "Water Supply Department") whenever name resolution succeeds, only falling back to the raw code on failure. `HRMSUtil.getCurrentDepartment` returns the employee's raw HRMS **code** (e.g. "WS"). The new `ser.additionaldetails->>'department' IN (...)` filter joins these two, so under normal MDMS config it matches nothing — non-bypass employees get an empty result set / zero count for complaints that are genuinely theirs. This is the opposite of the class's own comment ("matching HRMS's code directly").

2. **`/_plainsearch` is a complete, unscoped bypass.** `PGRService.plainSearch()` (line 295) never calls `applyEmployeeDepartmentScope`. An employee restricted on `/v2/request/_search` can hit the unmodified `/v2/request/_plainsearch` with the same criteria and get every department's complaints back.

3. **No backward-compatibility switch — this is default-on for every deployment that upgrades.** `PGRService.applyEmployeeDepartmentScope` is invoked unconditionally for every `USERTYPE_EMPLOYEE` search/count call; the only escape hatch is the role-based `pgr.department.scope.bypass.roles` list, and the shipped `application.properties` narrows that from the code's own default (`PGR_ADMIN,SUPERUSER,MDMS_ADMIN,HRMS_ADMIN,STADMIN,SUPERVISOR,PGR_SUPERVISOR`) down to just `CMS_SCREENING_OFFICER`. There's no `pgr.department.scope.enabled`-style master flag to opt out entirely and get the pre-PR unrestricted-search behavior back — anyone who doesn't want this filtering yet (or hits issue #1 above) has no clean way to disable it short of bypassing every role their employees hold, which is fragile and easy to leave a role off of by accident.

   Suggest gating the whole feature behind a boolean property (default `false`, or default `true` only once #1 is fixed and rollout is deliberate), checked before `employeeDepartmentScopeService.applyScope(...)` is invoked:
   ```java
   if (!config.isDepartmentScopeEnabled()) return true;
   ```
   That lets existing deployments upgrade without a silent behavior change, and gives a real kill switch if #1 recurs in production.

4. **`application.properties` desyncs from the analytics bypass list.** The narrowed `pgr.department.scope.bypass.roles=CMS_SCREENING_OFFICER` only affects this new search/count scoping — `PrincipalScopeResolver`'s `TENANT_WIDE_ROLES` (used for analytics) is a separate hardcoded constant, still treating `PGR_ADMIN`/`SUPERVISOR`/etc. as tenant-wide. A `PGR_ADMIN` account now sees full tenant-wide analytics counts but gets empty/wrong-scoped complaint search results — contradicting the class-level comment claiming "the two stay consistent."

5. **Fail-closed on HRMS outage is indistinguishable from "no complaints."** `EmployeeDepartmentScopeService.fetchDepartments` catches any HRMS-call exception (timeout, 5xx, connection failure) and returns `emptySet()`, which `applyScope` treats identically to "employee genuinely has no department" — denying the search. This is logged at INFO only. Before this PR, employee search had no HRMS dependency at all; now an HRMS outage silently presents as empty results rather than an error, which will be hard to distinguish from real "no complaints" during an incident.

### Minor

6. **No fallback for complaints with missing/"NA" department.** Complaints predating consistent department population (or explicitly "NA", per the class's own javadoc) will never match any employee's department-code list and become invisible to every non-bypass employee, with no signal that results were filtered rather than empty.

7. **`HRMSUtil.getCurrentDepartment` duplicates existing boilerplate.** It's structurally identical to the pre-existing `getSupervisorUuid` (build URL → fetch → null-check/log → try/catch JsonPath), differing only in the JsonPath expression and return type; a third variant (`getDepartment`) already exists. Worth extracting a shared helper — this copy-paste pattern is plausibly how the code/name mismatch in #1 went unnoticed.

### Recommendation

Fix #1 (code vs. name) and #2 (`_plainsearch` bypass) before merge — as written, the feature either shows employees nothing or can be trivially routed around. Add the master enable/disable flag (#3) so this ships as an opt-in change rather than a silent behavior break for every existing deployment, and reconcile the bypass-role list with `PrincipalScopeResolver` (#4). #5–#7 are worth addressing in the same pass but aren't release-blocking on their own.

---
🤖 Review generated with [Claude Code](https://claude.com/claude-code)

backend/pgr-services/src/main/java/org/egov/pgr/repository/rowmapper/PGRQueryBuilder.java
```diff
@@ -137,6 +137,16 @@ public String getPGRSearchQuery(RequestSearchCriteria criteria, List<Object> pre
             addToPreparedStatement(preparedStmtList, serviceRequestIds);
         }
 
+        // Server-resolved department scope (see EmployeeDepartmentScopeService) — never bound
+        // from the request body. additionaldetails.department is populated with the raw MDMS
+        // department code (see PGRService#getDepartmentFromMDMS), matching HRMS's code directly.
+        Set<String> departmentCodes = criteria.getDepartmentCodes();
+        if (!CollectionUtils.isEmpty(departmentCodes)) {
+            addClauseIfRequired(preparedStmtList, builder);
+            builder.append(" ser.additionaldetails->>'department' IN (").append(createQuery(departmentCodes)).append(")");
```
vinothrallapalli-eGov, line 146: **Blocking: this filter compares a NAME to a CODE, so it will essentially never match.**

The comment above says `additionaldetails.department` is "populated with the raw MDMS department code... matching HRMS's code directly", but `PGRService#getDepartmentFromMDMS` actually stores the *resolved MDMS display name* (e.g. "Water Supply Department") whenever name-resolution succeeds, only falling back to the raw code on failure. `HRMSUtil.getCurrentDepartment` returns the employee's raw HRMS code (e.g. "WS"). Under normal MDMS config these never match, so non-bypass employees get an empty result set / zero count for complaints that are genuinely theirs — the feature silently defeats itself.

Also: no fallback for complaints where this field is missing/"NA" (per the class's own javadoc) — those become permanently invisible to every non-bypass employee with no signal that results were filtered.

backend/pgr-services/src/main/java/org/egov/pgr/service/PGRService.java
```diff
@@ -261,10 +269,28 @@ public ServiceRequest update(ServiceRequest request){
...
+        if (!USERTYPE_EMPLOYEE.equalsIgnoreCase(requestInfo.getUserInfo().getType()))
+            return true;
+
+        String scopeTenantId = criteria.getTenantId() != null
+                ? criteria.getTenantId() : requestInfo.getUserInfo().getTenantId();
+        return employeeDepartmentScopeService.applyScope(requestInfo, scopeTenantId, criteria);
+    }
+
 
     public List<ServiceWrapper> plainSearch(RequestInfo requestInfo, RequestSearchCriteria criteria) {
```
vinothrallapalli-eGov, line 298: **Blocking: `plainSearch()` never calls `applyEmployeeDepartmentScope`, so it's a complete, unscoped bypass of the new access control.**

An employee restricted on `/v2/request/_search` can hit `/v2/request/_plainsearch` with the same criteria and get every department's complaints back.

backend/pgr-services/src/main/resources/application.properties
```diff
@@ -127,6 +127,9 @@ egov.url.shortner.endpoint=/egov-url-shortening/shortener
 citizen.allowed.search.params=serviceRequestId,ids,mobileNumber,applicationStatus,tenantId
 employee.allowed.search.params=serviceRequestId,ids,mobileNumber,serviceCode,applicationStatus,tenantId
 
+#Roles that bypass employee department-based search/analytics scoping (see EmployeeDepartmentScopeService, PrincipalScopeResolver)
+pgr.department.scope.bypass.roles=CMS_SCREENING_OFFICER
```
vinothrallapalli-eGov: This narrows the bypass list from the code's own default (`PGR_ADMIN,SUPERUSER,MDMS_ADMIN,HRMS_ADMIN,STADMIN,SUPERVISOR,PGR_SUPERVISOR`) down to just `CMS_SCREENING_OFFICER`. `PrincipalScopeResolver`'s `TENANT_WIDE_ROLES` (used for analytics) is a separate hardcoded constant unaffected by this property, so `PGR_ADMIN`/`SUPERVISOR` accounts now see full tenant-wide analytics counts but get empty/wrong-scoped complaint search results — contradicting the class-level comment that claims the two lists "stay consistent".

backend/pgr-services/src/main/java/org/egov/pgr/service/EmployeeDepartmentScopeService.java
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * Pure HRMS department lookup for this one employee — CURRENT assignment(s) only, so a past
+     * (ended) assignment can't widen access. Empty (never null) if unresolved.
+     */
+    private Set<String> fetchDepartments(RequestInfo requestInfo, String tenantId, User user) {
+        try {
+            List<String> departments = hrmsUtil.getCurrentDepartment(user.getUuid(), requestInfo, tenantId);
+            return CollectionUtils.isEmpty(departments) ? Collections.emptySet() : new LinkedHashSet<>(departments);
+        } catch (Exception e) {
```
vinothrallapalli-eGov, line 75: Any HRMS-call exception (timeout, 5xx, connection failure) is caught here and treated identically to "employee genuinely has no department" — denying the search, logged at INFO only. Before this PR, employee search had no HRMS dependency at all; now an HRMS outage silently presents as empty results indistinguishable from "no complaints exist", which will be hard to catch during an incident.

backend/pgr-services/src/main/java/org/egov/pgr/util/HRMSUtil.java
```diff
@@ -65,6 +66,35 @@ public List<String> getDepartment(List<String> uuids, RequestInfo requestInfo,St
...
     }
 
+    /**
+     * Departments from the employee's CURRENT assignment(s) only (isCurrentAssignment==true) —
+     * unlike {@link #getDepartment}, which returns every assignment's department, past or
+     * present. Used for department-scoped search, where an ended assignment must not widen
+     * access. Returns an empty list (never throws) if the employee has no current assignment
+     * with a department, or the HRMS response can't be parsed.
+     */
+    public List<String> getCurrentDepartment(String uuid, RequestInfo requestInfo, String tenantId) {
```
vinothrallapalli-eGov, line 76: Nit: this is structurally identical to `getSupervisorUuid` below (build URL → fetch → null-check/log → try/catch JsonPath), differing only in the JsonPath expression and return type — and a third variant (`getDepartment` above) already exists. Worth extracting a shared helper; this copy-paste pattern is plausibly how the code/name mismatch flagged above went unnoticed.

backend/pgr-services/src/main/java/org/egov/pgr/service/PGRService.java
```diff
@@ -261,10 +269,28 @@ public ServiceRequest update(ServiceRequest request){
...
         Integer count = repository.getCount(criteria);
         return count;
     }
 
+    /**
+     * Employee-only: restricts {@code criteria} to the searching employee's own department(s),
+     * unless they hold a department-scope-bypass role. Citizen/system callers are untouched.
+     * Returns false when the caller must see nothing (search/count should short-circuit).
+     */
+    private boolean applyEmployeeDepartmentScope(RequestInfo requestInfo, RequestSearchCriteria criteria) {
```
vinothrallapalli-eGov, line 287: Invert this to an **opt-in** "which roles get scoped" list instead of an opt-out "which roles bypass" list, defaulted to empty.

Concretely:
- Rename `pgr.department.scope.bypass.roles` → something like `pgr.department.scope.roles`, default `` (empty) in `application.properties`, with a comment explaining the knob, e.g.:
  ```properties
  # Role codes to restrict to their own HRMS department on complaint search/count.
  # Empty (default) = no employee role is department-scoped — today's unrestricted
  # search behavior is preserved on upgrade. Add role codes here only once the
  # department-code matching (see PGRQueryBuilder) is verified correct for this
  # tenant's MDMS config, to roll out scoping for those specific roles.
  pgr.department.scope.roles=
  ```
- Invert the check in `EmployeeDepartmentScopeService.applyScope`: `if (!hasAnyRole(user, config.getDepartmentScopeRoles())) return true;` — only callers whose role IS in the configured list get scoped; everyone else is untouched.

This solves the backward-compatibility gap more directly than a separate enable/disable flag: the same property naturally acts as the kill switch (empty = fully off, matching pre-PR behavior for every deployment on upgrade), and there's no risk of forgetting to add a new employee role to a bypass list — you only ever have to remember to *add* a role when you deliberately want it scoped, not enumerate every role that shouldn't be.

Still need #1 (name/code mismatch) and #2 (`/_plainsearch` gap) fixed before adding any role to that list, and the analytics `PrincipalScopeResolver.TENANT_WIDE_ROLES` should probably be reconciled with whatever list ends up driving this so the two don't desync again.

backend/pgr-services/src/main/java/org/egov/pgr/config/PGRConfiguration.java
```diff
@@ -192,6 +192,11 @@ public MappingJackson2HttpMessageConverter jacksonConverter(ObjectMapper objectM
     @Value("${employee.allowed.search.params}")
     private String allowedEmployeeSearchParameters;
 
+    // Department scope — roles that bypass employee department-based scoping (complaint search
+    // AND the analytics row-scope, see PrincipalScopeResolver) and may see every department.
+    @Value("${pgr.department.scope.bypass.roles:PGR_ADMIN,SUPERUSER,MDMS_ADMIN,HRMS_ADMIN,STADMIN,SUPERVISOR,PGR_SUPERVISOR}")
+    private List<String> departmentScopeBypassRoles;
```
pradeepkumarcm-egov: Add a **`departmentScopeRoles`** configuration to define the roles for which department-based filtering should be applied. This will preserve the default product behavior while allowing department-specific filtering only for the configured roles. Any additional roles requiring this validation can simply be added to this configuration.

pradeepkumarcm-egov (approved):

subhashini-egov:
@priyanshu-egov can you please review coderabbit and vinoth's comments and resolve? Ping when it is ready for resolution.

PR #1157, opened 2026-07-10
feat(audit): audit-service parity — keep on k8s, add to compose (item #13)

singhalkarun:
## What changed (scope reversed after review)

This PR originally **disabled** the audit-service query API on k8s (deployment-parity item #13), on the assumption it was unused by CCRS.

Review feedback (@subhashini-egov): *audit-service is the mutation-audit trail — the only way to trace mutations across entities — and must not be disabled.*

So the scope is reversed: instead of removing audit-service, **bring it to both stacks** so audit is at parity everywhere.

### k8s
- Reverted the `coreservices-helmfile.yaml` change — `audit-service` is `installed: true` again.

### compose (new)
- Added the `audit-service` container.
- Added a one-shot `audit-service-migration` (published `egovio/audit-service-db:v2.9.2-4a60f20`) that creates `audit_service_schema` / `eg_audit_logs`, connected **directly to `postgres-db`** (not the pgbouncer `postgres` alias, which runs in transaction mode and breaks Flyway session locks).
- Compose already had the persister audit config (`configs/egov-persister/audit-service-persister.yml`) but **no `eg_audit_logs` table**, so audit capture there was non-functional. This completes the pipeline.

## Verification (smoke-tested against the live stack)
- Migrator applied 2 migrations and created `eg_audit_logs` (confirmed present).
- `audit-service` booted DB-connected and reports `{"status":"UP"}` at `/audit-service/actuator/health`.

## Follow-up (not in this PR)
- Full audit **capture** (records flowing into `eg_audit_logs`) additionally needs services to **emit** audit events via the egov-tracer — a broader per-service config, tracked separately. Schema + service + persister config are now in place.

> Note: the branch name (`chore/disable-unused-audit-service`) predates this scope reversal.

### DIFF
--- a/devops/deploy-as-code/charts/core-services/coreservices-helmfile.yaml
+++ b/devops/deploy-as-code/charts/core-services/coreservices-helmfile.yaml
@@ -35,8 +35,14 @@ releases:
     installed: true
     <<: *default
 
+  # Disabled: the audit *query* API is unused by CCRS. Its only UI consumer is the
+  # Utilities "Audit History" page, reachable only via a link from the Workbench
+  # module's MDMS view — and CCRS intentionally omits Workbench (see digit-ui
+  # App.js: enabledModules = ["Utilities","PGR"]). Audit *capture* is unaffected —
+  # audit-service-persister still runs on both stacks, so records keep persisting.
+  # (Deployment-parity item #13. Re-enable if a direct/compliance API consumer surfaces.)
   - name: audit-service
-    installed: true
+    installed: false
     <<: *default
 
   - name: egov-enc-service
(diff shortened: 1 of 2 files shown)


subhashini-egov:
@singhalkarun we should not disable the audit service. This will be needed for compliance purposes. What is the goal?

subhashini-egov (changes requested):
We should not disable the audit-service. This is the only way we will be able to trace mutations across entities. What's the goal of this PR? Is it to cull out unused things? We should leave this in for compliance.

subhashini-egov (approved):

Issue #1140, opened 2026-07-10
[Config] ke.nairobi ward boundary geometry not seeded + MAP_TENANT unset → citizen locality auto-fill silently fails


KDwevedi:
@KDwevedi — surfacing a **config/data gap** (not a frontend bug) found while investigating #920. The citizen create-complaint map does not auto-fill the **County / locality** cascade from the dropped pin. The frontend auto-fill code (CCRS#491 `BoundaryComponent.findWardPath`) is present and correct; it silently no-ops because the environment is misconfigured.

## Root cause (verified live on naipepea.digit.org boundary-service)
1. **No boundary geometry is seeded for `ke.nairobi`.** `POST /boundary-service/boundary/_search?tenantId=ke.nairobi&codes=...` returns **0** for every boundary — County (`NAIROBI_CITY`), SubCounty (`NAIROBI_CITY_MAKADARA`), and Ward (`NAIROBI_CITY_VIWANDANI`, …). The ADMIN **relationship** tree exists (County→SubCounty→Ward, incl. `NAIROBI_CITY_VIWANDANI`), but the **geometry** entities are empty.
2. **`MAP_TENANT` is not configured** in `globalConfigs.js` (only `stateTenantId=ke.nairobi`).

## Consequence chain
- `useTenantBoundaries` finds no usable geometry (MAP_TENANT unset + zero `ke.nairobi` geometry) → falls back to the **static bundled Nairobi-wards GeoJSON**.
- The static wards carry **different codes/names** than the seeded `NAIROBI_CITY_*` hierarchy.
- `resolveWard` (point-in-polygon on the static wards) returns a ward whose code doesn't match any ADMIN-tree node.
- `findWardPath` finds no match → **silent no-op → County/locality never auto-fills.** (Pincode is unaffected — it comes from Nominatim.)

Also spotted (data cleanup): stub nodes `COUNTY_001` / `SUBCOUNTY_001` / `WARD_001` and duplicate child nodes in the ADMIN tree.

## Fix
1. **Seed real geometry** for the `ke.nairobi` ADMIN wards (polygons under codes `NAIROBI_CITY_*`) so `boundary/_search` returns them. Tracks the known "ward geometry" gap.
2. **Set `MAP_TENANT=ke.nairobi`** in naipepea `globalConfigs.js` so the map uses the real wards instead of the static fallback.
3. Clean up the `*_001` stub + duplicate boundary nodes.

[...]

Issue #1003, opened 2026-07-01
[Feature] CMS: Allow ADMIN to edit RAINMAKER-PGR.MapConfig from the configurator UI


KDwevedi:
Split out from #882 (the localization half of that issue is handled in #1002; this is the remaining **feature** ask).

## Description
Expose `RAINMAKER-PGR.MapConfig` as an editable master in the configurator UI so an ADMIN can change the citizen-map appearance per tenant without a manual MDMS write.

## Fields to edit
`RAINMAKER-PGR.MapConfig` (MDMS v2, seeded at the map tenant):
- `baseMapTheme` — enum `voyager` / `light` / `dark` / `osm` (base tile theme; added in #894)
- `wardHighlightColor` — hex colour for the ward fill/outline (e.g. `#FFA74F`)
- (optional/advanced) raw `tileUrl` + `tileAttribution` override

## Where it's consumed today
- `useMapConfig` / `useWardHighlightColor` hooks in `digit-ui-esbuild/products/pgr` read this master at render for the complaint-location map (base tiles + ward colour).
- `MAP_TENANT` (globalConfigs) selects which tenant's config/boundaries the map uses.

## Acceptance criteria
- ADMIN can view + edit `baseMapTheme` and `wardHighlightColor` for the map tenant from the configurator.
- Save writes `RAINMAKER-PGR.MapConfig` via MDMS v2 `_create`/`_update` and busts the relevant cache.
- The citizen map reflects the change after a refresh.
- Sensible defaults shown when the master is absent (`voyager` + legacy orange).

## Notes
- Ward **geometry** and boundary **names** are tenant data (not part of this master) — see #882 / #1002. This issue is only the theme/colour config editor.

_Filed as the feature follow-up to #882._

KDwevedi:
Picked this up and it turned out to be bigger than the theme half. Audit + design proposal (and open questions) are in a discussion, with a draft PR behind it:

- Discussion: https://github.com/egovernments/Citizen-Complaint-Resolution-System/discussions/1163
- Draft PR: #1162

[...]

PR #1162, opened 2026-07-13
feat(maps): make RAINMAKER-PGR.MapConfig the single source of truth for map tooling + starting position

KDwevedi:
**Draft, for design review + hands-on trial.** Verified end-to-end against the live Bomet box; still draft pending the reviews on the [design discussion](https://github.com/egovernments/Citizen-Complaint-Resolution-System/discussions/1163). Companion to that discussion; subsumes the theme-only ask in #1003.

## Problem

Map configuration lived in **four** places and only one was runtime-editable — and the split fell exactly along the two halves of the feature request ("configure maps tooling **and starting map position**"):

| Layer | Carried | Editable? |
|---|---|---|
| MDMS `RAINMAKER-PGR.MapConfig` | `baseMapTheme`, `tileUrl`, `wardHighlightColor` | yes |
| `globalConfigs.js` (Ansible, deploy-time) | `MAP_CENTER`, `MAP_TENANT`, `HIERARCHY_TYPE` | no |
| Vite build env (`digit-ui-v2`) | `VITE_CITIZEN_MAP_DEFAULT_LAT/_LNG` | no |
| Hardcoded constants | zoom levels, dashboard tiles, Nominatim scope | no |

## What this does

Makes `RAINMAKER-PGR.MapConfig` the single source of truth, resolving every field **MDMS → globalConfigs → built-in default** so existing deployments are unaffected. Fields: `baseMapTheme` / `tileUrl` / `tileAttribution` / `wardHighlightColor` (already there) **+** `center`, `defaultZoom`, `minZoom`, `maxZoom`, `boundaryTenantId`, `geocodeCountryCodes`, `searchViewbox`.

**Two bugs fixed along the way:**
- The citizen address search was hardcoded to a Nairobi bounding box (`countrycodes=ke` + `bounded=1`), which silently discarded every result outside Nairobi — breaking address search for every other tenant. Now config-driven; no viewbox is sent unless one is set.
- `useTenantBoundaries` read its config in a `useEffect` with `[]` deps, so it couldn't react to config resolving. Fixed.

## Configurator: two ways in

**1. Auto-derived at onboarding.** Phase 2 already fetches the OSM boundary polygons, so the centroid/bbox of the onboarded area *is* the correct start centre, zoom and search extent. `deriveMapPosition` computes them at the end of the shared post-create pipeline and upserts MapConfig — the operator types nothing. (Sanity check: Bomet's boundaries derive to `(-0.80, 35.30)`, matching the hand-set `map_center` host_var `(-0.78, 35.34)`.) Zoom is fitted in projected space and clamped; boundaries without geometry are skipped; the upsert only updates a record the tenant *owns*, else shadow-creates, so it never rewrites a parent record inherited down the mdms-v2 tenant tree.

**2. A dedicated management editor** (`/manage/map-config`) — replacing the generic 12-input scroll whose ra-core save path silently swallowed submits (verified: theme changes never persisted). The new editor is self-managed with a direct `mdmsUpdate` (reliable save, same guard as above), and:
- fields grouped into Basemap / Starting position / Ward boundaries / Address search cards, each with a plain-language description;
- base map theme is a dropdown of the four presets, not a free-text box;
- per-field info-tooltips instead of paragraph help; country-code help gives real examples (`ke`, `mz`, `in`);
- a **live Leaflet preview** of the chosen basemap, ward colour and start position;
- **set the search area and start point by framing the map** ("Set search area to this view" / "Set start point to map centre") — no typing coordinates.

## MDMS schema, migration + docs

Adds the `RAINMAKER-PGR.MapConfig` schema (keyed on a stable `code`, not the colour). Because mdms-v2 schema codes are **immutable** over the API (`DUPLICATE_SCHEMA_CODE` / HTTP 501), a box that already carries the earlier hand-registered colour-keyed schema needs a DB-level fix: the Flyway migration `local-setup/docker/db-migrations/sql/egov-mdms/V20260715000000__…` (own history table) re-keys the legacy record to `DEFAULT` **preserving its colour**, rewrites the rogue schema **in place** to the correct definition for every tenant that has it, and registers the schema for any tenant with data but no schema. It's self-contained (no dependency on a default-data-handler restart), guarded, and idempotent — a no-op on a fresh/correct box. The embedded definition is generated from `schema/RAINMAKER-PGR.json` so the two can't drift. Dry-run-verified live in a rollback transaction (rogue schema → `x-unique:["code"]` with the full definition; record → `DEFAULT`).

Docs: **`docs/map-config.md`** — record shape, the MDMS→globalConfigs→default resolution table, consumers (incl. the Nominatim bounded-search gotcha and the `cacheTime` hard-refresh), how it's set, and the migration story.

## Reviewer decisions applied (from #1163)

- **Per-tenant, no root assumption** — the hook resolves the acting city; mdms-v2's own tenant-tree inheritance does the rest.
- **`hierarchyType` removed** from MapConfig — it's a boundary construct; belongs to a future default-boundary-hierarchy master (it's read in 8 PGR places today with inconsistent defaults). Consumers keep reading globalConfigs meanwhile.
- **Dashboard keeps its own map config.**
- **Q5/Q6 (@subhamegov): derive by default, expose as override** — implemented; the "frame on the map" capture is a first cut of his advanced visual editor.

## Verification

- configurator `vitest run` — **59 pass** (15 new: boundary-derivation math + `upsertMapConfig` tenant guard); `vite build` green; `tsc -b` clean for all touched files (one pre-existing error in an untouched file the Dockerfile already tolerates).
- **Live on Bomet** (`ke` / scratch tenants): create → edit → **save now persists**; auto-derivation, the theme dropdown, the live preview and the map-frame capture all confirmed; test record restored afterward. digit-ui bundle confirmed to read the new fields.

Refs #1003, #1140, #874.


KDwevedi:
Design discussion for this change (rationale, current-state audit, and the open questions): https://github.com/egovernments/Citizen-Complaint-Resolution-System/discussions/1163

Please take the design debate to the discussion — this PR is the concrete artifact to review against it.

subhashini-egov:
Left a comment on the discussion board.

KDwevedi:
## Live end-to-end verification + a deployment finding

Installed the configurator deps and actually ran things this time:

- `vitest run` — **59 pass** (15 new): the boundary-derivation math (bbox union, MultiPolygon nesting, out-of-range coord rejection, zoom clamping, zero-area viewbox omission, and Bomet's extent reproducing the operator's hand-set `map_center`) and the `upsertMapConfig` tenant guard (update only when the tenant owns the record; shadow-create, never rewrite, an inherited parent record).
- `vite build --base=/configurator/` — **green** (the app bundles with these changes).
- `tsc -b` — clean for every file I touched. (One pre-existing error remains in `NotificationProviderList.tsx`, which I didn't touch; the Dockerfile already tolerates it by building via vite, not tsc.)
- **Live round-trip against bomet** (`ke.testcitya` scratch tenant, employee auth): `_create` → `_update` (`defaultZoom` 10→14) → deactivate, each verified by read-back. My `upsertMapConfig` payloads are accepted by the real mdms-v2.

### Deployment finding that needs a decision

Probing the live box surfaced platform behaviour that changes the migration story (not the code):

1. **A schema code is immutable.** `schema/v1/_create` on an existing code returns `DUPLICATE_SCHEMA_CODE`; `schema/v1/_update` returns **HTTP 501**. You cannot change a master's `x-unique`/shape via API.
2. **`uniqueIdentifier` is server-derived from `x-unique`** — the value posted is ignored.
3. bomet already has a **rogue, hand-registered** MapConfig schema (not in `develop`) with `x-unique: ["wardHighlightColor"]` and only that one property — created 10ms after the `#22394D` data record. It's the colour-as-identity antipattern, schema-level.

Consequently, on **any box that already carries that rogue schema**, this PR's `x-unique: ["code"]` schema is **rejected**, and records keep keying on the colour. On a **fresh** box the PR's schema registers cleanly and everything behaves as designed. `additionalProperties` is open on the rogue schema, so the *runtime* still reads `center`/`zoom`/`viewbox` correctly on bomet — only edit-in-place / singleness is broken there.

**Migration for an existing box is therefore a DB step, not an API one**: delete the old `eg_mdms_schema_definition` row for `RAINMAKER-PGR.MapConfig` and deactivate the `#22394D` data record, then let DDH register the new schema. Flagging for the reviewers on the discussion — this is the concrete shape of the "singleton keying / migration" question (#1 there).

_All verification was local + against a scratch tenant; nothing was written to a production tenant, and the probe record was deactivated._

KDwevedi:
## Migration for existing boxes added

Flyway migration `local-setup/docker/db-migrations/sql/egov-mdms/V20260715000000__mapconfig_recode_from_colour_key.sql`, wired into `migrate-all.sh` under its own history table (`egov_mdms_schema_version`) so it never touches the mdms-v2 service's own flyway history.

It does two things, in order:

1. **Re-keys** the legacy `code`-less data record to the stable `DEFAULT` key, *preserving* the configured `wardHighlightColor` — deleting it would silently revert a deliberately-themed map to the default orange (Bomet's is `#22394D`). Deactivates any extra legacy rows first so the re-key can't collide on the PK.
2. **Drops** the rogue colour-keyed schema so the default-data-handler re-registers the correct `code`-keyed definition at startup. Keeps the schema definition in one place (DDH's JSON), not duplicated in SQL.

Guarded (`x-unique @> ["wardHighlightColor"]` for the schema; `not (data ? 'code')` for the record) and idempotent, so a fresh box — correct schema, or none — is a no-op.

**Verified by dry-run against live Bomet in a rollback transaction:**

```
=== BEFORE ===  schema-count 16 | legacy-data: ke #22394D {"wardHighlightColor":"#22394D"}
UPDATE 0   (no extra legacy rows to deactivate)
UPDATE 1   (re-key)
DELETE 16  (rogue schemas, all tenants — bootstrap had copied it)
=== AFTER  ===  schema-count 0 | ke: DEFAULT {"code":"DEFAULT","wardHighlightColor":"#22394D"}
ROLLBACK
```

So the colour is preserved and the record ends up correctly keyed, with the schema cleared for DDH to re-register.

One operational caveat, noted in the dir's README: `docker-compose.egov-digit.yaml` pulls a **pre-built** db-migrations image, so on an existing box this file only takes effect once that image is rebuilt — or by applying the SQL directly (it's idempotent; the README has the one-liner + the DDH restart).

KDwevedi:
**Deploy finding — the migration this PR ships no longer has an execution path after #1142.**

Deployed this branch (merged onto develop `7030a905f`) on the Bomet box. #1142 retired the unified `db-migrations` runner: `docker-compose.migrations.yml` overrides it with `entrypoint: ["true"]`, so `migrate-all.sh` — including the `egov-mdms` slot this PR adds — never executes on any box deployed with the new overlay set. `V20260715000000__mapconfig_recode_from_colour_key.sql` therefore only runs via the README's manual-psql path, on fresh and existing boxes alike.

**Fix to make before un-drafting:** port the cleanup to the item-10 convention — a dedicated migrator init-container in `docker-compose.migrations.yml` (history table `egov_mdms_schema_version`, gated on `db-history-normalize` like the others) — and drop the now-dead `migrate-all.sh` slot.

Data point: the SQL itself is verified live. Applied manually on Bomet after deploying this branch: `UPDATE 1` (ke record re-keyed to `DEFAULT`, `wardHighlightColor` preserved), `DELETE 16` (rogue colour-keyed schemas), then a `default-data-handler` restart registered the correct `x-unique: ["code"]` schema at `ke`. Matches the earlier dry-run exactly.

subhamegov:
@KDwevedi Please close this PR. We will open another issue for the additional piece I mentioned.

subhashini-egov (approved):

PR #1175, opened 2026-07-13
fix(db): never let a migrator replay from V1 over a populated dump (#1142 follow-up)

singhalkarun:
## Problem

#1142 re-baked the repo's own dump to the K8s `<service>_schema` history names, and
assumed every service team would rename theirs at source. They have not, and nothing
in the deploy checks. The fast-path overlay loads *whatever* file sits at
`db/full-dump.sql`.

Measured against a team-supplied dump (57 tables, 97,250 rows): the migrators find no
history under the names they expect, replay from V1 (2017) against populated tables, and

**97,250 rows → 26,377. 73% of the data destroyed.**

| Table | Before | After |
|---|---|---|
| `message` | 70,835 | **0** |
| `eg_enc_symmetric_keys` | 27 | **0** |
| `eg_enc_asymmetric_keys` | 27 | **0** |

Nine migrators crash with 42P07 — that crash is the only reason their data survived.
The two that do *not* crash are the ones that do the damage: `egov-localization` and
`egov-enc-service` open their V1 migration with `DROP TABLE IF EXISTS`, so they drop,
recreate empty, and **exit 0**. A green deploy that ate the data.

The encryption keys are the unrecoverable part: `eg_user.name` and `mobilenumber` are
ciphertext, and those keys were the only thing that could decrypt them. No re-seed
brings that PII back.

## Fix

A `db-history-normalize` init container, gated ahead of every migrator in compose — so
Ansible inherits it (it already runs these compose files, **no playbook change**) and a
bare `docker compose up` is covered too. It reads a shared map, renames legacy history
tables to canonical names, rebuilds empty orphan tables, and **aborts the deploy** on
anything it cannot prove safe.

| Canonical | Legacy alias | Data tables | Action |
|---|---|---|---|
| present | — | — | no-op |
| — | present | — | **rename** alias → canonical |
| present | present | — | **abort** — ambiguous |
| — | — | absent | no-op — fresh install |
| — | — | present, **0 rows** | **drop**; the migrator rebuilds them |
| — | — | present, **has rows** | **abort** — cannot prove which migrations are applied |

The one destructive path (`DROP` of empty orphan tables) takes an `ACCESS EXCLUSIVE`
lock and re-counts rows *inside the same transaction*, so a table that gains a row
between the check and the drop raises and rolls back rather than losing data.

**Verified:** with the normalizer, the same dump comes through a full migrator run
**byte-identical** — every pre-existing table matches on both row count and content
checksum, and all 11 migrators exit 0.

## Tests

- **Integration proof** — real dump + the real pinned `egovio/*-db` images → zero data change.
- **Negative test** — skip the normalizer → assert the data IS destroyed. If that ever
  stops failing, the guard has been silently disabled and the positive test alone would
  not tell us.
- Decision-table unit tests (12), idempotency, fresh-install.
- CI: a migrator whose `SCHEMA_TABLE` is missing from the map now fails the build.

## Notes for review

- `egov-otp` is expected to **apply**, not no-op: the dump ships `eg_token` with no
  history at all, so the normalizer drops the empty table and its migrator rebuilds it.
  Previously this hit 42P07 and blocked boot.
- The map carries 16 services, not the 13 in the spec — `egov-indexer`, `novu-bridge`
  and `digit-config-service` are needed so the map can replace the CI check's hardcoded
  `BASELINE_FRESH` set.
- **Not** in scope: verifying the dump's *schema* matches what its history claims. The
  rename assumes it. That is the schema-drift check, and it is its own piece of work.

Spec: `docs/superpowers/specs/2026-07-13-ansible-flyway-history-normalization-design.md`
Plan: `docs/superpowers/plans/2026-07-13-flyway-history-normalization.md`

### DIFF
--- a/.github/scripts/check-flyway-dump-alignment.py
+++ b/.github/scripts/check-flyway-dump-alignment.py
@@ -45,25 +45,36 @@ DUMP = ROOT / "local-setup/db/full-dump.sql"
 COMPOSE = ROOT / "local-setup/docker-compose.egov-digit.yaml"
 FAST_PATH = ROOT / "local-setup/docker-compose.fast-path.yml"
 MIGRATIONS = ROOT / "local-setup/docker-compose.migrations.yml"
+MAP = ROOT / "local-setup/db/flyway-history-map.yml"
 
-# ── Transitional allowlists (drive these to empty as item #10 lands) ──────────
-# Enabled services that legitimately create their history fresh: they claim a
-# table absent from the dump AND their data tables are not in the dump, so Flyway
-# baselines from empty with no 42P07 risk. Remove an entry only if that service's
-# schema gets baked into the dump (then it must align instead).
+
+def _load_history_map() -> dict:
+    with open(MAP) as fh:
+        return yaml.safe_load(fh) or {}
+
+
+_HISTORY_MAP = _load_history_map()
+
+# Services that legitimately create their history fresh: their data tables are not
+# in the dump, so Flyway baselines from empty with no 42P07 risk. Declared in the
+# map (`baseline_fresh: true`) so the map and this check cannot drift apart.
 BASELINE_FRESH = {
-    "egov_indexer_schema",         # egov-indexer      — no eg_indexer* in dump
-    "digit_config_service_schema", # digit-config-svc  — no config tables in dump
-    "novu_bridge_schema",          # novu-bridge       — no novu tables in dump
+    spec["canonical"] for spec in _HISTORY_MAP.values() if spec.get("baseline_fresh")
 }
 
-# History tables the dump ships whose owning service currently has Flyway OFF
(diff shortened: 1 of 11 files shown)


singhalkarun:
Auto-merged into `feat/item10-db-migration-parity` when its commits were pushed there — the guard and the Ansible wiring belong with the overlay they protect, so the team reviews the whole thing as one piece in #1142.

**Correcting a claim in this PR's description.** It said *"Ansible inherits the guard — it already runs these compose files."* **That was wrong.** The playbook never copied `docker-compose.migrations.yml` and never added it to `compose_files`, so an Ansible deploy ran none of the per-service migrators and none of the guard. Fixed in #1142 (`fix(ansible): actually deploy the DB-migration overlay`); end-to-end verification posted there.

Issue #432, opened 2026-04-22
Nai Pepea: Inbox issues


aarushibaser-lab:
**Describe the bug**
1. The inbox is full of resloved complaints. Resolved complaints should not be visible, it is not helpful. Need to see unresolved complaints only. 
2. Also SLA sorting is not working. I know there is atleast one unresolved complaint which I created a few mibutes back.
3. The filter card shows status as a filter but no fields below

**To Reproduce**
1. Go to Search Complaint (the inbox)
2. See error

**Expected behavior**
1. I expected to see unresolved complaints only. 
2. The SLA is to be populated and sortable if you click on the SLA. 
3. A field under status which allows me to select from a dropdown. 

**Screenshots**
Attached

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser: Arc

<img width="1910" height="946" alt="Image" src="https://github.com/user-attachments/assets/7a25a94e-0875-4552-ac83-d89007691930" />

KDwevedi:
UI half landed in [theflywheel/digit-ui-esbuild#47](https://github.com/theflywheel/digit-ui-esbuild/pull/47). Three changes:

1. **Default to open/actionable states.** Inbox previously returned every state including resolved/rejected/cancelled. Now defaults to PENDINGFORASSIGNMENT, PENDINGFORREASSIGNMENT, PENDINGATLME, PENDINGATSUPERVISOR when the operator hasn't checked anything in the Status filter.
2. **Status filter dropdown populated.** The hook used to return `statusMap: []`, so the WorkflowStatusFilter rendered an empty card. It now fetches the PGR business-service workflow (`/egov-workflow-v2/egov-wf/businessservice/_search?businessServices=PGR`) and maps the 11 returned states into `{statusid, state, businessservice}` shape.
3. **Sort icons removed from un-sortable columns.** Only `applicationStatus` is currently a valid `sortBy` value in pgr-services' `RequestSearchCriteria.SortBy` enum. Showing sort icons on Complaint No, Locality, Owner, SLA was misleading — the click silently no-op'd. Set `disableSortBy: true` on those four.

[...]

PR #1212, opened 2026-07-14
fix(pgr-inbox): make employee inbox column-header sort actually work (#922)

### LINKED ISSUE #922: [Bug] CMS: Employee: Inbox sort options not working
# Description

Employee: Inbox sort options not working 

---

# Steps to Reproduce

1. Login to CSR or GRO
2. Go to inbox
3. Click on column headers in inbox to check if sorting order changes 

---

# Expected Behavior

Sort button clicks but its non functional 

---

# Actual Behavior

https://github.com/user-attachments/assets/14b009fc-6d5e-48f9-a7a8-eef98321a9dc

vinothrallapalli-eGov:
Closes #922.

## RCA

Clicking a column header in the employee PGR inbox (CSR/GRO) had no effect, and once made clickable, the sort direction icon didn't update either. Investigation turned up **six independent bugs stacked on top of each other** — no single fix would have made sort work correctly; all had to be fixed together:

1. **Every column was `disableSortBy: true`.** `PGRSearchInboxConfig.js` had disabled sort on all five columns as a mitigation applied for the earlier issue #432 ("SLA sort inconsistent with pagination"). That mitigation hid the affordance entirely instead of fixing the underlying sort mechanism — an overcorrection that left the feature reported as "broken" again.
2. **No header click was ever wired to a network request.** The generic `ResultsDataTableWrapper` (used by the inbox's `react-data-table-component`-based table) had no `onSort` handler at all — clicking a header updated only the library's internal (purely visual) sort-direction state, never touched `tableForm`, and never triggered a refetch.
3. **A latent case-sensitivity bug**, unrelated to the header click itself: an existing "sort" toggle button in the search/filter form (`SearchComponent.js`) unconditionally seeds `tableForm.sortOrder` with a lowercase `"asc"` on mount. pgr-services' `RequestSearchCriteria.SortOrder` is a Java enum matched case-exactly, so a stray lowercase value 400s *every* inbox search — independent of whether the user ever touched a column header.
4. **The actual root cause, in the backend:** `PGRService.search()` unconditionally re-grouped and re-sorted the final result list by `createdTime DESC` in Java, *after* `PGRQueryBuilder` had already built a correct SQL `ORDER BY` for whatever `sortBy` was requested (`locality` / `applicationStatus` / `serviceRequestId` / `sla`). This silently discarded the SQL-level ordering on every call. Confirmed via Postgres query logs: the SQL sent to the DB was always correct (`ORDER BY ads.locality DESC` when requested), but the HTTP response order never reflected it. This means the elaborate SQL-sort work landed in prior PRs for #432 (SLA CASE-expression ordering, etc.) **never actually took effect end-to-end** — it was dead code from day one.
5. **A crash found during live verification**: once sort was actually enabled, clicking a header crashed the page. `CustomSVG.SortUp`/`SortDown` were defined with positional args `(style, onClick, fill, width, height)` instead of a destructured props object — used as `<CustomSVG.SortUp .../>`, React's implicit second (legacy context) argument landed in the `onClick` position as `{}`, and clicking triggered React's "Expected `onClick` listener to be a function, instead got a value of `object`" invariant. This bug was always there but dormant, since sort was never clickable before (bug #1).
6. **A follow-up reported after the first round of fixes: the sort icon never switched direction, always showing the same (up) arrow.** `ResultsDataTableWrapper` unmounts `<ResultsDataTable>` (and the underlying `<DataTable>`) on every `isLoading`/`isFetching` transition — which now happens on every sort click, since sorting triggers a real refetch. `react-data-table-component` only reads `defaultSortFieldId`/`defaultSortAsc` **once, on mount** to seed its internal "which column, which direction" state, and ignores prop changes afterward. So every sort-triggered remount wiped that internal memory and re-seeded it from the (unset) static default — the icon reset to the same direction after every single click, no matter how many times the operator clicked.

## Impact

- **User-facing:** Employees (CSR/GRO) could not sort the inbox by any column — clicking a header silently did nothing, and even after the core fix, the sort icon gave misleading feedback (never changing direction), which would have looked like a fresh bug to any reviewer/QA.
- **Backend:** `PGRService.search()`'s dead re-sort meant *no* caller of `/pgr-services/v2/request/_search` could ever get results ordered by anything other than `createdTime DESC`, regardless of the `sortBy` param sent — a latent bug affecting any other integration relying on this endpoint's documented sort support, not just this UI.
- **Blast radius of the fix:** Frontend changes to `ResultsDataTableWrapper`/`ResultsDataTable` are in the shared `digit-ui-components` package, but the new `sortServer`/`onSort`/`defaultSortFieldId`/`defaultSortAsc` wiring is a no-op for any other consumer that doesn't set a column's `sortKey` (verified: no other config in the repo currently sets `column.sortFunction`, so no other screen's row-ordering behavior changes; `defaultSortFieldId` falls back to `undefined`/no-op the same way `defaultSortAsc` already did). The `CustomSVG.SortUp`/`SortDown` fix is behavior-preserving for every existing call site (none of them pass an `onClick` prop today).

## Fix

1. `PGRSearchInboxConfig.js` — replaced blanket `disableSortBy: true` with a `sortKey` per sortable column (`serviceRequestId`, `locality`, `applicationStatus`, `sla`), matching pgr-services' `RequestSearchCriteria.SortBy` enum. Left `disableSortBy: true` only on "Owner" (assignee), which has no backend sort equivalent. Also added `defaultSortBy`/`defaultSortOrder` (`sla`/`ASC`) matching the backend's own default, so the header icon is correct even before the first click.
2. `ResultsDataTableWrapper.js` — added a real `handleSort(column, direction)` wired to the table's `onSort`, which writes `sortBy`/`sortOrder` into `tableForm` and resets to page 1, triggering a genuine server-side refetch. Set `sortServer={true}` so the table library doesn't also attempt a misleading client-side, per-page-only sort. Removed the dummy `(rowA, rowB) => 0` `sortFunction` fallback, which had the side effect of forcing `sortable: true` on every column (including ones marked `disableSortBy`) via the library's `sortable: column.sortable || !!column.sortFunction` check. Gave sortable columns a stable `id` (their `sortKey`), and now source `defaultSortFieldId`/`defaultSortAsc` from our own persisted `tableForm.sortBy`/`sortOrder` (which survives the remount, since it lives in the `InboxContext` reducer) instead of a static config value — so every remount re-seeds the library with the *current* sort state and the icon stays accurate.
3. `ResultsDataTable.js` — forwarded the new `sortServer`/`onSort`/`defaultSortFieldId` props through to the underlying `react-data-table-component`.
4. `UICustomizations.js` (`PGRInboxConfig.preProcess`) — read `sortBy`/`sortOrder` from `tableForm` (falling back to the existing default of `sla`/`ASC`, most-urgent-first) instead of hardcoding them; uppercase `sortOrder` defensively to absorb the lowercase `"asc"` written by the unrelated search-form toggle.
5. `PGRService.java` (`search()`) — removed the unconditional `TreeMap`-based re-sort by `createdTime DESC` that was discarding the SQL query's actual `ORDER BY`. The query builder already defaults to `ORDER BY ser_createdtime DESC` when no `sortBy` is given, so default (no explicit sort) behavior is unchanged; explicit `sortBy` now finally takes effect.
6. `CustomSVG.js` — fixed `SortUp`/`SortDown` to accept a destructured props object (matching every other icon in the file) instead of positional arguments, eliminating the `onClick={}` crash on header click.

## Test plan

- [x] Backend: rebuilt `pgr-services` and confirmed via direct API calls (and Postgres query logs) that `sortBy=locality`/`serviceRequestId` with `sortOrder=ASC|DESC` now produces genuinely different, correctly-grouped/ordered results (previously always returned `createdTime DESC` regardless of the param).
- [x] Frontend: rebuilt `digit-ui-esbuild` and hot-swapped into a running local-setup stack; logged in as an employee and drove the PGR inbox with a headless browser.
  - Confirmed clicking "Locality" fires `POST .../pgr-services/v2/request/_search?...sortBy=locality&sortOrder=ASC...` and the table re-renders with rows grouped by locality (previously: no network request, no reorder).
  - Confirmed no more page crash on header click (previously: React `onClick`-listener-is-object invariant violation).
  - Confirmed the "Owner" column remains non-interactive (no sort icon, no click), consistent with it having no backend sort support.
  - Confirmed default (no explicit sort) inbox load behavior is unchanged (SLA-ascending, most urgent first).
  - Confirmed the sort icon now genuinely toggles: clicked "Locality" three times in a row and verified (via the SVG's computed CSS transform/opacity, and visually via screenshots) that the arrow alternates ↓ / ↑ / ↓ instead of staying fixed.
- [x] Automated regression: added `tests/employee/pgr-inbox-sort-922.spec.ts` (Playwright, drives the real deployment end-to-end, no mocking). Seeds fresh complaints into two distinct ward-level boundaries, then verifies: clicking "Locality" fires `sortBy=locality&sortOrder=ASC` and the row order is independently re-checked via the PGR API (raw `address.locality.code`, not the localized display text); a second click toggles to `DESC` and the order flips; the header's sort-icon CSS class alternates asc/desc across three repeat clicks; no page error is thrown; "Current Owner" renders no sort icon. All 4 cases pass against the live local-setup stack with this fix deployed — see the comment below for the run output and screenshots (`docs/qa/pgr-inbox-sort-922/`).
- [ ] Manual QA in a deployed environment: log in as CSR/GRO, click each sortable column header (Complaint No, Locality, Current Status, SLA), verify ascending/descending toggle (icon *and* row order) and that the full paginated result set (not just the current page) reflects the new order across page-size changes.
- [ ] Regression: confirm other `digit-ui-components` consumers of `ResultsDataTableWrapper`/`ResultsDataTable` (e.g. HRMS, workbench, sandbox inboxes) are unaffected — none currently set `column.sortFunction`, so their sort-icon-toggles-but-doesn't-reorder behavior is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added server-side sorting for complaint search results.
  * Users can sort supported columns, including complaint number, locality, status, and SLA days remaining.
  * Sorting selections now persist during pagination and refresh results from the selected page.

* **Bug Fixes**
  * Search results now respect the selected sort column and direction instead of always sorting by creation time.
  * Improved default handling for sort settings to ensure consistent result ordering.


vinothrallapalli-eGov:
## Automated regression test + verification screenshots

Added `tests/integration-tests/tests/employee/pgr-inbox-sort-922.spec.ts` — a Playwright spec that drives the real deployment end-to-end (no mocking) and locks in this fix. Ran it against the running local-setup stack with the actual fix deployed:

```
npx playwright test tests/employee/pgr-inbox-sort-922.spec.ts --project=chromium --no-deps
```

```
Running 4 tests using 1 worker

  ✓  clicking Locality sorts ascending server-side, by the raw code, with no page error @p0 (8.9s)
  ✓  clicking Locality again toggles to descending, in both the request and the row order @p0 (10.4s)
  ✓  the sort icon toggles direction across repeated clicks, instead of staying on one arrow @p0 (9.5s)
  ✓  "Current Owner" has no sort affordance — pgr-services has no sort-by equivalent for it @p0 (4.5s)

  4 passed (35.1s)
```

Screenshots below are the actual final-state captures Playwright took during this passing run (`docs/qa/pgr-inbox-sort-922/`), not manually staged:

**1. Locality sorted ascending — request + row order both correct**
locality ascending

**2. Second click toggles to descending — row order flips, icon flips**
locality descending toggle

**3. Icon direction toggles across repeated clicks (regression for the remount-resets-sort-state bug)**
icon direction toggle

**4. "Current Owner" has no sort icon (no backend sort-by equivalent) — SLA column still shows its default arrow for comparison**
owner column no sort icon

Note: `--no-deps` skips this suite's shared UI-login `setup` project — its login-page selectors (`#username`) don't match this particular local-setup deployment's login form, an unrelated pre-existing gap, not something this spec needs (it authenticates via API token injection, same as its sibling specs in `tests/employee/`).

---

**Update:** the `--no-deps` workaround above is no longer needed. Root-caused the underlying issue: this suite's shared `setup` project (configurator UI login, used only by admin/configurator specs) was hard-failing on deployments that don't serve the configurator at all (`GET /configurator/login` 404s here), and Playwright treats a failed dependency as blocking *every* dependent project — silently skipping employee/citizen specs too. Fixed in `tests/fixtures/auth.setup.ts` to skip gracefully (not fail) when the route isn't reachable, since employee/citizen specs authenticate independently and never read its output anyway. Re-ran the full invocation without `--no-deps` — `setup` now shows as skipped, and all 4 `pgr-inbox-sort-922.spec.ts` tests still pass:

```
Running 7 tests using 1 worker
  ✓  citizen-setup
  -  setup (skipped: configurator not reachable on this target)
  ✓  lifecycle-setup
  ✓  clicking Locality sorts ascending server-side, by the raw code, with no page error @p0
  ✓  clicking Locality again toggles to descending, in both the request and the row order @p0
  ✓  the sort icon toggles direction across repeated clicks, instead of staying on one arrow @p0
  ✓  "Current Owner" has no sort affordance — pgr-services has no sort-by equivalent for it @p0

  1 skipped
  6 passed (37.1s)
```

subhashini-egov (approved):

PR #1239, opened 2026-07-15
fix(pgr): restore consent checkboxes on citizen create — explicit opt-in (CCSD-1979 revisited)

Hari-egov:
## Requirement clarified

Consents must be **shown at citizen create** (explicit opt-in); they are hidden **only on the details/view screens** (CCSD-1988 — untouched by this PR). The earlier CCSD-1979 change removed them from create per the product decision recorded on that ticket; this restores them per the clarified requirement.

## What's restored (and improved)

- The two consent checkboxes (TRUTHFULNESS, DATA_PROCESSING) back on step 3, above the confidential toggle, with required markers — now **localized** via `PGR_CONSENT_TRUTHFULNESS_LABEL` / `PGR_CONSENT_DATA_PROCESSING_LABEL` (en + pt seeds added; PT: *"Declaro que as informações prestadas são verdadeiras e exactas."* / *"Autorizo o tratamento dos meus dados pessoais para a tramitação desta manifestação."*) instead of the old hardcoded English.
- **Submit gate restored**: both consents required once an authority/template is in play.
- Payload now records the consents the citizen actually ticked (was: implicit auto-accept of both).

## Browser-verified (local, full citizen wizard walk)

- Step 3 renders both checkboxes with `*`; Submit stays disabled until description + mandatory dynamic field (`Institute Name`) + **both consents**.
- Unchecking either consent re-disables Submit; re-checking enables it.
- Details/view screens unchanged — consents still hidden there.

## Env note

The two new localization keys ride the DDH seeds for new tenants; existing envs get them from the migration runner (seeds missing keys only) or a one-off upsert.

Jira: CCSD-1979

### DIFF
--- a/digit-ui-esbuild/products/pgr/src/pages/citizen/Create/CreatePGRFlowV2.tsx
+++ b/digit-ui-esbuild/products/pgr/src/pages/citizen/Create/CreatePGRFlowV2.tsx
@@ -263,9 +263,9 @@ function fieldsFromSchema(schema: any): TemplateField[] {
 }
 
 // Complaint-level declarations confirmed before submit (BRD).
-const REQUIRED_CONSENTS: ReadonlyArray<{ code: string; label: string }> = [
-  { code: "TRUTHFULNESS", label: "I declare that the information provided is true and accurate." },
-  { code: "DATA_PROCESSING", label: "I consent to my data being processed to handle this complaint." },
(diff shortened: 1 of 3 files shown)


Hari-egov:
Superseded by the combined PR #1241 (both Moz QA fixes consolidated per review preference).

PR #1243, opened 2026-07-15
perf(digit-ui): gzip + Cache-Control on the /digit-ui bundle (~7MB → ~2MB)

Hari-egov:
## What & why

The digit-ui SPA ships as a single, non-code-split bundle (`index.js` ~7-8 MB) plus ~1.5 MB CSS, served **uncompressed** with **no `Cache-Control`**. Cold loads (fresh WebView / new device / after a deploy) pull the whole thing raw over HTTP/1.1 — the reported "app takes 15-20s but browser 2s" gap is exactly this (browser has it warm-cached; a cold client re-downloads ~9 MB).

This adds, **scoped to the `/digit-ui` location only** (no Kong / API route touched):

| | Before | After |
|---|---|---|
| `index.js` on the wire | 7.2 MB | **2.1 MB** — `Content-Encoding: gzip` (3.3×) |
| Cache-Control | *(none)* | `no-cache` → ETag revalidation, cheap `304` |
| Stale-bundle-after-deploy risk | present (heuristic caching) | removed |

`Cache-Control` is deliberately `no-cache`, **not** `immutable`/long-max-age: the filename is not content-hashed, so long caching would strand users on an old build. Pure nginx config — **no bundle rebuild, no app redeploy, reload-reversible.**

## Verified locally, end-to-end (this session)

Confirmed on the running local stack at `:80`, not just in theory:
- `content-encoding: gzip` + `cache-control: no-cache` at the edge; **7.2 MB → 2.1 MB** transferred.
- Conditional GET returns **`HTTP 304`, 0 bytes** (no re-download when unchanged).
- **globalConfigs.js injection intact** — `sub_filter` runs before gzip, so the config the SPA depends on is unaffected (verified in served `index.html`).
- **Kong / API untouched** — gzip lives inside `location /digit-ui` only; proxied API responses are byte-identical (an earlier draft had it server-level + `application/json`, which *would* have compressed API JSON — corrected to location-scoped).
- CSS in the same location also compresses.

## Two serving modes (discovered during local testing — important for deploy)

- **Disk-serve** (ansible static/vendored mode — **pilot & prod**): host nginx serves `/digit-ui` from disk (`alias /opt/digit-ui-esbuild/build/`). gzip goes **inside that location** — this PR’s `nginx-site.conf.j2` change. No proxy trick needed.
- **Proxy-to-container** (local compose, and any box where the host `proxy_pass`es `/digit-ui`): the container gzips fine, but the host must fetch identity and gzip at the edge — add `proxy_set_header Accept-Encoding "";` alongside the gzip block. This is spelled out in the runbook; the container-side configs in this PR (`playbook-deploy.yml`, `local-setup/nginx/digit-ui.conf`) carry gzip for that path.

## Files
- `local-setup/ansible/templates/nginx-site.conf.j2` — **pilot/prod (disk-serve) path**
- `local-setup/ansible/playbook-deploy.yml` + `local-setup/nginx/digit-ui.conf` — container-mode, kept consistent
- `docs/ops/digit-ui-compression.md` — **full deploy / verify / rollback runbook** (mode detection, ansible + manual paths, mctd HTTP-only note)

## Deploy (summary; full version in the runbook)
1. Merge → `git pull` on the box.
2. `sudo grep -RnE "location .*/digit-ui|alias|proxy_pass" /etc/nginx/` → confirm disk-serve vs proxy.
3. Re-render via ansible (`--tags nginx`) **or** hand-add the gzip block to the live `/digit-ui` location.
4. `nginx -t && systemctl reload nginx` — no rebuild.
5. Verify with the runbook curls; hard-refresh the UI. Rollback = restore backup + reload.

Follow-up (separate, build-level, needs a browser regression pass): esbuild content-hashing (`[name]-[hash]`) + `immutable`, then code-splitting.

### DIFF
--- a/docs/ops/digit-ui-compression.md
+++ b/docs/ops/digit-ui-compression.md
@@ -42,17 +42,29 @@ sudo grep -RnE "location .*/digit-ui|alias|proxy_pass" /etc/nginx/ | grep -i dig
 
 ## Deploy — ansible-managed box (recommended)
 
-```bash
-# 1. on your workstation / CI: merge this PR, then on the box:
-cd ~/Citizen-Complaint-Resolution-System && git pull
+Use `deploy.sh` — it regenerates the inventory (`inventory/hosts.yml`, which is
+NOT checked in) from `inventory/host_vars/` and then runs the playbook. The
+host-nginx render tasks carry a `nginx` tag, so a surgical reload is:
 
-# 2. re-render the host nginx site + reload (regenerates from nginx-site.conf.j2):
-#    run the site-config play (adjust inventory/limit to the target host)
-ansible-playbook -i local-setup/ansible/inventory.yml \
-  local-setup/ansible/playbook-deploy.yml --tags nginx --limit <host>
-#    (if the play has no nginx tag, run the nginx template task or the relevant play;
-#     it ends with `nginx -t && systemctl reload nginx`.)
+```bash
+# 1. pull the merged change onto wherever you run ansible from:
+cd <repo> && git pull
+
+# 2. re-render just the host nginx site config + reload, for one host:
+cd local-setup/ansible
+./deploy.sh <host> --tags nginx
+#   deploy.sh → ansible-playbook -i inventory/hosts.yml --limit <host> \
+#               playbook-deploy.yml --tags nginx
+#   the tagged tasks render templates/nginx-site.conf.j2 and notify the
+#   `Reload nginx` handler (which runs `nginx -t` then reloads).
 ```
+
+> ⚠️ **`nginx_preserve_vhost` hosts skip this.** On any host with
+> `nginx_preserve_vhost: true` in its `host_vars` (hand-crafted vhost — e.g.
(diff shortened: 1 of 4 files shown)


Hari-egov:
Good catch — fixed in `966d9f44`. The runbook's headline command was indeed a silent no-op:
- `-i local-setup/ansible/inventory.yml` → corrected to the sanctioned entrypoint `cd local-setup/ansible && ./deploy.sh <host> --tags nginx` (deploy.sh regenerates the real `inventory/hosts.yml` from host_vars, then runs the playbook).
- `--tags nginx` matched nothing → added `tags: ['nginx']` to the four host-nginx site tasks (render / enable / remove-default / flush-handlers), so it's now a real surgical target that re-renders `nginx-site.conf.j2` and fires the `Reload nginx` handler.

Also added a caveat the runbook was missing: hosts with `nginx_preserve_vhost: true` (e.g. Bomet) have `when: not nginx_preserve_vhost` on the render task, so ansible reports `changed=0` and touches nothing — those must use the manual path. Same fix cherry-picked to #1244.

Issue #1169, opened 2026-07-13
[Enhancement] Use Dashboard in Portuguese


subhamegov:
**Role(s):** As a Citizen, Employee, Supervisor, Account Administrator, whoever is permitted to access the piublished version of the restricted dashboard
**Action:** I want the dashboard to automatically load in the configured default language and allow me to switch languages without losing my context.
**Rationale:** The Complaint Management dashboard should respect tenant-specific localization settings to provide a consistent user experience across multilingual deployments such as Mozambique and Bomet.
**Benefit Hypothesis:** Enables multilingual deployments without code changes, improves usability for local users, and aligns the dashboard with the platform-wide localization framework.

**Acceptance Criteria:**
* All dashboard labels, KPI names, descriptions, filters, tooltips, error messages, empty states, and static UI text are retrieved from the Localization Service.
* The dashboard loads in the default locale configured for the tenant in Global Configuration.
* Portuguese text is displayed correctly in the Mozambique tenant.
* Portuguese text is available and displayed correctly in the Bomet tenant when Portuguese is selected or configured.
* Users can switch between supported languages without refreshing the page.
* Switching languages does not reset selected filters, date ranges, dashboard configuration, KPI selections, or navigation context.
* The complete dashboard updates to the selected language after the locale is changed.
* Translation keys are used consistently across dashboard components.
* Missing translation keys fall back to the configured fallback locale.
* Translation keys or unresolved localization codes are not displayed to users.
* Existing supported languages continue to work without regression.
* New languages can be added through the Localization Service without requiring dashboard code changes.

**Non-Functional:**
* Translation lookup shall not introduce noticeable delay during dashboard loading.
* No dashboard functionality shall regress due to localization.
* Translation implementation shall support future languages without code changes.
* All tests are automated and run
* Load localization data as per review by Claude code for now.

PR #1258, opened 2026-07-15
feat(dashboard): resolve the route/card role gate from MDMS (dss.DashboardConfig) with code fallback

KDwevedi:
## Motivation

The supervisor dashboard's nav gate is hardcoded: `digit-ui-esbuild/products/dashboard/roles.js` ships `DASHBOARD_ROLES = ["SUPERVISOR","PGR_SUPERVISOR","GRO","DGRO","PGR_LME","PGR_ADMIN","SUPERUSER"]`, and both `Module.js` (route guard) and `DashboardCard.js` (home card) gate on it. Deployments with a custom role taxonomy — e.g. the `CMS_*` roles on moz — are locked out of the dashboard entirely unless they rebuild the FE bundle. Discussion #784 ("Frontend gates") flagged exactly this. Moving the list to MDMS also gives the G7 configurator a natural editing surface for it later.

## What changes

- **New MDMS master `dss.DashboardConfig`** (state root tenant, single record):
  ```json
  { "id": "default", "allowedRoles": ["SUPERVISOR", "PGR_SUPERVISOR", "GRO", "DGRO", "PGR_LME", "PGR_ADMIN", "SUPERUSER"] }
  ```
  Schema: `local-setup/db/notif-mdms-seed/schemas/dss.DashboardConfig.json` (follows the `dss.KpiDefinition`/`dss.DashboardPack` conventions). Seed data: `ansible/nairobi-mdms/mdms/dss/DashboardConfig.json` — mirrors today's hardcoded list, so seeded deployments behave identically.
- **FE:** `roles.js` keeps `DASHBOARD_ROLES` as the **fallback** and gains `useDashboardAccess()` — fetches `dss.DashboardConfig` at the state root (`Digit.ULBService.getStateId()`) via the same MDMS v1-compat search the rest of the UI uses (`Digit.Hooks.useCustomMDMS` → `POST /egov-mdms-service/v1/_search`), cached per session via react-query (`staleTime/cacheTime: Infinity`, shared cache entry across both consumers).
- Both `Module.js` and `DashboardCard.js` gate on the shared hook, so route and card keep agreeing. While the gate resolves, the route holds on the `Loader` and the card renders nothing — no flash of redirect/card.
- Docs (`docs/dashboard-configuration/`) updated: the open-the-view knob is now MDMS, no rebuild.

## Semantics

- Record present with a non-empty `allowedRoles` array → those role codes.
- Record absent / fetch error / malformed shape → the built-in `DASHBOARD_ROLES`, **silently**. A deployment that never seeds the master is byte-for-byte backward compatible.

## Explicitly NOT a security boundary

This gates navigation surfaces only (home card + `/employee/dashboard` route). The data plane is already enforced server-side by the analytics catalog RBAC (`rbac.visibleTo`) and scope RBAC — a user who deep-links past this gate still sees only what the API grants their token.

## Deployment notes

- Backend untouched.
- Validated live on bomet: schema registered + `default` record seeded at tenant `ke`, branch temp-deployed, bundle serves the MDMS-resolved gate.
- moz applicability: once its dashboard port lands (#1169), seeding a `dss.DashboardConfig` record with the `CMS_*` roles unlocks the dashboard there with no rebuild.
- **Fresh compose installs** (review follow-up): `local-setup/db/full-dump.sql` now seeds the `dss.DashboardConfig` schema definition + the `default` record at tenant `ke` (`dashboard-config-schema-001` / `dashboard-config-data-001`, matching the hand-seed idiom of `theme-config-*-001`), so new deployments get the master out of the box. Validated by loading the dump into a scratch postgres:14 container (`ON_ERROR_STOP=1`) — both rows land and the JSON parses.
- The record shape is additive: `numberFormat` arrives via #1272, which extends the same `default` record — per the seed-convergence rule, whichever PR merges later updates the dump row.

## Summary by CodeRabbit

* **New Features**
  * Dashboard view access roles can now be configured via MDMS (`dss.DashboardConfig.allowedRoles`) without frontend rebuild.
  * Added a tenant default dashboard configuration (`id: "default"`) with role allowlisting.
  * Route and card access checks now share the same MDMS-driven authorization logic.

* **Bug Fixes**
  * Prevented dashboard card/route UI from flashing while access permissions are still resolving.
  * Added fallback to built-in dashboard roles when MDMS config is missing or malformed.

* **Documentation**
  * Updated dashboard configuration, access, embedding, and setup guidance to match MDMS-based gating.
  * Added/clarified schema validation rules for `dss.DashboardConfig`.

### DIFF
--- a/ansible/nairobi-mdms/mdms/dss/DashboardConfig.json
+++ b/ansible/nairobi-mdms/mdms/dss/DashboardConfig.json
@@ -1,0 +1,17 @@
+[
+  {
+    "tenantId": "ke",
+    "data": {
+      "id": "default",
+      "allowedRoles": [
+        "SUPERVISOR",
+        "PGR_SUPERVISOR",
+        "GRO",
+        "DGRO",
+        "PGR_LME",
+        "PGR_ADMIN",
+        "SUPERUSER"
+      ]
+    }
+  }
+]
--- a/digit-ui-esbuild/products/dashboard/DashboardCard.js
+++ b/digit-ui-esbuild/products/dashboard/DashboardCard.js
@@ -1,14 +1,17 @@
 import { EmployeeModuleCard } from "@egovernments/digit-ui-react-components";
 import React from "react";
 import { useTranslation } from "react-i18next";
-import { DASHBOARD_ROLES } from "./roles";
+import { useDashboardAccess } from "./roles";
 
 const DashboardCard = () => {
   const { t } = useTranslation();
 
-  // Same tenant-agnostic check as the DashboardModule route guard (roles live
-  // at the state root tenant), so the card and the route always agree.
-  if (!Digit.UserService.hasAccess(DASHBOARD_ROLES)) {
+  // Same MDMS-resolved, tenant-agnostic check as the DashboardModule route
+  // guard (shared useDashboardAccess hook — one react-query cache entry), so
+  // the card and the route always agree. Render nothing while it resolves so
+  // the card doesn't flash in or out.
+  const { allowed, loading } = useDashboardAccess();
+  if (loading || !allowed) {
     return null;
   }
 
--- a/digit-ui-esbuild/products/dashboard/Module.js
+++ b/digit-ui-esbuild/products/dashboard/Module.js
@@ -3,15 +3,17 @@ import { Redirect } from "react-router-dom";
 import { Loader } from "@egovernments/digit-ui-react-components";
 import AdminDashboard from "./src/AdminDashboard";
 import DashboardCard from "./DashboardCard";
-import { DASHBOARD_ROLES } from "./roles";
+import { DASHBOARD_ROLES, useDashboardAccess } from "./roles";
 
 export { DASHBOARD_ROLES };
 
 // Mounted by core AppModules at /{contextPath}/employee/dashboard INSIDE the
 // employee chrome (topbar + sidebar). AppModules already guarantees a logged-in
-// session, so the only guard needed here is the role check — tenant-agnostic
-// (see roles.js) — before rendering the dashboard in embedded mode (which
-// suppresses its standalone shell: internal sidebar, login gate, 100dvh layout).
+// session, so the only guard needed here is the role check — resolved from
+// MDMS (dss.DashboardConfig) with the roles.js list as fallback, checked
+// tenant-agnostically (see roles.js) — before rendering the dashboard in
+// embedded mode (which suppresses its standalone shell: internal sidebar,
+// login gate, 100dvh layout).
 const DashboardModule = ({ stateCode }) => {
   // Lazy-load this module's localization bundles into the host i18next
   // (same pattern as PGRModule): rainmaker-dashboard for the dashboard's own
@@ -24,11 +26,14 @@ const DashboardModule = ({ stateCode }) => {
     language: Digit.StoreData.getCurrentLanguage(),
     modulePrefix: "rainmaker",
   });
-  if (!Digit.UserService.hasAccess(DASHBOARD_ROLES)) {
-    return <Redirect to={`/${window?.contextPath}/employee`} />;
-  }
-  if (isLoading) {
+  const { allowed, loading: accessLoading } = useDashboardAccess();
+  // While the MDMS-backed gate resolves, hold on the Loader (never the
+  // Redirect) so an eventually-allowed role doesn't flash away from the route.
+  if (accessLoading || isLoading) {
     return <Loader />;
+  }
+  if (!allowed) {
+    return <Redirect to={`/${window?.contextPath}/employee`} />;
   }
   return <AdminDashboard embedded />;
 };
--- a/digit-ui-esbuild/products/dashboard/roles.js
+++ b/digit-ui-esbuild/products/dashboard/roles.js
@@ -1,6 +1,64 @@
-// Roles allowed to open the supervisor dashboard. Checked tenant-agnostically
-// (role CODE only, via Digit.UserService.hasAccess) because employee roles live
-// at the state root tenant ("ke") while the working tenant may be a city tenant —
-// Digit.Utils.didEmployeeHasAtleastOneRole filters by current tenant and would
-// wrongly hide the dashboard there.
+// FALLBACK roles allowed to open the supervisor dashboard, used whenever the
+// dss.DashboardConfig MDMS record is absent or unreadable (fresh tenants,
+// MDMS down, schema not registered) so existing deployments keep working
+// unchanged. Deployments with a custom role taxonomy (e.g. CMS_* roles)
+// override this via MDMS instead of a rebuild — see useDashboardAccess below.
+//
+// Checked tenant-agnostically (role CODE only, via Digit.UserService.hasAccess)
+// because employee roles live at the state root tenant ("ke") while the working
+// tenant may be a city tenant — Digit.Utils.didEmployeeHasAtleastOneRole filters
+// by current tenant and would wrongly hide the dashboard there.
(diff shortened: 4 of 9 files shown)


KDwevedi:
**Convergence note — three in-flight `dss.DashboardConfig` writers** (this PR, #1272, #1282):

- **Shared cache is real, refactor still owed.** `useDashboardAccess` here and #1272's `useDashboardConfig` issue the identical `useCustomMDMS(stateId, "dss", [{ name: "DashboardConfig" }])` call — same react-query key `[tenantId, moduleName, masterDetails]` (options/select don't enter the key), `staleTime`/`cacheTime` Infinity on both — so it's one request per session whichever merges first. Follow-up once both are in: refactor `useDashboardAccess` to consume `../useDashboardConfig`. (Already noted on #1272; recording it here so it's on both PRs.)
- **Merge order is schema-safe in any order.** #1272 carries this PR's full record (`id` + `allowedRoles`) plus optional `numberFormat`, and its schema keeps `required: ["id","allowedRoles"]` — so #1272-first cannot produce a record this PR's schema rejects. Whichever of the two merges **second** hits an add/add conflict on `ansible/nairobi-mdms/mdms/dss/DashboardConfig.json` + `local-setup/db/notif-mdms-seed/schemas/dss.DashboardConfig.json`: resolve by taking #1272's superset versions wholesale.
- **#1282 composes cleanly with either** (verified via `git merge-tree`: no conflicts against this branch or #1272). It only *documents* `departmentScoping`; absence = `enforced`, so adding `"departmentScoping": "enforced"` to the seed after it lands is cosmetic.
- **Live-box note:** bootstrap copy is create-if-absent — a tenant seeded before `numberFormat` lands needs the mdms-v2 `_update` runbook from #1272's description; re-running seeds won't add the field.

ansible/nairobi-mdms/mdms/dss/DashboardConfig.json
```diff
@@ -0,0 +1,17 @@
+[
+  {
+    "tenantId": "ke",
```
vinothrallapalli-eGov, line 3: @KDwevedi The newly added master data and schema are not included in the `full-dump.sql` file. These new  schema and data won't seed for new deployments.

vinothrallapalli-eGov (changes requested):

ansible/nairobi-mdms/mdms/dss/DashboardConfig.json
KDwevedi: Good catch — fixed in e6f9233: `full-dump.sql` now seeds both the `dss.DashboardConfig` schema and the `default` record at `ke` (verified by loading the full dump into a scratch postgres, clean). #1272 later extends the same record with `numberFormat`; whichever merges second updates the dump row.

Hari-egov:
@vinothrallapalli-eGov the fix for your comment (e6f9233) checks out — I verified it independently, ready for re-review:

**Dump seeding verified:**
- Loaded this branch's `full-dump.sql` into a scratch postgres:14 with `ON_ERROR_STOP=1` — clean load, zero errors.
- Both rows land correctly: `dashboard-config-schema-001` in `eg_mdms_schema_definition` and `dashboard-config-data-001` in `eg_mdms_data`, both at tenant `ke`, valid jsonb — schema `required: ["id","allowedRoles"]`, record carries all 7 roles, byte-identical to the ansible/notif-mdms-seed files.
- `ke` is the right state root for fresh installs (the dump ships the full `ke` MDMS set incl. `ke.nairobi`), the row idiom matches the existing `theme-config-*-001` hand-seed convention, and there are no duplicate-key collisions. `seed.sh` only registers TemplateBinding/ProviderDetail, so the dump is the sole fresh-install seed path — no double-seed conflict.

**On the red `test` check:** Local Setup CI has been failing on *every* branch since 2026-06-24 — `static/deployment-contracts.test.ts:50` still expects the old task name *"…seed INTERNAL_USER system user on state_root for HRMS"* after the playbook task was renamed to *"…(retry until tenant UserValidation is live)"*, plus the pgr-tenant/pgr-workflow smoke suites fail on the CI stack on all branches. Not caused by this PR — raising it separately.

**Merge coordination:** whichever of #1258/#1272 merges second takes #1272's superset versions of the two DashboardConfig files and adds `numberFormat` to the dump row (#1272 doesn't touch `full-dump.sql` yet).

local-setup/db/full-dump.sql
```diff
@@ -19763,6 +19763,7 @@ theme-config-data-001	pg	themeconfig	common-masters.ThemeConfig	{"id": "themecon
...
 a51e8099-4433-4409-b0ac-5b5873d78957	pg.citya	HRMS|All	CMS-BOUNDARY.HierarchySchema	{"hierarchy": "ADMIN", "department": "All", "moduleName": "HRMS", "lowestHierarchy": "Locality", "highestHierarchy": "Zone"}	t	system	system	1781074295688	1781074295688
 91e3b381-dfc4-4dfd-9768-3f193e57c088	pg.citya	CRS_BOUNDARY_DATA	CRS-ADMIN-CONSOLE.adminSchema	{"title": "CRS_BOUNDARY_DATA", "properties": {"numberProperties": [{"name": "CRS_LAT", "type": "number", "isRequired": true, "description": "Latitude", "orderNumber": 2}, {"name": "CRS_LONG", "type": "number", "isRequired": true, "description": "Longitude", "orderNumber": 3}], "stringProperties": [{"name": "CRS_BOUNDARY_CODE", "type": "string", "isRequired": true, "description": "Boundary Code", "orderNumber": 1, "freezeColumn": true}]}, "campaignType": "all"}	t	system	system	1781074295688	1781074295688
+dashboard-config-data-001	ke	default	dss.DashboardConfig	{"id": "default", "allowedRoles": ["SUPERVISOR", "PGR_SUPERVISOR", "GRO", "DGRO", "PGR_LME", "PGR_ADMIN", "SUPERUSER"]}	t	system-mdms-seed	system-mdms-seed	1770614666000	1770614666000
```
vinothrallapalli-eGov, line 19766: The default data should be loaded with 'pg' tenant.

KDwevedi: Done in 45a70b2d8 — the `default` record + its schema now also seed at `pg` (`dashboard-config-data-002` / `dashboard-config-schema-002`), matching the dump's system-seed idiom (same as `theme-config-*-001` at `pg`) and the ansible template's default `ui_state_tenant_id: pg`. I kept the `ke` pair alongside because `local-setup/nginx/globalConfigs.js` pins `stateTenantId = "ke"`, so the compose UI's gate query (`Digit.ULBService.getStateId()`) resolves against `ke` — dual-root seeding of the same master is already the dump's pattern (e.g. `ACCESSCONTROL-ROLES.roles` schema at `pg`/`statea`/`ke`). Re-validated the full dump on a scratch postgres:14 with `ON_ERROR_STOP=1`; both roots return the record with `allowedRoles` parsed. Happy to drop the `ke` pair instead if you'd rather the dump stay pg-only.

Hari-egov:
@KDwevedi one open item left — CodeRabbit's minor finding: add `minLength: 1` to `allowedRoles.items` so MDMS rejects empty role strings at validation time instead of the hook silently filtering them at runtime. If you take it, please apply it in all four copies so they stay consistent: both dump schema rows (`dashboard-config-schema-001` / `-002`), `local-setup/db/notif-mdms-seed/schemas/dss.DashboardConfig.json`, and flag it on #1272 so its superset schema picks it up too. If you'd rather not, resolve the thread with a brief reason — either way works, it's non-blocking.

@vinothrallapalli-eGov both your comments are addressed in 45a70b2d (record + schema now seed at both `pg` and `ke`, re-validated with a clean scratch-postgres load) — could you re-review/approve when you get a chance? The red `test` check is the repo-wide Local Setup CI breakage tracked in #1296, unrelated to this PR.

vinothrallapalli-eGov (approved):

PR #1260, opened 2026-07-15
Fix/pgr department search api

priyanshu-egov:
feat(pgr): add admin cross-department complaint search endpoint
- Add AdminComplaintSearchService to handle SUPERUSER cross-department search with department filter resolution
- Add AdminComplaintSearchController with POST /pgr-services/v2/request/_admin/_search endpoint
- Add AdminSearchCriteria and AdminSearchResponse models for admin search requests and paginated results
- Extend PGRQueryBuilder to support lastModifiedTime sorting option
- Enhance EmployeeDepartmentScopeService to resolve department names to codes for complaint matching
- Update RequestSearchCriteria with skipEmployeeDepartmentScope flag to prevent auto-scoping in admin searches
- Update PGRService to expose count method and support department scope skipping
- Update MDMSUtils to cache department name-to-code mappings for efficient resolution
- Enforce pagination limits (default 20, max 50) and validate required tenantId parameter
- Department filters match both MDMS codes and names, with multiple departments OR'd together


vinothrallapalli-eGov:
@priyanshu-egov This is customisation changes, can't merge to product. Keep these changes in customization branch.

Hari-egov:
Ran this branch locally (jar baked onto \`develop-0a4abaa\`, swapped into the compose stack) and exercised the endpoint through the gateway against real mz.igsae data. Findings:

**Working ✓**
- Gateway pass-through: \`/pgr-services/v2/request/_admin/_search\` routes fine via Kong — no gateway config needed.
- \`departmentCode\` multi-value OR works, matching both codes and names (e.g. \`IP_OPS\` matched a row whose stored department is the display name "Intellectual Property Operations").
- \`departmentCode=NA\` cleanly returns the unresolved-department complaints (the "No department" bucket).
- \`serviceRequestId\`, \`fromDate\`/\`toDate\`, \`sortBy=lastModifiedTime\`, \`sortOrder\` all behave; \`offset\` paging returns distinct, correctly-ordered pages.
- Existing \`_search\` unaffected (regression checked).

**Bug ✗ — \`totalCount\` echoes the page size, not the total**
```
limit=2  → totalCount=2
limit=5  → totalCount=5
limit=20 → totalCount=20
limit=50 → totalCount=50   (same unfiltered tenant query each time)
```
Looks like the count query receives the criteria with limit/offset applied. The FE pagination ("Showing 1–20 of N", page numbers) needs the true filtered total — could you make the count path ignore limit/offset (mirroring \`_search\`+\`_count\` semantics)?

Also worth noting for the FE (no action needed): the stored \`additionalDetail.department\` value is sometimes a CODE (\`EDUCATION_OPS\`) and sometimes a NAME ("Education Sector Operations") depending on how the complaint was created — the FE will normalize for display.

Hari-egov:
More findings from FE integration testing against a local build of this PR (64-complaint dataset at `mz.ige`):

1. **Sort order not applied on the final result at larger limits** — `sortBy=createdTime&sortOrder=ASC&limit=3` returns the oldest rows correctly, but the same query with `limit=51` returned a mid-range row (13 Jul) first, and even within a page same-day rows come back unordered (13:19 → 11:05 → 10:44 for ASC). Looks like the ORDER BY is applied in an inner/id query and the enriched outer result isn't re-sorted.
2. **`departmentCode=NA` doesn't match complaints whose `additionalDetail` has no `department` key** — only a literal `"department": "NA"` matches. Legacy/enrichment-skipped complaints without the key become unfilterable. Suggest treating absent/null as NA.
3. **`limit` is clamped to 50** (expected behaviour, just noting it interacts with the totalCount bug already reported — a client can't over-fetch past 50 to detect the next page; the FE now probes the next offset instead).
4. FYI: `additionalDetail.department` sent by the client on `_create` is overwritten by type-routing enrichment — fine for real flows, worth knowing for test-data setups.

Issue #1251, opened 2026-07-15
[Bug] Mozambique dashboard: number format should follow EU convention (56,5% and 52.560)


subhashini-egov:
## Reported

- Reporter: @subhashini-egov
- Source: Sprint 13 end-of-sprint triage (2026-07-14)

## Symptom

On the Mozambique deployment dashboard:
- Percentages render as `56.5%` — should render as `56,5%` (comma decimal)
- Counts render as `52,560` — should render as `52.560` (period thousands)

Mozambique/Lusophone convention: `.` for thousands, `,` for decimal.

## Related work — please review before starting

**Existing feature issue: #1213 [Enhancement] [MVP] Configure the default number display format for a tenant** (assignee @KDwevedi, milestone Release 2.20). That issue covers the full tenant-level number-format-config capability. The fix path for this bug is likely that feature landing.

## Ask

- P0 because Mozambique pilot users are currently seeing incorrect number formats in dashboards.
- If #1213 is on track to land in Sprint 14 for the mz tenant, this issue closes when that lands.
- If #1213 slips, we need an interim Mozambique-only hardcoded fix.

## Classification

- **Feature**: feature:internationalization
- **Area**: area:frontend
- **Priority**: P0
- **Target**: Sprint 14

Issue #1213, opened 2026-07-14
[Enhancement] [MVP] Configure the default number display format for a tenant


subhamegov:
# Current Behavior

How does this work today?
Only supports one format configurable from the backend.
---

# Proposed Improvement

As a tenant administrator,
I want to configure the default number display format for my tenant,
so that numbers are displayed according to the convention used in my country

As a citizen or an employee end user
I want to see default numbers displayed in the number display format I am comfortable with
So that I dont feel out of place.

---

# Why It Matters

Number formatting conventions vary across countries. For example, Portuguese-speaking countries commonly use a period as the thousands separator and a comma as the decimal separator, while English-speaking countries use the opposite convention.

For the MVP, number formatting should be configured once at the tenant level and applied consistently, irrespective of the language selected by the user.

Reference: 
Google sheets custom display format code for Lusuphone countries: #,##0.00
Google sheets custom display format code for Francophone countries: # ##0,00

---

# Acceptance Criteria

- The Configurator provides a user-friendly field for selecting the tenant’s default number display format.
- Administrators can choose from predefined, human-readable options instead of entering a technical format code.
- Each option displays an example using the same sample number.
- The configured format is applied consistently to all applicable numeric values, including KPIs, charts, tables, tooltips and downloaded reports where supported.
- Changing the user interface language does not change the number display format. (This is applicable for the MVP story only) 
- The configuration does not change the underlying numeric value or calculation precision. It only changes how the value is displayed.
- Display values in drop downs or filters, hover states on KPIs must also display using this number format

Definition of Done
- The number display format can be configured without editing code or configuration files.
- The selected format is reflected consistently across supported dashboard components.
- Automated tests cover all three predefined formats.
- Existing calculations, sorting, filtering and aggregation continue to use numeric values rather than formatted strings.
- Configuration and administrator documentation are updated.

Out of Scope
- Selecting a different number format for each language.
- Allowing administrators to create arbitrary custom number format patterns.
- Different formatting rules for individual dashboard widgets or fields.

# Classification

> Add these labels before submitting. See [CONTRIBUTING.md](../CONTRIBUTING.md) for the full list.

- **Feature**: 
- **Area**: 
- **Priority**:

subhamegov:
cc: @aarushibaser-lab

PR #1272, opened 2026-07-16
feat(dashboard): tenant-configurable number display format (#1213, fixes #1251)

KDwevedi:
Implements #1213 — locale-configurable number display format for the supervisor dashboard. Fixes #1251.

> **Requirement change (2026-07-17):** the mask is now resolved **per LOCALE, not per tenant**. On a multi-locale tenant (bomet: `en_IN` + `fr_FR` + `pt_PT`) each user sees numbers in their **selected language's** convention; switching the TopBar language reformats every number in place. The original tenant-wide behavior survives as the legacy string form (one mask for all locales).

## What

A per-locale display mask, `dss.DashboardConfig.numberFormat` (state root tenant, record id `default`), applied to **every number the dashboard renders** — KPI values and deltas, table cells, bar/line/stacked value labels, axis ticks, chart tooltips, map hovers, and duration strings. The mask contributes **separators only**; decimal counts stay per-KPI (`viz.format` semantics), and unit suffixes (`%`, `pp`, `/5`, `h`, `hr/hrs/day/days`) stay at the call sites.

`numberFormat` accepts two shapes:

- **Object keyed by locale code** (canonical), with optional `default` — resolved as `numberFormat[language] ?? numberFormat.default ?? unconfigured`:

  ```json
  { "en_IN": "#,##0.00", "pt_PT": "#.##0,00", "fr_FR": "# ##0,00", "default": "#,##0.00" }
  ```

- **String** (legacy, kept for back-compat): one mask applied for every locale — i.e. the tenant-wide behavior this PR originally shipped.

| Selected language | Mask | 12348248 renders as |
|---|---|---|
| en_IN | `#,##0.00` | `12,348,248.00` |
| pt_PT | `#.##0,00` | `12.348.248,00` |
| fr_FR | `# ##0,00` | `12 348 248,00` |

Parser rule (`products/dashboard/src/utils/numberFormat.js`): placeholder chars `[#0]` are stripped **before** separator detection; first remaining char = grouping separator, last = decimal separator; a lone separator in the 3-placeholder thousands position (`#,##0`) is grouping, anything else (`0,00`) is decimal. Invalid masks — and locales with no entry and no `default` — resolve to "unconfigured".

## Unconfigured = unchanged, guaranteed

`formatNumber` returns `null` when no mask is configured, and every one of the ~12 call-site files branches `masked ?? <existing expression>` — so tenants without the field (or locales without a mask) render **byte-identically** to today, including Apex formatters that must keep returning numbers. This also fixes #1251 for masked tenants: table integer cells were `String(Math.round(v))` with no grouping at all.

## Render gate + re-prime on language switch

The parsed mask lives in a module-level store (same pattern as `i18n/localeRuntime.js`) because the presentation configs are plain modules hooks can't reach. `AdminDashboard` primes the store **synchronously during render** — `setNumberFormatMask(resolveNumberFormatMask(config.numberFormat, language))` — and holds the dashboard (one session-cached MDMS request, `retry: false`) until the config query settles, so the first painted frame is already masked.

`AdminDashboard` subscribes to the locale runtime (`useDashboardT`), so a language switch re-renders it and re-runs the same synchronous prime with the new locale's mask; parents render before children, so the store already holds the new mask when any tile re-renders.

**Apex redraw correctness:** react-apexcharts (1.4.1) decides whether to call `updateOptions` by comparing `JSON.stringify(options)` — which **drops formatter closures** — so a mask-only change would never redraw baked data labels / axis ticks / tooltips. Every Apex chart whose options bake `formatNumber` closures (`StackedBarChart`, `LineChart`, `DepartmentBarChart`, `HorizontalBarChart`) embeds a stringifiable `_numberFormatStamp` (the resolved mask string) in its memoized options and deps on it. All other surfaces (KpiTile, DashboardTable, PieChart, ComplaintsAtRiskTable, Leaflet map tooltips) format at render/redraw time and already re-render on language change from the l10n work.

## Relationship to #1258

- Ships `products/dashboard/useDashboardConfig.js` as **the** shared loader for this master (identical react-query key to #1258's `useDashboardAccess`: same `useCustomMDMS(stateId, "dss", [{name:"DashboardConfig"}])`, `staleTime`/`cacheTime` Infinity, `retry: false`, mdmsv2 falsy, and its cjson shape-guards verbatim). When #1258 lands, `useDashboardAccess` should be refactored to consume this hook — one fetch per login either way, merge order doesn't matter.
- Seed files carry the **full** record (`id` + `allowedRoles` + `numberFormat`) and the full schema since #1258 is unmerged; whoever lands second resolves a trivial JSON overlap.
- Closes a latent #1258 gap: `dss.DashboardConfig` is now in digit-mcp `tenant_bootstrap`'s copy list — nothing seeded this master on new tenant roots before, so the role gate and the mask would both have silently fallen back to built-ins.

## CSV export stays raw

Display-only mask: the CSV export keeps raw dot-decimal values (commented at the export site) so Excel/import pipelines keep parsing them as numbers. Dates are untouched (already locale-driven). Durations DO take the mask decimal separator ("2,5 hrs") — they're numbers.

## Seeds / live-tenant runbook

The ke reference seed now carries the object form with bomet's three real conventions (`en_IN`/`pt_PT`/`fr_FR` + `default` en mask), and the `dss.DashboardConfig` schema allows `oneOf` [string, locale-keyed object]. To change a live tenant, **update the existing record only** via mdms-v2 (never re-create the schema):

```
POST /mdms-v2/v2/_update/dss.DashboardConfig
uniqueIdentifier: "default"   (tenantId = state root, e.g. "ke")
data: { "id": "default", "allowedRoles": [...unchanged...],
        "numberFormat": { "en_IN": "#,##0.00", "pt_PT": "#.##0,00", "fr_FR": "# ##0,00", "default": "#,##0.00" } }
```

Then hard-refresh the dashboard (the config is session-cached).

## Validation

- `npm run build` (digit-ui-esbuild) green; digit-mcp `npm run build` (tsc) green.
- 23 unit tests (`node --test products/dashboard/src/utils/numberFormat.test.js`): the original 16 parser/formatter tests untouched, +7 for per-locale resolution (exact match, `default` fallback, string legacy, malformed objects), an end-to-end separator flip across a simulated en→pt→fr→en language switch, and the options-stamp lifecycle.
- Schema validated with ajv: object seed and legacy string both pass; number / empty object / non-string entries reject.
- Smoke: under `#.##0,00`, `formatNumber(52560, "integer")` → `52.560`; `formatNumber(56.5, "percentOneDecimal")` → `56,5` (suffix at call site).
- Grep gate: no remaining `toFixed`/`toLocaleString`/`Math.round` producing displayed numbers outside `formatNumber` fallbacks, except (intentional): header/as-of **dates** (out of scope), pie slice pct (integer 0–100 — mask-invariant), SVG path coords / layout math / axis-bound and index computations (not display strings), and the dead pre-inversion exports in `kpiDisplay.js` (`parseKpiListItems`/`buildKpiContextText` — no render path).

## Seed convergence with #1258

`numberFormat` is an additive field on the same `dss.DashboardConfig` `default` record that #1258 seeds (including the `full-dump.sql` rows for fresh compose installs) — whichever PR merges later updates the ansible/nairobi-mdms record, the seed schema, and the `full-dump.sql` row to the merged shape.


KDwevedi:
Merged `develop` into the branch (4233c3072) to clear the conflict with #1283: both sides added one import line after `evaluateCompose` in `KpiTile.jsx` — kept both (`applyGroupByToColumns` from develop, `formatNumber` from this branch). No semantic overlap (#1283 touches `renderTable`/`groupBy`; this branch touches `formatDeltaDisplay`/`applyFormat`). Verified on the merged tree: 16/16 `numberFormat` unit tests pass and the full `npm run build` (digit-ui-esbuild) is green. PR is mergeable again.

Hari-egov:
@KDwevedi the post-#1258 convergence on this branch looks complete (superset seeds, `numberFormat` in both dump rows, `roles.js` → `useDashboardConfig`). Two small things to close this out: (1) optionally fold in CodeRabbit's `minLength: 1` on `allowedRoles.items` from #1258 — it merged without it, and this PR's schema copies are the natural place; (2) checks are green apart from the repo-wide #1296, so please get a re-review requested so this can merge.

KDwevedi:
@Hari-egov folded in — 18b7572da adds `minLength: 1` to `allowedRoles.items` in both schema copies this PR carries (`local-setup/db/notif-mdms-seed/schemas/dss.DashboardConfig.json` + the two dump schema rows). Review requests are already pending with the codeowner pool, so this is ready whenever someone picks it up.

vinothrallapalli-eGov (approved):

PR #1274, opened 2026-07-16
feat(migration): banner + gzip phases in the unified runner

Hari-egov:
## ccrs-migrate: `banner` + `gzip` phases

Extends the unified runner with the two remaining env-setup steps, following its standards (idempotent, `--dry-run` aware, continue-on-error with stable error codes + remediation, **never overwrites existing data**).

### `banner` (runs in the default pipeline)
- **Schema**: registers `tenant.citymodule` from the DDH seed when absent (definition includes `bannerImage`). When the schema exists **without** `bannerImage`, reports `CITYMODULE_SCHEMA_DRIFT` with remediation → `docs/migration/fix-citymodule.sh` (#1210) — MDMS has no schema-update API and the runner never touches the DB.
- **Rows**: creates *missing* citymodule rows only (Workbench/PGR/HRMS from the seed); existing rows untouched.
- **Banner value**: `--banner-url <url>` (env `BANNER_URL`) fills the PGR row's `bannerImage` **only when empty**; an existing value is never overwritten (a differing value is reported).

### `gzip` (opt-in: `--gzip`)
- Always **probes** `<host>/digit-ui/index.js` with `Accept-Encoding: gzip` (the runbook's own verify) — `OK` when `content-encoding: gzip` comes back, so it doubles as a post-deploy check from anywhere.
- When not enabled **and the script runs on the serving box** (config auto-discovered under `/etc/nginx`, or passed via `--nginx-conf`): timestamped backup → inserts the runbook's gzip + `Cache-Control: no-cache` block inside the `/digit-ui` location (adds `proxy_set_header Accept-Encoding ""` in proxy mode) → `nginx -t` with **automatic rollback** on failure → reload → re-probe.
- Anywhere else: `PARTIAL` + remediation → `docs/ops/digit-ui-compression.md` (`./deploy.sh <host> --tags nginx` on ansible boxes). Warns about the `add_header` scope-replacement caveat when the target file already sets headers.

### Verified
- `--phases auth,banner --dry-run` against local: schema detected (with `bannerImage`), row plan printed, current bannerImage reported.
- `--phases auth,gzip --gzip --dry-run` against local **and** pilot: both correctly detected as already enabled (`content-encoding: gzip · cache-control: no-cache`).

### DIFF
--- a/docs/migration/ccrs-migrate.cjs
+++ b/docs/migration/ccrs-migrate.cjs
@@ -766,14 +766,24 @@ async function phaseBanner() {
   let wanted = [];
   try { wanted = JSON.parse(JSON.stringify(readJson(SEED.citymoduleRows)).split('{tenantid}').join(CFG.state)); }
   catch (e) { return record('banner', OUTCOME.FAILED, `cannot read citymodule row seed: ${e.message}`, 'SEED_FILE_MISSING', `Expected ${SEED.citymoduleRows}.`); }
-  const { rows } = await mdmsSearch(CFG.state, SCHEMA);
-  const byCode = new Map((rows || []).filter((m) => m.isActive).map((m) => [m.data && m.data.code, m]));
+  // NB: mdmsSearch().rows strips each record down to .data — the FULL rows
+  // (id/uniqueIdentifier/isActive, required by _update) come back in .raw.
+  const fullRows = (schemaCode) => mdmsSearch(CFG.state, schemaCode).then((r) => (r.raw || []).filter((m) => m.isActive));
+  let live2 = await fullRows(SCHEMA);
+  let byCode = new Map(live2.map((m) => [m.data && m.data.code, m]));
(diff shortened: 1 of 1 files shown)


Hari-egov:
Both 🔴 findings addressed:

- gzip location targeting: the picker now iterates ALL `location …/digit-ui` blocks, skips `return 30x` redirect stubs and prefers blocks containing `alias/root/proxy_pass/try_files` — verified against `nginx-site.conf.j2` (the `location = /digit-ui` stub is skipped, `location /digit-ui/` is chosen).
- `{ rows }` vs full records: fixed in 5c3c62bc (phaseBanner reads `.raw`, duplicate creates labeled honestly, one-pass banner set) — verified end-to-end on local incl. the write path.
- Also added the 3s persister grace after schema create before row creates (consistent with fix-citymodule.sh).

Issue #1280, opened 2026-07-16
Complaints carry no department: department-scoped employees see structurally empty dashboards


KDwevedi:
## Gap (promoted from #1166's residual-state notes — it needs its own tracking)

On the Mozambique deployment, 32 of 34 complaint facts have an EMPTY `department_code`, and the HRMS assignment taxonomy (e.g. `central`) is disjoint from the department codes that do appear in facts (e.g. `DEPT_5`). The analytics ABAC resolver scopes every employee to their HRMS department — by design, even SUPERVISOR/SUPERUSER when an assignment exists — so **any normally-configured employee sees an empty dashboard regardless of catalog, localization, or UI work**. The current demo works only because a test user's assignment was hand-blanked.

## Why it matters

Department scoping (#908 Part B/C) is only as good as the department data. Until complaints are department-enriched and the HRMS↔complaint department taxonomies are aligned, supervisor-scoped dashboards on affected deployments are structurally empty — this silently invalidates dashboard rollouts.

## Probable fix path

1. Department enrichment at complaint creation (ServiceDefs → department mapping into the complaint / facts grain), plus backfill for existing complaints.
2. One department taxonomy shared by HRMS assignments and complaint service definitions per tenant — related to #901 (department master per tenant).
3. A bootstrap/onboarding validation that flags disjoint taxonomies before go-live.

Related: #1166, #901, #908.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

KDwevedi:
Consolidating into a personal gap backlog for sprint prioritisation (KDwevedi/Citizen-Complaint-Resolution-System#27); will be re-filed/scheduled individually once prioritised.

Issue #1111, opened 2026-07-09
[P2] Dashboard: group by complaint hierarchy levels (Category / Sub-Type / Leaf)


subhashini-egov:
## Summary

Add a "group by complaint hierarchy level" option to the Configurable Dashboard widgets so operators can pivot data at any level of the tenant's complaint tree — top-level Category, mid-level Sub-Type, or leaf — instead of only at the leaf.

Priority: **P2**.

## Context

Complaint types on newer deployments are structured as a **hierarchy** (schema `RAINMAKER-PGR.ComplaintHierarchy` + `RAINMAKER-PGR.ComplaintHierarchyDefinition` in MDMS). Nodes carry a `path` like `PGR.DelayInAttending/NoSkilledStaffAvail`, so a single complaint has multiple valid grouping levels:

| Level | Example path segment | Example bucket |
|---|---|---|
| 1 (Category) | `PGR` | "Public Grievance" |
| 2 (Sub-Type) | `PGR.DelayInAttending` | "Delay in attending" |
| 3 (Leaf) | `PGR.DelayInAttending/NoSkilledStaffAvail` | "No skilled staff available" |

## Current behaviour

Widgets that break down complaints by type (bar charts, tables, KPI tiles) group at the **leaf** level only. On a tenant with 200+ leaf nodes the resulting charts are cluttered and unreadable, and it's impossible to answer questions like "which category has the highest volume this week?" without ad-hoc SQL.

## Expected behaviour

1. Every complaint-type-grouping widget gains a **Group by level** control in its widget config panel — a dropdown enumerated from the deployment's `ComplaintHierarchyDefinition.levels` (so tenants with 2 levels see 2 options, tenants with 4 see 4).
2. The default level is **Category (Level 1)** for aggregated overview widgets; **Leaf** for detail-oriented tables — configurable per widget template.
3. When Level = 1 or 2, the widget rolls up all descendant leaf complaints into the parent bucket (sum of counts, weighted averages for SLA/resolution metrics).
4. Drilldown: clicking a bar/row at Level 1 opens the same widget filtered to that node and re-grouped at Level 2 (progressive disclosure).
5. Filter/legend labels come from the localised name in MDMS, not the raw `path` code.
6. Works on both hierarchy-enabled deployments (Bomet ke) and legacy flat deployments (Ethiopia) — on flat deployments the control shows only "Leaf" and hides the dropdown.

## Acceptance criteria

[...]

PR #1282, opened 2026-07-16
feat(analytics): group KPIs by complaint-hierarchy level — query-time rollup (#1111 PR1)

KDwevedi:
Implements #1111 (PR 1 of 2) — dashboard type-dimension KPIs can now be grouped by any complaint-hierarchy level at query time, via a new `hierLevel` KPI param. PR 2 adds the per-widget FE "Group by" control.

> **STACKED ON #1267** (`feat/dashboard-metrics-server`). This branch is cut from that PR's head, so its commits appear in this diff until #1267 merges — at which point this diff collapses to the 7 `#1111` commits. Review from `def2bb91b` onward.

## What changes

A KPI request (or a def's declared default) may now carry `params.hierLevel`:

- `"leaf"` / absent — today's per-subtype buckets, byte-identical queries (no-op).
- `"1"`..`"12"` — every `service_code` dimension is re-grouped by the Nth segment of the materialized `complaint_node_path`, **aliased back `AS service_code`** so `viz.dimensionKey`, sort, columns and FE labels (`COMPLAINT_HIERARCHY.<code>`) all keep working with zero FE change.

The rollup is **backend, query-time, in SQL** — `coalesce(nullif(split_part(complaint_node_path,'.',least(N,complaint_depth)),''), service_code)` — so pre-aggregation `limit`s truncate AFTER summing into level buckets (a level-1 view of a 200-leaf tenant is not a truncated leaf list) and every measure recomputes over raw rows: weighted averages and count-FILTER ratios, never an average of leaf averages (verified for all 6 type-details measures).

## Decisions (from the reviewed plan)

| # | Decision | Choice |
|---|---|---|
| 1 | Rollup location | Backend, query-time — FE rollup is wrong under pre-aggregation `limit` and cannot reconstruct weighted avgs |
| 2 | Mechanism | Derived expr over the existing `complaint_node_path` (no new MV column); N clamps to row depth; NULL path falls back to leaf |
| 3 | Catalog shape | New param `{name:"hierLevel", allowed:[...], default:...}` mirroring `window` — not forked dimension names, not viz-only |
| 4 | Wire protocol | Expression aliased `AS service_code`, grouped by ordinal — FE contract unchanged; `service_group` dim dropped at non-leaf (it collapses into a duplicate of the level bucket) |
| 5 | Level count | `params.allowed` seeded generously (`leaf,1..4`); the PR2 dropdown enumerates `ComplaintHierarchyDefinition.levels` at runtime — never hardcoded |
| 6 | MVP UI | PR1 = catalog defaults only (level-1 on the two overview charts fixes the leaf clutter day one); in-UI control is PR2 |
| 7 | Drilldown (issue AC 4) | **Excluded** — click-to-filter is a hierarchy-FILTER interaction; hierarchy-aware filters were built and deliberately abandoned by the project owner |

## Injection surface (R1)

There is **no generic expression support** in the planner. The level expr template is fixed Java in `AnalyticsCatalog`; `hierLevel` must strictly parse to an int in 1..12 before any SQL assembly. The composer hands the planner an object-dimension marker gated on a **per-JVM nonce** that request bodies and MDMS defs (parsed JSON) can never carry — a forged object dimension is rejected as `unknown_column` before SQL is built. The inline-query grammar and PII gate are untouched.

## R5 — data-hygiene migration (`V20260716000000`)

Legacy flat imports (`complaints.categories.*`) registered single hierarchy nodes whose **code itself contains dots**, so splitting their "path" fabricates a garbage `complaints` level-1 bucket. Live exposure at authoring time: **20 of 645 fact rows across 3 tenants** (13 further NULL-path rows already take the leaf fallback). The migration recreates both grains (MV-recreate idiom, supersedes `V20260708000000`) with one change: `complaint_node_path` is NULL when the matched node's own code contains `'.'` — those rows roll up as themselves via the existing leaf fallback instead of polluting level buckets. Validation criterion: no `complaints` bucket at level 1.

## Seeds (R6) + live-update runbook

Seed changes upstream-first so the repo is a superset of live `ke`: `cl_table_complaint_type_details` window gains `"all"` + `default:"all"`; `cl_chart_over_time_created_daily` gains `viz.valueKey:"created"`. Then `hierLevel` (allowed `["leaf","1","2","3","4"]`): defaults `"1"` on `cl_chart_complaints_by_type` / `cl_chart_open_by_type_stage`, `"leaf"` on `cl_table_complaint_type_details` / `cl_table_subtype_performance` / `cl_table_recurring_ward_subtype`. Row-grain KPIs (at-risk, map pins) excluded. Versions bumped; saved layouts are kpiId-keyed and unaffected.

**Ops runbook for live tenants — never bulk-replace `dss.KpiDefinition`.** Live records carry hand-enrichments in both directions. Per record, for each of the 5 KPIs above (+ the 2 enrichment targets below):

1. `_search` the record by `uniqueIdentifier` (schema `dss.KpiDefinition`, tenant `ke`).
2. **Merge into the fetched JSON** (read-merge-write): append the `hierLevel` param to `data.params`, bump `data.version`; leave every other field exactly as fetched.
3. `_update` the single record with the merged payload.
4. In the same pass, push the seed-only enrichments live: `viz.seriesLabelKey` (`DASHBOARD_SERIES_FILED`) on `cl_chart_complaints_by_type`, and the `channelMap[].labelKey`s on `cl_chart_open_by_channel`.
5. Refresh the MVs after the migration lands so the R5 path change takes effect, then spot-check: level-1 bucket sums == sum of level-2 buckets == sum of leaf buckets on `cl_chart_complaints_by_type`, and no `complaints` bucket at level 1.

## Caveat documented (R4): `ideal_sla_ms` at non-leaf

`avg(sla_target_ms)` inside a level bucket averages heterogeneous per-subtype SLA targets — a correctly weighted mean of the bucket's complaints' targets, but not "the category's SLA" (no such thing exists). Documented in `10-kpi-catalog.md`; the type-details table ships with default `"leaf"` (unchanged UX), and the column relabel/hide reacts in PR2 where the FE can respond to the active level.

## PR2 preview (FE, separate review surface)

Per-widget "Group by" select in the AdminDashboard widget header / `DataTableChrome.headerActions`, options enumerated from `ComplaintHierarchyDefinition` filtered by the deployment's `hierarchyType` (ke carries both `PGR` and `PGR_TEST` — naive fetch shows the wrong dropdown), localized level names, per-tile localStorage overrides merged into that tile's params before companion-ref spreads.

## Tests

`mvn test` (maven:3.9-eclipse-temurin-17): **107 tests, 0 failures, 2 skips** (pre-existing `SeedFixtureDriftTest` skips). New: 16 composer/planner tests (leaf no-op, level rewrite, `service_group` + sort drop, strict int-parse incl. injection strings, NULL-path fallback + depth clamp in generated SQL, SQL snapshot with ordinal GROUP BY, events grain, forged-marker rejection, out-of-range genuine marker) + 7 allow-list validation tests (hierLevel + window regression, open/undeclared params, null edges).

## Also in this PR: tenant-configurable department scoping — `dss.DashboardConfig.departmentScoping` (#1280)

Implements the team-call decision to make the analytics department-scope axis configurable per tenant, for deployments whose complaint data carries no departments (#1280): there, every department-scoped officer resolves a department that matches no fact row (or hits the fail-closed sentinel) and sees zero on every tile.

**What**: `dss.DashboardConfig` gains an OPTIONAL `departmentScoping` field:

| value | effect |
|---|---|
| `"enforced"` (default; field/record absent) | today's behavior — HRMS department resolution, `department_code IN (...)`, fail-closed sentinel for unresolvable constrained employees |
| `"disabled"` | `PrincipalScopeResolver` skips HRMS department resolution entirely for employees — no department filter, no sentinel. Tenant scoping and citizen self-scope untouched |

**Fail-safe default**: anything that is not an explicit `"disabled"` (case-insensitive, trimmed) — missing dss module, missing record, missing field, malformed value, MDMS error — resolves to `"enforced"`. The lookup (`KpiCatalogService.isDepartmentScopingDisabled`) never throws, so it can never flip the resolver's HRMS-error path either way.

**Mechanics**: read at the tenant's state root through the existing `fetchMaster` MDMS path in `KpiCatalogService`, cached in-memory for 5 minutes per state root (same idiom as the `/packs` recordCount cache, injectable clock for expiry tests) — a config flip applies within 5 minutes, no redeploy.

**Warning** (also in `docs/dashboard-configuration/20-packs-and-rbac.md` Layer 1): `"disabled"` widens data visibility — every employee on the tenant sees tenant-wide aggregates. Intended as a temporary bridge until department enrichment lands in the complaint data.

**Seeds**: the `DashboardConfig` seed file lives on #1258/#1272's branches, not this one, so no seed change here — the field composes additively with whichever of those merges first (`"departmentScoping": "enforced"` can be added to that record then; absence already means enforced).

**Tests**: enforced default (no record → dept scoping as today, incl. fail-closed sentinel), disabled skips the HRMS call, cache hit / expiry / flip pickup, malformed values → enforced, citizen self-scope untouched. Suite now **125 tests, 0 failures, 2 skips**.

## Summary by CodeRabbit

* **New Features**
  * Added complaint hierarchy drilldown via `hierLevel` (including non-leaf rollups) and enhanced KPI regrouping.
  * Added `complaintPath` subtree filtering (with `serviceCode` leaf support) and hierarchy-aware regrouping.
  * Improved analytics observability: optional `x-trace-id`, server-side query metrics, and per-request slow-query logging (top slow queries).
  * Extended analytics pack responses with `packId`, `persona`, and `recordCount`.
  * Introduced configurable department-scoping disable mode for Layer 1 row-scope behavior.
* **Bug Fixes**
  * Fixed hierarchy path handling for dotted hierarchy codes (preventing incorrect rollups).
  * Corrected daily “created” chart key mapping.
* **Documentation**
  * Updated KPI catalog docs, parameter validation rules, and server observability/dept-scoping guidance.

---

## Addendum: `complaintPath` subtree filter param (+ `paramsIgnored` envelope) — 33f738c88

Adds the small composer param the FE one-widget complaint-type tree filter needs (pairs with the upcoming FE tree-filter PR; validation verdict "BUILD-WITH-CHANGES", required changes #2 and #4).

**Subtree filter.** `params.complaintPath` = an **interior** hierarchy node's dot-path (e.g. `SANITATION.SEWAGE`). The composer injects a delimiter-guarded `subtree` predicate on `complaint_node_path` — `(complaint_node_path = ? OR complaint_node_path LIKE ? || '.%')` — on the grains that carry the path column (facts/events), riding the same prefix-filterable allowlist, bound params and LIKE-escaping as the inline path's `starts_with`. The `'.'` guard means `PGR` can never match a `PGRX` sibling; the `=` arm keeps complaints filed AT a mixed interior+serviceable node in its own subtree. Values are sanitized against the path alphabet (`[A-Za-z0-9._/-]`, ≤ 256 chars) — anything else is a hard `invalid_param`; empty/absent is a no-op. Composes with `hierLevel` (subtree filter = WHERE, level rollup = GROUP BY — orthogonal, covered by a combined test) and never widens ABAC (row-scope is still injected on top; asserted in tests).

**Daily-grain honesty (`paramsIgnored`).** The daily grain has no `complaint_node_path`, so the filter cannot apply to the 4 daily-grain widgets. Instead of `ward`-style silent skipping, the result envelope now reports it: `"paramsIgnored": ["complaintPath"]` on the per-query result (batch entries, the single arm, and compose refs) whenever a supplied param couldn't be applied and the caller must know. The FE shows a "filter not applied" indicator from this field rather than presenting an unfiltered number as filtered.

**Leaf vs interior split.** `serviceCode` (exact `service_code = ?`) remains the param for **leaf** selections — it works on every grain including daily. `complaintPath` is for interior nodes only; both are documented in `ANALYTICS-QUERY-API.md` §params and `10-kpi-catalog.md` §4.

**Seeds.** `complaintPath` declared (free-form validated string — no default, no `allowed` list) on the same 5 type-dimension KPIs that carry `hierLevel`; versions 1.1.0 → 1.2.0.

**Tests.** 16 new (`KpiQueryComposerComplaintPathTest`): subtree SQL snapshot, LIKE-escape of `_`, sanitizer rejections + live path shapes, daily-grain `paramsIgnored` + collector dedupe + ward-stays-silent, `hierLevel`+`complaintPath` combined, ABAC-still-applied, planner `subtree` op gating. Full pgr-services suite green: 141 tests, 0 failures (maven:3.9-eclipse-temurin-17).

---

## Addendum: two fixes found in integration testing (bomet deploy) — bffe6e039, 885324955

Both surfaced while deploying this PR to the bomet integration environment:

1. **Seed schema violation (blocker) — bffe6e039.** The bare `{"name":"complaintPath"}` params entries violated the live/registered `dss.KpiDefinition` schema, which requires `default` + `allowed` on every params entry — mdms `_update` 400'd (`required key [allowed] not found`) and fresh `tenant_bootstrap` seeding failed identically. All 5 entries are now `{"name":"complaintPath","default":"","allowed":[]}`; verified live that the server treats an empty `allowed` list as unvalidated free-form and an empty-string `default` as a no-op, and new unit tests in `AnalyticsServiceParamValidationTest` pin both semantics. (Supersedes the "no default, no `allowed` list" wording in the previous addendum's Seeds paragraph.)

2. **Migration heuristic widening — 885324955.** Bomet data showed a pollution shape the R5 migration (`V20260716000000`: NULL `complaint_node_path` when the node's own code contains `.`) misses: clean-code nodes parented under dotted legacy codes (e.g. node `DamagedRoad`, stored path `complaints.categories.DamagedRoad.DamagedRoad`, depth 4) — 30 live rows survived as a garbage `complaints` level-1 bucket. Since `V20260716000000` is already applied on bomet, a NEW migration `V20260717000000__hier_path_null_on_dotted_parent_codes.sql` (same MV-recreate idiom, otherwise byte-identical) widens the cnp predicate in both grains: also NULL when the matched node's `parentCode` contains `.`. Migration-rationale note updated in `10-kpi-catalog.md`.

Full pgr-services suite green after both: 144 tests, 0 failures, 0 errors (maven:3.9-eclipse-temurin-17).

### DIFF
--- a/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsMetrics.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsMetrics.java
@@ -75,7 +75,7 @@ public class AnalyticsMetrics {
             queryDuration.record(tookMs, attrs);
             queryRows.add(rowCount, attrs);
         } catch (RuntimeException e) {
-            log.debug("analytics metrics record failed (ignored): {}", e.toString());
+            log.debug("analytics metrics record failed (ignored)", e);
         }
     }
 
--- a/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsMetrics.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsMetrics.java
@@ -46,14 +46,22 @@ public class AnalyticsMetrics {
     private final LongCounter queryRows;
 
     public AnalyticsMetrics() {
-        Meter meter = GlobalOpenTelemetry.getMeter(METER_NAME);
+        this(GlobalOpenTelemetry.getMeter(METER_NAME));
+    }
+
+    /** Visible for tests: lets a test pass an SDK meter to assert instrument shape. */
+    AnalyticsMetrics(Meter meter) {
+        // NOTE: no unit on purpose. The metric NAMES already carry their unit (.ms),
+        // and the collector's prometheus exporter appends a unit-derived suffix when
+        // unit is set — validated live on bomet: unit "ms" surfaced as
+        // pgr_analytics_query_duration_ms_milliseconds_*. Omitting the unit keeps the
+        // scraped names exactly pgr_analytics_query_duration_ms_* /
+        // pgr_analytics_query_rows_total (same fix as the client's dashboardMetrics.js, #1268).
         this.queryDuration = meter.histogramBuilder("pgr.analytics.query.duration.ms")
                 .setDescription("Duration of one executed analytics SQL query (per batch entry / compose source)")
-                .setUnit("ms")
                 .build();
         this.queryRows = meter.counterBuilder("pgr.analytics.query.rows")
                 .setDescription("Rows returned by executed analytics SQL queries")
-                .setUnit("{row}")
                 .build();
     }
 
--- a/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsMetrics.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsMetrics.java
@@ -29,8 +29,10 @@ import org.springframework.stereotype.Component;
  *   <li>{@code pgr.analytics.query.rows} — counter of rows returned.</li>
  * </ul>
  * Attributes: {@code kpi_id} (the KPI id, {@code inline} for inline-grammar queries),
- * {@code grain}, {@code tenant}. Cardinality is bounded: kpi_id by the MDMS catalog
- * (tens), grain by the 3 grains, tenant by the tenant tree.
+ * {@code grain}, {@code tenant} (the STATE-ROOT tenant, never the raw request
+ * tenantId — see {@link QueryTelemetry#stateRootOf}). Cardinality is bounded: kpi_id
+ * by the MDMS catalog (tens), grain by the 3 grains, tenant by the deployment's
+ * state roots.
  */
 @Component
 @Slf4j
@@ -72,7 +74,9 @@ public class AnalyticsMetrics {
      * @param kpiId  the KPI id the query was resolved from, or {@code "inline"} for
      *               inline-grammar queries
      * @param grain  the grain the planner resolved ({@code facts|events|daily})
-     * @param tenant the request tenantId (scope tenant, not per-row)
+     * @param tenant the STATE-ROOT tenant (callers must pre-normalize via
+     *               {@link QueryTelemetry#stateRootOf} — the raw request tenantId is
+     *               attacker-controlled and would blow up label cardinality)
      */
     public void recordQuery(String kpiId, String grain, String tenant, long tookMs, long rowCount) {
         try {
--- a/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsMetrics.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsMetrics.java
@@ -1,0 +1,85 @@
+package org.egov.pgr.analytics;
+
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.metrics.DoubleHistogram;
+import io.opentelemetry.api.metrics.LongCounter;
+import io.opentelemetry.api.metrics.Meter;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Component;
+
+/**
+ * OTEL metrics for the dynamic analytics endpoints (#1110 — dashboard render-lag
+ * instrumentation, server half).
+ *
+ * <p>Thin wrapper over {@link GlobalOpenTelemetry}: when the service runs under the
+ * OpenTelemetry javaagent (the deploy.sh stack attaches it and sets
+ * {@code OTEL_METRICS_EXPORTER=otlp}), the agent bridges this to its SDK and the
+ * instruments export to the collector → Prometheus. Without the agent (unit tests, CI,
+ * bare {@code java -jar}) {@code GlobalOpenTelemetry} is a no-op and every call here is
+ * free and side-effect-less — no config or conditional wiring needed.
+ *
+ * <p>Instruments (OTLP names; Prometheus surfaces them as
+ * {@code pgr_analytics_query_duration_ms_bucket/_sum/_count} and
+ * {@code pgr_analytics_query_rows_total}):
+ * <ul>
+ *   <li>{@code pgr.analytics.query.duration.ms} — histogram, one point per executed SQL
+ *       query (each batch entry, and each SOURCE query of a backend-composed KPI).</li>
+ *   <li>{@code pgr.analytics.query.rows} — counter of rows returned.</li>
+ * </ul>
+ * Attributes: {@code kpi_id} (the KPI id, {@code inline} for inline-grammar queries),
+ * {@code grain}, {@code tenant}. Cardinality is bounded: kpi_id by the MDMS catalog
+ * (tens), grain by the 3 grains, tenant by the tenant tree.
+ */
+@Component
+@Slf4j
+public class AnalyticsMetrics {
+
+    static final String METER_NAME = "pgr-analytics";
+
+    static final AttributeKey<String> ATTR_KPI_ID = AttributeKey.stringKey("kpi_id");
+    static final AttributeKey<String> ATTR_GRAIN  = AttributeKey.stringKey("grain");
+    static final AttributeKey<String> ATTR_TENANT = AttributeKey.stringKey("tenant");
+
+    private final DoubleHistogram queryDuration;
+    private final LongCounter queryRows;
+
+    public AnalyticsMetrics() {
+        Meter meter = GlobalOpenTelemetry.getMeter(METER_NAME);
+        this.queryDuration = meter.histogramBuilder("pgr.analytics.query.duration.ms")
+                .setDescription("Duration of one executed analytics SQL query (per batch entry / compose source)")
+                .setUnit("ms")
+                .build();
+        this.queryRows = meter.counterBuilder("pgr.analytics.query.rows")
+                .setDescription("Rows returned by executed analytics SQL queries")
+                .setUnit("{row}")
+                .build();
+    }
+
+    /**
+     * Record one executed analytics query. Never throws — telemetry must not be able to
+     * break the query path.
+     *
+     * @param kpiId  the KPI id the query was resolved from, or {@code "inline"} for
+     *               inline-grammar queries
+     * @param grain  the grain the planner resolved ({@code facts|events|daily})
+     * @param tenant the request tenantId (scope tenant, not per-row)
+     */
+    public void recordQuery(String kpiId, String grain, String tenant, long tookMs, long rowCount) {
+        try {
+            Attributes attrs = Attributes.of(
+                    ATTR_KPI_ID, orDash(kpiId),
+                    ATTR_GRAIN, orDash(grain),
+                    ATTR_TENANT, orDash(tenant));
+            queryDuration.record(tookMs, attrs);
+            queryRows.add(rowCount, attrs);
+        } catch (RuntimeException e) {
+            log.debug("analytics metrics record failed (ignored): {}", e.toString());
+        }
+    }
+
+    private static String orDash(String v) {
+        return (v == null || v.isEmpty()) ? "-" : v;
+    }
+}
--- a/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsService.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsService.java
@@ -236,8 +236,11 @@ public class AnalyticsService {
      * > the def's baked query (a defaulted param flows through {@link KpiQueryComposer#mergeParams}
      * exactly like a caller-sent one, and the C1 window allow-list check runs on the EFFECTIVE
      * params). Returns {@code reqParams} untouched (possibly null) when no default applies.
+     * An EMPTY-STRING default is a no-op (skipped like an absent one) — the dss.KpiDefinition
+     * schema requires {@code default} on every params entry, so free-form params (complaintPath)
+     * are seeded with {@code "default":""}. Package-private for tests.
      */
-    private JsonNode withDeclaredDefaults(KpiDefinition def, JsonNode reqParams) {
+    JsonNode withDeclaredDefaults(KpiDefinition def, JsonNode reqParams) {
         List<KpiDefinition.KpiParam> declared = def.getParams();
         if (declared == null || declared.isEmpty()) return reqParams;
         com.fasterxml.jackson.databind.node.ObjectNode merged = null;
@@ -261,7 +264,9 @@ public class AnalyticsService {
      * def's declared params carrying a non-empty {@code allowed} list, an effective (caller-sent or
      * defaulted) value must be in that list. Out-of-list (incl. arbitrary {@code last_Nd} windows,
      * out-of-range {@code hierLevel}s) → {@code invalid_param}. No-op for params the def declares
-     * without an allow-list (open values) and for undeclared params (composer vocabulary applies).
+     * without an allow-list — absent OR empty {@code "allowed":[]} both mean unvalidated free-form
+     * (the dss.KpiDefinition schema requires the key, so complaintPath is seeded with an empty
+     * list) — and for undeclared params (composer vocabulary applies).
      * Runs on BOTH the normal kpiId path and the compose path (each calls this on its own def).
      */
     void validateAllowedParams(KpiDefinition def, JsonNode reqParams) {
--- a/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsService.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsService.java
@@ -268,15 +268,24 @@ public class AnalyticsService {
      * (the dss.KpiDefinition schema requires the key, so complaintPath is seeded with an empty
      * list) — and for undeclared params (composer vocabulary applies).
      * Runs on BOTH the normal kpiId path and the compose path (each calls this on its own def).
+     * Malformed shapes are rejected outright: a non-object {@code params} node, or an object/array
+     * value for a declared allow-listed param, is {@code invalid_param} (never a silent bypass).
      */
     void validateAllowedParams(KpiDefinition def, JsonNode reqParams) {
-        if (reqParams == null || def.getParams() == null) return;
+        if (reqParams == null || reqParams.isNull()) return;
+        if (!reqParams.isObject())
+            throw new IllegalArgumentException("invalid_param: params must be an object");
+        if (def.getParams() == null) return;
         for (KpiDefinition.KpiParam p : def.getParams()) {
             if (p == null || p.getName() == null) continue;
             List<String> allowed = p.getAllowed();
             if (allowed == null || allowed.isEmpty()) continue;
             if (!reqParams.hasNonNull(p.getName())) continue;
-            String requested = reqParams.get(p.getName()).asText();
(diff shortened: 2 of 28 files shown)


subhashini-egov (approved):

subhashini-egov:
@KDwevedi Is this added to the ansible and docker compose?

KDwevedi:
@subhashini-egov Yes on both counts, with no compose change needed: the catalog seed ships in `ansible/nairobi-mdms/mdms/dss/KpiDefinition.json`, and the MV migration runs via Flyway inside the pgr-services image on startup — existing compose/ansible pick both up as-is. (Already-running deployments additionally apply the seed changes to their live MDMS records via the read-merge-write runbook in the PR body, since MDMS bootstrap is create-if-absent.)

backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+        try {
+            Attributes attrs = Attributes.of(
+                    ATTR_KPI_ID, orDash(kpiId),
+                    ATTR_GRAIN, orDash(grain),
+                    ATTR_TENANT, orDash(tenant));
+            queryDuration.record(tookMs, attrs);
+            queryRows.add(rowCount, attrs);
+        } catch (RuntimeException e) {
+            log.debug("analytics metrics record failed (ignored): {}", e.toString());
+        }
```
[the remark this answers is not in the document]
KDwevedi: Fixed — both sites, on the base PR #1267 where this code originates (this branch is now rebased onto it): the `AnalyticsMetrics.recordQuery` catch was already fixed in 70bd0d6ab (exception object passed to `log.debug`), and the `QueryTelemetry.resolveTraceId` silent catch is fixed in 01b1ac2c0 — the class is now `@Slf4j` and the ignored exception is logged at TRACE while keeping the deliberate fall-through-to-header semantics.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsService.java
```diff
@@ -218,25 +246,26 @@ private JsonNode withDeclaredDefaults(KpiDefinition def, JsonNode reqParams) {
...
-                            "invalid_param: window '" + requested + "' is not allowed for KPI '" + def.getId()
-                                    + "'; allowed=" + allowed);
-                return;
-            }
+            if (p == null || p.getName() == null) continue;
+            List<String> allowed = p.getAllowed();
+            if (allowed == null || allowed.isEmpty()) continue;
+            if (!reqParams.hasNonNull(p.getName())) continue;
+            String requested = reqParams.get(p.getName()).asText();
+            if (requested.isEmpty()) continue;
```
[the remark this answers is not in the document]
KDwevedi: Fixed in 3a2e59bc8 — `validateAllowedParams` now rejects a non-object `params` node and object/array values for declared allow-listed params with `invalid_param`, per your suggested shape. Non-text scalars (e.g. `hierLevel: 2`) still match the allow-list by text form, and the empty-string continue is kept deliberately: it is the unset-default sentinel required by the seeded `{"default":"","allowed":[]}` shape. 13 tests in AnalyticsServiceParamValidationTest now pin all of this.

backend/pgr-services/src/test/java/org/egov/pgr/analytics/AnalyticsMetricsTest.java
```diff
@@ -0,0 +1,27 @@
...
+ * throw — when the OpenTelemetry javaagent is NOT attached (unit tests, CI, bare
+ * java -jar). GlobalOpenTelemetry returns its no-op implementation here, so these
+ * calls exercise exactly the agent-less production path.
+ */
+public class AnalyticsMetricsTest {
+
+    @Test
+    public void constructsAndRecordsWithoutAgent() {
+        AnalyticsMetrics metrics = new AnalyticsMetrics();
+        assertDoesNotThrow(() -> metrics.recordQuery("open-complaints", "facts", "ke", 42L, 7L));
```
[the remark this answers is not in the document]
KDwevedi: Respectfully declining this one. Agent-less, `GlobalOpenTelemetry` returns the OTel API's built-in no-op `MeterProvider` — asserting that a no-op "produces no telemetry" would re-test OpenTelemetry's own contract, not this component's; the component contract these tests pin is exactly "never throws without an agent" (no test in this module registers a global SDK, so the global is deterministically the no-op here). The direct observation you're asking for does exist on the arm where it's meaningful: this branch is now rebased onto #1267's 277fed2ba, whose `instrumentNamesCarryUnitAndUnitFieldStaysEmpty` drives `recordQuery` through an explicit `SdkMeterProvider` + `InMemoryMetricReader` and asserts the exact instruments/points emitted.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/KpiCatalogService.java
```diff
@@ -103,6 +113,49 @@ public Optional<KpiDefinition> getDef(String kpiId, String tenantId) {
...
+            if (disabled)
+                log.info("dss.DashboardConfig.departmentScoping=disabled at {} — employee department scoping OFF",
+                        stateRoot);
+            deptScopingCache.put(stateRoot, new Object[]{disabled, now + CONFIG_TTL_MS});
+            return disabled;
+        } catch (Exception e) {
+            // Fail-safe: any unexpected error means "enforced" (current behavior), never a throw
+            // that could flip the caller's HRMS-error path to deny-all.
+            log.warn("departmentScoping lookup failed for tenant {}; treating as enforced. Cause: {}",
+                    tenantId, e.toString());
```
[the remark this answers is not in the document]
KDwevedi: Fixed in 533aa5074 — the exception object is now passed to `log.warn` so the stack trace / root cause is preserved; the fail-safe treat-as-enforced behavior is unchanged.

backend/pgr-services/src/main/resources/db/migration/main/V20260717000000__hier_path_null_on_dotted_parent_codes.sql
```diff
@@ -0,0 +1,439 @@
...
+-- path still carries the legacy pollution — e.g. node 'DamagedRoad' with path
+-- 'complaints.categories.DamagedRoad.DamagedRoad' (depth 4). Splitting those
+-- paths on '.' survives R5 and still fabricates a garbage 'complaints' level-1
+-- bucket (30 live fact rows on bomet at authoring time).
+--
+-- Fix: the cnp CTE now emits NULL complaint_node_path when the matched node's
+-- own code contains '.' OR its parentCode contains '.'; those rows take the
+-- SAME leaf fallback the query-time level expression already has for
+-- flat/legacy tenants (coalesce(nullif(split_part(...)),''), service_code) —
+-- i.e. they roll up as themselves instead of polluting level buckets.
```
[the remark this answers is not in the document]
KDwevedi: Fixed in 4c64e2454 — valid finding: `chwalk`'s `built_path` (and a master-provided `path`) carries the whole ancestor chain, so a dotted grandparent (`legacy.dotted → CleanParent → CleanLeaf`) still polluted clean descendants. Both `cnp` CTEs now NULL `complaint_node_path` when a dotted code or parentCode appears ANYWHERE in the node's `chwalk` ancestry; the node itself is the depth-1 walk row, so the new predicate strictly contains the old node+parent check, and flagged rows keep the existing leaf `service_code` fallback. Implementation note: I used an `EXISTS` over the walk rows instead of the suggested window-function-in-LATERAL — same semantics, keeps the `built_path` lateral byte-identical. The migration is new on this (unmerged) PR, so it was amended in place rather than superseded by another migration.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsService.java
```diff
@@ -464,6 +542,48 @@ private Long asOf(){
         catch (Exception e) { return System.currentTimeMillis(); }
     }
 
+    // ---- #1110: tenant record-count for /packs (record_count_tier tag source) ----
+
+    private static final long RECORD_COUNT_TTL_MS = 5 * 60_000L;
```
vinothrallapalli-eGov: @KDwevedi Make cache time configurable, in future it should be changeable as per requirement.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/KpiCatalogService.java
```diff
@@ -35,6 +35,16 @@ public class KpiCatalogService {
     private static final String DSS_MODULE = "dss";
     private static final String MASTER_KPI  = "KpiDefinition";
     private static final String MASTER_PACK = "DashboardPack";
+    private static final String MASTER_CONFIG = "DashboardConfig";
+
+    /** The only departmentScoping value that turns employee department scoping OFF (#1280). */
+    private static final String DEPT_SCOPING_DISABLED = "disabled";
+    private static final long CONFIG_TTL_MS = 5 * 60_000L;
```
vinothrallapalli-eGov: @KDwevedi Make configurable and make sure all places refer from same config instead of hard coding in many file

backend/pgr-services/src/main/resources/db/migration/main/V20260717000000__hier_path_null_on_dotted_parent_codes.sql
```diff
@@ -0,0 +1,454 @@
...
+--
+-- NOTE: this supersedes the events/facts definitions in V20260716000000 (which
+-- superseded V20260708000000, V20260629000000, V20260608000000 in turn).
+-- Recreate BOTH MVs in any future shape change — the originals cannot be
+-- edited without breaking deployed checksums.
+-- ============================================================================
+
+
+-- ---- grain 1: complaint_events ----------------------------------------------
+DROP MATERIALIZED VIEW IF EXISTS complaint_events CASCADE;
```
vinothrallapalli-eGov, line 36: @KDwevedi In the other sql file, you are doing DROP MATERIALIZED VIEW IF EXISTS complaint_events CASCADE; and create. Again in this file also dropping and creating, why two files required instead of keep both changes in one file

vinothrallapalli-eGov (changes requested):

backend/pgr-services/src/main/resources/db/migration/main/V20260717000000__hier_path_null_on_dotted_parent_codes.sql
KDwevedi: Good question — they can't be squashed because V20260716000000 is already APPLIED on the reference deployment: bomet ran it via the migration container during integration testing on 2026-07-17, so its row exists in Flyway's schema history. Flyway treats applied migrations as immutable — editing (or folding it into) that file changes its checksum and every subsequent deploy on any box that ran it fails validation. V20260717000000 therefore supersedes it the only safe way: a new version that re-creates the MVs, which is idempotent by design in this DROP+CREATE MV-recreate idiom. Squashing would only be possible with a documented flyway repair step on every already-migrated box — not worth the operational risk for two small sequential files.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsService.java
KDwevedi: Fixed in 78236b098 — TTL is now `pgr.analytics.config-cache-ttl-ms` (application.properties + PGRConfiguration, default 300000 ms). Tests cover a non-default TTL.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/KpiCatalogService.java
KDwevedi: Fixed in 78236b098 — both caches (recordCount in AnalyticsService and departmentScoping here) now read the single `pgr.analytics.config-cache-ttl-ms` property from PGRConfiguration; the per-file constants are gone.

Hari-egov:
@KDwevedi thanks for the quick turnaround on the TTL config (`78236b09`) and the Flyway immutability rationale on the two-migrations question. Please resolve the addressed threads and re-request review from @vinothrallapalli-eGov so this can be closed out.

KDwevedi:
@Hari-egov done — resolved the three addressed threads (TTL config 78236b0, single-config refactor, Flyway two-file rationale) and re-requested review from @vinothrallapalli-eGov.

vinothrallapalli-eGov (approved):

PR #1294, opened 2026-07-17
fix(ansible): post-bootstrap survives a pre-seeded (live-dump) database

Hari-egov:
## What
The documented 0→1 clone flow swaps `local-setup/db/full-dump.sql` for a live environment's dump. On such a deploy, ADMIN already exists under the state enc key, so the MCP `user_only` re-provision errors on the existing row — and the old `failed_when` asserts killed an otherwise healthy deploy at the very last stage.

## Changes (playbook-deploy.yml only)
- The two **post-bootstrap ADMIN re-provision** tasks now only require the provisioning shim to be reachable (HTTP 200), instead of asserting `admin_user_provisioned`.
- The authoritative gate is a **real ADMIN login**: the existing configurator-i18n login already covers the state root, and a new probe task now verifies ADMIN authenticates at the **city tenant** (retries 6×10s, `no_log`), which previously had no login verification at all.

## Verified
Fresh local 0→1 with the pilot dump pre-seeded: recap `ok=113 failed=0`, ADMIN+SUPERUSER and pilot employees log in, 9.6k pt_PT localization keys, 84-node IGE hierarchy, P-2026-* complaints present.

### DIFF
--- a/local-setup/ansible/playbook-deploy.yml
+++ b/local-setup/ansible/playbook-deploy.yml
@@ -2354,9 +2354,12 @@
         status_code: [200]
         timeout: 60
       register: mcp_rekey_root
-      failed_when: >-
-        mcp_rekey_root.status != 200 or
-        not (mcp_rekey_root.json.admin_user_provisioned | default(false))
+      # A DB pre-seeded from a LIVE environment dump (instead of the stock pg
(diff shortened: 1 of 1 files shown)


Hari-egov:
Closing — not required.

PR #1297, opened 2026-07-17
fix(gatus): review health dashboard — add missing services, unmask Postgres, gate optional stacks

singhalkarun:
## Why

The Gatus dashboard covered 14 of ~30 services, and two defects meant it reported health it wasn't measuring.

**"PostgreSQL" was monitoring PgBouncer.** The check pointed at `tcp://postgres:5432`, but `postgres` is a network alias for PgBouncer in Compose and a Service selecting `app: pgbouncer` in k8s. The database itself (`postgres-db`) was never probed, so a dead database behind a live pooler would have shown green.

**The k3s tier could never start.** `k8s/tools/gatus.yaml` mounted a ConfigMap named `gatus-config` that was not defined anywhere in the repo. The pod would sit waiting for a volume that never materialises.

**Optional stacks were invisible or permanently red.** Five stacks are profile-gated in `docker-compose.egov-digit.yaml` — `otp`, `search`, `keycloak`, `mcp`, `notifications` — and none were monitored. That cuts both ways: `mcp` is active on a normal deploy and was unmonitored, while a check for an inactive profile would sit red forever.

## What changed

- **Postgres and PgBouncer are now checked separately**, so neither can mask the other.
- **The k3s `gatus-config` ConfigMap is authored**, in `tools/gatus.yaml` beside the Deployment — matching how the rest of the k8s config in this tree is kept.
- **Coverage goes from 14 to 49 endpoints**: MinIO, mdms-backend, localization, filestore, hrms, bndry-mgmnt, url-shortening, both nginx proxies, a localization API test, and the five optional stacks.
- **Optional endpoints are gated** on `enabled: ${GATUS_PROFILE_*}`, so a stack is monitored exactly when it's deployed.
- **Both tiers define all 49 endpoints identically**, zero tier-only entries, so "deployed in both or neither" holds by construction. k3s pins every toggle off, as it has no manifests for any optional stack yet.

## Notes for reviewers

**Gatus defaults `enabled` to true when the variable is unset** — it fails *open*. Every compose file therefore passes each toggle explicitly; the `${VAR:-false}` defaults are what make a forgotten toggle mean "off" rather than "red". They aren't redundant.

**Profiles exist only in `egov-digit.yaml`.** In the other four compose files these services are ungated and always start, so those files pin the toggles to true/false to match what they actually deploy. That's why the env blocks differ per file.

**`GATUS_SMS_PORT` exists because `egov-notification-sms` listens on 8005 under `egov-digit.yaml` but 8080 in the other four**, and all five share one config. I ran both: `SERVER_PORT` is honoured either way and the context path is `/notification-sms` regardless, so this is cosmetic drift rather than a defect, and normalising it is left as a separate call.

**Endpoint URLs were taken from each service's own container healthcheck, then probed against a running stack** rather than inferred. Where a healthcheck was disabled I ran the image to find the real endpoint — that turned up actuators on `egov-notification-sms` and `user-otp` that a first pass had written off as TCP-only. `egov-bndry-mgmnt` answers 415 on every path tried, so its TCP probe stands.

## Verification

Run against the live Compose stack:

- Toggles off → 29 checks, no optional-stack entries.
- `GATUS_PROFILE_MCP=true` → 2 added checks, both pass (mcp is genuinely deployed).
- `GATUS_PROFILE_OTP=true` with the trio deployed → all 3 pass, sms port at 8005.
- `GATUS_PROFILE_OTP=true` without them → all 3 fail, which is why the gate exists.
- Both tiers mechanically diffed: 49 endpoints each, zero drift, nothing tier-only.

## Bonus fix: egov-url-shortening has been down for days

Adding the dashboard immediately surfaced a real, live bug, so it's fixed here too.

`egov-url-shortening` reports `{"status":"DOWN"}` and its DataSource never opens:

```
RuntimeException: Driver io.opentelemetry.instrumentation.jdbc.OpenTelemetryDriver
  claims to not accept jdbcUrl, jdbc:postgresql://postgres-db:5432/egov
```

Both images this service is pinned to (`maven-jdk21-983e8b2` and `v2.9.2-4a60f20`) bake `spring.datasource.driver-class-name=OpenTelemetryDriver` into their `application.properties`. That driver's `acceptsURL` only matches `jdbc:otel:` URLs, so Hikari rejects a plain `jdbc:postgresql://` one outright and every DB-backed request fails. Sibling services take a plain URL happily because their images don't pin the driver — this is specific to url-shortening.

`egov-digit.yaml` and `registry.yml` both passed a plain URL; `docker-compose.yml` and `deploy.yaml` already use the `jdbc:otel:` form, which is why the fault only shows on the egov-digit deploy.

Adding the prefix is the whole fix. Verified on both images: plain URL reproduces `{"status":"DOWN"}` with the driver error; `jdbc:otel:` gives `{"status":"UP"}` and `POST /shortener` returns a working short URL, proving the pool really reads and writes.

`SPRING_FLYWAY_URL` deliberately keeps its plain URL — Flyway resolves its own driver from the URL and must reach `postgres-db` directly rather than the transaction-mode pooler.

A follow-up commit then closes the inconsistency the fix left behind: this was the 1 service of 19 whose app datasource went straight to `postgres-db`, bypassing the pooler that keeps Postgres' 200 connection slots from being exhausted by 18 services' Hikari pools. It now uses the `postgres` (pgbouncer) alias like every peer, and `depends_on` names both — pgbouncer for the datasource, postgres-db for Flyway.

That move has a real hazard worth knowing about: pgbouncer 1.25.1 here runs with `max_prepared_statements` unset (so disabled), while pgjdbc promotes a statement to a server-side prepared statement after `prepareThreshold=5` executions — which under transaction pooling can land on a backend that never prepared it. Tested rather than assumed: 25 sequential then 60 concurrent writes through the pooler gave all 200s, 85 rows persisted, zero prepared-statement or connection-closed errors, and Postgres reported no direct backends from the app.

## CI guard

`.github/workflows/gatus-coverage.yml` now fails a PR when a deployed service has no Gatus check, when a check points at a host no tier deploys, or when the two tiers' catalogues drift. This gap is silent by construction — a missing check looks exactly like a healthy one — so the guard has to be mechanical. Adding a service now forces a decision: monitor it, or list it in `EXEMPT` with a reason.

Matching is by URL host with aliases resolved, because names and hosts aren't the same thing: `tcp://postgres:5432` reaches **pgbouncer** via a compose alias, and k8s `kafka`/`redpanda` share a selector. Comparing names to URLs naively reports both as unmonitored.

The guard immediately found **audit-service** — always-on in `egov-digit.yaml`, real actuator endpoint, monitored by nothing. Now checked, gated on `GATUS_AUDIT` since it exists in neither `docker-compose.yml` nor the k3s tier.

Observability plumbing (grafana/prometheus/loki/tempo/promtail/otel-collector) and openbao are `EXEMPT` with reasons rather than monitored: their failure costs visibility, not serving. promtail exposes no port and openbao binds `127.0.0.1`, so neither is probeable anyway, and nothing declares `depends_on` openbao.

The self-test asserts every detector fires on its own failure mode — a green check that cannot fail is worthless.

## Review fixes

**The profile toggles were dead on the deploy path that matters.** ansible renders `.env` from `digit.env.j2`, which emitted no GATUS key at all, so every `${GATUS_PROFILE_*:-false}` fell through to `false` on every provisioned tenant — a tenant with `enable_mcp: true` watched nothing. `digit.env.j2` now derives all five from the same `enable_*` host_vars that feed `compose_profiles`, so the toggle and the profile share one source of truth. (Note ansible never deploys `docker-compose.yml`; its stack is `egov-digit.yaml` plus overlays.)

**Stale headers** claiming the OTP/proxy endpoints were absent from k3s are corrected — they were present and gated off, and following the comment literally would have produced duplicates.

## Out of scope (pre-existing, untouched)

- **The `IDGen API - Generate ID` probe burns a real complaint ID every 60s.** Verified: the sequence is at 8607 against 52 real complaints, so ~8,555 user-visible PGR IDs have been consumed. Pre-existing and unchanged by this PR (which only mirrors the existing catalogue into k3s). The clean fix — a throwaway `idName` — needs an MDMS `IdFormat` seeded per tenant, since idgen rejects unknown names despite `AUTOCREATE_NEW_SEQ=true`. Raising separately.

- `docker-compose.db-migrations.yml` does not parse: duplicate `image` key at line 1113. Confirmed identical on `develop`.


local-setup/docker-compose.egov-digit.yaml
```diff
@@ -1662,6 +1662,15 @@ services:
...
       GATUS_CONFIG_PATH: /config/config.yaml
+      # Profile-gated stacks: keep these beside COMPOSE_PROFILES in .env.
+      # Unset means off -- gatus enables an endpoint when its variable is missing.
+      GATUS_PROFILE_OTP: ${GATUS_PROFILE_OTP:-false}
+      GATUS_PROFILE_SEARCH: ${GATUS_PROFILE_SEARCH:-false}
+      GATUS_PROFILE_MCP: ${GATUS_PROFILE_MCP:-false}
+      GATUS_PROFILE_KEYCLOAK: ${GATUS_PROFILE_KEYCLOAK:-false}
+      GATUS_PROFILE_NOTIFICATIONS: ${GATUS_PROFILE_NOTIFICATIONS:-false}
+      GATUS_PROXIES: "true"
+      GATUS_SMS_PORT: "8005"
```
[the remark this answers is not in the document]
singhalkarun: Confirmed and fixed in a4300e96.

Verified the claim before acting: `grep -rn GATUS ansible/` returned zero matches, and the live `/opt/digit/.env` on our deploy box has 23 keys, none of them GATUS. So all five resolved to `false` on every ansible-provisioned tenant — and that box proves the impact, since it runs mcp (`digit-mcp`/`mcp-postgres` up and healthy) while `GATUS_PROFILE_MCP` was never set. The PR's own motivating example, unfixed. Good catch.

`digit.env.j2` now derives all five from the same `enable_*` host_vars that feed `compose_profiles`, so the toggle and the profile share one source of truth and cannot drift:

```
GATUS_PROFILE_OTP={{ enable_otp_services | default(false) | lower }}
GATUS_PROFILE_SEARCH={{ enable_search_stack | default(false) | lower }}
GATUS_PROFILE_MCP={{ enable_mcp | default(false) | lower }}
GATUS_PROFILE_KEYCLOAK={{ enable_keycloak | default(false) | lower }}
GATUS_PROFILE_NOTIFICATIONS={{ enable_novu | default(false) | lower }}
```

Rendered the template against three tenant fixtures to confirm: `enable_mcp: true` → `GATUS_PROFILE_MCP=true`; nothing set → all `false`.

`GATUS_PROXIES`/`GATUS_SMS_PORT`/`GATUS_AUDIT` are hardcoded per compose file rather than read from the environment, so they were never affected.

local-setup/k8s/tools/gatus.yaml
```diff
@@ -1,3 +1,454 @@
+# Gatus health dashboard for the k3s tier.
+#
+# Kept in sync with the compose tier: local-setup/gatus/config.yaml.
+# Shared services must have identical names, groups, URLs and conditions in both files.
+# The compose tier additionally monitors egov-otp, user-otp, egov-notification-sms,
+# egov-user-proxy and egov-workflow-proxy; those have no manifest under local-setup/k8s/
+# and are deliberately absent here. Add them once the manifests exist.
```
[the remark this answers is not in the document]
singhalkarun: Correct — fixed in a4300e96. The header predated the gated entries and I never updated it; the same commit that wrote "deliberately absent" had already added all five, switched off.

The header now states that both tiers define the same catalogue, that the optional stacks and proxies are **defined here and switched off** via the Deployment env, and that landing the manifests means flipping a toggle to `"true"` — explicitly *not* adding endpoint blocks, which was the landmine you identified.

Also added a CI guard (`.github/workflows/gatus-coverage.yml`) that fails on tier drift, so the two catalogues can no longer diverge unnoticed and a duplicate-name append would be caught mechanically rather than by comment-reading.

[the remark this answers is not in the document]
singhalkarun: Fixed in a4300e96 — both headers were stale. `k8s/tools/gatus.yaml` now says the checks remain defined but disabled, and `gatus/config.yaml` no longer claims any entry is compose-only (nothing is: both tiers carry all 50 endpoints, only runtime enablement differs).

Added a CI check that enforces this mechanically rather than by comment: `.github/scripts/check-gatus-coverage.py` fails on any endpoint present in one tier and not the other.

local-setup/k8s/tools/gatus.yaml
```diff
@@ -1,3 +1,454 @@
...
+          Content-Type: application/json
+        body: |
+          {
+            "RequestInfo": {"apiId": "digit", "ver": "1.0"},
+            "idRequests": [{"tenantId": "pg", "idName": "pgr.servicerequestid"}]
+          }
+        interval: 60s
+        conditions:
+          - "[STATUS] == 200"
+          - "[BODY].responseInfo.status == SUCCESSFUL"
```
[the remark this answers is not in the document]
singhalkarun: Verified the finding is real, but deferring it — not fixing it here.

It reproduces exactly as described. Two calls against a live stack:

```
call 1 -> PG-PGR-2026-07-17-008605
call 2 -> PG-PGR-2026-07-17-008606
```

And the damage is already measurable: `seq_eg_pgr_id.last_value` is **8607** against **52** rows in `eg_pgr_service_v2`. The probe has burned roughly **8,555 user-visible complaint IDs** on that box, so the next real complaint gets ~008608. Not theoretical.

Deferring for two reasons:

1. **Pre-existing.** This check is unchanged from the current `gatus/config.yaml` on develop; this PR only mirrors the existing catalogue into the k3s ConfigMap. Fixing it is a behaviour change to a live check that predates this work.
2. **The obvious fix is not config-only.** Pointing it at a throwaway `idName` is the right shape — it keeps the functional "can idgen actually mint an ID" coverage while isolating the burn — but idgen rejects unknown names despite `AUTOCREATE_NEW_SEQ=true`:
   `400 BAD_REQUEST: No Format is available in the MDMS for the given name and tenant`
   So it needs an `IdFormat` entry seeded in MDMS per tenant, which is a seed change with its own blast radius rather than a one-line edit here.

The alternative — dropping the probe and relying on the existing `ID Generation Service` health check — loses the functional coverage that makes this check worth having.

Raising it separately so it gets the MDMS change it needs. Flagging explicitly rather than closing silently, since the ID gaps are user-visible.

local-setup/gatus/config.yaml
```diff
@@ -1,5 +1,9 @@
 # Gatus Configuration for DIGIT Core Services
 # https://github.com/TwiN/gatus
+#
+# Kept in sync with the k3s tier: local-setup/k8s/tools/gatus.yaml (gatus-config ConfigMap).
+# Shared services must have identical names, groups, URLs and conditions in both files.
+# Entries marked "compose-only" have no k8s manifest and so exist here only.
```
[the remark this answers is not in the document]
singhalkarun: Fixed in a4300e96 — that header was stale for the same reason as the k3s one. It no longer claims anything is compose-only; nothing is. Both tiers define all 50 endpoints identically and only runtime enablement differs, which the header now says, along with a pointer to the CI check that enforces it.

.github/workflows/gatus-coverage.yml
```diff
@@ -0,0 +1,50 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  gatus-coverage:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
```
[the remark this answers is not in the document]
singhalkarun: Correct — fixed in fef4640f. Verified: `pull_request` watched 9 paths, `push` only 4, dropping `docker-compose.migrations.yml`, `docker-compose.migrations.ansible.yml`, `docker-compose.fast-path.yml`, the checker script, and the workflow itself. All three of those compose files are live inputs to `COMPOSE_FILES`, so the point about a direct push to develop skipping the guard is right — and worse, a push that broke a detector would not have re-run the `--self-test`.

Both lists are now identical (9 paths each, asserted).

One deliberate deviation from the suggested fix: I duplicated the list rather than sharing it via a YAML anchor. GitHub Actions does not expand anchors, so `paths: *gatus_paths` would silently drop the push filter — strictly worse than the original gap. The duplication carries a comment saying it must stay in sync and why an anchor is not used.

local-setup/k8s/tools/gatus.yaml
```diff
@@ -1,3 +1,470 @@
...
```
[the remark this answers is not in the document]
singhalkarun: Verified, and the k3s-specific point is a fair one — but deferring, consistent with the same finding raised by CodeRabbit on this PR (thread above).

It reproduces exactly as described:

```
call 1 -> PG-PGR-2026-07-17-008605
call 2 -> PG-PGR-2026-07-17-008606
```

and the burn is measurable: `seq_eg_pgr_id.last_value` is **8607** against **52** rows in `eg_pgr_service_v2` — roughly **8,555 user-visible complaint IDs** consumed by monitoring on that box. Confirmed `[BODY].responseInfo.status == SUCCESSFUL` does mean a real ID was minted, so this is not a side-effect-free branch.

You are right that this PR is what first brings the probe to k3s, rather than it being purely pre-existing. Weighed against that: the k3s tier has no Gatus running at all today (its ConfigMap was never defined — this PR authors it), so nothing regresses at merge; and the clean fix is not available here. A dedicated `idName` is the right shape — it keeps the functional "can idgen actually mint an ID" coverage while isolating the burn — but idgen rejects unknown names despite `AUTOCREATE_NEW_SEQ=true`:

```
400 BAD_REQUEST: No Format is available in the MDMS for the given name and tenant
```

so it needs an `IdFormat` seeded in MDMS per tenant — a seed change with its own blast radius, not a config edit in this PR. The cheap alternative (dropping the probe and leaning on the existing `ID Generation Service` health check) throws away the functional coverage that justifies the check.

Tracking separately so it gets the MDMS change it needs. Flagged rather than closed silently, since the ID gaps are user-visible on complaint numbers.

vinothrallapalli-eGov (approved):

local-setup/k8s/tools/gatus.yaml
```diff
@@ -1,3 +1,470 @@
...
+# The optional stacks (otp/search/keycloak/mcp/notifications) and the two nginx proxies
+# have no manifests under local-setup/k8s/ yet, so their checks are DEFINED here and
+# switched OFF via the GATUS_PROFILE_* / GATUS_PROXIES values in the Deployment env
+# below. When those manifests land, flip the matching toggle to "true" -- the endpoint
+# blocks already exist, so do not add new ones.
+# CI enforces tier parity (.github/workflows/gatus-coverage.yml).
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: gatus-config
```
KDwevedi, line 16: 🔴 **Blocker: this ConfigMap collides with the one `Tiltfile.k8s:23` already generates, killing the whole k8s tier at load.**

`Tiltfile.k8s` line 23 creates `gatus-config` dynamically (`kubectl create configmap gatus-config --from-file=config.yaml=gatus/config.yaml -n digit --dry-run=client`) and line 56 loads this file — so `tilt up -f Tiltfile.k8s` (the only documented deploy path; `setup-k8s.sh` ends by pointing at it) now sees the same kind/name/namespace twice. Reproduced under tilt v0.36.0:

```
Error in k8s_yaml: Duplicate YAML: ConfigMap gatus-config (Namespace: digit)
```

Tiltfile evaluation aborts **before deploying anything**, so every resource in the tier fails to load, not just gatus. Neither the new coverage guard (static YAML) nor the advisory tilt-test would catch it.

This also means the PR-body claim "the k3s tier could never start" doesn't hold: Tilt has always materialised this ConfigMap, so the base gatus pod started fine on the real path — the missing-ConfigMap failure only existed on a raw-`kubectl` path nothing in the repo uses.

Fix: pick one source of truth. Deleting `Tiltfile.k8s:23` is probably right, since the k8s tier now legitimately needs its own env pinning — but note what the old arrangement bought: the CM was generated from the compose config, so tier drift was impossible by construction. Worth a line in the header acknowledging the trade.

.github/scripts/check-gatus-coverage.py
```diff
@@ -0,0 +1,357 @@
...
+
+def compose_targets(paths):
+    """{reachable dns name -> service name} for every compose service.
+
+    Includes network aliases, because a Gatus URL names the DNS host, and the
+    host is often an alias rather than the service (postgres -> pgbouncer).
+    """
+    names, aliases = {}, {}
+    for p in paths:
+        if not p.exists():
```
KDwevedi, line 137: 🟠 **Missing source-of-truth file is silently skipped (false green).** Mutation-tested: renaming `docker-compose.yml` away entirely → `OK: 79 compose services … rc=0`. Same for `fast-path.yml`. A rename, a path typo in a future `COMPOSE_FILES` edit, or a directory restructure silently shrinks the guard's scope to nothing it complains about.

One-line fix: a missing `COMPOSE_FILES` entry should be a hard error, not a `continue`.

.github/scripts/check-gatus-coverage.py
```diff
@@ -0,0 +1,357 @@
...
+    # not break serving. Listens on 127.0.0.1 only, so Gatus could not reach it.
+    "openbao": "deploy-time secrets store: no runtime dependents, binds 127.0.0.1 only",
+    # Tooling, not part of the serving stack.
+    "jupyter": "dev tool, not a serving dependency",
+    "gatus": "the monitor itself",
+}
+
+# Suffixes that mark generated one-shot migration containers. These are created
+# per schema-owning service by the migrations overlay and all exit on success.
+EXEMPT_SUFFIXES = ("-migration", "-migrations", "-seed", "-init")
```
KDwevedi, line 108: 🟠 **The suffix exemption tests the name, not the behavior the docstring promises (false green).** Mutation-tested: a long-lived service named `session-init` with `restart: unless-stopped` and published ports passes clean (`OK: 84 compose services … rc=0`). Any future service ending in `-init`/`-seed`/`-migration`/`-migrations` is exempt regardless of whether it's a one-shot — the "forgotten by construction" gap this guard exists to close, reopened for a whole name class. (All 20 current suffix-hits are genuine one-shots, so no live impact today.)

Cheap hardening: apply the suffix exemption only when the service also has no `ports`/`expose`, or has `restart: "no"`.

.github/scripts/check-gatus-coverage.py
```diff
@@ -0,0 +1,357 @@
...
+            continue
+        doc = yaml.safe_load(_stub_vars(p.read_text())) or {}
+        for svc, body in (doc.get("services") or {}).items():
+            names[svc] = svc
+            nets = (body or {}).get("networks")
+            if isinstance(nets, dict):
+                for net in nets.values():
+                    for a in ((net or {}).get("aliases") or []):
+                        aliases[a] = svc
+    names.update(aliases)
```
KDwevedi, line 147: 🟠 **Alias collision silently hijacks coverage — and can re-hide the exact bug this PR fixes.** `names.update(aliases)` lets an alias clobber a real service's entry. Mutation-tested: adding an `evil-svc` with network alias `postgres-db` → still `OK: 83 compose services` (count unchanged — the new service appeared and the real database vanished as a required target in one silent move). The `tcp://postgres-db:5432` check now "covers" the impostor and the actual postgres is no longer required to be monitored: the pgbouncer-masking failure mode, reintroducible via one copy-pasted networks block with zero CI signal.

Fix: `raise` on an alias/service-name collision instead of clobbering.

.github/scripts/check-gatus-coverage.py
```diff
@@ -0,0 +1,357 @@
...
+#                               playbook-deploy.yml "Compute compose file stack",
+#                               compose_files: -f docker-compose.egov-digit.yaml ...).
+#
+# The overlays add no long-lived services of their own (they add *-migration one-shots
+# and override existing services), but they are listed so that a future long-lived
+# service added there cannot slip past this check.
+#
+# registry.yml / deploy.yaml / db-migrations.yml are variants of the above and would
+# only contribute duplicate names.
+COMPOSE_FILES = [
```
KDwevedi, line 60: 🟠 **Five compose files are neither scanned nor watched by the workflow — including three this PR itself adds toggle blocks to.** `deploy.yaml`, `registry.yml`, `db-migrations.yml` are excluded as "variants", and `bomet.yml`/`core.yml` aren't mentioned at all; `gatus-coverage.yml`'s path filters omit all five, so a PR touching only those files never even triggers the run. Mutation-tested: a fake service added only to `docker-compose.deploy.yaml` → rc=0 — and `deploy.yaml` is a real usage path (the digit-mcp docs deploy with it). Today the delta is benign (only `*-seed`/`db-seed` one-shots are unique to the unscanned files — I diffed all five), but the "cannot slip past this check" comment overstates the perimeter.

Either add the remaining files to `COMPOSE_FILES` + the workflow paths, or scope the comment to what the guard actually covers. Two independent reviewers converged on this one.

KDwevedi (commented):
Deep review (multi-agent, findings verified by running the code rather than reading it). One blocker and four majors below, all with cheap fixes. Everything else checked out remarkably well — all 50 endpoint host/port/paths verified against each target's own healthcheck in every compose file, tier parity confirmed zero-drift by mechanical diff, the fail-open `enabled` semantics confirmed empirically against gatus:latest (unset AND empty string both enable), `digit.env.j2` render-tested (undefined vars and boolean casing safe), and the jdbc:otel root cause confirmed by extracting `application.properties` from the pinned image. Two narrative nits worth fixing alongside: both tiers carry **50** endpoints, not 49; and the openbao exemption reason ("binds 127.0.0.1") is wrong — the in-container listener is `0.0.0.0:8200`, only the host publish is loopback.

local-setup/k8s/tools/gatus.yaml
singhalkarun: Confirmed and fixed in #1303. Rendered both objects: identical \`v1 ConfigMap name=gatus-config ns=digit\`. Deleted the Tiltfile generation and kept the manifest, since the k3s tier now needs its own toggle pinning — with the header note on what the generated-from-compose arrangement bought.

Your correction of the PR body was right, but both of us were wrong about the tier. `Tiltfile.k8s:20` has pointed at `configs/persister/` since `09a8ec2e` (2026-05-13) renamed it to `configs/egov-persister/` — so Tilt aborts at the *first* ConfigMap, before reaching gatus. It has not materialised `gatus-config`, or anything else, since May. Nothing in the repo ever `kubectl apply`s `k8s/tools/gatus.yaml` either (`Tiltfile.k8s:56` is its only consumer). Fixed both in #1303; the tier was broken for a third reason neither of us named.

.github/scripts/check-gatus-coverage.py
singhalkarun: Reproduced and fixed in #1303 — now a hard error. It was worse than a shrunken scope: with *every* compose source missing it reported `0 services, rc=0`, a guard certifying an empty perimeter.

.github/scripts/check-gatus-coverage.py
singhalkarun: Fixed in #1303. Exemption now requires the service to publish no `ports`/`expose`, defaulting to "must be monitored" when unknown — the name is a hint, the absent port is the evidence. Audited all 25 current suffix-hits: none publish a port, so the perimeter is unchanged today. One note — a `restart: "no"` test would have been wrong, as 3 of the 25 leave `restart` unset; the ports test is the safe one.

.github/scripts/check-gatus-coverage.py
singhalkarun: Fixed in #1303 — an alias that shadows a service name, or that two services claim, is now a hard error. Legitimate aliases (`postgres` → `pgbouncer`, repeated across overlay files) still resolve. This was the sharpest of the five: it silently reintroduces the exact pgbouncer-masking bug the PR set out to fix.

.github/scripts/check-gatus-coverage.py
singhalkarun: Fixed in #1303 — took the "add them" option. All 10 compose files are now scanned (83 → 88 services; everything unique to the five was a `*-seed` one-shot, as you found, so all exempt). The `paths:` filters are globs rather than an enumeration, since an enumeration that drifts fails silently — the workflow never triggers, so the guard cannot report what it was never run to see. The script also now fails if `COMPOSE_FILES` misses a file present on disk, so the perimeter is derived rather than remembered.

PR #690, opened 2026-05-31
theme(complaint-form): postal validator + LocalityPicker leaf-only — #478

KDwevedi:
## Summary

Two commits closing both halves of #478.

- **postal-code validator** — surfaces "Enter a valid 5-digit postal code" on the complaint create form for invalid `1234`, accepts `00100`.
- **LocalityPicker leaf-only** — `BoundaryDropdown` filters its option list to leaf boundary types only so an operator cannot file a complaint at County / Sub-County / Ward level (closes the actively-unresolved item from Gurjeet's 2026-05-19 retest).

## Validation video (live bomet 2026-05-31)

- **Bundled flow** — https://github.com/KDwevedi/Citizen-Complaint-Resolution-System/raw/screenshots-2026-05-31/videos/2026-05-31/configurator-bundle-bomet.mp4

The honest-drive section in `demo-configurator-bundle-bomet.spec.ts` reads the boundary picker option list and asserts no non-leaf admin labels (`County`, `Sub-County`, `Ward`, etc.) appear — flips red on a regression of the leaf-only filter.

## Test plan

- [ ] Complaint create with invalid pincode `1234` shows the help text.
- [ ] Postal `00100` accepted, complaint submits.
- [ ] Boundary picker on the complaint form lists only leaf boundary labels (no admin levels).

PR #1315, opened 2026-07-20
fix(pgr): postal-code pattern as single source of truth + dynamic localized messages

### LINKED ISSUE #722: [Bug]  Postal Code validation issue

vinothrallapalli-eGov:
## Summary

Fixes #722 (re-reported 2026-07-15 as still broken after prior fixes landed).

The employee form, legacy citizen `FormExplorer`, and citizen v2 flow each
re-implemented postal-code validation slightly differently. A redundant
`postalCodeLength` / `postalCodeErrorMessage` config could drift out of
sync with the actual `postalCodePattern`, silently reintroducing the
digit-count mismatch the ticket reports — e.g. an operator bumping
`postalCodeLength` alone had no effect, because nothing ever read that
field; only `postalCodePattern` drove validation.

- `core_postal_configs.postalCodePattern` is now the **only** postal-code
  knob — a regex already expresses length, starting-digit constraints,
  and alnum shapes in one place, so there's no separate field that can go
  stale relative to it. `postalCodeLength` / `postalCodeErrorMessage` are
  removed from `host_vars`, the ansible template, and every consumer.
- New shared `digit-ui-esbuild/products/pgr/src/utils/postalCode.js`
  centralizes pattern validation and derives the error message from the
  pattern's digit count, resolved through i18next
  (`CS_COMPLAINT_POSTALCODE_INVALID_ERROR_LEN` / `_GENERIC`) instead of a
  hand-maintained override string. `createComplaintForm.js`,
  `FormExplorer.js`, and `CreatePGRFlowV2.tsx` all use it now instead of
  three divergent copies of the same logic.
- Seeded the new localization keys in both in-repo locale packs
  (`default`, `en_IN`) so the length-aware message is actually served
  per-locale, not just an English code-level fallback.
- Fixed `configurator/src/admin/validation.ts`'s `CORE_POSTAL_CODE_CONFIGS`
  globalConfigs key typo (never set by the playbook, so DIGIT Studio's own
  postal validator silently always used its hardcoded 5-digit default),
  merged with an equivalent fix that had already landed upstream on
  `develop` independently (dual-key fallback for the legacy name).

## Upgrade / release notes

- **Localization key rollout ordering.** The employee create-complaint field now
  points at `CS_COMPLAINT_POSTALCODE_INVALID_ERROR_GENERIC` with no code-level
  fallback at that call site (FormComposerV2 calls `t(error)` directly). On an
  existing install the field shows the raw key until default-data-handler
  re-seeds the `rainmaker-pgr` localization module **and** the `digit-redis`
  `messages` hash is flushed. The dynamic submit-time messages
  (`..._ERROR_LEN` / `..._GENERIC` via `getPostalCodeErrorMessage()`) fall back
  to correct English strings in the meantime.
- **Silent message change for tenants that had set `postalCodeErrorMessage`.**
  The per-tenant message override is removed; the message is now always derived
  from the pattern. E.g. Maputo (`^[0-9]{4}$`) previously pointed at
  `CS_COMPLAINT_POSTALCODE_INVALID_ERROR` — a key seeded nowhere in this repo,
  so it rendered raw — and now gets "Please enter a valid 4-digit postal code".
  Stale `postalCodeLength` / `postalCodeErrorMessage` keys left in host_vars are
  rendered through `to_json` verbatim and simply ignored.
- **MDMS is now the PRIMARY postal knob — migration step for non-5-digit
  deployments.** A `fieldType: "postalCode"` row in
  `common-masters.FormValidations` outranks host_vars
  `core_postal_configs` everywhere (PGR flows + DIGIT Studio, both reading
  `window.__DIGIT_FORM_VALIDATIONS.postalCode`, a channel named after the
  master). DDH seeds a default **5-digit** row at tenant creation, so any
  deployment whose postal shape isn't 5 digits (e.g. a 4-digit Maputo-style
  tenant) must set its MDMS row (DIGIT Studio > FormValidations, live regex
  tester included) as part of adopting this — the host_vars pattern alone no
  longer wins on newly-seeded tenants. host_vars remains the fallback for
  tenants without the row (dump-booted stacks, pre-FormValidations tenants).
  The `common-masters.FormValidations` schema is also registered in
  `local-setup/db/full-dump.sql` now (verified: full dump loads clean into
  postgres 16 with ON_ERROR_STOP).
- **Forward-port.** This lands on `master`; `develop` is behind on
  `CreatePGRFlowV2.tsx` / `_example.yml` and doesn't carry this fix — it needs
  to ride the next `master` → `develop` sync so it isn't reverted.

## Test plan
- [x] `npx tsc --noEmit` clean in `configurator/`
- [x] `esbuild` syntax check clean on all 5 touched `digit-ui-esbuild` files
- [x] All 3 touched localization JSON files parse
- [x] `globalConfigs.js.j2` renders valid JS via Ansible's Jinja2 templating (checked with `postalCodePattern` set and unset)
- [x] Manual + Playwright verification on a running local-setup stack, including a deliberately divergent MDMS 4-digit row over a globalConfigs 6-digit pattern — employee and citizen v2 forms both followed the MDMS rule, showed "Please enter a valid 4-digit postal code" (length-interpolated, in real time on the employee form), with no native browser bubble; all 11 postal UI tests green against the live stack
- [x] `npm run test` in `configurator/` — 78/78 passing
- [x] `npm run build` (esbuild bundle) clean after the type:"number" → type:"text" flips
- [x] `npx tsc --noEmit` clean in `tests/integration-tests/` (new rendered-error spec)
- [x] Node sanity harness: legacy `CORE_POSTAL_CODE_CONFIGS` key now reachable behind an ansible-rendered empty `CORE_POSTAL_CONFIGS`; MDMS channel wins when present; `isPostalCodeNumeric()` true/false for digit-only/alnum patterns

## Summary by CodeRabbit

- **Bug Fixes**
  - Postal-code validation and localized error text are now consistent across citizen and complaint creation flows, derived from the configured `postalCodePattern` with safe fallback for malformed patterns.
  - When the pattern implies a digit-only length, the UI shows the expected `{N}`-digit message; otherwise it shows a generic invalid-postal-code message.
  - Citizen postal-code inputs now preserve alphanumeric and dash-suffixed values (and apply a higher maxlength).
- **Documentation**
  - Setup guidance updated to treat `postalCodePattern` as the single source of truth.
- **Localization**
  - Added length-specific and generic invalid postal-code messages.
- **Tests**
  - Added Playwright regression coverage for input preservation and length-specific error messaging.


subhashini-egov (changes requested):
@vinothrallapalli-eGov Please ensure you address all code review comments.

digit-ui-esbuild/products/pgr/src/utils/postalCode.js
```diff
@@ -0,0 +1,62 @@
...
+// (`CS_COMPLAINT_POSTALCODE_INVALID_ERROR_LEN`, seeded per-locale in the
+// `rainmaker-pgr` localization module — see
+// utilities/default-data-handler/src/main/resources/localisations/*/rainmaker-pgr.json).
+// For non-numeric-length patterns (e.g. UK alnum) we fall back to a
+// generic, still-localized message instead of asserting a wrong digit
+// count.
+
+export function getPostalCodePattern() {
+  const getConfig = typeof window !== "undefined" ? window.globalConfigs?.getConfig : undefined;
+  const cfg = getConfig?.("CORE_POSTAL_CONFIGS") || getConfig?.("CORE_POSTAL_CODE_CONFIGS") || {};
```
KDwevedi: **The legacy `CORE_POSTAL_CODE_CONFIGS` fallback is unreachable.**

`globalConfigs.js.j2:74` renders `var corePostalConfigs = {{ core_postal_configs | default({}) | to_json }}`, so `getConfig("CORE_POSTAL_CONFIGS")` always returns at least `{}` — which is truthy, so the `||` never falls through to the legacy key. `configurator/src/admin/validation.ts:100` has the same problem with `??` (`{}` isn't nullish either).

So `5ce9df8`'s stated goal — "picks up the legacy `CORE_POSTAL_CODE_CONFIGS` fallback" — doesn't hold on any ansible-rendered globalConfigs. Select on the field, not the object:

```js
const cfg =
  [getConfig?.("CORE_POSTAL_CONFIGS"), getConfig?.("CORE_POSTAL_CODE_CONFIGS")]
    .find((c) => c?.postalCodePattern) || {};
```

digit-ui-esbuild/products/pgr/src/configs/CreateComplaintConfig.js
```diff
@@ -216,16 +218,36 @@ export const CreateComplaintConfig = {
...
-                  // to the legacy 5-digit default when the host hasn't set it.
-                  pattern: new RegExp(
-                    window?.globalConfigs?.getConfig?.("CORE_POSTAL_CONFIGS")?.postalCodePattern || "^[0-9]{5}$"
-                  ),
+                  // No configured postal pattern needs more than this many
+                  // digits — caps unbounded typing client-side (RenderFormFields
+                  // reads populators.validation.maxlength, lowercase) instead of
+                  // relying solely on the pattern check, which only surfaces
+                  // its error on submit.
+                  maxlength: 16,
```
KDwevedi, line 242: **This field is `type: "number"` (line 209), so the shared validator now accepts shapes the input physically can't hold.**

The PR deliberately un-restricted the citizen v2 input so UK/US/dash shapes can be typed, and `_example.yml` now advertises `^[A-Z]{1,2}...` and `^[0-9]{5}(-[0-9]{4})?$` as supported. But a `number` input won't take a letter, space or dash — and `createComplaintForm.js` calls `isPostalCodeValid()` against the *full* configured pattern and toast-and-returns. On such a tenant this field becomes unfillable **and** blocking.

Same for the legacy citizen path: `steps-config/locationDetails.js:46` is also `type: "number"`, and `FormExplorer.js` now validates it with the same shared helper.

Either flip both fields to `type: "text"` (the `maxlength` truncation in `TextInput.js` works for text too), or state in `_example.yml` that non-numeric patterns are citizen-v2-only. As it stands the PR's single-source-of-truth claim holds for the *rule* but not for the *surfaces*.

Minor, same file: `locationDetails.js:50` puts its cap at `populators.maxlength`, which nothing reads — `RenderFormFields.js:81` reads `populators.validation.maxlength`, as you correctly did here. That legacy field has never actually been capped.

configurator/src/admin/validation.ts
```diff
@@ -84,20 +102,17 @@ function resolvePostalRule(): { pattern: RegExp; message: string } {
   const patternStr =
     (mdmsRule?.pattern as string | undefined) ||
     (globalRule?.postalCodePattern as string | undefined);
```
KDwevedi, line 115: **This resolver prefers MDMS (`__DIGIT_USER_VALIDATION.postalCode`) over globalConfigs; the new `utils/postalCode.js` ignores MDMS entirely.**

That's two different precedence orders for one rule. A tenant that authors the postal rule in MDMS — which master's `common-masters.FormValidations` split (`ce10afcf4`) actively encourages — gets one pattern in DIGIT Studio and a different one in the PGR create-complaint flows. That is #722's exact failure mode, just moved up a layer.

Since the PR's thesis is "one knob", either teach `getPostalCodePattern()` to read `window.__DIGIT_USER_VALIDATION?.postalCode?.pattern` first (same order as here), or document explicitly that globalConfigs wins for PGR and MDMS is Studio-only.

digit-ui-esbuild/products/pgr/src/pages/citizen/Create/CreatePGRFlowV2.tsx
```diff
@@ -711,17 +667,15 @@ function Step2Location({ data, patch, t }: StepBodyProps) {
...
-          error={showPostalError ? postalErrorText(t) : undefined}
+          error={showPostalError ? getPostalCodeErrorMessage(t) : undefined}
         >
           <Input
             id="postal-code"
             type="text"
-            inputMode="numeric"
-            pattern="[0-9]*"
-            maxLength={7}
+            maxLength={16}
```
KDwevedi, line 692: Dropping `inputMode="numeric"` / `pattern="[0-9]*"` alongside the digit-strip is a mobile-keyboard regression for every digit-only tenant — which is all the real deployments (KE 5-digit, MZ 4-digit, IN 6-digit). Citizens now get a full alpha keyboard for a numeric field.

`utils/postalCode.js` already knows whether the pattern is digit-only (`getPostalCodeDigitLength()` returns non-null exactly then). Export it and make the hint conditional:

```tsx
inputMode={isPostalCodeNumeric() ? "numeric" : "text"}
```

That keeps the alnum fix without costing the common case.

utilities/default-data-handler/src/main/resources/localisations/default/rainmaker-pgr.json
```diff
@@ -2201,6 +2201,18 @@
         "module": "rainmaker-pgr",
         "locale": "default"
     },
+    {
+        "code": "CS_COMPLAINT_POSTALCODE_INVALID_ERROR_LEN",
+        "message": "Please enter a valid {{length}}-digit postal code",
```
KDwevedi, line 2212: This is the **first `{{ }}` interpolated message in either locale pack** — nothing else in `rainmaker-pgr.json` uses one, so this path has no precedent in the repo.

It should work: `packages/libraries/src/translations/index.js` initialises i18next with interpolation enabled (`escapeValue: false`). But that same config chains two post-processors over every resolved value (`ReactPostprocessor` + the custom `templatePostprocessor`, whose `replaceLiterals()` does its own brace matching), and `getPostalCodeErrorMessage()` decides "missing key" by comparing the result to the key string.

Nothing in the PR demonstrates the rendered output: the new Playwright spec only asserts input preservation, and the manual-verification box in the test plan is still unchecked. One assertion that the visible error reads "…valid 4-digit postal code" on a 4-digit tenant would be worth more than the input-preservation test, since that's the actual thing #722 is about.

KDwevedi (commented):
Right direction and the blast radius is safe — but the consolidation stops one layer short: the *validator* is now shared, the *input widgets* it validates are not, and on a non-numeric tenant that combination hard-blocks the form. Inline comments carry the specifics; the rest is below.

Reviewed head `5ce9df8` against `upstream/master`, plus consumers, the i18n bootstrap, the form-render plumbing and deployed host_vars.

### Worth fixing before merge

1. **Inputs weren't consolidated with the validators** — see the inline comment on `CreateComplaintConfig.js`. The employee field and the legacy citizen field (`steps-config/locationDetails.js:46`) are both still `type: "number"`, so on any tenant using the alnum/dash patterns this PR's own `_example.yml` now advertises, those two forms can't hold a value the shared validator accepts.
2. **New drift vector, one layer up** — see the inline comment on `configurator/src/admin/validation.ts`: MDMS-first there, MDMS-never in `utils/postalCode.js`.
3. **The legacy-key fallback is unreachable** — see the inline comment on `utils/postalCode.js:23`.

### Second-order effects

- **Inert `maxlength` on the legacy citizen field.** `locationDetails.js:50` sets `populators.maxlength: 7`, but nothing reads `populators.maxlength` — `RenderFormFields.js:81` reads `populators.validation.maxlength`. So that field has never had a cap, and the two configs now disagree on where the key lives. (The employee cap added in `5ce9df8` *is* real: `TextInput.js:322` special-cases `type === "number"` + `maxlength` and truncates.)
- **Upgrade ordering.** `CreateComplaintConfig.js:250` points at `..._GENERIC` with no code-level fallback, unlike `getPostalCodeErrorMessage()` which falls back to English. On an existing install the field renders the raw key until DDH re-seeds *and* the `digit-redis` `messages` hash is cleared. Worth a line in the PR body / release notes.
- **Mobile keyboard regression.** See the inline comment on `CreatePGRFlowV2.tsx`.
- **Allowlist matching.** `pincodeAllowlistOk()` normalizes by stripping leading zeros and compares case-sensitively — correct for digits, wrong now that `"sw1a 1aa"` can be typed. Only bites tenants running alnum postal codes *and* a `pincode` allowlist, so low priority, but it's downstream of this change.
- **Deployed configs degrade cleanly** — confirmed. The template passes `core_postal_configs` through `to_json` verbatim, so host_vars still carrying `postalCodeLength`/`postalCodeErrorMessage` (e.g. Maputo: `^[0-9]{4}$`, length 4) keep rendering and are simply ignored. Maputo actually improves: its `postalCodeErrorMessage` pointed at `CS_COMPLAINT_POSTALCODE_INVALID_ERROR`, which is seeded **nowhere** in this repo — that surface was rendering the raw key, and now gets "valid 4-digit postal code". Still a silent behavior change for anyone who had set a real message; call it out in the notes.
- **No backend coupling** — no vendored Java service reads a pincode pattern, so this is FE + ansible only. Good.

### Process

- The two red checks are **not** this PR: `Local Setup CI` fails on an expired TLS cert for `registry.preview.egov.theflywheel.in` (`certificate has expired ... after 2026-07-25`). Unrelated infra.
- Base is `master` and the head is 126 commits behind it (still auto-mergeable; `validation.ts` hasn't drifted underneath). `develop` doesn't have this fix and is behind `master` on `CreatePGRFlowV2.tsx` / `_example.yml` — worth confirming the forward-port path so this doesn't get reverted on the next sync.
- `subhashini-egov`'s CHANGES_REQUESTED is still standing with no unresolved threads left; needs a re-review request once the above lands.

digit-ui-esbuild/products/pgr/src/utils/postalCode.js
vinothrallapalli-eGov: Fixed in adc453bbe — both call sites now select on the field, not the object: `[getConfig?.("CORE_POSTAL_CONFIGS"), getConfig?.("CORE_POSTAL_CODE_CONFIGS")].find((c) => c?.postalCodePattern)`, exactly as suggested (same fix in `validation.ts`, which had the identical problem with `??`). Verified in a Node harness that the legacy key is now reachable behind an ansible-rendered empty `CORE_POSTAL_CONFIGS = {}`.

digit-ui-esbuild/products/pgr/src/configs/CreateComplaintConfig.js
vinothrallapalli-eGov: Fixed in adc453bbe — both fields flipped to `type: "text"`: this employee field and the legacy citizen one in `steps-config/locationDetails.js`, so all three surfaces can now hold any shape the shared validator accepts. The employee cap keeps working through the native `maxLength` attr (`TextInput.js` passes `populators.validation.maxlength` straight through for text inputs). Also took your minor point: `locationDetails.js`'s inert `populators.maxlength: 7` is now a real `populators.validation.maxlength: 16`, with a comment noting the old key was never read.

configurator/src/admin/validation.ts
vinothrallapalli-eGov: Fixed in adc453bbe — went with option 1: `getPostalCodePattern()` now reads `window.__DIGIT_USER_VALIDATION?.postalCode?.pattern` first, then globalConfigs (both keys, by field), then the 5-digit default — the same order as `resolvePostalRule()` here, and the doc comment cross-references it. Note nothing populates the `postalCode` channel yet (on master, `useMobileValidation` publishes only mobile/name/email from `FormValidations`), so this is zero-behavior-change today; when a `postalCode` fieldType row gets wired into that hook, both layers pick it up with identical precedence.

digit-ui-esbuild/products/pgr/src/pages/citizen/Create/CreatePGRFlowV2.tsx
vinothrallapalli-eGov: Fixed in adc453bbe — exported `isPostalCodeNumeric()` from `utils/postalCode.js` (non-null `getPostalCodeDigitLength()`, as suggested) and the input now sets `inputMode`/`pattern` conditionally: digit-only tenants (KE 5, MZ 4, IN 6) keep the numeric pad, alnum/dash tenants get the full keyboard. Still no keystroke filtering in either mode, so the shared validator remains the sole gate. Also fixed the downstream `pincodeAllowlistOk()` point from your review body: comparison is now case-folded, and leading zeros are stripped only from purely numeric values.

utilities/default-data-handler/src/main/resources/localisations/default/rainmaker-pgr.json
vinothrallapalli-eGov: Fair — added the assertion in adc453bbe: a second spec in `postal-code-alnum-input.spec.ts` reads the tenant's effective pattern in-page (same resolution order as `utils/postalCode.js`), skips unless it's `^[0-9]{N}$`-shaped, types N+1 digits, and asserts the *visible* error text matches "valid N-digit postal code" with the tenant's actual N. That exercises the full `{{length}}` interpolation path through the i18next init with both post-processors — a mangled placeholder, a raw key leaking through, or a digit count derived from the wrong quantifier all fail it. The manual-verification box on a 4-digit tenant is still open; the upgrade-ordering caveat for the `_GENERIC` key (DDH re-seed + `digit-redis` `messages` flush) is now called out in the PR body.

vinothrallapalli-eGov:
@KDwevedi thanks for the thorough pass — all five inline comments are addressed in adc453bbe (replies on each thread). On the review-body points:

- **Inputs consolidated with the validators**: employee + legacy citizen postal fields are now `type: "text"`, so every surface can hold what the shared validator accepts; the citizen v2 field regains a *conditional* numeric-keyboard hint via the new `isPostalCodeNumeric()`.
- **Inert legacy `maxlength`**: moved to `populators.validation.maxlength: 16` (and documented that the old key was never read).
- **Upgrade ordering + Maputo message change**: now called out in a new "Upgrade / release notes" section in the PR body, including the DDH re-seed + `digit-redis` `messages`-hash flush requirement for the `_GENERIC` key and the forward-port note for `develop`.
- **Allowlist matching**: `pincodeAllowlistOk()` now case-folds and only strips leading zeros from purely numeric values.

Validation: configurator `tsc` + 78/78 vitest, esbuild bundle build clean, `tsc --noEmit` clean on the new integration spec, plus a Node harness confirming the legacy-key fallback is reachable and MDMS precedence matches `validation.ts`.

@subhashini-egov this should now cover all outstanding review comments — requesting a re-review.

digit-ui-esbuild/products/pgr/src/utils/postalCode.js
```diff
@@ -0,0 +1,91 @@
...
+// (`CS_COMPLAINT_POSTALCODE_INVALID_ERROR_LEN`, seeded per-locale in the
+// `rainmaker-pgr` localization module — see
+// utilities/default-data-handler/src/main/resources/localisations/*/rainmaker-pgr.json).
+// For non-numeric-length patterns (e.g. UK alnum) we fall back to a
+// generic, still-localized message instead of asserting a wrong digit
+// count.
+
+export function getPostalCodePattern() {
+  if (typeof window === "undefined") return "^[0-9]{5}$";
+  const mdmsPattern = window.__DIGIT_USER_VALIDATION?.postalCode?.pattern;
```
KDwevedi: Precedence now matches `resolvePostalRule()` — but worth knowing this branch is **currently unreachable**: nothing writes `__DIGIT_USER_VALIDATION.postalCode`. `hooks/pgr/useMobileValidation.js:126-138` mirrors only `.mobile`, `.name` and `.email` onto that channel, and the master it reads has no postal field. (The same is true of `validation.ts`'s step 1, which has documented an MDMS channel that never existed since #690.)

So this is a no-op on every live tenant today — no behaviour change, and no ordering hazard from the hook writing after first render, since the key is never written at all. Fine to land as forward-compatible wiring.

The one-line follow-up that makes it real, alongside the existing name/email mirrors:

```js
if (mdmsConfig.postalRegex) {
  window.__DIGIT_USER_VALIDATION.postalCode = { pattern: mdmsConfig.postalRegex };
}
```

Worth noting in the PR body either way: the day that master gains a postal field, MDMS will silently outrank every tenant's `core_postal_configs` with no host_vars change. That's the intended precedence (same as mobile/name/email), just a surprise if it isn't written down.

tests/integration-tests/tests/citizen/postal-code-alnum-input.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    // N+1 digits can never satisfy ^[0-9]{N}$, so the field must flag it.
+    await postalInput.click();
+    await postalInput.fill('');
+    await postalInput.pressSequentially('1'.repeat(n + 1), { delay: 50 });
+
+    // The visible error must carry the tenant's real digit count — not a raw
+    // localization key, not a mangled "{{length}}" placeholder, and not a
+    // hardcoded count from some other tenant's pattern.
+    const errorText = page.getByText(new RegExp(`valid\\s+${n}[\\s-]?digit postal code`, 'i'));
```
KDwevedi: `getByText()` resolves to *all* matching nodes, and `toBeVisible()` throws a strict-mode violation on more than one. The message is currently rendered once (the `Field` error slot), but if the wizard ever also surfaces it in a toast or a summary step this fails as a strict-mode error rather than a clean assertion — which reads like a real regression when it isn't. `.first()` keeps the assertion honest:

```ts
const errorText = page.getByText(new RegExp(`valid\\s+${n}[\\s-]?digit postal code`, 'i')).first();
```

KDwevedi (commented):
Re-reviewed at `adc453b`. All five points are addressed, and I verified the mechanisms rather than just the diff:

- **`type: "number"` → `"text"` on both configs** — the cap survives the switch: `TextInput.js:345` applies `maxLength={props.maxlength}` unconditionally, so the `type === "number"` truncation special-case at line 322 is no longer needed and text gets the native attribute. Payload shape is unchanged too — `event.target.value` was always a string, so `formPayloadToCreateComplaint`'s `pincode` is the same as before. The two forms are now genuinely fillable on an alnum/dash tenant.
- **`locationDetails.js`** — cap moved under `populators.validation.maxlength`, which is the key `RenderFormFields.js:81` actually reads. That field now has a real cap for the first time.
- **Legacy-key fallback** — `.find((c) => c?.postalCodePattern)` in both `postalCode.js` and `validation.ts` reaches the legacy key past the `{}` the ansible template always renders. Correct.
- **`inputMode`** — `isPostalCodeNumeric()` is the right seam, and digit-only tenants keep the number pad.
- **Allowlist `norm()`** — case-fold plus numeric-only zero-stripping is right; `"0100" ≡ "100"` for numeric codes while a leading zero stays significant in an alnum code.
- **The `{{length}}` test** now proves the actual #722 deliverable end-to-end, and it resolves the pattern the same way the app does instead of assuming a tenant. This is the assertion that was missing.

Two small things below, neither blocking. CI is still red on `test` / `tilt-test` for the same unrelated reason (expired TLS cert on `registry.preview.egov.theflywheel.in`, `after 2026-07-25`) — confirmed on this commit's run.

One stale-comment cleanup if you're touching the branch again: `tests/integration-tests/tests/specs/pgr-postal-code.spec.ts` still says "the postalCode field is also type=number, which silently ignores its pattern attribute" and does `POSTAL_CODE_VALID.split('-')[0]` to dodge a hyphen the field can now hold. Harmless, but it documents behaviour this commit just changed — and on `mz.maputo` the full `0101-03` sample is now actually enterable.

From my side this is good to merge once the two notes are resolved or waved off; `subhashini-egov`'s CHANGES_REQUESTED still needs a re-request.

KDwevedi (approved):
Approving at `adc453b` — every point from my first pass is addressed, and I verified the mechanisms (maxLength survives the `number` → `text` switch, the cap now sits on the key `RenderFormFields` actually reads, the legacy-key fallback is reachable, and the `{{length}}` message is proven end-to-end by the new spec).

The two items in my follow-up review are non-blocking and can land as a follow-up: the MDMS branch being unreachable until `useMobileValidation` mirrors `postalCode` (a no-op today, no behaviour change), and the `getByText(...).first()` strict-mode guard. Same for the stale comments in `tests/specs/pgr-postal-code.spec.ts`.

Red CI is the expired `registry.preview.egov.theflywheel.in` TLS cert, unrelated to this change.

digit-ui-esbuild/products/pgr/src/utils/postalCode.js
vinothrallapalli-eGov: Resolved for real in eb72731a6 — the channel is no longer forward-compatible wiring, it's live: `useMobileValidation` (digit-ui) and a new `usePostalRule` hook (configurator) both mirror a `fieldType: "postalCode"` FormValidations row, and per a follow-up review ask the postal rule now rides `window.__DIGIT_FORM_VALIDATIONS` — a channel named after the master, keyed by fieldType — rather than piggybacking on `__DIGIT_USER_VALIDATION`. DDH seeds a default 5-digit row (matching the code fallback), the schema is registered in `full-dump.sql` (verified loading the full dump into postgres 16 with ON_ERROR_STOP), and a Studio schema descriptor makes the row authorable. Precedence was proven live on a local stack with a deliberately divergent MDMS 4-digit row over a globalConfigs 6-digit pattern — both employee and citizen surfaces followed MDMS. The rollout caveat you asked to have written down (MDMS silently outranking `core_postal_configs`) is now in the PR body's upgrade notes, including the migration step for non-5-digit deployments.

tests/integration-tests/tests/citizen/postal-code-alnum-input.spec.ts
vinothrallapalli-eGov: Fixed in efcef0ee7 — `.first()` added exactly as suggested, with a comment explaining the strict-mode rationale. Also took your stale-comment note from the review body: `tests/specs/pgr-postal-code.spec.ts` no longer claims the field is type=number, uses `POSTAL_CODE_VALID` verbatim (mz.maputo's full `0101-03` is now enterable), and its expected-message resolution derives the length-interpolated `..._LEN` text since a raw key can no longer appear in the DOM. Both postal suites (11 UI tests, including a new real-time-typing case) pass against a live local stack — including the `{{length}}` interpolation assertion you noted was the missing proof.

subhashini-egov (approved):

PR #1316, opened 2026-07-20
Release v2.12

vinothrallapalli-eGov:
## Summary

Promotes `develop` to `master` for the v2.12 release — comparing against the [v2.11](docs/release-notes-v2.11.md) tag.

See [docs/release-notes-v2.12.md](docs/release-notes-v2.12.md) (draft, pending final review) for the full feature/enhancement/bug-fix breakdown, covering Complaint Hierarchy (N-Level), Dashboard & Analytics, Configurator, PGR/Complaints, Mobile Number Validation, Keycloak/Auth, Notifications (Novu/WhatsApp), MCP Tools, Chatbot, and Observability.

Note: this is a large diff (`develop` has been the active branch for an extended period since `master` was last updated) — the previous attempt at this release PR (#1089) was closed after CodeRabbit's automated review skipped it for exceeding its file-count limit. This PR carries the same scope; review here is expected at the release-notes level rather than a line-by-line diff.

## Test plan
- [ ] Confirm `docs/release-notes-v2.12.md` reflects everything merged since it was drafted (2026-06-30) through the current `develop` HEAD
- [ ] Confirm CI is green on `develop` before merging
- [ ] Tag `master` as `v2.12` after merge, matching the `v2.11` precedent

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1321, opened 2026-07-20
fix(config-service): serialize jsonb criteria filter values with Jackson

KDwevedi:
## Context
Surfaced by the Claude bot review on the **v2.12 release PR #1316** (CodeRabbit skipped that PR for exceeding its file-count limit). Fixing on `develop` so it flows into the release.

## 🔴 Bug: nested criteria filter values crash the jsonb cast
This release widened `ConfigDataCriteria.criteria` and `ConfigDataResolveRequest.ResolveParams.criteria` from `Map<String,String>` to `Map<String,Object>` so filter values can be numbers, booleans, or nested objects/arrays for jsonb containment queries.

But `QueryUtil.preparePartialJsonStringFromFilterMap` built the JSON string by hand and only special-cased `String` values (`val instanceof String`); every other type fell through to `sb.append(val)` → `Object.toString()`. Jackson deserializes a nested JSON object into a `LinkedHashMap` and an array into an `ArrayList`, whose `toString()` renders **Java** syntax (`{a=1}`, `[a, b]`), not JSON.

That malformed literal is bound into `data @> CAST(? AS jsonb)`, so a request like:

```json
POST /config/data/_search
{ "criteria": { "tenantId": "pg", "config": { "nested": 1 } } }
```

produces `{"tenantId":"pg","config":{nested=1}}` → Postgres `invalid input syntax for type json` → an unhandled **500** on both `/config/data/_search` and `/config/data/_resolve`, for exactly the value shapes the `Object` widening was meant to support. Scalar filters (string/number/boolean) happen to render valid tokens, so most existing callers are unaffected.

## Fix
Replace the hand-rolled string building with `ObjectMapper.writeValueAsString(filterMap)` (mirroring `ConfigDataRepository.toJsonString`), which serializes nested Map/List values as valid JSON at any depth. Empty/null map still short-circuits to `{}`.

## Test plan
- [ ] CI green
- Added `QueryUtilTest`: scalars, quote-escaping, and the nested object/array regression (asserts parseable JSON and that the old `nested=1` rendering is gone).

### DIFF
--- a/backend/digit-config-service/src/main/java/org/egov/config/utils/QueryUtil.java
+++ b/backend/digit-config-service/src/main/java/org/egov/config/utils/QueryUtil.java
@@ -1,11 +1,15 @@
 package org.egov.config.utils;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.StringJoiner;
 
 public class QueryUtil {
+
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
     private QueryUtil() {}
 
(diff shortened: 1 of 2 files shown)


KDwevedi:
@coderabbitai review

vinothrallapalli-eGov (approved):

Issue #521, opened 2026-05-05
[Bug] Escalate button not working for freshly created application


pradeepkumarcm-egov:
**Describe the bug**

Search for records that are pending assignment. When a complaint is selected, the Escalate option appears in the action dropdown; however, clicking it does not trigger any further action.

 

**Expected behavior**

It should give option to select user to whom the complaint to be forward.

**Screenshots**
 
https://github.com/user-attachments/assets/0a10cf5e-3d21-4629-b664-a5faf5ac122c

Gurjeet-eGov:
@ChakshuGautam 
after complaint creation the **escalate** button is non functional at GRO login 

https://github.com/user-attachments/assets/a2eeed43-4ec7-4615-b400-574f903aad1e

Gurjeet-eGov:
Unresolved  

https://github.com/user-attachments/assets/4292ee71-b013-4876-a982-2bda22a0d3a9

KDwevedi:
Handoff note (root cause already traced, reassigning to @ChakshuGautam):

The Escalate option renders nothing because **`products/pgr/src/pages/employee/PGRDetails.js` has no `ESCALATE` entry in `ACTION_CONFIGS`** (it only defines ASSIGN / REOPEN / REJECT / RESOLVE / REASSIGN, lines 15–197). On select, `getUpdatedConfig()` (line ~368) does `configs.find(c => c.actionType === 'ESCALATE')` → `null`, and `PGRWorkflowModal.js:48` (`if (!config) return null`) renders nothing — so the button is a visible no-op.

**Fix:** add an `ESCALATE` entry mirroring `REASSIGN` (a `PGRAssigneeComponent` picker + mandatory comments). The dynamic role population in `getUpdatedConfig`/`computeAssigneeRoles` already injects the correct forward-state assignee roles, and `handleActionSubmit` already maps `SelectedAssignee.uuid → workflow.assignes`. Add one localization key (`CS_ACTION_ESCALATE`, en_IN + sw_KE). Confirm the exact action code the live PGR BusinessService emits for the dropdown option (expected `ESCALATE`; `SLA_ESCALATE` is the AUTO_ESCALATE/scheduler action and won't appear in a human dropdown).

This also unblocks the remaining item on #490 (LME action availability in the assigned state).

KDwevedi:
[...]

PR #1375, opened 2026-07-22
fix(configurator): admin PGR create/list/assign/workflow bugs (two-level tenant)

dhruv-1001:
## Admin (configurator) + PGR seed bug fixes

Bugs surfaced by the integration test suite on a two-level (root/city) deployment
(Maputo). Most are in the admin SPA / its data-provider, plus one fix to the
shipped PGR workflow seed. Verified locally.

Test-suite changes are **not** in this PR — they are in #1376.

---

### 1. Complaint create stamped the wrong tenant

**What:** `service.tenantId` was set to the session/root tenant rather than the
tenant that owns the selected boundary.
**Why it mattered:** a root admin creating a complaint against a city boundary got
`INVALID_BOUNDARY_CODE`, because the boundary does not exist at the root.
**Fix:** derive the city tenant from the boundary and stamp that.

### 2. Department / tenant lists filtered `isActive` after pagination

**What:** the data provider requested a page of rows and *then* dropped inactive
ones client-side.
**Why it mattered:** on a tenant with soft-deleted rows the whole page could be
inactive, so the grid rendered empty under a "showing 1–N of M" header.
**Fix:** push `isActive` to the server so pagination runs over the filtered set.

### 3. Assign-to-employee sent the wrong key and merged a null `serviceCode`

**What:** the assignee was sent under `assignees`; PGR binds `assignes`. Separately,
a text-only edit merged a `null` `serviceCode` into the update payload.
**Why it mattered:** the assignment silently did nothing, and the null `serviceCode`
caused an NPE in pgr-services.
**Fix:** correct key name, plus a null guard on the edit merge.

### 4. Assignee dropdown listed the root tenant's employees

**What:** the employee picker searched HRMS at the session tenant.
**Why it mattered:** for a root admin working a city complaint, the list contained
people who cannot act on it, and omitted the ones who can.
**Fix:** `DigitFormSelect` gains a tenant filter and hrms `getOne` honors it, so the
picker lists the **complaint's city** employees.

### 5. Workflow actions read from the root tenant's PGR config (#521)

**What:** available actions were fetched against the session tenant.
**Why it mattered:** city-only actions (e.g. `ESCALATE`) were missing from the action
bar.
**Fix:** read the workflow from the complaint's city tenant.

### 6. Wizard-created tenants had no egov-enc-service key

**Files:** `configurator/src/api/services/tenantBootstrap.ts`,
`configurator/src/pages/Phase1Page.tsx`, `configurator/src/pages/Phase4Page.tsx`

**What:** egov-enc-service holds exactly one symmetric key per tenant, and every
write carrying PII is sealed with it. `bootstrapStateRoot()` generated a key for a
brand-new state root, but the wizard's normal path creates a city under an
**existing** root and skips bootstrap entirely — so no key was ever generated.

**Why it mattered:** every tenant the wizard created was structurally incapable of
accepting employees. Phase 4 always failed on it with
`400 UNKNOWN_ERROR "Unknown error occurred in encryption process"`.

**Fix:** `registerEncKey` is extracted and exported as `ensureEncKey()`, and Phase 1
calls it on the tenant-creation path. Deliberately narrow so it can never rotate a
live tenant's key:

- called only on the creation path — skipped on the duplicate branch, and for a
  fresh state root bootstrap has already done it;
- `POST /crypto/v1/_generatekey` is create-if-missing, not rotation. Verified against
  a throwaway tenant: three calls returned `created: true / false / false` with a
  single DB row whose `key_id` and `secret_key` never changed;
- no "regenerate key" affordance is exposed anywhere in the UI. Rotation would strand
  already-encrypted data (#622) and remains a deliberate ops action.

Failure is non-fatal — the tenant is created and Phases 2–3 work — so it raises a
non-blocking warning rather than rolling the operator back.

Also in `Phase4Page.tsx`: the employee-create loop continues past a row failure, so
the enclosing try/catch never observed per-row errors and the complete screen could
only report "0 created, 1 failed" with no reason (the Banner even promised a "failure
list below" that did not exist). Per-row `ApiClientError.firstError` is now recorded
and rendered.

### 7. PGR workflow seed referenced per-tenant state UUIDs

**File:** `utilities/default-data-handler/src/main/resources/PgrWorkflowConfig.json`

**What:** three transitions referenced workflow **state UUIDs** captured from the
tenant the seed was originally taken from.

**Why it mattered:** state UUIDs are minted per tenant, so on any other tenant those
references resolve to nothing. `PENDINGATLME --FORWARD--> 04752227-…` was a dangling
transition — confirmed live on `mz.maputo` before the fix.

**Fix:** reference the state **names**, which the workflow service resolves per
tenant:

| was | now | where |
| --- | --- | --- |
| `04752227-e267-43e3-bfd8-6e48bdbfd97a` | `PENDINGATSUPERVISOR` | the `FORWARD` `nextState`, and a `currentState` |
| `e80e3dd6-8336-4586-b84a-c191dcd80c98` | `RESOLVED` | the `RESOLVEBYSUPERVISOR` `nextState` |

The `ESCALATE` row already used a name and is unchanged.

### 8. Complaint-types list had no Department filter

**File:** `configurator/src/resources/complaint-types/ComplaintTypeList.tsx`

**What:** the list rendered a Department column but offered no way to filter by it,
unlike `DepartmentList` and `DesignationList`, which both ship a `FilterBar`.

**Why it mattered:** Department is the axis admins actually slice this list by
("what does Public Works own?"), and there was no way to do it.

**Fix:** added a `filters` array with a `ReferenceFilterInput` for department,
declared `alwaysOn` so it is visible without going through the Add-filter popover.
Note that passing `filters` to `DigitList` swaps out its built-in search box for the
`FilterBar`, so `SearchFilterInput` is re-declared alongside it — otherwise free-text
search would have disappeared from the page.

### 9. Minor UX

Create → Show redirect after creating a record.

### 10. Complaint edits silently discarded themselves — data loss

**Files:** `configurator/src/resources/complaints/LocalityPicker.tsx`,
`configurator/src/resources/complaints/ComplaintHierarchyCascade.tsx`,
`configurator/packages/data-provider/src/providers/dataProvider.ts`

Editing a complaint returned `200`, advanced the workflow, and left the record
untouched — description, status, everything. `egov-persister` logged **825**
`null value in column "locality" of relation "eg_pgr_address_v2"` rows in 24 hours; the
address and service UPDATEs share one transaction, so the service row rolled back with
the address, and after nine retries the message was dropped. This affected real
operators, not only tests.

**Why:** a Radix Select interaction, not our state handling. Inside a `<form>`, Radix
renders a hidden form-bubble `<select>` whose `<option>`s are contributed by
`SelectItem`s living in the *portalled content* — not mounted while the dropdown is
closed. Its effect runs on every controlled-value change; with no options mounted the
assignment cannot take and it emits `onValueChange("")`. On the edit form the
hierarchy/type selects go `"" → resolved default` as soon as the boundary queries
settle, so both echo `""` unprompted on **every page load**, and ra-core's
`defaultParse` turns `''` into `null`.

- Both components now ignore an empty `onValueChange` payload (Radix forbids an empty
  `SelectItem` value, so `""` is only ever the echo).
- `dataProvider` gains `mergePgrAddress()` — defence in depth: `locality.code` is
  restored from the loaded record when the incoming one is empty, since the column
  forbids null and clearing a locality is not a supported edit. Other address fields
  merge as before, because nulling a landmark *is* a real edit.
- `stripLocalitySidecars()` drops the picker's `__h`/`__t` navigation keys, which were
  leaking into PGR's address contract via `useInput` on `` `${source}__h` ``.
- `perPage` raised for boundaries/hierarchies. The provider fetches everything and
  slices client-side, so `perPage: 1000` against 1256 nodes was a pure truncation knob.
  This produced the visible tell (the Boundary control showing its placeholder) but was
  **not** the null source — a competing hypothesis that the evidence refuted.

**Verified:** an edit persists through a re-search, pgr and `egov-wf` agree on the
state, and `egov-persister` logs zero errors over the window.

### 11. The configurator offered soft-deleted complaint types

**Files:** `configurator/src/api/services/mdms.ts`,
`configurator/src/resources/complaints/ComplaintHierarchyCascade.tsx`,
`configurator/src/resources/complaint-hierarchies/ComplaintHierarchyShow.tsx`

Selecting one produced an opaque `400 INVALID_SERVICECODE ... not present in MDMS`.
`mdmsService.search()` never sent `isActive: true` and then did
`.map(record => record.data)`, discarding the record-level flag entirely — so no caller
could filter on it. The cascade filtered on `data.active`, a *different* field, still
`true` on deleted rows. Four deleted types sorted **first** (`order: 1`) ahead of the two
real ones.

`search()` now requests active rows (opt out via `includeInactive`), re-filters
client-side as a fallback, and preserves the flag as `_isActive`. Every caller was
reviewed: none legitimately wants inactive rows — migrating or verifying against a
deleted ServiceDef would be a false pass. **Behaviour change worth noting:** the bulk
importers use these lists for duplicate detection, so a soft-deleted department is no
longer treated as a duplicate; the create then surfaces MDMS's own "uid taken" error
instead of being silently skipped. `ComplaintHierarchyShow`'s pager assumed a short page
meant the last page, which stops holding once `search()` can drop rows — it now breaks on
an empty page.

### 12. The complaints Department filter did not filter

**File:** `configurator/packages/data-provider/src/providers/dataProvider.ts`

Selecting a department left the grid byte-identical and the outbound request unchanged.
`ComplaintList` declares `source="additionalDetail.department"`; react-hook-form nests
the dotted path, while the provider read the flat key and found `undefined`.

Fixed **generically** rather than by renaming the source: `flattenFilterSources()`
additionally publishes every nested leaf under its dotted path, so any future dotted
source works too. The nested shape is kept intact (ra-core's filter-form reset depends on
it) and a flat key passed by a caller is never overwritten. The declaration is
deliberately left dotted so the path stays exercised as a regression guard.

### 13. Date filters were discarded before the request was built

**File:** `configurator/packages/data-provider/src/providers/dataProvider.ts`

`DateFilterInput` drives an `<input type="date">`, so the value is the string
`"2026-07-20"`; the provider gated on `typeof === 'number'` and dropped it silently.

`toEpochMs()` coerces `number | Date | YYYY-MM-DD | epoch-string`, parsing the date parts
in **local** time — `new Date(str)` is spec'd as UTC midnight and lands on the wrong day
east of UTC. `'end'` resolves to `23:59:59.999` so an inclusive To date still matches
same-day complaints. `fromDate` is anchored to `0` when only `toDate` is set, because
`PGRQueryBuilder` throws `INVALID_SEARCH` on a to-without-from — latent before, since both
values were always dropped.

---

## Known-open, diagnosed, **not** addressed by this PR

Investigated in the same pass and understood, but each needs a decision or a fix that is
out of scope here. Their tests in #1376 are **deliberately red** rather than skipped, so
they stay visible.

- **Complaint-type create is broken.** `ComplaintTypeCreate.tsx` omits `hierarchyType`
  and `levelCode`, both of which the schema requires → `400 INVALID_REQUEST_REQUIRED`.
  The fix is ~2 lines, but needs a product decision on what `levelCode` should default to,
  or whether it should derive from the selected parent.
- **`_search` silently caps at 200 rows** while `_count` reports the true total (e.g. 203).
  Requesting 250 or 300 still yields exactly 200. Backend; needs a decision on whether the
  cap is intended and, if so, on making the counter agree with it.
- **A ward-scoped CSR receives the full boundary tree.** `boundary-service` ignores the
  jurisdiction when resolving boundary-relationships. Backend, outside the configurator.
- **pgr-services accepts an `ASSIGN` with an empty `assignes` array** and moves the
  complaint to `PENDINGATLME`. Backend validation gap, noticed while diagnosing the above.

### DIFF
--- a/configurator/packages/data-provider/src/client/DigitApiClient.ts
+++ b/configurator/packages/data-provider/src/client/DigitApiClient.ts
@@ -491,7 +491,10 @@ export class DigitApiClient {
   async pgrUpdate(service: Record<string, unknown>, action: string, options?: {
     comment?: string; assignees?: string[]; rating?: number;
   }): Promise<Record<string, unknown>> {
-    const workflow: Record<string, unknown> = { action, assignees: options?.assignees || [], comments: options?.comment };
+    // The PGR workflow POJO binds `assignes` (single-e), NOT `assignees`. Sending
+    // `assignees` silently drops the assignee, so an ASSIGN runs with no target and
+    // pgr-services 400s with a NullPointerException. Send the key the server reads.
+    const workflow: Record<string, unknown> = { action, assignes: options?.assignees || [], comments: options?.comment };
     if (options?.rating !== undefined) workflow.rating = options.rating;
 
     const data = await this.request<{ ServiceWrappers?: Record<string, unknown>[] }>(this.endpoint('PGR_UPDATE'), {
--- a/configurator/packages/data-provider/src/client/DigitApiClient.ts
+++ b/configurator/packages/data-provider/src/client/DigitApiClient.ts
@@ -187,11 +187,16 @@ export class DigitApiClient {
   // --- MDMS v2 ---
 
   async mdmsSearch(tenantId: string, schemaCode: string, options?: {
-    limit?: number; offset?: number; uniqueIdentifiers?: string[];
+    limit?: number; offset?: number; uniqueIdentifiers?: string[]; isActive?: boolean;
   }): Promise<MdmsRecord[]> {
     const criteria: Record<string, unknown> = { tenantId, limit: options?.limit || 100, offset: options?.offset || 0 };
     if (schemaCode) criteria.schemaCode = schemaCode;
     if (options?.uniqueIdentifiers) criteria.uniqueIdentifiers = options.uniqueIdentifiers;
+    // Push isActive to the server so limit/offset paginate over the FILTERED set.
+    // Filtering active rows client-side after a raw page is fetched drops rows the
+    // page already spent, so a window of soft-deleted rows renders empty under a
+    // "showing 1-N of M" footer (MDMS v2 honors this criterion).
+    if (options?.isActive !== undefined) criteria.isActive = options.isActive;
 
     const data = await this.request<{ mdms?: MdmsRecord[] }>(this.endpoint('MDMS_SEARCH'), {
       RequestInfo: this.buildRequestInfo(), MdmsCriteria: criteria,
--- a/configurator/packages/data-provider/src/providers/dataProvider.ts
+++ b/configurator/packages/data-provider/src/providers/dataProvider.ts
@@ -870,7 +870,11 @@ export function createDigitDataProvider(client: DigitApiClient, tenantId: string
         return { data: normalizeRecord(users[0], config) };
       }
       if (config.type === 'workflow-bs') {
-        const services = await client.workflowBusinessServiceSearch(tenantId, [String(params.id)]);
+        // Honor a meta.tenantId override — a complaint's PGR workflow (actions
+        // like ESCALATE) lives at the CITY tenant; the root/session tenant's PGR
+        // config differs, so reading it there hides city-only actions.
+        const wfTenant = (params.meta as Record<string, unknown> | undefined)?.tenantId as string | undefined;
+        const services = await client.workflowBusinessServiceSearch(wfTenant || tenantId, [String(params.id)]);
         if (!services.length) throw new Error(`Workflow business service not found: ${params.id}`);
         return { data: normalizeRecord(services[0], config) };
       }
--- a/configurator/packages/data-provider/src/providers/dataProvider.ts
+++ b/configurator/packages/data-provider/src/providers/dataProvider.ts
@@ -794,7 +794,10 @@ export function createDigitDataProvider(client: DigitApiClient, tenantId: string
         if (!hasClientFilter) {
           const tenant = pickTenant(tenantId, filter);
           const offset = (page - 1) * perPage;
-          const raw = await client.mdmsSearch(tenant, config.schema!, { limit: perPage, offset });
+          // isActive:true so the server paginates over active rows only. The
+          // client-side .filter below stays as a defensive fallback for any MDMS
+          // that ignores the criterion (degrades to old behavior, never worse).
+          const raw = await client.mdmsSearch(tenant, config.schema!, { limit: perPage, offset, isActive: true });
           const data = raw.filter((r) => r.isActive).map((r) => normalizeMdmsRecord(r, config));
           const sorted = clientSort(data, field, order);
           const total = raw.length >= perPage ? offset + perPage + 1 : offset + data.length;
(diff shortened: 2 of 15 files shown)


dhruv-1001:
@claude review.

vinothrallapalli-eGov (approved):

PR #1383, opened 2026-07-22
fix(pgr): respect configured highest/lowest boundary level and seed m…

Lokendra-egov:
…issing localization (CCRS#721)

The citizen/employee complaint-location dropdown only capped the bottom of the boundary cascade (PGR_BOUNDARY_LOWEST_LEVEL) and ignored PGR_BOUNDARY_HIGHEST_LEVEL entirely, so tenants with a hierarchy deeper than intended (e.g. Maputo's Município -> District_Municipio -> Bairro -> Quarteirão, capped at Bairro) still showed every level.

- digit-ui-esbuild BoundaryComponent.js: cap the rendered cascade at both the configured highest and lowest level, not just the lowest.
- crs_dataloader dataloader_ui.py: fix update_hierarchy_schema() being called with highest_hierarchy=levels[-1] (the deepest level) instead of levels[0] (the top level) at both call sites.
- digit-mcp xlsx-loader.ts: the xlsx boundary-onboarding phase created boundary entities/relationships but never wrote localization messages, so deployments using this path never got translated boundary names or level labels for any level, not just the deepest.

Verified against a live docker-compose stack: seeded a real 4-level Maputo-style hierarchy, confirmed HierarchySchema highest/lowest values, confirmed localization lands in egov-localization, and confirmed in a live browser that the location dropdown now stops at the configured highest and lowest level with real translated labels.

## Summary by CodeRabbit

- **New Features**
  - Added automatic English localization for boundary names, codes, and hierarchy levels during boundary setup, and included created localization keys in the boundary phase output.

- **Bug Fixes**
  - Boundary hierarchy display now correctly honors both configured highest and lowest hierarchy caps.
  - Hierarchy setup now consistently treats the first level as the highest when syncing hierarchy schema updates.


digit-ui-esbuild/products/pgr/src/components/BoundaryComponent.js
```diff
@@ -102,27 +102,40 @@ const BoundaryComponent = ({ t, config, onSelect, userType, formData, readOnly }
...
+  // happened to return as root, regardless of the configured highest level
+  // (egovernments/CCRS#721).
   // `lowestLevelCapped` records whether PGR_BOUNDARY_LOWEST_LEVEL was BOTH
   // configured AND matched in this tree (the cap actually applied). The
   // childless-node-is-leaf fallback keys off it: only a deployment that
   // declared a lowest level treats a childless mid-tree node as fileable.
   // Unconfigured deployments keep strict deepest-level-only leaf semantics so a
   // County seeded without children can't become fileable (egovernments/CCRS#478).
   const { effectiveHierarchy, lowestLevelCapped } = useMemo(() => {
+    const configuredHighest = window?.globalConfigs?.getConfig?.("PGR_BOUNDARY_HIGHEST_LEVEL");
```
vinothrallapalli-eGov: @Lokendra-egov This should first be read from the **Hierarchy-Schema** master, with the global configuration used as a fallback if no value is defined there.

Also, consider a scenario where the hierarchy is configured with three boundary levels, but for some boundaries only two levels are available and the third level does not exist. The implementation should handle such cases gracefully without assuming that all configured levels are always present.

digit-ui-esbuild/products/pgr/src/components/BoundaryComponent.js
```diff
@@ -102,27 +102,40 @@ const BoundaryComponent = ({ t, config, onSelect, userType, formData, readOnly }
   }, [tenantId]);
   const hierarchyType = window?.globalConfigs?.getConfig("HIERARCHY_TYPE") || "ADMIN";
```
vinothrallapalli-eGov: @Lokendra-egov The hierarchy type should read from hierarchy schema master

vinothrallapalli-eGov (changes requested):

digit-ui-esbuild/products/pgr/src/components/BoundaryComponent.js
Lokendra-egov: Done

digit-ui-esbuild/products/pgr/src/components/BoundaryComponent.js
Lokendra-egov: Fixed

vinothrallapalli-eGov:
@Lokendra-egov Please resolve the merge conflicts

vinothrallapalli-eGov (approved):

PR #1389, opened 2026-07-22
feat(theme): add Bomet Blue and Maputo Green as OOTB themes (CCRS#1035)

Lokendra-egov:
Adds two new out-of-the-box UI themes selectable per tenant via the existing MDMS ThemeConfig mechanism (isActive record wins, no theme-picker plumbing needed):

- Bomet Blue formalizes Bomet's real, already-live production theme colors as a committed preset (previously undocumented anywhere in the repo, only known via a test-fixture comment).
- Maputo Green is a fresh Mozambique-flag-inspired palette (deep green, gold, red), since Maputo has no existing theme to match.

Neither Bomet nor Maputo's deploy had the common-masters.ThemeConfig MDMS schema registered at all (it only existed in the Kenya-only nairobi-mdms submodule), so this also adds the schema to default-data-handler's generic common-masters seed and a small opt-in ansible task (seed_theme_preset: bomet-blue / maputo-green) that seeds it at both the state root and city tenant, mirroring how the reference kenya-green theme is seeded today.

Verified live against a running docker-compose stack: schema-validated both presets, unit-tested applyTheme() against them, seeded via the new script (including idempotent re-run and preset-swap deactivation), and confirmed in a real browser that both citizen and employee UI pick up the exact seeded colors.

## Summary by CodeRabbit

- **New Features**
  - Added Bomet Blue and Maputo Green out-of-the-box theme presets.
  - Added configurable theme seeding during deployment, supporting both root and city tenants.
  - Theme seeding is idempotent and ensures only the selected theme remains active.

- **Documentation**
  - Added setup instructions and example configuration for enabling theme presets.

- **Tests**
  - Added coverage validating preset availability and expected theme color properties.


vinothrallapalli-eGov:
@Lokendra-egov Please resolve the code rabbit comments

digit-ui-esbuild/src/theme/presets/bomet-blue.json
```diff
@@ -0,0 +1,116 @@
+{
+  "version": "3",
+  "code": "bomet-blue",
+  "name": "Bomet Blue",
+  "colors": {
+    "grey": { "bg": "#E6E6E6", "light": "#FAFAFA", "lighter": "#F2F2F2", "disabled": "#C5C5C5" },
+    "link": { "hover": "#1565A8", "normal": "#1B85D2" },
```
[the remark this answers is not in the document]
Lokendra-egov: fixed

KDwevedi (commented):
We are adding themes as hardcoded presets in the FE code, which is largely parallel to the configurator based theming path. Is this deliberate?
Seems like we're creating two parallel sources of truth for theming here, one via configurator one basis presents in FE.

Lokendra-egov:
@KDwevedi re: "adding themes as hardcoded presets in the FE code... two parallel sources of truth" —

Not a duplicated runtime engine: `applyTheme()` is called from exactly two places in `digit-ui-esbuild/src/index.js` — once synchronously on boot with the bundled `default.json` (a placeholder before any network call resolves), and once later via `window.Digit.applyTheme()` in `StoreService.digitInitData()` with whatever MDMS returns for that tenant's `common-masters.ThemeConfig`. The new `digit-ui-esbuild/src/theme/presets/{bomet-blue,maputo-green}.json` files are **not** imported by the running app at all — only `applyTheme.test.js` (schema/render validation) and the new `seed-theme.py` read them. `seed-theme.py`'s own docstring says it directly: "Single source of truth: reads the SAME committed JSON... for the schema and the data." This is the same seed-JSON-into-MDMS pattern default-data-handler already uses for every other reference master (departments, designations, etc.) — not a new parallel system for theming specifically.

That said, digging into this turned up a real, more precise version of the concern: `seed_theme_preset` is a *persistent* host_vars setting, and `seed-theme.py` unconditionally re-asserts the named preset as active every time it runs. Since the ansible task's only gate was `when: seed_theme_preset is defined` (no "first bootstrap only" check), **every subsequent routine `./deploy.sh <tenant>` run — for any unrelated reason — would silently revert any theme change made afterward through the Configurator** (a color tweak, or switching to a bespoke non-preset theme). That's a genuine two-sources-of-truth conflict, just at the redeploy layer rather than the runtime layer.

Fixed in 315ed881: added a marker file (`theme-seed/.seeded-preset`) on the target recording which preset was last seeded by this mechanism. The seed now only (re)runs on first bootstrap, on a deliberate change to `seed_theme_preset` itself, or when `seed_theme_preset_force: true` is explicitly set — never on a routine redeploy where nothing about the theme config changed. README updated with the new behavior.

subhashini-egov:
@Lokendra-egov @KDwevedi - I didn't check the files themselves. Just saw the comments. So broadly, we should have a single source of truth which is MDMS. And we seed the MDMS data with out of the box presets. We shouldn't be adding this anywhere else - hardcoded or otherwise.

configurator/src/admin/themeEditor/ThemeConfigCreate.tsx
```diff
@@ -0,0 +1,19 @@
...
+/** Same tabbed/live-preview editor Edit already uses (ThemeConfigFormBody),
+ *  just mounted inside DigitCreate instead of DigitEdit — there's no schema
+ *  reason a new theme needs a different (worse) form than an existing one.
+ *  `version: '3'` defaults every new theme onto the current designer
+ *  taxonomy rather than a stale v1/v2 shape. */
+export function ThemeConfigCreate() {
+  const resource = useResourceContext() ?? '';
+  const label = getResourceLabel(resource);
+  return (
+    <DigitCreate title={`Create ${label}`} record={{ version: '3', colors: {} }}>
```
KDwevedi: Creating through this component is unsafe when the tenant already has a theme. `DigitCreate` ultimately calls `mdmsCreate`, which hard-codes `isActive: true`, while this path does not deactivate the previous active ThemeConfig. The runtime then applies `ThemeConfig[0]`, so two active rows make the effective theme nondeterministic and the newly created theme may not appear. Please create inactive until explicit activation, or atomically activate the new row and deactivate the old one.

local-setup/ansible/README.md
```diff
@@ -521,6 +521,36 @@ enable_search_stack: true   # ~2 GB RAM extra
 The playbook resolves this into `docker compose --profile search …`
 flags. Setting back to `false` sweeps the running search containers.
 
+### Set the OOTB theme (issue #1035)
+
+MDMS (`common-masters.ThemeConfig`) is the single source of truth for theme
+colors — full stop. There is no committed preset file anywhere in this repo
```
KDwevedi: Blocking: this removes the actual presets and their seed/selection mechanism. CCRS#1035 asks for Bomet Blue and Maputo Green as named OOTB themes selectable per tenant. A blank manual ThemeConfig form requires operators to reconstruct the palette and provides no way to select either named preset, so a fresh Bomet/Maputo deployment still cannot obtain the requested themes. Please keep MDMS as the runtime source of truth while shipping seedable preset data into MDMS.

configurator/src/admin/themeEditor/ThemeConfigCreate.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { describe, it, expect, beforeAll } from 'vitest';
+import { render, screen } from '@testing-library/react';
+import { CoreAdminContext, ResourceContextProvider, TestMemoryRouter, type DataProvider } from 'ra-core';
+import { QueryClient } from '@tanstack/react-query';
+import { MdmsResourceCreate } from '../MdmsResourceCreate';
+
+function makeDataProvider(): DataProvider {
+  return {
+    getList: async () => ({ data: [], total: 0 }),
+    getOne: async (_resource, params) => ({ data: { id: params.id } }),
```
KDwevedi: This new test breaks the Configurator production build. After building the local data-provider package, `npm run build` fails with TS7006 for `_resource` and `params` here and in the create/update/delete callbacks below (eight errors total). Because the object is only cast after construction, these callbacks receive no contextual `DataProvider` typing. Please type the mock methods or contextually type the provider object.

utilities/default-data-handler/src/main/resources/schema/common-masters.json
```diff
@@ -708,5 +708,37 @@
...
+          "type": "string"
+        },
+        "name": {
+          "type": "string"
+        },
+        "version": {
+          "type": "string"
+        },
+        "colors": {
+          "type": "object"
```
KDwevedi: This does not enforce the runtime theme contract despite the description above. MDMS accepts any `colors` object and requires only `code`; the editor also accepts `#RGB` / `#RRGGBBAA` without blocking submit, while `digit-ui-esbuild/src/theme/schema.json` permits only six-digit hex values and `applyTheme()` rejects the entire record when validation fails. A theme created through the newly documented manual-only path can therefore save successfully but do nothing at runtime. Please align the MDMS/form validation with the runtime schema.

Lokendra-egov:
@KDwevedi  Addressed all four review comments:

1. Create now deactivates any other active theme for the tenant right after a successful save, so it can't leave two active rows around.
2. Fixed the TS7006 errors in ThemeConfigCreate.test.tsx that were breaking the production build.
3. ColorInput blocks submit on invalid hex now (3 or 8-digit), and the MDMS ThemeConfig schema enforces the same 6-digit hex pattern that the runtime schema.json expects.
4. Restored the Bomet Blue and Maputo Green preset files and seed-theme.py that got removed in the earlier refactor commits. MDMS stays the only runtime source of truth; these are just ops data used to seed it.

Tested all of this live on a real local stack, not just unit tests: brought up the full docker compose stack, ran seed-theme.py against real MDMS (had to start egov-persister first; it wasn't running, and updates were silently not persisting), and clicked through Theme Config Create in the browser to confirm the dedup behavior and the hex validation.

One thing to flag before this goes out: MDMS schema update returns 501 in this version, so schemas can't be upgraded in place. A tenant that already has an older ThemeConfig schema registered won't automatically pick up the new hex validation; only a fresh schema registration will. It's worth checking whether the target tenant already has this schema before assuming the new validation is active there.

local-setup/scripts/seed-theme.py
```diff
@@ -0,0 +1,232 @@
...
+LOGIN_TENANT = os.environ.get("DIGIT_LOGIN_TENANT", TENANT)
+_here = os.path.dirname(os.path.abspath(__file__))
+SCHEMA_FILE = os.environ.get("SCHEMA_FILE", os.path.join(_here, "theme-seed", "common-masters.json"))
+DATA_FILE = os.environ.get("DATA_FILE", os.path.join(_here, "theme-seed", "bomet-blue.json"))
+BASIC = "Basic ZWdvdi11c2VyLWNsaWVudDo="  # egov-user-client: (empty secret)
+
+THEME_CODE = "common-masters.ThemeConfig"
+
+
+def _post(path, body, tok=None, headers=None):
```
subhashini-egov: **`tok` parameter in `_post` is accepted but silently unused — misleads callers.**

The function signature takes `tok=None`, but the body never uses it. DIGIT's auth pattern is correct here (the token goes in `RequestInfo.authToken` in the JSON body, not an HTTP header), but the dead parameter suggests to every future caller that passing a token does something:

```python
def _post(path, body, tok=None, headers=None):  # tok accepted but ignored
```

All six call sites at the bottom of the file pass `tok` as a positional arg, adding noise. Either remove the parameter entirely and document why auth is body-only, or use it to set an `Authorization: Bearer` header for a real defense-in-depth (Kong also accepts bearer tokens on most routes). At minimum, drop it from the signature so callers don't pass a no-op argument:

```python
def _post(path, body, headers=None):  # auth is in RequestInfo.authToken in body
```

local-setup/scripts/seed-theme.py
```diff
@@ -0,0 +1,232 @@
...
+            return "dup"
+        print("    ! %s row FAILED (%s): HTTP %s %s" % (code, row_code, e.code, blob))
+        return "failed"
+
+
+def main():
+    print("seed-theme: tenant=%s url=%s" % (TENANT, URL))
+    tok = token()
+
+    schemas = {s.get("code"): s for s in json.load(open(SCHEMA_FILE))}
```
subhashini-egov: **`open()` without `with` and without `encoding=` — resource leak and locale-sensitive.**

Two calls:
```python
schemas = {s.get("code"): s for s in json.load(open(SCHEMA_FILE))}
preset = json.load(open(DATA_FILE))   # line 170
```

Both open a file handle that is never explicitly closed (the dict/object comprehension holds a reference to the json reader while iterating, then the handle becomes garbage-collectable but is not deterministically closed). On CPython this is usually fine, but it's a resource leak that also prevents the temp file from being deleted on Windows if the GC hasn't run yet.

More concretely: without `encoding='utf-8'`, `open()` uses `locale.getpreferredencoding(False)`, which can be `cp1252` on some Windows systems or a non-UTF-8 locale on minimal Linux installs. The preset JSON files contain non-ASCII characters (e.g. the schema description's em-dash).

Fix:
```python
with open(SCHEMA_FILE, encoding='utf-8') as f:
    schemas = {s.get('code'): s for s in json.load(f)}
# ...
with open(DATA_FILE, encoding='utf-8') as f:
    preset = json.load(f)
```

local-setup/ansible/playbook-deploy.yml
```diff
@@ -4302,6 +4302,142 @@
...
+      # state_root — state_root is a compose boot-pin that intentionally
+      # diverges from the logical tenant tree for MCP-onboarded countries
+      # (e.g. Maputo: state_root=pg, state_tenant_id=mz — the theme belongs
+      # under mz, not the pg boot pin).
+      when: seed_theme_preset is defined and theme_seed_should_run | bool
+      tags: ['themes', 'theme-seed']
+      ansible.builtin.set_fact:
+        theme_seed_root_tenant: "{{ state_tenant_id | regex_replace('\\..*$', '') }}"
+        theme_seed_city_tenant: "{{ ui_state_tenant_id | default(city_tenant | default(state_tenant_id ~ '.citya')) }}"
+
```
subhashini-egov, line 4398: **`.citya` fallback is a fabricated tenant that doesn't exist — the city seed step will always fail if both `ui_state_tenant_id` and `city_tenant` are undefined.**

```yaml
theme_seed_city_tenant: "{{ ui_state_tenant_id | default(city_tenant | default(state_tenant_id ~ '.citya')) }}"
```

If neither `ui_state_tenant_id` nor `city_tenant` is defined in `host_vars`, this resolves to e.g. `mz.citya` or `ke.citya` — a tenant that has never been bootstrapped. The city seed step then runs `seed-theme.py` at that tenant, fails with a login or MDMS error, and the `failed_when` condition (`rc != 0 or 'DONE' not in stdout`) propagates a misleading failure for every deploy that doesn't define those vars.

Better options:
1. **Default to the root tenant** — seeding twice at the same tenant is idempotent (the second run hits the duplicate path and skips), so it's safe: `default(theme_seed_root_tenant)`
2. **Skip the city step when the var is absent** — add `when: ui_state_tenant_id is defined or city_tenant is defined` to the city seed tasks
3. **Require at least one** — add a `assert` task before the seed that fails with a clear message if both are undefined

configurator/src/admin/widgets/ColorInput.tsx
```diff
@@ -12,10 +24,34 @@ interface ColorInputProps extends InputProps {
...
+  const { id, field, fieldState, isRequired } = useInput({ ...inputProps, validate: combinedValidate });
+  // `isTouched` only flips on blur, so gating on it alone leaves the field
+  // showing no error while the user is still typing an invalid value.
+  // `isDirty` flips on the first change, giving immediate feedback (mirrors
+  // DigitFormInput's convention).
+  const hasError = fieldState.invalid && (fieldState.isDirty || fieldState.isTouched);
+  // ra-core v5 wraps validator errors as `@@react-admin@@${JSON.stringify(msg)}`
+  // before storing them in react-hook-form state.
+  const rawError = fieldState.error?.message;
+  const errorMessage = rawError?.startsWith('@@react-admin@@')
```
subhashini-egov: **Magic number `15` in `rawError.slice(15)` will silently drift if the ra-core prefix ever changes.**

```typescript
const errorMessage = rawError?.startsWith('@@react-admin@@')
  ? (() => { ... JSON.parse(rawError.slice(15)) ... })()
  : rawError;
```

`'@@react-admin@@'.length` is 15, so this is currently correct, but the hardcoded literal creates a gap: if ra-core ever changes the prefix (e.g. to `@@ra@@` or adds a version indicator), `startsWith` catches the change correctly but `slice(15)` parses from the wrong offset and surfaces a garbled error message with no runtime signal that something is wrong.

Use the prefix length directly:
```typescript
const RA_PREFIX = '@@react-admin@@';
const errorMessage = rawError?.startsWith(RA_PREFIX)
  ? (() => { ... JSON.parse(rawError.slice(RA_PREFIX.length)) ... })()
  : rawError;
```

KDwevedi (commented):
@Lokendra-egov thanks for addressing the earlier round — my 4 comments above are now stale against the current diff and I'm marking them resolved.

Holding off on approval though: Subhashini left 4 fresh comments on this diff today (`seed-theme.py` unused `tok` param, `open()` without `with`/`encoding=`, the `.citya` fabricated-tenant fallback in `playbook-deploy.yml:4393`, and the magic-number `slice(15)` in `ColorInput.tsx`) that are still unresolved and don't look addressed in the latest commit. Once those land I'm happy to approve — none of them look architecturally contentious, just fix-and-push.

local-setup/scripts/seed-theme.py
Lokendra-egov: Removed in 78598497 — took the "drop it from the signature" option rather than wiring up a bearer header, since the body-only auth is already correct and an extra header would be a behaviour change to every route for no functional gain.

```python
def _post(path, body, headers=None):
    # No auth argument by design: DIGIT carries the token in the request BODY as
    # RequestInfo.authToken (see ri()), not in an Authorization header, so every
    # caller has already embedded it via ri(tok) before getting here.
```

All six call sites updated to stop passing the no-op argument. `tok` is still a real parameter on the five functions that build a body with `ri(tok)`, so nothing else changed — I re-ran an AST scan over the file afterwards to confirm no unused parameters remain.

local-setup/scripts/seed-theme.py
Lokendra-egov: Fixed in 78598497 — both reads now use `with` and an explicit encoding:

```python
with open(SCHEMA_FILE, encoding="utf-8") as f:
    schemas = {s.get("code"): s for s in json.load(f)}
...
with open(DATA_FILE, encoding="utf-8") as f:
    preset = json.load(f)
```

The locale point is the one that would actually have bitten: the preset files carry non-ASCII, so on a `cp1252` or non-UTF-8 default this would have thrown `UnicodeDecodeError` at parse time rather than degrading quietly.

local-setup/ansible/playbook-deploy.yml
Lokendra-egov: Good catch — fixed in 78598497 with your option 2 (skip), not option 1 (default to root).

Option 1 works and is safe, but it would report a city seed that never happened: `theme_seed_city_tenant == theme_seed_root_tenant` makes the second run a no-op that still prints as a city-scope seed, and the README would keep claiming both scopes were covered. Skipping is the honest outcome — with neither var set there is no city tenant to seed.

```yaml
# No fabricated fallback here. A guessed `<root>.citya` names a tenant
# that was never bootstrapped, so the city seed would fail login/MDMS
# and — via failed_when — take the whole deploy down on every host that
# sets neither var. Empty means "no city tenant configured", and the
# city seed step is skipped instead of inventing a target.
theme_seed_city_tenant: "{{ ui_state_tenant_id | default('', true) or city_tenant | default('', true) }}"
```

The city task is gated on `theme_seed_city_tenant | length > 0`, and a new debug task fires in the skip case so the operator isn't left wondering why only the root was seeded:

> theme-seed: neither ui_state_tenant_id nor city_tenant is set, so '<preset>' was seeded at <root> only. Set one of those vars if a city-scoped ThemeConfig is wanted.

Note the `| default('', true)` form is deliberate — plain `default()` would not substitute when `ui_state_tenant_id` is defined but empty, which the resolver at ~L4037 can produce.

`README.md` L519-532 updated so the two-tenant claim is conditional on a city tenant actually being configured, rather than unconditional.

configurator/src/admin/widgets/ColorInput.tsx
Lokendra-egov: Fixed in 78598497:

```typescript
const RA_ERROR_PREFIX = '@@react-admin@@';
...
const errorMessage = rawError?.startsWith(RA_ERROR_PREFIX)
  ? (() => {
      const encoded = rawError.slice(RA_ERROR_PREFIX.length);
      ...
      } catch { return encoded; }
    })()
  : rawError;
```

Hoisted the slice into `encoded` so the fallback path can't drift from the parse path either — previously both branches repeated the literal independently.

One thing worth flagging separately: `configurator/src/admin/DigitFormInput.tsx:53` has a byte-identical copy of this block, `slice(15)` included. It's outside this PR's diff so I've left it alone here to keep the change scoped to the review, but the same drift risk lives there and it's worth a follow-up.

local-setup/ansible/files/theme-presets/maputo-green.json
```diff
@@ -0,0 +1,116 @@
...
+    "info": "#234E70",
+    "info-dark": "#1E425F",
+    "warning": "#9E5F00",
+    "warning-dark": "#9E5F00",
+    "brand": "#FFCE00",
+    "brand-on": "#046A38",
+    "surface-header": "#046A38",
+    "surface-page": "#FAFAFA",
+    "selected-bg": "#FFF9E0",
+    "primary-1": "#046A38",
```
subhashini-egov, line 38: **`primary-1` (#046A38) doesn't match `primary.dark` (#035C2F) — unlike Bomet Blue, where both are `#1565A8`.**

Bomet Blue's preset is internally consistent: `primary-1` and `primary.dark` are the same hex value, so components on the v1/v2 path (reading `colors.primary.dark`) and those on the v3 path (reading `colors["primary-1"]`) see identical colors.

In Maputo Green:
- `colors.primary.dark` = `#035C2F` (line 9 of this file — a slightly deeper forest green)
- `colors["primary-1"]` = `#046A38` (this line — the brighter brand green)

These differ by about 12 luminance points. Header and sidebar background (`#046A38`) will match `primary-1` but not `primary.dark`. If any existing component reads the nested path on a v3 record, it will render a noticeably different shade from every v3-aware sibling in the same layout.

If the intent is a single authoritative dark green for this preset, align them. If `primary.dark` is deliberately a deeper accent distinct from the brand green, a comment here would prevent it from being treated as an accidental inconsistency during future edits.

local-setup/scripts/seed-theme.py
```diff
@@ -0,0 +1,237 @@
...
+    # caller has already embedded it via ri(tok) before getting here.
+    data = json.dumps(body).encode()
+    h = {"Content-Type": "application/json"}
+    if headers:
+        h.update(headers)
+    req = urllib.request.Request(URL + path, data=data, headers=h)
+    return urllib.request.urlopen(req, timeout=40)
+
+
+def token():
```
subhashini-egov, line 68: **`urllib.error.URLError` and `socket.timeout` are not caught — a network failure produces an unhandled Python traceback.**

`token()` calls `urlopen()` directly. If Kong is not yet up, if DNS fails, or if the 40-second timeout fires, Python raises `urllib.error.URLError` (wrapping `socket.timeout`) — not `urllib.error.HTTPError`. The downstream helpers (`schema_exists`, `search_rows`, `_set_active`, `create_row`) each catch `HTTPError` but nothing catches `URLError`.

In the Ansible context this still fails the task (rc ≠ 0, caught by `failed_when`), but the operator sees a raw Python traceback in the task output rather than a meaningful message. A minimal wrapper in `main()` makes the failure mode readable:

```python
try:
    tok = token()
except (urllib.error.URLError, OSError) as e:
    sys.exit("ERROR: could not reach DIGIT at %s: %s" % (URL, e))
```

The same guard applies to `schema_exists` — a `URLError` there also propagates unhandled. `OSError` covers `socket.timeout` on Python < 3.3 (not a concern here, but costs nothing to include).

configurator/src/admin/themeEditor/ThemeConfigCreate.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const deactivateSiblings = async (created: RaRecord) => {
+    const newCode = (created as { code?: unknown }).code;
+    const rows = await digitClient.mdmsSearch(tenantId, SCHEMA, { limit: 200 });
+    const stale = rows.filter(
+      (r) =>
+        r.isActive &&
+        r.tenantId === tenantId &&
+        (r.data as { code?: unknown } | undefined)?.code !== newCode,
+    );
+    for (const row of stale) {
```
subhashini-egov, line 41: **Sequential deactivation loop; `Promise.all` would run the updates in parallel.**

```typescript
for (const row of stale) {
  await digitClient.mdmsUpdate(row, false);
}
```

Each `await` serialises the next network round-trip. A tenant that has accumulated N previously-seeded theme rows takes N × RTT before the create callback returns and the form navigates away. The updates are independent, so they can run concurrently:

```typescript
await Promise.all(
  stale.map((row) =>
    digitClient.mdmsUpdate(row, false).catch((e) => {
      const staleCode = (row.data as { code?: unknown } | undefined)?.code ?? row.uniqueIdentifier;
      toast({
        title: 'Could not deactivate previous theme',
        description: `${staleCode}: ${e instanceof Error ? e.message : String(e)}`,
        variant: 'destructive',
      });
    }),
  ),
);
```

In practice N is almost always 1, so this is a correctness-of-abstraction point rather than an observed perf problem — but the parallel form is also cleaner because the toast-on-error logic stays inside the `map` and the outer function needs no `try/catch`.

subhashini-egov (commented):
## Round 2

All four comments from the previous round are addressed — the `RA_ERROR_PREFIX` constant, `with open(…, encoding='utf-8')`, the `_post` signature cleanup with the body-auth note, and the Ansible `.citya` fallback are all cleanly fixed. Three new issues below.

subhashini-egov (approved):

Issue #1380, opened 2026-07-22
[Bug]  The boundary dropdown is not being auto-populated using the selected map.


pradeepkumarcm-egov:
# Description### 

Employee Portal

1. Log in as an **Employee** and select the boundary using the **Map**.
2. The **Boundary** dropdown is **not auto-populated** based on the selected location on the map.

Citizen Portal

1. Log in as a **Citizen**.
2. The boundary is **defaulting to Bomet County**, irrespective of the user's selected location.
3. The maps available in the **Employee Portal** are **not visible** in the **Citizen Portal**.
4. Upon selecting a location, the application continuously invokes APIs, causing the map to repeatedly reposition and shift from one location to another.

 

---
 

# Screenshots / Logs

<img width="1470" height="956" alt="Image" src="https://github.com/user-attachments/assets/bccac6c3-4992-4bd9-9c40-b9135c5fd38d" />

<img width="1470" height="956" alt="Image" src="https://github.com/user-attachments/assets/eb89ee75-23d2-45dc-98ed-08851354e0b3" />

https://github.com/user-attachments/assets/b4fba44f-2349-4cd5-a1a4-0f4f230540ce

KDwevedi:
## Investigation results

Read-only diagnosis across both environments in this report. Summary: **this issue contains at least four independent defects**, in two different environments, and they need splitting. One is already fixed, one has a PR up, two need decisions.

---

### 0. The two halves of this report are from different environments

The Employee screenshot is `cms-pilot.digit.org/digit-ui/employee/pgr/create-complaint` (the Mozambique pilot). The Citizen screenshot is **`bometfeedbackhub.digit.org/digit-ui/citizen/...`** — a different deployment. The Citizen findings therefore do not all describe the Mozambique pilot, which is why "defaulting to Bomet County" looked inexplicable there. Worth splitting before assigning.

---

### 1. Employee: boundary dropdown not auto-populated — CONFIRMED, PR up

Root cause is a level mismatch, not a broken control.

`useTenantBoundaries.js` resolves a dropped pin against the **deepest level present in the boundary tree**, ignoring `PGR_BOUNDARY_LOWEST_LEVEL`:

```js
const maxDepth = nodes.reduce((max, n) => (n.depth > max ? n.depth : max), -1);
const leaves = nodes.filter((n) => n.depth === maxDepth);
```

[...]

PR #1392, opened 2026-07-22
fix(pgr): resolve complaint-map pins at the configured boundary level

KDwevedi:
## Problem

The complaint map resolves a dropped pin against the **deepest level present in the boundary tree**, ignoring `PGR_BOUNDARY_LOWEST_LEVEL` — the level the boundary cascade actually files at.

`useTenantBoundaries.js`:
```js
const maxDepth = nodes.reduce((max, n) => (n.depth > max ? n.depth : max), -1);
const leaves = nodes.filter((n) => n.depth === maxDepth);
```

Two consequences:

**1. The map and the cascade can disagree.** The map resolves to the deepest level and writes it to `GeoLocationsPoint.ward`; `BoundaryComponent.findWardPath` then searches for a node of that same deeper type, because `targetType` was computed from the **uncapped** `boundaryHierarchy` rather than `effectiveHierarchy`. On a tenant that seeds the deepest level only patchily, the lookup finds nothing and the boundary dropdowns silently stay empty after dropping a pin.

**2. Deeper is not better for coverage.** Mozambique seeds `Municipio` under only **66 of 158** `Distrito`s, so resolving at `Municipio` leaves every pin outside a municipality unresolvable — while `Distrito` tiles the whole territory. The operator already declares the intended level; the map just wasn't reading it.

Reported as part of egovernments/Citizen-Complaint-Resolution-System#1380 (boundary dropdown not auto-populated from the map).

## Change

- `useTenantBoundaries` selects the resolution level from `PGR_BOUNDARY_LOWEST_LEVEL` when that level exists in the tree.
- `BoundaryComponent` computes `targetType` from `effectiveHierarchy` (the capped one), so map and cascade target the same level.
- New `utils/boundaryLevels.js` — accent-insensitive level-name comparison, shared by both.

## Why accent-insensitive

Operators author `PGR_BOUNDARY_LOWEST_LEVEL` in `host_vars` with the tenant's own orthography, while the loader writes the unaccented form into the tree. On the Mozambique pilot, byte-for-byte:

| source | value | length |
|---|---|---|
| `globalConfigs.js` | `Município` | 10 |
| `boundary_relationship.boundarytype` | `Municipio` (hex `4d756e69636970696f`) | 9 |

The existing `String(k).toLowerCase() === String(configuredLowest).toLowerCase()` misses — and the miss is **invisible**: the cap simply never applies, with no error anywhere. Both sides read correct to a human.

## Backwards compatibility

Unchanged when the key is unset, or when it names a level absent from the tree. Both fall back to the previous deepest-level selection — the second with a `console.warn` naming the levels the tree *does* have, so the fallback is not silent.

## Verification

Parse/bundle-checked all three files with the repo's own esbuild.

Level matcher unit-tested (9 cases, all pass): accented vs plain, case-only, whitespace, genuinely-different levels, and empty/null which must **not** match.

Selection logic replayed against the **live `mz.ige` tree** (235 boundaries, fetched from `boundary-relationships/_search`):

| `PGR_BOUNDARY_LOWEST_LEVEL` | selected | renderable geometry | path |
|---|---|---|---|
| unset | 66 | 56 | fallback: deepest depth (**identical to today**) |
| `Municipio` | 66 | 56 | configured level |
| `Município` | 66 | 56 | configured level (**was a miss before**) |
| `Distrito` | 158 | 157 | configured level |
| `Provincia` | 11 | 0 | configured level |
| `Bairro` (absent) | 66 | 56 | fallback + warning |

## Not in scope

Two related findings from the same investigation, deliberately left out:

- **The Mozambique boundary geometry is largely placeholder** — 99 of 235 boundaries share one identical 1°×1° square at Null Island, and no `Municipio` has geometry of its own (each is a copy of its parent `Distrito`). This PR lets an operator pick a level that *has* data; it cannot create data. Note `hasRealGeometry` does not reject that square (it only rejects spans < ~0.001°), so it renders and participates in point-in-polygon.
- **The citizen map draws no overlay while the employee map does**, because the map reads MapConfig from the *session* tenant (a state root, which carries no boundaries) rather than the tenant the complaint files under. The fix needs the citizen authority-picker flow, which does not exist on `master`.

Both are being raised separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved boundary-level matching across differences in capitalization, spacing, and diacritical marks.
  * Fixed map-based location autofill to resolve ward paths using the correct configured hierarchy level.
  * Added fallback behavior when the configured lowest boundary level is unavailable, preventing boundary selection failures.

### DIFF
--- a/digit-ui-esbuild/products/pgr/src/components/BoundaryComponent.js
+++ b/digit-ui-esbuild/products/pgr/src/components/BoundaryComponent.js
@@ -2,6 +2,7 @@ import { Loader } from "@egovernments/digit-ui-components";
 import { Field as V2Field, Select as V2Select } from "@egovernments/digit-ui-components-v2";
 import React, { useEffect, useMemo, useState } from "react";
 import { useTranslation } from "react-i18next";
+import { sameLevelName } from "../utils/boundaryLevels";
 
 // Humanize a boundary-type code for use as a graceful fallback when its
 // localization key isn't seeded: "SUB_COUNTY" -> "Sub County", "bairro" ->
@@ -117,9 +118,10 @@ const BoundaryComponent = ({ t, config, onSelect, userType, formData, readOnly }
   const { effectiveHierarchy, lowestLevelCapped } = useMemo(() => {
     const configuredLowest = window?.globalConfigs?.getConfig?.("PGR_BOUNDARY_LOWEST_LEVEL");
     if (!configuredLowest) return { effectiveHierarchy: boundaryHierarchy, lowestLevelCapped: false };
-    const idx = boundaryHierarchy.findIndex(
-      (k) => String(k).toLowerCase() === String(configuredLowest).toLowerCase()
-    );
+    // Accent-insensitive: operators author this in host_vars with the tenant's
+    // own orthography, which need not match how the loader spelled the level in
+    // the tree. A miss here is silent — the cap just never applies.
+    const idx = boundaryHierarchy.findIndex((k) => sameLevelName(k, configuredLowest));
     return idx >= 0
       ? { effectiveHierarchy: boundaryHierarchy.slice(0, idx + 1), lowestLevelCapped: true }
       : { effectiveHierarchy: boundaryHierarchy, lowestLevelCapped: false };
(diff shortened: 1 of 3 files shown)


vinothrallapalli-eGov (approved):

KDwevedi:
## Added: citizen boundary cascade scoped to the filing tenant (74d0ab7)

I originally left the citizen-side tenant fix out of this PR, on the grounds that the fix needed the release branch's authority picker. **That was wrong** — there is a distinct and more basic version of the bug on `master`, and the fix belongs here.

### The bug on master

`BoundaryComponent` resolved its tree from `ULBService.getCurrentTenantId()`:

```js
const tenantId =
  user?.info?.type === "EMPLOYEE" && user?.info?.tenantId
    ? user?.info?.tenantId
    : window?.globalConfigs.getConfig("STATE_LEVEL_TENANT_ID");
```

That returns the user's own tenant **only for `EMPLOYEE`**. Every citizen falls through to `STATE_LEVEL_TENANT_ID` — the state root — no matter which city they are filing in.

Meanwhile the citizen create flow already computes the filing tenant:

```js
const tenantId =
  Digit.SessionStorage.get("CITIZEN.COMMON.HOME.CITY")?.code ||
  Digit.ULBService.getCurrentTenantId();
```

and uses it for `useCreateComplaint(tenantId)` and the create payload. `useMapConfig` resolves the map from that **same expression**. The boundary cascade was the only one of the three not using it.

`CITIZEN.COMMON.HOME.CITY` is set on Keycloak citizen SSO login and by the location picker, so the divergence is reachable, not theoretical.

Two consequences:

- A state root that carries no boundary tree of its own renders an **empty cascade**, while an employee — whose session tenant *is* the city — sees boundaries fine. That is precisely the citizen/employee asymmetry reported in #1380.
- Where the root *does* carry a tree, the citizen picks a boundary from one tenant's tree and the complaint is created in another.

### The change

Pass the filing tenant explicitly rather than re-deriving it from session identity — `StepBodyProps` carries `tenantId`, and `Step2Location` hands it to the cascade through the `config.tenantId` prop the component now honours.

### Blast radius

- **Fallback unchanged.** `config?.tenantId || Digit.ULBService.getCurrentTenantId()` — employees and every caller that doesn't pass the prop behave exactly as today.
- **The filing tenant is not changed.** Which tenant a complaint is created under is untouched, and `ComplaintsList` retrieves with the same expression the create path uses, so a filer's existing and future complaints remain visible exactly as before.
- **One data note for reviewers:** on a deployment where the state root and the city tree use *different* boundary codes, complaints filed before this change carry root-tree codes while new ones carry city-tree codes. Anything keyed on boundary code (inbox filters, ward dashboards, escalation routing) would see both vintages. This PR does not migrate that, and shouldn't — flagging it so it is a decision rather than a surprise.

Both files parse+bundle clean under the repo's esbuild.

Issue #1166, opened 2026-07-13
moz: five serial unblocks needed to surface the supervisor dashboard (with data) for an employee


KDwevedi:
Getting the supervisor dashboard to actually appear — with data — for one employee (**EMP486**, city tenant **mz.igsae**) on the moz deployment required five unblocks **in series**. Each layer hid the next one. Recording them in order, since together they describe everything a deployment needs beyond "the code ships the dashboard".

Baseline that already worked before any of this: the dashboard module was in the served UI bundle, the analytics backend answered, and the analytics materialized views were populated and auto-refreshing. Everything below sat on top of that.

### 1. The KPI catalog didn't exist in MDMS
- **Symptom:** `catalog/_search` returned `{"tiles":[],"total":0}` for every principal; the dashboard route rendered an empty shell.
- **Cause:** no `dss.KpiDefinition` / `dss.DashboardPack` schemas or data had ever been seeded at the state tenant. The repo's dss schema files were also stale (their `viz.kind`/`viz.format` enums reject 26 of the 28 catalog entries).
- **Unblock:** registered the canonical dss schemas, seeded 28 KpiDefinition entries + 1 DashboardPack adapted from the reference deployment's seed (tenant re-pointed), with 3 small data fixes (a missing `viz.valueKey`, a null `titleKey`, a missing param default). RBAC audit as part of the seed: two roles referenced by the seed don't exist on this deployment (`PGR_SUPERVISOR`, `PGR_ADMIN`) — remapped `visibleTo` to the deployment's existing `CMS_SUPERVISOR`, dropped the redundant one, and added `CMS_CASE_MANAGER` to the pack. **No new roles were created.**

### 2. The dashboard route and home card are code-gated on a hardcoded role list
- **Symptom:** direct navigation to `/employee/dashboard` bounced back to the employee home; no card anywhere.
- **Cause:** `products/dashboard/roles.js` hardcodes `DASHBOARD_ROLES = [SUPERVISOR, PGR_SUPERVISOR, GRO, DGRO, PGR_LME, PGR_ADMIN, SUPERUSER]` and both the route and the card check it. This deployment's entire role taxonomy is `CMS_*`, so no real employee passes the gate. No MDMS/roleactions write can fix this.
- **Unblock:** granted the employee the *existing* `SUPERVISOR` role. **Follow-up worth its own issue:** the allowed-roles list should be configuration (MDMS/globalConfig), not code.

[...]

Issue #631, opened 2026-05-29
[Feature] CCRS SaaS: Dashboard


KDwevedi:
- Configurable (data widgets, resizable in a grid, some custom filters) dashboards for citizen, varying employee personas, admin persona tracking KPIs.
- To scope out User level vs Role level data customisations on dashboards basis KPIs.

---

## Progress

**Done**
- Configurable widget/grid dashboard with KPI tracking for employee/supervisor personas — built (demo live via #866, pending merge).
- Config-driven KPI analytics engine (V2 grains + dynamic query API) — merged & live.
- User-level vs role-level data scoping (RBAC) — **design complete and reviewed**: see the [design comment](https://github.com/egovernments/Citizen-Complaint-Resolution-System/issues/631#issuecomment-4775492466).

**Remaining**
- [ ] **RBAC & role/user-level data scoping — implementation:** tracked in sub-issue #908 (full design in the linked comment above; recommended first PR = Part A, identity/auth).
- [ ] Wire the V2 metric grains into the refresh scheduler so dashboards show fresh data (they don't auto-refresh today).

KDwevedi:
## Dashboard data architecture — scoping notes

Grounded in a real-source sweep of `pgr-services`, `egov-workflow-v2`, `egov-hrms`, the deployed compose, and the DSS stack. Full per-KPI accounting with `file:line` citations lives in [KDwevedi/Citizen-Complaint-Resolution-System#15](https://github.com/KDwevedi/Citizen-Complaint-Resolution-System/issues/15).

### 1. The shape of the problem: a three-gate funnel

Every KPI has to clear three gates, **in order**. It's deliverable only if all three pass:

1. **Recorded** — the underlying datapoint is actually written to a DB. *(write-path)* If it isn't, no amount of dashboarding conjures it.
2. **Aggregable** — a mechanism rolls raw rows up (count / group-by / percentile / delta) **without pulling every row to the client**. *(aggregation tier)*
3. **Fetchable & scoped** — an API exposes that aggregate, scoped to the viewer's context (tenant / boundary / role / user).

Where today's KPIs fail: a few fail gate 1 (never recorded — escalation source, acknowledgement timestamp, risk score); **most fail gate 2** (data is recorded, but the only aggregation surface that exists is Chakshu's `/v2/dashboard`); gate 3 scoping is exactly what "user-level vs role-level" is about.

[...]

PR #1400, opened 2026-07-22
feat(dashboard): enable-dashboard.sh installer + tenant_bootstrap gap warnings

KDwevedi:
## Why

Enabling the supervisor dashboard on a running deployment had **no script**. The runbook on #631 documented it, `tenant_bootstrap` covered only new state roots, and the 315-message localization packs existed solely as TypeScript constants in `digit-mcp/src/tools/dashboard-l10n-seed.ts` — reachable from bootstrap and from nowhere else. An operator could hand-copy the catalog JSON; they had no path at all to the strings.

This came out of asking why the dashboard is invisible on **moz** (#1166). Every config layer is missing there, and `tenant_bootstrap` cannot fix it: bootstrap *copies* `dss.*` from a source root, and moz's source (`pg`) has no `KpiDefinition` rows — so it copies nothing, reports success, and the root gets a working shell with an empty catalog.

## What

**`local-setup/scripts/enable-dashboard.sh`** — the installer form of the #631 runbook, modelled on `enable-notifications.sh`. Seven resumable idempotent steps: register the dss schemas → seed 39 defs + pack (roles remapped) → `dss.DashboardConfig` → sidebar action → l10n packs + cache-bust → oauth token flush → verify end-to-end. Seeds from the **repo files**, so it does not depend on another tenant already being correct.

Catalog data is read from `ansible/nairobi-mdms/mdms/dss/` rather than copied into the seed dir — a second copy is how "works on bomet, empty on the repro box" happens (`60-operations.md` §4).

**Preflight is read-only and refuses to write** when it finds what seeding cannot fix:

| Check | Why it is not the naive version |
|---|---|
| A KPI whose entire **non-empty** role ceiling is unheld | Mirrors `KpiDefinition.isVisibleTo`: `PUBLIC` is additive, and an *empty* ceiling means "all authenticated roles". The naive "every referenced role must have holders" check fails even on bomet, where `PGR_ADMIN` and `PGR_SUPERVISOR` have **zero holders** |
| Schema-as-data records | POSTing a schema body to `/v2/_create/<schema>` stores the JSON Schema *as* the record. It occupies the uid the real record needs, so every later seed fails as a duplicate and the master stays broken while looking populated. Live on `mz` for both `dss.KpiDefinition` and `dss.DashboardPack`. `--repair` deactivates them |
| Facts with no `department_code` (#1280) | Every department-scoped employee gets empty tiles. `DEPARTMENT_SCOPING=disabled` is the documented temporary widening |

**`tenant_bootstrap`** gets the two gaps this uncovered:
- **Empty source is silent.** Zero source records land in no counter — not `copied`, `skipped` or `failed` — so a root can be bootstrapped to an empty catalog and report success. Now emits an explicit `warnings[]` entry for schemas where empty is a defect (`dss.*`, `INBOX.*`, `roleactions`), surfaced at the top level of the response. Deliberately excludes operator-owned masters (`ComplaintHierarchy`) where empty is correct.
- **Corrupt records propagate.** A schema-as-data row at the source was copied verbatim to every tenant bootstrapped from it. Now skipped with a warning.

Plus: the three dss schema files move out of `notif-mdms-seed/` (which never referenced them) into the new `dss-mdms-seed/`, and `export-l10n.mjs` derives the message packs as JSON with `--check` for drift.

## Validation (live, not theoretical)

**bomet** (`ke`, reference deployment) — preflight passes; verifier reproduces the runbook's recorded numbers exactly:
```
[ OK ] dss.KpiDefinition: 39 active records
catalog/_search — anonymous 10, admin 25
[ OK ] /packs returns 25 tiles for the admin's role
_query scope: {"tenantId":"ke","level":"state","departments":["DEPT_2"]}
verifying locales: en_IN fr_FR pt_PT
[ OK ] en_IN / fr_FR / pt_PT — every catalog titleKey resolves
```
(Locales come from `StateInfo.languages`, not from our own packs — otherwise a locale the deployment offers but we never seeded, like bomet's `fr_FR`, would be reported clean.)

**moz** (`mz`, the broken case) — preflight correctly reports the corruption, the `CMS_*` taxonomy mismatch and 16/18 unscoped facts, and **writes nothing**:
```
[WARN] schema-as-data rows detected: dss.DashboardPack/default, dss.KpiDefinition/default
[FAIL] these rows will shadow the real records — re-run with --repair
[FAIL] preflight failed — nothing was written
```
With the moz `ROLE_MAP`, every target role resolves to a live audience (`CMS_CASE_MANAGER` 240, `CMS_SUPERVISOR` 121, …) and the run proceeds.

`tsc --noEmit` clean on digit-mcp.

## Notes for review

- **Roles default to canonical CCRS** (`PGR_*`/`GRO`/`DGRO`/`SUPERVISOR`/`SUPERUSER`), matching the repo seed and the FE's own fallback gate, with `ROLE_MAP` for deployments that use their own taxonomy. The script never invents a role — preflight validates holders against `eg_userrole_v1`. Worth knowing: two canonical roles (`PGR_ADMIN`, `PGR_SUPERVISOR`) have zero holders even on bomet, so the seed's canonical taxonomy is partly aspirational today.
- **The l10n JSON is derived**, with the `.ts` staying the source of truth because bootstrap seeds from it. Inverting that (JSON as source, `.ts` importing it) is the cleaner end state but changes bootstrap's data path, so it is deliberately out of scope here.
- This does **not** fix moz. It makes fixing moz a scripted operation — and moz still needs an FE rebuild for #1258 (the served bundle has zero refs to `dss.DashboardConfig`, so ~480 `CMS_*` users stay outside the hardcoded gate) and an answer to what wiped its catalog in July.
- Runbook on #631 graduates into `docs/dashboard-configuration/` separately; this PR points the ops doc at the script and drops the "seed manually" instruction.

Refs #631, #1166, #1280

## Summary by CodeRabbit

- **New Features**
  - Added a dashboard enablement tool for running deployments, including validation, seeding, repair, updates, dry runs, and end-to-end verification.
  - Added dashboard schemas, KPI definitions, dashboard packs, configuration, and localized content for supported languages.
  - Added dashboard configuration options for role access, number formatting, and department scoping.

- **Bug Fixes**
  - Tenant bootstrap now reports missing source KPI catalogs and skips corrupted records instead of silently copying invalid data.

- **Documentation**
  - Updated dashboard setup guidance, troubleshooting steps, seed asset references, and localization maintenance instructions.


singhalkarun:
## Review: ran the script against a live stack

Nice piece of work — the seven-step split, the refuse-to-write preflight, and the "seed from repo files, not from another tenant" decision are all the right calls. Preflight caught real problems on my box and correctly wrote nothing.

I exercised it on a local compose stack (state root `pg`, `egovio/mdms-v2:maven-jdk21-9f83afb`). Everything below is **read-only** — `--dry-run` for steps 1–5, read-only for step 7, nothing written anywhere.

Two things block merge, plus one open question. Details and how I checked, so you can shoot any of it down.

### What checks out

Container names, ports (18094/18096/18000), `/mdms-v2/health`, all six tables it reads, the 4 `pgr_mv_*` matviews, step 7's `jsonb_path_query` SQL on PG16, the `_create` / `_update` / `schema/v1/*` paths, both localization routes (`_upsert` → 400 `tenantId: must not be empty`, `cache-bust` → 200), the wrapped-vs-flat seed handling (39 defs + 1 pack), `export-l10n.mjs --check` clean, `bash -n` clean, `jq 1.7` has `_nwise`. On the MCP side, `warnings[]` at the top level and `isSchemaDocument` both address real observed failures.

---

### Blocker 1 — `POST /mdms-v2/v2/_search/<schemaCode>` doesn't exist on this mdms-v2

Three call sites use the path form: `enable-dashboard.sh:505` (repair), `:585` (`update_record`), `:653` (step 4 template).

```
POST /mdms-v2/v2/_search/dss.DashboardConfig
  → 400 NoResourceFoundException: "No static resource v2/_search/dss.DashboardConfig."

POST /mdms-v2/v2/_search   (schemaCode in MdmsCriteria)
  → 200, returns the record
```

The repo already documents this — `tests/integration-tests/tests/utils/probes.ts:423`:

> MDMS v2 search. The schemaCode goes in the CRITERIA, not the path — `POST /mdms-v2/v2/_search/<schema>` 404s ("No static resource").

Every other caller uses the no-suffix form (`digit-ui-v2/.../endpoints.ts:7`, the integration tests); the workbench usages of the suffix form are commented out. The image is the tag five `local-setup/docker-compose*.yml` files pin, so this isn't an exotic build.

Three consequences:

1. **Step 3 aborts the run on any tenant that already has `dss.DashboardConfig`.** `_create` → duplicate → `update_record` → broken search → `return 1` → `err "dss.DashboardConfig exists but update failed"`. My box has that record, so it dies there.
2. **`--repair` silently doesn't repair.** The search returns an error object, so `$cur.mdms[0]` is null. I ran the exact jq: `null + {isActive:false}` yields `{"isActive":false}` and exits **0** — no jq error, so the `|| warn` fallback at `:510` never fires. It POSTs a bodyless update, warns once, moves on. That's the flag the description says is needed for moz.
3. **Step 4 skips the sidebar grant and returns success.** It logged `no existing roleactions row to mirror — skipping sidebar grants` while there are **368 active roleactions rows** at `pg`. Step 7 doesn't verify grants, so the script can print "Dashboard enabled" with no sidebar link.

Three-line fix.

---

### Blocker 2 — the schema step 1 registers rejects 24 of the 39 records step 2 seeds

Step 1 registers `dss-mdms-seed/schemas/dss.KpiDefinition.json`; step 2 seeds `ansible/nairobi-mdms/mdms/dss/KpiDefinition.json`. Validated offline with `jsonschema` (Draft7):

```
dss.KpiDefinition:   39 records, 24 FAIL
dss.DashboardPack:    1 record,   0 FAIL
dss.DashboardConfig:  1 record,   0 FAIL
```

The schema allows `viz.kind ∈ [scalar, bar, rankedList, line, map, dow]`. The catalog uses nine values outside it:

```
x7 number-tile-sparkline   x2 data-table       x1 sla-risk-table
x7 number-tile-delta       x1 pie              x1 table
x3 stacked-bar             x1 histogram        x1 horizontal-bar
                                          + viz.format 'ratingOutOfFive'
```

It's datable drift. The schema was written in `ac0289d3` on **2026-06-25** and hasn't been touched since; `number-tile-sparkline` entered the data on **2026-06-29**; **30 commits** have touched the catalog data in that window. This PR moves the schema file with **0 line changes**, so it carries the staleness forward rather than introducing it.

Failure path: `seed_records:559-575` → no `"mdms"` in the response → `failed++` → `[[ $failed -eq 0 ]] || return 1` → `main:917` exits 1.

**Why the bomet run wouldn't have caught this:** step 1 greps for `DUPLICATE|already exists` and logs `already registered` (`:470`). On bomet the dss schemas already existed, so the older, broader *live* schema stayed in force and step 2 succeeded. Register-then-seed on a virgin tenant — the moz case this script exists for — wasn't exercised end to end.

**Where I could be wrong, and it matters:** I'm inferring that mdms-v2 enforces `enum`. What I can actually show is (a) your own `b7cc109f` — *"made mdms `_update` 400 (`required key [allowed] not found`) and broke fresh tenant_bootstrap seeding identically (found in the bomet integration deploy)"* — which proves `required` enforcement, and (b) on my box, 80 live rows under three populated enum-bearing registered schemas have **zero** enum violations. If you know enums aren't validated, this downgrades. But the schema and the dataset still contradict each other either way, and the PR is the natural place to reconcile them.

**Could you dump bomet's registered `dss.KpiDefinition` and diff it against the repo file?** That one artifact settles this and the next item in one shot.

---

### Open question — `x-unique` is absent, but the logic assumes it

`dss.KpiDefinition` and `dss.DashboardPack` declare no `x-unique`; `dss.DashboardConfig` declares `["id"]`. mdms-v2 derives the uid from it — confirmed on my box: `dss.DashboardConfig` → uid `default`, `ACCESSCONTROL-ROLEACTIONS` (`["rolecode","actionid"]`) → 64-char hash.

But `update_record:583` searches `uniqueIdentifiers:[<record.id>]`, and the header's "re-runs report existing records as skipped" both need `x-unique: ["id"]`. I couldn't determine what mdms-v2 does with none without writing to a live stack. If it rejects, the `:224` retry loop burns ~15s × 40 records before reporting 40 failures; if it mints fresh uids, every re-run duplicates the catalog. Noting that `:224` already retries on `"Values defined against unique fields cannot be empty"`, so you may have met this.

Probably the same root cause as Blocker 2 — one stale schema file, two symptoms.

---

### Medium

- **Defaults hard-fail on a stock box.** Preflight blocked on `cl_chart_over_time_open_daily` and `cl_table_ward_open_daily`: `PGR_ADMIN`, `PGR_SUPERVISOR` and `TICKET_REPORT_VIEWER` all have 0 holders, against `GRO 13, DGRO 11, PGR_VIEWER 11, PGR_LME 8, SUPERUSER 6, SUPERVISOR 5`. The description acknowledges the taxonomy is partly aspirational, but an operator on the default path hits a stop and has to guess `ALLOW_EMPTY_ROLES=true`. Worth a line in the ops doc.
- **`--dry-run` skips auth (`:913`), so step 4 still fires a live search with `{"RequestInfo":,...}`** — malformed, 400, and it produces the misleading "seed ACCESSCONTROL-ROLEACTIONS first" advice. That's what sent me down the wrong path initially. Either mint in dry-run or skip step 4's probe.
- **No postgres reachability check.** With a wrong `DB_CONTAINER` you get a confident "0 holders" for every role and "found 0 matviews" rather than "can't reach postgres". It does still refuse to write, so it's misleading rather than dangerous — `enable-notifications.sh:729` has an explicit `pg_isready` require worth copying.
- **Step 7 ends red even after a clean seed**, where `StateInfo.languages` lists a locale you don't ship a pack for. Mine is `en_IN,ka_IN` against `en_IN`+`pt_PT`. That's the documented intent — just flagging that the run exits 1.

### Minor

- `--only` with no argument → raw `line 884: $2: unbound variable`.
- `--from step99` (typo) → runs nothing, **exits 0**.
- `--help` prints `set -euo pipefail` — `usage()` hardcodes `sed -n '2,75p'`, one line long.
- `mdms-tenant.ts:1596` hardcodes *"The dashboard will render an empty catalog. Seed it … with enable-dashboard.sh"* for all four entries in `SCHEMAS_EMPTY_IS_A_PROBLEM`, but `INBOX.InboxQueryConfiguration` and `ACCESSCONTROL-ROLEACTIONS.roleactions` aren't dashboard, and the script doesn't seed INBOX.
- The two corruption detectors differ slightly: `isSchemaDocument` type-checks `$schema`/`properties`; the preflight SQL at `:291` uses jsonb `?` (key presence only).
- `set -e` is suppressed inside every step because `main:917` calls them as `"$s" || {...}`, so `set -euo pipefail` buys less than the header implies.

---

### Caveats on my end

My `pgr-services` image has no analytics tier at all (`unzip -l /app/app.jar | grep -ci analytics` → 0), so preflight correctly blocked and I couldn't exercise anything past it against a working dashboard. Worth noting the gate reports `HTTP 400` there — it's actually a 404 masked by tracer's catch-all `@ControllerAdvice`, and since the check treats any non-2xx as "old image", a genuine 400 from a rejected body would get the same "upgrade your image" advice.

Findings are from one deployment; bomet may differ — which is rather the point of Blocker 2.

<sub>Verification run with Claude Code against a local compose stack. Read-only throughout.</sub>

KDwevedi:
Both blockers reproduced against bomet. Fixed in `eb70f7c`.

### Blocker 1 — `_search` path form

```
POST /mdms-v2/v2/_search/dss.DashboardConfig   → 400
POST /mdms-v2/v2/_search  (schemaCode in criteria) → 200
```

Documented at `probes.ts:423`; three call sites here used the wrong form. Consequence 2 is the real one: `null + {isActive:false}` is valid jq that exits 0, so `--repair` posted a bodyless update and reported success without repairing anything.

Fixes: `mdms_search`/`mdms_first_record` on the correct form; an empty readback in `--repair` is now a hard stop; step 4 cross-checks the DB before advising, distinguishing "no roleactions master" from "search failed".

### Blocker 2 — stale schema files

Dump of bomet's registered schema vs the repo file:

```
                  live   repo
viz.kind enum      15      6
x-unique         ["id"]  absent   (KpiDefinition, DashboardPack)
39 records         0 fail  24 fail
```

The missing `x-unique` also answers the open question — it is why uid derivation worked on bomet and would not on a virgin tenant. Replaced all three files with bomet's registered definitions, keeping `numberFormat` (#1213) and `departmentScoping` from the repo side.

On why bomet did not catch it: step 1 logs "already registered" when the schema exists, so the older live schema stays in force and only the from-scratch path fails. Preflight now validates the seed data against the schema files offline.

### Other findings (all in `eb70f7c`)

- postgres reachability check (`pg_isready`)
- `--dry-run` no longer fires a live search with a malformed RequestInfo
- the analytics probe no longer asserts "old image"; a missing route can surface as 400 via tracer's ControllerAdvice
- step 7 warns rather than fails on a locale offered in StateInfo with no shipped pack
- `--only` with no arg exits 2 instead of an unbound-variable crash; an unknown step exits 2 instead of running nothing; `--help` prints the full header
- steps run under an `ERR` trap instead of `"$s" || {…}`, which had suppressed errexit
- the corruption SQL type-checks both markers, matching `isSchemaDocument`
- `SCHEMAS_EMPTY_IS_A_PROBLEM` carries a per-schema remedy; it was pointing INBOX/roleactions at `enable-dashboard.sh`, which does not seed them

### Exercised live on bomet

Schema registration (3 already-registered), catalog seed (40 records, all duplicate-detected — confirms uid derivation), the update path (`DashboardConfig/default`, same row id, content unchanged), step 4 (4 grants present). Still unexercised: register-then-seed on a virgin root, and `--repair` against a real corrupt row — both covered in later comments.

KDwevedi:
## Runbook — `enable-dashboard.sh`

Seeds the dashboard config (schemas, KPI catalog, role gate, sidebar action, localization) on a running deployment. Builds and deploys nothing. Run against a state root; a city under an enabled root inherits automatically. Graduates into `docs/dashboard-configuration/` on merge. `--help` prints this header.

### Prerequisites

`--only step0` checks all of these and writes nothing:

- CCRS checkout with the seed files (`REPO_ROOT=` if run elsewhere); `jq`, `python3`, `curl`
- Postgres, mdms-v2 (`:18094`, direct — never Kong), localization (`:18096`), redis reachable
- pgr-services serving `/v2/analytics`; the four `pgr_mv_*` matviews present
- An admin login at the root (`SUPERADMIN` on a DDH box, `ADMIN` on an MCP box)
- Sidebar action `4557` present in the ACCESSCONTROL action masters
- A UI bundle embedding `products/dashboard/` (not script-checkable)

Know the box's reset path before seeding. A nightly cron or re-converge can undo the seed.

### Run

```bash
DASHBOARD_TENANT=<root> ./local-setup/scripts/enable-dashboard.sh --dry-run   # writes nothing
DASHBOARD_TENANT=<root> ./local-setup/scripts/enable-dashboard.sh
```

Steps: 0 preflight · 1 schemas · 2 catalog (39 defs + pack) · 3 DashboardConfig · 4 sidebar grants · 5 localization + cache-bust · 6 token flush (invalidates all sessions) · 7 verify. Flags: `--only`, `--from`, `--to`, `--repair`, `--update`, `--dry-run`. Re-runs skip existing records unless `--update`.

### Roles

Default is canonical CCRS (`PGR_*`/`GRO`/`DGRO`/`SUPERVISOR`/`SUPERUSER`). It is partly aspirational — `PGR_ADMIN` and `PGR_SUPERVISOR` hold nobody even on bomet. Remap to real roles:

```bash
ROLE_MAP="PGR_SUPERVISOR=CMS_SUPERVISOR,PGR_LME=CMS_CASE_MANAGER" \
DASHBOARD_ALLOWED_ROLES="SUPERVISOR,SUPERUSER,GRO,DGRO,CMS_SUPERVISOR" ...
```

`ROLE_MAP` rewrites the catalog `visibleTo`, the pack personas, and the gate. Preflight errors if a KPI's whole non-empty role ceiling is unheld, or if no gate role has holders. An empty `visibleTo` means all authenticated roles; `PUBLIC` is additive, not a ceiling.

### Blockers

| Symptom | Cause | Fix |
|---|---|---|
| `schema-as-data rows detected` | a JSON Schema POSTed as a record; squats the uid a real record needs | `--repair` |
| `N records violate the schema` | schema files and catalog data drifted | reconcile `dss-mdms-seed/schemas/` with `ansible/nairobi-mdms/mdms/dss/` |
| `action id 4557 does not exist` | sidebar action not seeded | seed it in both ACCESSCONTROL action masters, then `--only step4` |
| every grant rejected | ACCESSCONTROL-ROLES has hashed uids; the roleactions `rolecode` x-ref can't resolve | grant via DDH/dataloader, or re-seed roles with code-valued uids |
| token mint 500 ("encryption process") | tenant has no `eg_enc_symmetric_keys` row | provision the enc key |
| tiles render, values empty | department scoping (#1280), or MVs stale | `DEPARTMENT_SCOPING=disabled`; check `DashboardRefreshScheduler` |
| titles show `DASHBOARD_*` | l10n not seeded or not cache-busted | `--only step5` |

### The script cannot

Rebuild the frontend (the role gate reads `DashboardConfig.allowedRoles` only from a bundle with #1258; older bundles use a hardcoded list), seed the ACCESSCONTROL action masters, fix department enrichment, or survive a frontend redeploy.

---

## Deployments assessed

**bomet (`ke`)** — fully enabled. 39 defs, catalog anon 10 / admin 25, 25 tiles, three locales resolve. The verify numbers above come from here.

**moz (`mz`)** — not yet fixed. Catalog empty plus two corrupt schema-as-data rows; the FE bundle predates #1258 so `CMS_*` roles (~480 users) cannot reach the dashboard; `rainmaker-dashboard` messages absent; 16/18 facts have no `department_code`. Invocation:

```bash
DASHBOARD_TENANT=mz \
ROLE_MAP="PGR_SUPERVISOR=CMS_SUPERVISOR,PGR_LME=CMS_CASE_MANAGER,PGR_ADMIN=SUPERUSER" \
DASHBOARD_ALLOWED_ROLES="SUPERVISOR,SUPERUSER,GRO,DGRO,CMS_SUPERVISOR" \
DEPARTMENT_SCOPING=disabled \
./local-setup/scripts/enable-dashboard.sh --repair
```

`--repair` clears the two corrupt rows; `ROLE_MAP` maps the catalog to moz's `CMS_*` taxonomy; `CMS_SUPERVISOR` (121 holders) is the intended gate audience; `DEPARTMENT_SCOPING=disabled` because most facts are unscoped; add `ADMIN_USER=ADMIN` if `SUPERADMIN` does not authenticate. This reaches the ~26 legacy-role users; the FE rebuild is what reaches the 480 `CMS_*` users.

**ovh-8c24g (`pg`)** — dev box, two platform-shaped gaps. `ke` has no encryption key, so `ke` logins 500; use `pg`. ACCESSCONTROL-ROLES uses hashed uniqueIdentifiers, so roleactions grants cannot go through the v2 API (grant via DDH/dataloader).

vinothrallapalli-eGov (approved):

Issue #1276, opened 2026-07-16
[Bug] Employee dashboard: unable to add KPIs


subhashini-egov:
## Symptom

On the Configurable Dashboard, an employee cannot add a KPI to their dashboard layout. The "Add KPI" flow does not complete — the tile either fails to attach to the grid, or the picker returns no selectable KPIs, or the action errors silently.

## Expected behaviour

An employee opens their dashboard, selects "Add KPI" (or the equivalent widget-picker entrypoint), chooses from the list of KPIs their role is scoped to, and the new KPI tile appears in the grid at the next available slot. The layout persists on reload.

## Reproduction (please confirm)

1. Log in as an employee (GRO or supervisor persona)
2. Navigate to the dashboard
3. Attempt to add a KPI to the current layout
4. Observe: KPI does not attach

The exact point of failure is worth capturing — is it:
- The picker doesn't open?
- The picker opens but lists no KPIs?
- A KPI can be selected but nothing appears on the grid after confirming?
- The tile appears momentarily then disappears?
- Console / network error surfaces?

Whichever it is, the reproduction and the failing endpoint (or console error) should be attached before implementer picks this up.

## Suspected surface area

- Widget-config schema — whether the employee persona has KPI-add capability enabled
- KPI catalog scoping — whether `PrincipalScopeResolver` (or the RBAC layer) is filtering the catalog to empty for the employee's roles
- Layout-persist API — whether the save endpoint accepts the new tile config
- Frontend state — whether the picker's selection is wired to the grid mutation

Related recent work:
- #908 (Dashboard data scoping & KPI access control) — if that has landed partially, its RBAC filter may be filtering the KPI catalog too aggressively
- #1166 (Moz supervisor dashboard unblocks) — different persona, same class of "can't get KPIs onto the dashboard"

## Acceptance criteria

- [ ] Employee (GRO, supervisor personas at minimum) can add a KPI to their dashboard from the KPI catalog
- [ ] The added tile persists across reloads
- [ ] The tile scopes data correctly to the employee's role + tenant (does not leak cross-tenant data)
- [ ] Regression test added to the citizen/employee test suite that exercises the add-KPI path

## Parent epic

Part of **#631** — [Feature] CCRS SaaS: Dashboard.

subhashini-egov:
@KDwevedi This just landed in this sprint (bug bash ticket)

KDwevedi:
[...]

PR #1407, opened 2026-07-22
Fix dashboard Portuguese filter/table labels (#1108)

### LINKED ISSUE #1108: [P0] Dashboard: localise to Portuguese and populate language dropdown from DIGIT-configured languages

soniikajal:
Fixes #1108.

## Summary
- Stop English (`en_IN`) bleed into Portuguese UI by reading locale packs via locale-strict `getResource` / side-cache instead of i18next `exists`/`t` (which ignore `fallbackLng: false`).
- Keep `dimensionLabel` contract: locale message → data-owned name (boundary localname / MDMS) → **raw code**. No regex humaniser — missing `pt_PT` translations stay visible.
- Make filter/table chrome (“All wards”, “All types”, complaint types, departments, wards) re-resolve on language switch; `ensureMessages()` side-fetches packs with auth + tenant (failed fetches no longer poison an empty cache).
- Add `seed-pt-filter-l10n.mjs` to seed empty `pt_PT` boundary/pgr packs (optional `--translate` via MyMemory).
- Regenerate `local-setup/db/dss-mdms-seed/l10n/{en_IN,pt_PT}.json` from `dashboard-l10n-seed.ts` (includes `DASHBOARD_COMMON_UPDATED` after #1400).
- Also recovers empty catalog layouts when pack `defaultLayout` is `[]` (#1276) and scopes saved layout by tenant/user.

## Review follow-ups addressed
- Dropped `humanizeTypeCode` / `humanizeDimensionCode` from the `dimensionLabel` chain.
- Regenerated l10n JSON packs (`export-l10n.mjs --check` clean).
- CodeRabbit: MyMemory docstring; `readSaved` per-key parse so corrupt scoped JSON still falls back to legacy; code-echo guard for boundary/department; no empty poison-cache on `_search` failure.

## Summary by CodeRabbit

* **New Features**
  * Added English and Portuguese localization for dashboard “Updated” labels.
  * Added localized boundary, complaint-type, department, and filter labels.
  * Added Portuguese dashboard localization seeding support.
  * Added tenant- and user-specific dashboard layout persistence.

* **Bug Fixes**
  * Improved localization loading, fallback behavior, and translation accuracy.
  * Prevented dashboard content and titles from clipping in longer translations.
  * Ensured dashboards populate when layout metadata is unavailable.
  * Improved recovery of invalid or incomplete saved layouts.
  * Improved handling of missing or untranslated dimension labels.


KDwevedi (changes requested):
The locale-strict `getResource` fix for the `en_IN` bleed is the right root cause, and the added tests are good. Two things before merge:

### 1. Drop the humanizer fallback — let the localisation default surface

`dimensionLabel`'s original contract is deliberate: **no code-owned fallback.** When a translation is missing, the raw code renders so the gap is visible and gets fixed. The new `humanizeTypeCode` / `humanizeDimensionCode` steps reverse that — they paper over a missing `pt_PT` translation with a regex-humanized code, which reads as "done" on screen while the translation is still absent. That's exactly the silent gap we want to avoid.

Please drop the humanizer steps and keep the chain: **locale message → data-owned name (boundary `localname` / MDMS display) → raw code.** Data-owned names are fine — they're operator-authored, not invented. It's only the regex humaniser we don't want. (`useFilterOptions`'s docstring already says "then the raw code," which is the intent.)

### 2. Regenerate the l10n packs (drift with #1400)

This branch predates #1400, which added `local-setup/db/dss-mdms-seed/l10n/{en_IN,pt_PT}.json`, generated from `dashboard-l10n-seed.ts` by `export-l10n.mjs`. You added `DASHBOARD_COMMON_UPDATED` to the `.ts`; after rebasing onto `develop`, run `node local-setup/db/dss-mdms-seed/export-l10n.mjs` and commit the regenerated packs — otherwise `--check` drifts and a fresh `tenant_bootstrap` won't seed the new message.

### Minor (non-blocking)

- `ensureMessages()` now side-fetches `/localization/messages/v1/_search` in embedded mode too (was a no-op) — confirm that call carries auth + tenant in production.
- There are now three l10n seeders — this PR's `seed-pt-filter-l10n.mjs`, `enable-dashboard.sh` step 5, and `tenant_bootstrap`. Worth consolidating later.

digit-ui-esbuild/products/dashboard/src/hooks/useCatalogLayout.js
```diff
@@ -89,30 +96,44 @@ function normalizeItem(item, kpis) {
...
-    window.localStorage?.setItem(
-      STORAGE_KEY,
-      JSON.stringify(layout.map(({ i, x, y, w, h }) => ({ i, x, y, w, h })))
-    );
+    const scopedKey = storageKeyFor(getTenantId(), getUserUuid());
+    const payload = JSON.stringify(layout.map(({ i, x, y, w, h }) => ({ i, x, y, w, h })));
+    window.localStorage?.setItem(scopedKey, payload);
+    // Keep the legacy slot in sync so older builds still see the layout.
+    if (scopedKey !== STORAGE_KEY) {
+      window.localStorage?.setItem(STORAGE_KEY, payload);
```
KDwevedi: **Blocking — this reintroduces the #1276 cross-user clobber.**

We write the payload to the scoped key *and* the legacy global key. `readSavedLayout` falls back to the legacy key whenever the scoped one is empty, so the next user to sign in on this browser hydrates the previous user's layout — the exact failure `layoutStore.test.js` guards with *"two users on one browser no longer clobber each other (the #1276 shared-slot failure)"*. That test still passes because it exercises `layoutStore` directly, not this hook.

`layoutStore.js` also documents the invariant this breaks:

> Persisting always writes the scoped key, so the legacy slot is never mutated and stops mattering after the first save.

Please drop the legacy write. Backward compatibility is already covered by the read-side fallback.

digit-ui-esbuild/products/dashboard/src/hooks/useCatalogLayout.js
```diff
@@ -149,19 +174,35 @@ export function useCatalogLayout(kpis, packLayout) {
...
   useEffect(() => {
-    if (!seed.length) return;
+    if (!catalogReady) return;
     const saved = readSaved();
-    const source = saved !== null ? saved : seed;
+    // saved === null → first visit, use pack seed (may be empty).
+    // saved === [] with a non-empty seed → recover from the empty-defaultLayout
+    // era / failed hydrate that persisted an empty array and blocked the pack.
+    // saved with items → user's arrangement wins.
+    let source;
```
KDwevedi: **Blocking — "user cleared every tile" now silently un-clears itself.**

`saved === []` is treated as recoverable, replaced by the pack seed, and then persisted. `layoutStore.resolveInitialLayout` documents the opposite contract:

> the SAVED layout (the user's arrangement, **including an intentional empty one**) wins over the pack seed … so the seed does not re-add the removed tiles on reload.

Two contracts now disagree inside the same module tree, and the persist makes it unrecoverable — a user who removes every tile gets them all back permanently on the next load.

If the recovery is genuinely needed for layouts written during the empty-`defaultLayout` era, distinguish "never hydrated" from "user emptied" explicitly rather than overloading `[]` — e.g. version the stored payload (`{ v: 2, items: [] }`) and only recover unversioned empties. Whichever way it lands, `layoutStore`'s docstring and tests need to move with it.

digit-ui-esbuild/products/dashboard/src/hooks/useCatalogLayout.js
```diff
@@ -89,30 +96,44 @@ function normalizeItem(item, kpis) {
 }
 
 function buildSeedLayout(packLayout, kpis) {
-  return (packLayout || [])
-    .filter((item) => kpis[item.kpiId])
-    .map((item) =>
-      normalizeItem({ i: item.kpiId, x: item.x, y: item.y, w: item.w, h: item.h }, kpis)
+  const items = (packLayout || []).filter((item) => kpis[item.kpiId]);
+  const out = [];
+  for (const item of items) {
```
KDwevedi: **Blocking (design) — divergent duplicate implementations.**

This hook now carries its own `buildSeedLayout`, `normalizeItem`, `readSaved`, and persist logic while `layoutStore.js` exports `buildSeedLayout`, `reconcileLayout`, `resolveInitialLayout`, and `persistLayout` — which nothing imports any more. `layoutStore.test.js` is therefore testing code that no longer runs in the product, which is how the two issues above slipped past a green suite.

Please push the position-filling behaviour down into `layoutStore.buildSeedLayout` and have the hook call the shared helpers, so the tested module stays the shipped module.

digit-ui-esbuild/products/dashboard/src/hooks/useCatalogLayout.js
```diff
@@ -137,6 +158,10 @@ function persist(layout) {
 export function useCatalogLayout(kpis, packLayout) {
   const seed = useMemo(() => buildSeedLayout(packLayout, kpis), [packLayout, kpis]);
   const geom = useMemo(() => createCatalogDragGeometry(kpis), [kpis]);
+  // Catalog (not seed) is the hydration signal: /packs often returns
+  // defaultLayout: [] when no DashboardPack matches roles, while the user may
+  // still have a saved layout from Add-KPI (#1276).
+  const catalogReady = useMemo(() => Object.keys(kpis || {}).length > 0, [kpis]);
```
KDwevedi, line 70: Nit, but worth a test: switching the hydrate trigger from `seed.length` to `catalogReady` means hydration now fires as soon as any KPI is known, before the pack layout necessarily resolves. If `packLayout` arrives in a later render, `seed` changes and the effect re-runs — fine — but combined with the persist-on-recover above, an interim empty seed can be written to storage. A hook-level test around the ordering would be reassuring.

KDwevedi (commented):
Second pass. The i18n core is right and worth landing: `getResource`-based locale-strict lookup is the correct root cause for the `en_IN` bleed, the humanizer is gone from the `dimensionLabel` chain as asked, and the new tests mock `exists/t` to deliberately bleed English, which is exactly the right shape of regression test.

What I'd like resolved before merge, in order:

**Blocking**
1. `useCatalogLayout.persistPositions` dual-writes the legacy global key — that reintroduces the #1276 cross-user clobber the layout scoping was added to fix.
2. `saved === []` is now overwritten by the pack seed *and persisted*, which contradicts the documented `layoutStore.resolveInitialLayout` contract ("user cleared every tile" must survive a reload).
3. The hook now carries its own copies of `buildSeedLayout` / `normalizeItem` / read / persist while `layoutStore`'s tested versions go unused — the tested code is no longer the shipped code.

All three are layout-persistence, not #1108. Strong preference: split them into their own PR against #1276 and land the i18n half on its own.

**Should fix**
4. `findMessageLoose` full-bundle scan on the label hot path (perf).
5. `getLanguage()` reads localStorage on every call and makes it authoritative over `window.i18next.language`.
6. `DASHBOARD_COMMON_UPDATED` is a deploy dependency — needs the l10n upsert + cache-bust or cards render the raw key.
7. The seed script upserts English as `pt_PT` when MyMemory fails, which is worse than a visible gap.
8. The seed script is ke/Bomet-hardcoded.

Inline details below; the rest are nits.

digit-ui-esbuild/products/dashboard/src/i18n/localeRuntime.js
```diff
@@ -43,46 +44,110 @@ const readStoredLocale = () => {
...
+ * BOMET_CHEPALUNGU_KONGASIS while the pack keys BOMET_CHEPALUNGU_KONG_ASIS.
+ */
+export function findMessageLoose(code, lng) {
+  if (code == null || code === "" || !lng) return undefined;
+  const want = normalizeKey(code);
+  if (!want) return undefined;
+
+  const tryMap = (map) => {
+    if (!map) return undefined;
+    for (const [k, v] of Object.entries(map)) {
```
KDwevedi: **Perf — full-bundle linear scan on the label hot path.**

`Object.entries(map)` plus a `normalizeKey` on every key, per lookup. `dimensionLabel` calls this on every boundary miss, for the active locale *and* `en_IN`, for every cell, on every render. A 100-row table against a few-thousand-key boundary pack is on the order of a million string ops per paint, and misses are the common case on tenants with an unseeded `pt_PT` pack — i.e. exactly the case this PR targets.

Build the normalized index once per `(lng, bundle)` and cache it, invalidating on `languageChanged` / `added` (the same signals `subscribe` already listens to). The map lookup then goes back to O(1).

digit-ui-esbuild/products/dashboard/src/i18n/localeRuntime.js
```diff
@@ -43,46 +44,110 @@ const readStoredLocale = () => {
 };
 
 export function getLanguage() {
+  // Prefer Employee.locale — LocalizationService.changeLanguage writes it
+  // synchronously before the async bundle fetch finishes.
+  try {
+    const stored = window.localStorage.getItem("Employee.locale");
```
KDwevedi: Two concerns here.

**1. Source of truth inverted.** `Employee.locale` now wins over `hostI18next().language` unconditionally. When the dashboard is embedded, i18next is the authority — any host-driven language change that doesn't write that key (or a surface that writes `Citizen.locale`) leaves the dashboard rendering a different language than the chrome around it. The stated reason (`changeLanguage` writes it synchronously before the bundles land) argues for preferring it only while the host language is *unset*, not always. Consider `host?.language || stored` with the side-fetch covering the gap, or reconcile the two on the `languageChanged` signal.

**2. Cost.** This is a synchronous `localStorage.getItem` on every call, and `dimensionLabel` calls `getLanguage()` up to three times per label. Cache the value in module scope and refresh it in the `subscribe` callback.

Also, this now duplicates `readStoredLocale` above (which is reachable only when the read throws) — worth collapsing into one.

digit-ui-esbuild/products/dashboard/src/i18n/localeRuntime.js
```diff
@@ -111,25 +181,41 @@ export function ensureMessages() {
...
       delete standalone.pending[locale];
     });
 }
 
+/**
+ * Always side-fetch the active locale (and en_IN when different). Host
+ * useStore may race changeLanguage; the side-cache + getResource path is
+ * what actually makes "All wards" / "All types" flip to Portuguese (#1108).
+ */
+export function ensureMessages() {
```
KDwevedi, line 235: `ensureMessages` is no longer a no-op when embedded, and `useDashboardT` now calls it on *every* subscribe signal. The `messages`/`pending` cache keeps the steady state at two `_search` calls per session, which is fine — but the failure path deletes `pending` with no backoff, so a flapping localization service re-fires a pair of requests on every i18n event (and every card mounting `useDashboardT` participates). A retry cap or a short cooldown per locale would bound that.

The fix to stop caching `{}` on non-OK is right and worth keeping.

digit-ui-esbuild/products/dashboard/src/i18n/localeRuntime.js
```diff
@@ -111,25 +181,41 @@ export function ensureMessages() {
...
-      host.off("languageChanged", cb);
-      host.store?.off("added", cb);
-    };
+    unsubs.push(() => host.off("languageChanged", cb));
+    // addResources notifies the store, not i18n.on('added')
+    if (host.store?.on) {
+      host.store.on("added", cb);
+      unsubs.push(() => host.store.off?.("added", cb));
+    }
+    host.on?.("added", cb);
```
KDwevedi: `added` is now registered on both `host.store` and the i18next instance. If both emit, every bundle load fires the callback twice — and since the callback in `useDashboardT` also calls `ensureMessages()`, that's two re-renders and two cache probes per load. Harmless but wasteful; either pick the emitter that actually fires or coalesce the callback.

digit-ui-esbuild/products/dashboard/src/i18n/dimensionLabel.js
```diff
@@ -41,6 +78,33 @@ const CANDIDATES = {
...
+  return true;
+}
+
+function shouldTryEnIn(kind) {
+  if (kind === "boundary") return true;
+  if (kind === "complaintType") return getLanguage() === FALLBACK_LOCALE;
+  return false;
+}
+
+function usableFallbackText(kind, text, code) {
```
KDwevedi, line 100: The comment and the code disagree:

```js
// boundary-service often echoes the code as localname — treat as missing
return true;
```

The echo case is already handled by the `String(text) === String(code)` check two lines up, so this branch accepts everything. Reword the comment to say what the function does ("code-echo already rejected above; any other data-owned name is usable").

digit-ui-esbuild/products/dashboard/src/i18n/dimensionLabel.js
```diff
@@ -49,10 +113,45 @@ const CANDIDATES = {
  */
 export function dimensionLabel(code, kind, fallbackText) {
   if (code == null || code === "") return fallbackText !== undefined ? fallbackText : "";
+  const raw = String(code);
+  // Analytics null-buckets land as the literal "Unknown" — use the dashboard
+  // chrome key (pt: Desconhecido) instead of humanising the English word.
+  if (/^(unknown|null|undefined)$/i.test(raw.trim())) {
```
KDwevedi, line 123: Minor: this matches on the raw code for *every* kind, so a boundary or department legitimately named `Unknown` renders as the localized null-bucket label. Fine for analytics null buckets in practice — worth a one-line caveat in the docblock so it isn't debugged twice.

digit-ui-esbuild/products/dashboard/src/i18n/dimensionLabel.js
```diff
@@ -49,10 +113,45 @@ const CANDIDATES = {
...
+        const msg = translateInLocale(key, FALLBACK_LOCALE);
+        if (accept(msg)) return msg;
+      }
+    }
+  }
+  // Analytics ward codes sometimes drop underscores vs the pack
+  // (BOMET_CHEPALUNGU_KONGASIS vs …KONG_ASIS → Kong'asis).
+  if (kind === "boundary") {
+    const loose =
+      findMessageLoose(code, getLanguage()) || findMessageLoose(code, FALLBACK_LOCALE);
```
KDwevedi, line 149: Two scans per miss (active locale, then `en_IN`) — see the perf note on `findMessageLoose`. Once the normalized index is memoized this is cheap; today it is the most expensive path in a table render.

Separately, worth calling out on #1108 that boundary labels now *deliberately* fall back to `en_IN` (step 2 of the new chain). That is defensible for place names, but it is the opposite of this file's "gaps stay visible" doctrine, so QA will otherwise re-file "wards still show English in PT" as a bug.

digit-ui-esbuild/products/dashboard/src/components/CardUpdatedStamp.jsx
```diff
@@ -1,9 +1,13 @@
...
-const CardUpdatedStamp = ({ label }) => (
-  <span className="dashboard-card-updated tw-pointer-events-none tw-absolute tw-bottom-1 tw-right-5 tw-z-[2] tw-rounded tw-bg-surface tw-px-1 tw-text-[10px] tw-leading-tight tw-text-muted-foreground">
-    Updated {label}
-  </span>
-);
+const CardUpdatedStamp = ({ label }) => {
+  const { t } = useDashboardT();
+  return (
+    <span className="dashboard-card-updated tw-pointer-events-none tw-absolute tw-bottom-1 tw-right-5 tw-z-[2] tw-rounded tw-bg-surface tw-px-1 tw-text-[10px] tw-leading-tight tw-text-muted-foreground">
+      {t("DASHBOARD_COMMON_UPDATED", "Updated")} {label}
```
KDwevedi, line 14: **Deploy dependency worth flagging in the PR description.**

`translate` echoes the *key* on a miss, and the second argument is extraction-source only, never rendered. So on any environment where the new `DASHBOARD_COMMON_UPDATED` row has not been upserted, this card renders the literal `DASHBOARD_COMMON_UPDATED 2h ago` where it previously read `Updated 2h ago` — a visible regression for unseeded tenants.

The seed JSON ships in this PR, but running deployments need the localization upsert plus a cache-bust to land with (or before) the frontend. Please add that to the PR description / release checklist.

digit-ui-esbuild/products/dashboard/scripts/seed-pt-filter-l10n.mjs
```diff
@@ -0,0 +1,341 @@
...
+  const cache = {};
+  for (let i = 0; i < uniqueEnglish.length; i++) {
+    const text = uniqueEnglish[i];
+    const url = `https://api.mymemory.translated.net/get?q=${encodeURIComponent(text)}&langpair=en|pt`;
+    try {
+      const res = await fetch(url);
+      const data = await res.json();
+      cache[text] = data?.responseData?.translatedText || text;
+    } catch {
+      cache[text] = text;
```
KDwevedi: **This bakes English into the `pt_PT` pack.**

On a failed or rate-limited MyMemory call we fall back to the English string and then upsert it as a `pt_PT` message. That is worse than leaving the gap: the row looks seeded, `usableMessage` cannot detect it (it is not a code echo), and the missing translation becomes invisible to exactly the audit this PR is trying to make possible.

Skip the key on failure and report the count at the end instead.

Also worth pairing in the docblock: run `--dry-run --out=…` and eyeball the machine translations before upserting them into a live pack.

digit-ui-esbuild/products/dashboard/scripts/seed-pt-filter-l10n.mjs
```diff
@@ -0,0 +1,341 @@
...
+ * without quotes, or the raw JWT).
+ */
+import { writeFileSync } from "node:fs";
+import { pathToFileURL } from "node:url";
+
+const BASE = process.env.DIGIT_BASE || "http://127.0.0.1:18080";
+const TENANT = process.env.TENANT_ID || "ke";
+const LOCALE = "pt_PT";
+const SOURCE = "en_IN";
+const BOUNDARY_MODULE = "rainmaker-boundary-admin";
```
KDwevedi: This script is ke/Bomet-specific in a shared repo: `rainmaker-boundary-admin` hardcoded, `TENANT` defaulting to `ke`, `KE_ADMIN_` alias rows written into the pack, and a `BOMET_` prefix strip in `seedDisplayNameForWardCode`. The runtime half of this PR already derives the boundary module from `globalConfigs.HIERARCHY_TYPE` — please parameterize the same way (module, tenant, prefix list via env/flags) so the next tenant can use it unchanged.

Separately: `boundaryMessageRows` writes a synthetic `KE_ADMIN_<code>` duplicate for every entry, doubling the boundary namespace. If that alias is genuinely needed, a comment explaining which consumer requires it would help; otherwise drop it.

digit-ui-esbuild/products/dashboard/scripts/seed-pt-filter-l10n.mjs
```diff
@@ -0,0 +1,341 @@
...
+    byCode.set(m.code, m.message);
+    byNormalized.set(normalizeKey(m.code), m.message);
+  }
+  return { byCode, byNormalized };
+}
+
+function resolvePackMessage(code, index) {
+  if (index.byCode.has(code)) return index.byCode.get(code);
+  const want = normalizeKey(code);
+  for (const [k, message] of index.byNormalized) {
```
KDwevedi: `byNormalized` is a `Map` — this iterates it to find a key it could fetch directly. `return index.byNormalized.get(want) ?? null;` (with a `has` check to preserve the null contract). Same pattern as the runtime `findMessageLoose` scan; it matters here too because this runs inside the per-ward-code loop.

digit-ui-esbuild/products/dashboard/src/i18n/useDashboardT.js
```diff
@@ -20,7 +20,12 @@ export default function useDashboardT() {
   const [tick, bump] = useReducer((x) => x + 1, 0);
   useEffect(() => {
     ensureMessages();
-    return subscribe(bump);
+    return subscribe(() => {
+      // Language switch writes Employee.locale sync, then bundles arrive later.
+      // Re-fetch the active locale into the side-cache on every signal.
+      ensureMessages();
```
KDwevedi, line 29: Calling `ensureMessages()` from inside the subscribe callback means a successful side-fetch (`notifyStandalone`) re-enters this callback, which calls `ensureMessages()` again. The `messages[locale]` cache short-circuits it today, so there is no loop — but it is one cache-invalidation away from being one. A short comment naming the cache as the loop-breaker, or moving the re-fetch behind a `languageChanged`-only signal, would make that safe by construction rather than by coincidence.

KDwevedi:
Closing this PR as superseded after a current-base and related-dashboard-PR review.

The useful localization work should be re-applied in smaller PRs rather than rebasing this branch wholesale. The main findings are:

- The branch has a hidden current-base build failure: `useCatalogLayout` imports `getUserUuid` from `analyticsService`, but current `master`/`develop` no longer exports it. Git can auto-merge this file and still leave esbuild broken.
- The #1276 catalog/layout persistence work is unrelated to #1108 and is superseded by #1717. The empty-layout fallback also changes an intentionally empty/curated pack into “show every visible tile.”
- #1108 explicitly requires: “if a localisation key is missing in the selected locale, fall back to `en_IN` gracefully (don't render the raw key),” and its acceptance criteria say “No raw `LOCALIZATION_KEYS` visible in any language.” This branch's global locale resolver and tests instead make raw keys/codes the expected fallback. Dynamic dimension values may need a stricter policy, but that must not silently replace the dashboard chrome fallback contract.
- Porting this branch onto #1703 requires preserving the anonymous dashboard guards: disabled protected filter-option calls, the public catalog path, non-persistent public layouts, and public-runtime token isolation. Several of those files conflict directly.
- The header/CSS work is based on the old dashboard search and can resurrect behavior intentionally removed by #1474. Only the long-title wrapping should be re-applied to the current search-free header.
- `seed-pt-filter-l10n.mjs` is live-mutating operational tooling, optionally sends taxonomy strings to MyMemory, and automatically upserts output. It needs a separate ops review and an explicit output/review/apply workflow.

Planned superseding work:

1. Finish #1717 as the focused #1276 fix, including regression coverage for the existing chart (`4x6`) and table (`12x6`) defaults.
2. After the current dashboard/public stack lands, open a fresh #1108 PR from `master` containing only the locale resolver, filter/table/dimension-label behavior, and focused tests. It must preserve `en_IN` fallback for dashboard chrome and #1703's public-mode guards.
3. If still required, ship Portuguese title wrapping as a separate small UI PR against the current header and regenerate CSS from source.
4. Move localization seeding/backfill tooling to a separate operational PR with dry-run/output-first behavior and operator-reviewed writes.

The branch's existing dashboard tests pass, but they are stale relative to the current suite and do not exercise the synthetic-merge import failure or the public-dashboard integration. The two red GitHub checks are unrelated preview-registry certificate/telemetry failures.

Thanks for the work here—the focused pieces will be carried forward through the replacement PRs above.

Issue #1394, opened 2026-07-22
[Bug] dss.DashboardPack and dss.KpiDefinition schema/data not loaded during bootstrap


priyanshu-egov:
For a brand-new state-root tenant, the dashboard renders with no tiles —
`/packs` returns nothing and the analytics service logs "MDMS path not found
for dss.KpiDefinition", degrading gracefully to an empty view rather than
erroring. Per `docs/dashboard-configuration/60-operations.md` §3,
`dss.KpiDefinition` + `dss.DashboardPack` were supposed to be added to the
bootstrap allowlist in PR #1062, but on this tenant neither the schema
registration nor the seed data came through — confirmed by checking the live
mdms-v2 masters post-bootstrap.

This is also independently documented as a gap in
`docs/dashboard-configuration/30-view-access.md:173`: "`dss.*` is NOT copied
by `tenant_bootstrap` — a fresh tenant has actions but no
KpiDefinition/DashboardPack." — i.e. the two docs disagree on whether this
was actually fixed, which suggests the fix regressed or was incomplete
(e.g., schema registration missing even where data copy exists, or vice versa).

## Steps to reproduce
1. Bootstrap a new state-root tenant via `digit-mcp` `tenant_bootstrap`.
2. Check mdms-v2 for `dss.KpiDefinition` and `dss.DashboardPack` schema +
   records at the new tenant.
3. Open the employee dashboard for that tenant.

KDwevedi:
Fixed by #1408, which makes `tenant_bootstrap` seed `dss.KpiDefinition` + `dss.DashboardPack` from the repo (as a floor, like the l10n packs) instead of copying them from a source root — so a fresh tenant gets the catalog with no source and no manual step. #1400 is related: it adds the diagnostic warning when a source has no catalog, plus `enable-dashboard.sh` to seed it on an already-running deployment.

KDwevedi:
Fixed by #1408 (merged to develop 2026-07-23). Closing.

PR #1408, opened 2026-07-23
feat(bootstrap): seed dashboard catalog from repo, not a source tenant (#1394)

KDwevedi:
> **Draft — design review wanted before I validate live.** This changes what *every* `tenant_bootstrap` run does, so I want eyes on the mechanism first.
>
> **Stacked on #1400** (uses its canonical schema files and the `export-l10n.mjs` generator pattern). Base is `develop`, but the diff will show #1400's commits until it merges — review the two commits unique to this branch, or read #1400 first. I'll rebase onto `develop` once #1400 lands.

## The problem (#1394)

`dss.KpiDefinition` and `dss.DashboardPack` are **tenant-invariant** platform definitions — the bootstrap code's own comment says it: *"platform-level definitions… with no tenant identity inside."*

`tenant_bootstrap` seeds them by **copying from a source root tenant**. That needs a correct "golden source" to exist, and when it doesn't — a fresh root whose source has no catalog — it copies nothing, reports success, and the dashboard renders empty. That is #1394's root cause. #1400 made bootstrap *warn* about this; it did not fix it.

## The fix — same pattern the l10n packs already use

The localization packs solved the identical problem by embedding the repo data as a constant and seeding it as a floor (`DASHBOARD_L10N_PACKS`, `mdms-tenant.ts:15`). This does the same for the catalog and schemas.

- **`gen-dashboard-catalog.mjs`** generates `dashboard-catalog-seed.ts` from the repo source of truth (`local-setup/db/dss-mdms-seed/schemas` + `ansible/nairobi-mdms/mdms/dss`), with `--check` for drift — same shape as `export-l10n.mjs`.
- **Step 1b** registers the canonical `dss.*` schemas, create-if-absent.
- **Step 2b** seeds `dss.KpiDefinition` + `dss.DashboardPack` from the embedded catalog as a **floor**: a live record from the source copy or a prior run wins (first-writer-wins); a fresh/empty root gets the full 39 defs + pack.
- `dss.DashboardConfig` is **not** floored — `allowedRoles`/`numberFormat`/`departmentScoping` are tenant-specific and stay operator-owned (`enable-dashboard.sh`). Its schema is registered so bootstrap can seed it.

## Why this and not the alternatives

- **vs. copy-from-source (today):** invariant product data shouldn't depend on a peer tenant being correct. That coupling *is* the bug.
- **vs. a deploy/DDH step:** DDH is being removed, and the team stance is that `tenant_bootstrap` owns seeding — so the invariant data stays in bootstrap, just sourced from the repo rather than a tenant.
- **Floor, not replace:** create-if-absent preserves any KPI a source genuinely customized, and matches the l10n floor's first-writer-wins semantics.

## Open questions for review

1. **Floor vs. drop copy-from-source.** I kept the existing `dss.*` copy-from-source *and* added the floor (redundant but harmless, and it preserves source customizations). Should copy-from-source for `dss.*` be removed entirely now the repo is authoritative?
2. **Stale registered schema.** Create-if-absent won't reconcile a source that already registered a *stale* `dss.*` schema (mdms-v2 `_update` is 501). Bootstrap can't fix that from the API; it stays a DB-level op. Acceptable, or should bootstrap detect and flag it?
3. **`DashboardConfig` default.** Left entirely to the operator. Should bootstrap seed a canonical default record (operator overrides) so the nav gate works out of the box, or is per-tenant-only correct?

## Not yet done (deliberately)

No live bootstrap validation — holding it until the mechanism is agreed. Once it is, I'll bootstrap a fresh root on a dev box and confirm the catalog lands with no source, no manual step.

Fixes #1394

## Summary by CodeRabbit

* **New Features**
  * Added a complete dashboard catalog, including KPI definitions, dashboard packs, layouts, and validation schemas.
  * Added a guided installer to enable or repair dashboards on running deployments, with dry-run, update, verification, role, localization, and cache-management options.
  * Added dashboard localization resources for English and Portuguese.

* **Bug Fixes**
  * Bootstrap now detects corrupted schema records and missing dashboard data, reports warnings, and avoids copying invalid records.

* **Documentation**
  * Updated dashboard setup, operations, seed-data, and troubleshooting guidance.


KDwevedi:
Addressed the six findings.

**This PR (`f43f28f`)**
- **Source dss schemas winning (Major):** Step 1 copied the source's dss.* schemas before Step 1b registered the canonical ones, so the stale source definition won (schema code is immutable). Step 1 now skips the `DASHBOARD_CATALOG_SCHEMAS` keys, so Step 1b's canonical registration lands.

**#1400 (fixed on its branch in `d779c56`, merged in here)**
- **all-keys-resolve only checked titleKey (Major):** now unions titleKey + subtitleKey + labelKey + labelKeys[*]. Verified on bomet — en_IN and pt_PT resolve across all key families.
- **StateInfo languages query missing `isactive`:** filtered, both sites.
- **`--help` symptom→cause table overclaim** in both READMEs: reworded — `--help` is usage/flags; the runbook table is the PR comment that graduates into docs on merge.

**Not changing — `open` measure:** real in `cl_table_ward_performance` / `cl_chart_complaints_over_time`, but that's the canonical bomet-live catalog, shipped faithfully by both PRs by design. Whether `open` is produced by the grain at runtime needs verifying against the backend; a catalog-data fix is a separate concern from seeding it, so it's out of scope here. Worth its own issue if the column is genuinely empty in production.

vinothrallapalli-eGov (approved):

PR #1409, opened 2026-07-23
fix(pgr): break the reverse-geocode feedback loop in the complaint map

dhruv-1001:
Fixes the **continuous API invocation / map repeatedly repositioning** half of #1380 (symptom 4). Companion to #1392 (merged) and #1401, which cover the boundary-dropdown symptoms.

## Problem

`GeoLocations.js` syncs map state *from* the form with the **whole `formData` object** as the effect dependency:

```js
useEffect(() => {
  ... fetchAddress(lat, lng); ...
}, [formData, config.key]);
```

Every `onSelect` the component itself issues rebuilds `formData` through the wizard's `patch` (`CreatePGRFlowV2.tsx` Step1Map), so this effect re-runs after every fetch. The cycle:

1. A pin drop (or the mount-time seeding of the default centre) calls `fetchAddress` → Nominatim reverse geocode.
2. If Nominatim **fails, is rate-limited, or returns no `display_name`**, `fetchAddress` writes `{lat, lng, ward}` — **no `address`** — into the form.
3. New `formData` identity → the effect re-runs → saved point has no address → calls `fetchAddress` **again** → goto 2.

The loop is self-sustaining: the rapid-fire requests trip Nominatim's ~1 req/s rate limit, which guarantees the failure path, which keeps the loop alive. Each cycle also resets marker / selected-ward / search-box state, remounting the ward overlay — the visible "map keeps shifting" churn. Because the component seeds the default centre on mount, the loop can start with **no user interaction**, and a shared egress IP (typical hand-run deployment) makes the trigger condition routine.

## Change

Two guards in `GeoLocations.js`, nothing else:

- **Depend on values, not identity.** The sync effect now keys on the saved point's `lat` / `lng` / `address` instead of the `formData` object, so unrelated form writes — including the component's own round-trip — no longer re-run it.
- **Never re-request attempted coords.** A ref records the last reverse-geocode attempt (set *before* the request, so a throwing fetch still counts). The sync effect skips coords already attempted; one failed lookup stays one failed lookup.

## Blast radius

- A genuine restore (re-entering the map step with a pin saved but no address) still reverse-geocodes **once** — the ref only blocks repeats of the same coords.
- User actions (pin drop, suggestion select, search, locate-me) call `fetchAddress` directly and bypass the guard entirely — manual retry at the same spot still works.
- The success path is byte-identical: address, pincode extraction, ward resolution, `PGR_MAP_LOCATION` session save, and the `onSelect` payload are unchanged.

## Verification

Parse/bundle-checked with esbuild. Traced the loop statically: with the guard, the failure path terminates after one request (effect re-run hits the ref match and stops); the success path terminates exactly as before (saved address short-circuits the fetch branch).

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented repeated address lookup attempts when reverse geocoding fails or returns no address.
  - Improved restoration of saved location data, avoiding request loops and reducing unnecessary network activity.
  - Location forms now behave more reliably in rate-limit and service-error scenarios.

### DIFF
--- a/digit-ui-esbuild/products/pgr/src/components/GeoLocations.js
+++ b/digit-ui-esbuild/products/pgr/src/components/GeoLocations.js
@@ -144,6 +144,12 @@ const GeoLocations = ({ t, config, onSelect, formData }) => {
   const mapRef = useRef(null);
   const searchInputRef = useRef(null);
   const hasInitialized = useRef(false);
+  // Coords of the last reverse-geocode ATTEMPT (not success). The formData
+  // sync effect below must never re-request coords that were already tried:
+  // a failed / rate-limited Nominatim call writes {lat, lng} without an
+  // address back into formData via onSelect, and re-fetching on that write
+  // turns one failure into an infinite request loop (CCRS#1380 symptom 4).
+  const lastReverseAttempt = useRef(null);
 
   // Leaflet writes the stroke as an SVG DOM attribute, which doesn't resolve
   // CSS `var()`. Read the runtime accent at mount so the user-drawn polygon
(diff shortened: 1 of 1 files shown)


dhruv-1001:
Linking: this addresses **symptom 4 of #1380** — "Upon selecting a location, the application continuously invokes APIs, causing the map to repeatedly reposition and shift from one location to another."

Status of #1380 across PRs:
- Symptoms 1–2 (boundary dropdown not auto-populated from the map): #1392 (merged — requires `PGR_BOUNDARY_LOWEST_LEVEL` to be set to take effect) and #1401 (open).
- Symptom 3 (citizen portal shows no map overlay): data issue — placeholder boundary geometry; needs re-seeding, not a code change.
- Symptom 4 (continuous API calls / map repositioning): **this PR**.

dhruv-1001:
Linking: this addresses **symptom 4 of #1380** — "Upon selecting a location, the application continuously invokes APIs, causing the map to repeatedly reposition and shift from one location to another."

Coverage of #1380 across the three PRs:
- Symptoms 1–2 (boundary dropdown not auto-populated from the map / wrong default boundary): #1392 (merged) + #1401
- Symptom 4 (continuous API calls, map churn): **this PR**
- Symptom 3 (citizen sees no ward overlay): a data issue — most boundaries carry placeholder geometry (identical unit squares), so there is nothing renderable to draw; needs boundary re-seeding, not a frontend change.

vinothrallapalli-eGov (approved):

PR #1411, opened 2026-07-23
Merged from master

pradeepkumarcm-egov:


Hari-egov:
Closing per review: `release-v2.12-moz` already contains an equivalent-and-better version of this fix — 60c5198c (same geocodeAttemptRef guard strategy) plus hardening master lacks: resolveWard try/catch (QA #12), `response.ok` guard before `.json()`, and geolocation error-code toasts (e017f16c, 97633209). Merging could only preserve the status quo (conflicts resolved 'ours') or regress that QA-verified hardening ('theirs'). Also, the head branch is literally `master` — any commit landing there before merge would silently ride into the release branch. The useful follow-up is the opposite direction (port the moz hardening to master) — tracked by the open geo-loop PRs. The underlying map/boundary-overlay issue is being investigated separately.

Issue #1309, opened 2026-07-17
[Enhancement] Setup deployment using digit registry


dhruv-1001:
## Summary

Today the local/prod stack pulls its images from **five+ different sources**, which makes deploys fragile and hard to reason about. The `docker-compose.egov-digit.yaml` image refs currently come from:

- `registry.preview.egov.theflywheel.in/*` (Flywheel mirror) — ~26 refs
- `egovio/*` (Docker Hub) — ~15 refs
- public Docker Hub (`prom/`, `grafana/`, `mongo`, `postgres`, `openbao/…`, `quay.io/keycloak/…`) — ~8 refs
- `ghcr.io/novuhq/*` (Novu upstream) — 4 refs
- `ghcr.io/subhashini-egov/*` (personal namespace: `digit-mcp`, `default-data-handler`) — 2 refs
- one `*:local` tag with no registry at all (`otp-publisher`)

## Expectation from this ticket

1. **Every CCRS-owned image is published to and pulled from a single registry: the `egovio` Docker Hub org.** No more `ghcr.io/subhashini-egov/*`, no personal namespaces, no `:local`-only defaults, no reliance on the Flywheel mirror for CCRS images.
2. **A clean split between prod and local deploy:**
   - **Prod / default deploy:** pull-only from `egovio` — never builds on the box.
   - **Local dev deploy:** a separate, uniform mechanism where a developer can build and run **specific** services from source while pulling everything else from `egovio`.

### Scope (confirmed)

- **In scope:** only **CCRS-owned images** (source lives in this repo — `backend/`, `utilities/`, `digit-mcp/`, `frontend/`, `configurator/`, `local-setup/docker/`).
- **Out of scope:** third-party/dev-tooling images (postgres, redis, redpanda, minio, elasticsearch, kong, keycloak, openbao, mongo, grafana/loki/promtail/tempo/prometheus/otel, gatus, curl, nginx, novu-api/worker/ws) **and** the upstream eGov platform services (`egovio/egov-*`, whose source lives in other egov repos). These keep their current sources unchanged.

## Current state (findings)

Good news — most of the machinery already exists:

[...]

PR #1422, opened 2026-07-27
feat(deploy): nightly egovio image builds + local selective-build deploy (#1309)

dhruv-1001:
## What & why

Part of #1309. CCRS-owned images are built **nightly from `develop`** and published to a single registry — Docker Hub **`egovio`** — so deploys pull pre-built images instead of building on the box.

**Scope: CI/build only** (6 files, all under `.github/workflows/` + `build/`). The deploy-side half was split out to #1582.

## Changes

- **`nightly-build-develop.yml`** *(new)* — scheduled 19:00 IST + `workflow_dispatch`; calls the reusable pipeline against `develop`.
- **`build-images.yml`** *(new)* — the reusable build → manifest → cleanup pipeline. Builds each service per-arch on **native runners** (`ubuntu-latest` / `ubuntu-24.04-arm`), then stitches a multi-arch manifest.
- **`release-build.yml`** *(new)* — on release published, immutable version tag, no rolling tag, no prune.
- **`build.yml`** — fix the Docker Hub tag-lookup namespace (`edevops/…` → `egovio/…`; pushes always went to `egovio`).
- **`build-config.yml`** — enrol `novu-bridge-endpoint`.
- **`NIGHTLY-BUILDS.md`** — document the nightly as the canonical `egovio` driver.

## Two design decisions worth reviewing

**1. Tags are commit hashes, not dates — `develop-<sha8>`.**
Derived by slicing the SHA `enumerate` already resolved (the same SHA every build leg checks out), so a tag provably names the commit it was built from. Matches `build.yml`'s existing `<branch>-<hash>` shape, so a nightly and a manual dispatch of the same commit produce one tag rather than two.

> ⚠️ **The prune regex `^develop-[0-9a-f]{8,}$` is load-bearing.** 9 files (`local-setup/docker-compose.egov-digit.yaml` + 5 Helm `values.yaml`) pin **`develop-70916ea`** — a 7-char tag — in the same `egovio` repos the prune deletes from. A `{7,}` bound would have deleted live deployment pins.

Retention also had to stop sorting by tag name: dates sorted chronologically for free, hashes do not. It now sorts on the Hub's `last_updated`.

**2. Staging tags are scoped `<prefix>-staging-<run_id>-<run_attempt>-<arch>`.**
Both parts matter. `run_id` is stable across re-runs, so a **cancelled** attempt (which skips `cleanup` via `if: !cancelled()`) would leave tags that the next attempt of the *same run* would consume and stitch into a mixed-attempt manifest. The `manifest` job refuses to publish unless both arch tags from *this attempt* exist.

## Review findings addressed

| Finding | Fix |
|---|---|
| Staging tags not scoped to a run/attempt — manifest only checked a tag *existed* | Scoped by `run_id` + `run_attempt`; incomplete builds are skipped loudly |
| `release-build.yml` shared `staging_prefix: release` across concurrent releases | Resolved by the same scoping (deriving from the tag was rejected: release tag names aren't constrained to Docker's tag grammar) |
| `build.yml` `grep … \|\| true \| sort \| tail` precedence — sort/tail never saw grep's output → multi-line `LATEST_TAG` → `Invalid format` | Grouped the grep; arch suffixes excluded |
| Script injection via `inputs.ref` / `inputs.version_tag` | Passed through `env:`; resolved tag validated against Docker's tag grammar |
| Manifest matrix span ~19 runner VMs for one `imagetools` call each | One job looping over images; per-image isolation kept |
| `del()` returned the trailing `echo`'s status, so failed deletes reported success | HTTP-status based; 404 idempotent; job fails on real errors |
| Prune read only one `page_size=100` page | Follows `next` until exhausted |
| `docker/login-action@v2` unsupported | Both references → `@v4` |

Left open deliberately: the Docker Hub JWT/tag-listing duplication between `build-images.yml` and `build.yml`. A shared composite action is the right fix but belongs in its own PR.

## Validation

The workflow was **executed locally**, not just reviewed. `act` couldn't run it (`actions/checkout` insists on syncing a remote; the branch under test was local-only), so each `run:` block was extracted from the committed YAML and executed verbatim with the env a runner injects, pushing to a local registry so nothing could reach Docker Hub:

```
Resolve ref to an immutable SHA  -> 527a7b95fde3b6c9...
Flatten manifest + version tag   -> version_tag=develop-527a7b95
targets (real yq enumeration)    -> ['digit-ui-esbuild', 'configurator']
build legs (amd64)               -> OK, staging tags pushed
manifest (one arch only)         -> SKIPPED: "staging tags ... are incomplete"
manifest (both arches present)   -> Published 2 image(s)
```

That skip is the point — with an incomplete build the gate **refused to publish**. Both images were then pulled from the registry and ran, serving HTTP 200. The grep fix was verified against its real failure input (old → 3 lines, new → 1 correct tag, no-match → empty).

**arm64:** builds on native `ubuntu-24.04-arm` runners (free for this public repo), not qemu. Already proven in practice — `egovio/configurator` and `egovio/digit-ui-esbuild` carry `master-cae07a3-arm64` tags today, and their published manifests contain two real digests (`linux/amd64` + `linux/arm64`). The local run couldn't build arm64 (no qemu on that host), so its arm64 leg used an amd64 stand-in purely to exercise `imagetools create`.

## Note on the red CI

The failing checks are **pre-existing on `master`**, not introduced here — verified by running them against `master` alone. Summary in the PR comments.

## Known follow-ups

- Targets `master` (the default branch) because scheduled workflows only fire from there — on `develop` the cron would be dispatch-only.
- The first run must be a manual dispatch to mint the initial nightly tags.
- Prune/cleanup needs the token to have Docker Hub delete scope; the job now **fails** rather than warning if a delete is rejected.


dhruv-1001:
👋 Opening this as a **draft** to get feedback on the approach before we finalize. Summary of the intent and the open decisions:

### The idea
- **One source for CCRS-owned images: Docker Hub `egovio`.** A nightly GitHub Action builds every service listed in `build/build-config.yml` from `develop` and pushes `egovio/<name>:nightly-develop` (rolling) + `egovio/<name>:develop-YYYYMMDD` (immutable), keeping the newest 5 dated tags. Build and deploy stay **decoupled** — the action only builds/pushes; the server pulls on its own cadence.
- **Core/platform services keep fixed pinned versions** (`egovio/egov-*:…`), bumped by commit when needed — not part of the nightly.
- **Local dev** gets `local-deploy.sh` to build specific services from source and pull the rest.

### Decisions I'd like a sanity check on
1. **`develop` vs `master` for the nightly.** GitHub only fires `schedule:` cron from the **default branch (`master`)** — so on `develop` this runs **on-demand via `workflow_dispatch`** (and always builds `develop` via `ref: develop`). That's intentional *for now*: we can validate/produce images against develop by dispatching manually. **To get the automatic 7 PM nightly, this workflow needs to also land on `master`.** Is everyone OK with promoting just this workflow file to `master`, or do you prefer keeping it dispatch-only on develop until we're confident?
2. **`novu-bridge-endpoint` treated as CCRS** (our bridge code → built + moved to `egovio`). The upstream Novu images (`ghcr.io/novuhq/*`) are deliberately left where they are. Shout if you'd rather keep the bridge on flywheel.
3. **`inbox` and `token-exchange-svc` left on flywheel** — their tags don't exist on `egovio` and they're not built from this repo. Closing this needs the core team to publish those tags to `egovio`. OK to track as a follow-up?
4. **Retention = newest 5 `develop-YYYYMMDD` per image.** Enough for rollback headroom?
5. **Prune uses the `edevops` Docker Hub token** — it needs delete scope on the `egovio` org for pruning to actually remove tags (it warns-and-continues otherwise). Can someone confirm the token has that?
6. **Multi-arch (amd64+arm64)** via native runners — fine assuming this repo has arm runners available; happy to drop arm64 or use QEMU if not.

Vendor/infra images (postgres, redis, kong, grafana, etc.) and the ansible **server deploy path are untouched**. Feedback welcome before I take this out of draft. 🙏

subhashini-egov:
### Decisions I'd like a sanity check on

> 1. **`develop` vs `master` for the nightly.** GitHub only fires `schedule:` cron from the **default branch (`master`)** — so on `develop` this runs **on-demand via `workflow_dispatch`** (and always builds `develop` via `ref: develop`). That's intentional *for now*: we can validate/produce images against develop by dispatching manually. **To get the automatic 7 PM nightly, this workflow needs to also land on `master`.** Is everyone OK with promoting just this workflow file to `master`, or do you prefer keeping it dispatch-only on develop until we're confident? 
**We need to build nightly from develop. Yes, that's fine.**

> 2. **`novu-bridge-endpoint` treated as CCRS** (our bridge code → built + moved to `egovio`). The upstream Novu images (`ghcr.io/novuhq/*`) are deliberately left where they are. Shout if you'd rather keep the bridge on flywheel.
**Yes, novu-bridge is eGov code. Novu images can be left where they are.**

> 3. **`inbox` and `token-exchange-svc` left on flywheel** — their tags don't exist on `egovio` and they're not built from this repo. Closing this needs the core team to publish those tags to `egovio`. OK to track as a follow-up?
**Not sure what is this inbox image? token-exchange-svc for now, is ok on flywheel. But we should move that also to egovio.**

> 4. **Retention = newest 5 `develop-YYYYMMDD` per image.** Enough for rollback headroom?
**Yes. But ensure you pull the latest nightly from develop.**

> 5. **Prune uses the `edevops` Docker Hub token** — it needs delete scope on the `egovio` org for pruning to actually remove tags (it warns-and-continues otherwise). Can someone confirm the token has that?
**Not sure of this. Vinoth can help**

> 6. **Multi-arch (amd64+arm64)** via native runners — fine assuming this repo has arm runners available; happy to drop arm64 or use QEMU if not.

**We will need ARM runners. Internal environments mandate this now for cost efficiencies**

local-setup/docker-compose.egov-digit.yaml
```diff
@@ -1561,10 +1562,10 @@ services:
   # ===========================================================================
   digit-ui:
     # Parameterized so deploys can pin a nightly-develop (or other) build via
-    # DIGIT_UI_IMAGE; default keeps the prior pinned tag. NOTE: this is the
+    # DIGIT_UI_IMAGE; default tracks the egovio nightly. NOTE: this is the
     # legacy micro-ui container — the modern esbuild bundle is laid in via
     # build_digit_ui or served by host nginx (see build/NIGHTLY-BUILDS.md).
-    image: ${DIGIT_UI_IMAGE:-registry.preview.egov.theflywheel.in/digit-ui:pgr-fixes}
+    image: ${DIGIT_UI_IMAGE:-egovio/digit-ui:nightly-develop}
```
vinothrallapalli-eGov: **[correctness — confirmed]** This default (`egovio/digit-ui:nightly-develop`, same for `egovio/novu-bridge-endpoint:nightly-develop` at ~line 2049) points at a tag that doesn't exist on Docker Hub (verified 404), and the nightly workflow that would create it only fires from the default branch `master` while this PR merges to `develop`.

On merge, the canonical `docker compose up -d` fails for every developer and local-setup CI: digit-ui is not profile-gated and errors with `manifest for egovio/digit-ui:nightly-develop not found`. The gap persists until `nightly-build-develop.yml` reaches master AND its first run completes. This also moves the default from an immutable pin to a rolling tag, so any bad develop nightly thereafter breaks every fresh `up` with no change in this repo's checked-out code.

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,180 @@
...
+      - name: Create version (+ rolling) manifest
+        run: |
+          set -euo pipefail
+          IMG="${PUSH_NAMESPACE}/${{ matrix.image }}"
+          VERSION_TAG='${{ needs.enumerate.outputs.version_tag }}'
+          ROLLING_TAG='${{ inputs.rolling_tag }}'
+          SP='${{ inputs.staging_prefix }}'
+          # Both arch tags must exist (fresh from this run) or we refuse to publish
+          # — a half-built image must never become the version/rolling tag.
+          docker buildx imagetools inspect "$IMG:${SP}-staging-amd64" >/dev/null
```
vinothrallapalli-eGov: **[correctness — confirmed]** The manifest job runs on build failure (`if: !cancelled()` with `fail-fast: false` in the build matrix), and this `imagetools inspect` guard only checks that the staging tags *exist* — not that they came from this run.

Scenario: nightly run N's arm64 build of pgr-services fails on a flaky runner. The inspect of `…-staging-arm64` still succeeds because run N-1 left that tag on Docker Hub, so the job publishes `develop-YYYYMMDD` and moves `nightly-develop` to a manifest mixing today's amd64 with **yesterday's arm64** — contradicting the in-file comment that a half-built image must never become the version/rolling tag. Consider gating on `needs.build.result == 'success'` per-image, or embedding the run ID in the staging tags and verifying it.

.github/workflows/release-build.yml
```diff
@@ -0,0 +1,25 @@
...
+  group: release-${{ github.event.release.tag_name }}
+  cancel-in-progress: false
+
+jobs:
+  build:
+    uses: ./.github/workflows/build-images.yml
+    with:
+      ref: ${{ github.event.release.tag_name }}
+      version_tag: ${{ github.event.release.tag_name }}
+      staging_prefix: release
```
vinothrallapalli-eGov, line 30: **[correctness — confirmed]** All releases share the single `staging_prefix: release`, but the concurrency group is keyed per release tag (`release-<tag>`) — so two releases published close together run in parallel and overwrite each other's `egovio/<image>:release-staging-{amd64,arm64}` tags before the manifest jobs stitch them.

Publishing v2.12.0 and v2.12.1 in the same window can permanently publish an arch-mixed, wrong-version image under an immutable release tag (e.g. `pgr-services:v2.12.1` on amd64 silently running v2.12.0 code). Fix: derive the staging prefix from the tag (`release-${{ github.event.release.tag_name }}`), or use a single shared concurrency group for all releases.

.github/workflows/build.yml
```diff
@@ -165,13 +165,13 @@ jobs:
...
             -H "Authorization: JWT $TOKEN" \
-            "https://hub.docker.com/v2/repositories/$DOCKER_USERNAME/$SERVICE_NAME/")
+            "https://hub.docker.com/v2/repositories/egovio/$SERVICE_NAME/")
 
           if [ "$HTTP_CODE" -ne 200 ]; then
             NEXT_TAG="${BRANCH}-${COMMIT_HASH}"
           else
             EXISTING_TAGS=$(curl -s -H "Authorization: JWT $TOKEN" \
-              "https://hub.docker.com/v2/repositories/$DOCKER_USERNAME/$SERVICE_NAME/tags?page_size=100" \
+              "https://hub.docker.com/v2/repositories/egovio/$SERVICE_NAME/tags?page_size=100" \
```
vinothrallapalli-eGov, line 174: **[correctness — confirmed]** Switching the tag lookup to the fixed `egovio` namespace exposes a latent precedence bug here: `grep "^${BRANCH}-${COMMIT_HASH}" || true | sort -V | tail -n 1` parses as `grep … || (true | sort | tail)` — the sort/tail never run on the grep output.

Now that the lookup hits the namespace this workflow actually pushes to, a re-run for an already-built commit matches `dev-<hash>`, `dev-<hash>-amd64`, `dev-<hash>-arm64`, making `LATEST_TAG`/`NEXT_TAG` multi-line; `echo "tag=$NEXT_TAG" >> $GITHUB_OUTPUT` then writes bare non-key=value lines and the step hard-fails with `Unable to process file command 'output' … Invalid format`. Fix the grouping: `{ grep … || true; } | sort -V | tail -n 1` (and consider excluding `-amd64`/`-arm64` suffixes from the match).

vinothrallapalli-eGov (commented):
Automated high-effort review (multi-agent find + independent verification). 28 candidate findings were checked; 5 were empirically refuted and dropped. The 10 confirmed findings are attached as inline comments — 4 correctness issues in the publish pipeline, 4 in local-deploy.sh, and 2 CI-efficiency/duplication cleanups.

**Most severe:** the compose defaults now point at `egovio/*:nightly-develop` tags that don't exist yet (verified 404) and can't be created until the nightly workflow lands on `master`, so `docker compose up -d` breaks on merge.

local-setup/local-deploy.sh
```diff
@@ -0,0 +1,306 @@
...
+
+cd "$SCRIPT_DIR"
+if [ ${#CHOSEN[@]} -gt 0 ]; then
+  set -- docker compose -f "$BASE_COMPOSE" -f "$OVERRIDE_FILE" up -d --build
+else
+  set -- docker compose -f "$BASE_COMPOSE" up -d
+fi
+if [ ${#EXTRA_ARGS[@]} -gt 0 ]; then
+  info "${*} ${EXTRA_ARGS[*]}"
+  "$@" "${EXTRA_ARGS[@]}"
```
vinothrallapalli-eGov: **[correctness — confirmed]** `EXTRA_ARGS` from `--` land after `up -d`, but the documented primary use case (`-- --profile notifications`, docs/LOCAL-DEPLOY.md example 6) is a **global** compose flag, not an `up` flag — `docker compose … up -d --build --profile notifications` fails with `unknown flag: --profile` (verified on compose v5.3.1) and nothing starts.

Insert `EXTRA_ARGS` between `docker compose` and `up` (or add a first-class `--profile` option), since global flags are positional.

local-setup/local-deploy.sh
```diff
@@ -0,0 +1,306 @@
...
+    --list)
+      DO_LIST=1; shift ;;
+    --no-up)
+      NO_UP=1; shift ;;
+    --)
+      shift
+      while [ $# -gt 0 ]; do EXTRA_ARGS+=("$1"); shift; done
+      ;;
+    -h|--help)
+      sed -n '2,40p' "${BASH_SOURCE[0]}" | sed 's/^# \{0,1\}//'
```
vinothrallapalli-eGov: **[correctness — confirmed]** `sed -n '2,40p'` overshoots the header comment block, which ends at line 33 — so `--help` emits lines 34–40 (`set -euo pipefail`, `SCRIPT_DIR=`, `REPO_ROOT=`, `BASE_COMPOSE=`, `OVERRIDE_FILE=`, `BUILD_CONFIG=`) as raw shell source after the usage text. Trim the range to the actual header (`2,33p`) or stop at the first non-comment line.

local-setup/local-deploy.sh
```diff
@@ -0,0 +1,306 @@
...
+
+# ---------------------------------------------------------------------------
+# read_targets — flatten build/build-config.yml into TAB rows:
+#   <image-name> <work-dir> <dockerfile|-> <mode:maven|plain>
+# mode=maven iff the entry uses the shared build/maven/Dockerfile (repo-root
+# context + WORK_DIR arg); everything else is a plain build (context=work-dir).
+# Identical flatten to local-setup/ansible/files/nightly-build-push.sh.
+# ---------------------------------------------------------------------------
+read_targets() {
+  python3 - "$BUILD_CONFIG" "$MAVEN_DOCKERFILE" <<'PY'
```
vinothrallapalli-eGov: **[simplification — confirmed]** `read_targets()` is a byte-for-byte copy of the PyYAML flatten in `local-setup/ansible/files/nightly-build-push.sh` (the comment says so), and `build-images.yml` adds a third yq reimplementation — with **already-divergent** maven detection: the shell drivers use `endswith(maven_df)` while the workflow uses exact string equality, so a `./build/maven/Dockerfile` entry builds differently per driver.

Consider extracting one shared helper (e.g. `build/read-build-targets.py`) sourced by both shell scripts, with the workflow consuming the same output.

local-setup/local-deploy.sh
```diff
@@ -0,0 +1,306 @@
...
+DO_LIST=0
+NO_UP=0
+
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --build)
+      shift
+      [ $# -gt 0 ] || die "--build requires an argument (a service name or space-separated set)."
+      # Split the space-set into tokens and accumulate.
+      for tok in $1; do BUILD_SVCS+=("$tok"); done
```
vinothrallapalli-eGov: **[correctness — confirmed]** `for tok in $1` word-splits the `--build` argument unquoted, so it's also subject to pathname glob expansion in the caller's cwd — `./local-deploy.sh --build 'pgr-*'` from `local-setup/` expands against directory contents and produces a baffling *"'docker-compose.egov-digit.yaml' is not an image-name"* error instead of one about the actual input. `read -ra` into an array (or `set -f` around the split) keeps the space-separated semantics without globbing.

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,180 @@
...
+          elif [ -n "$DOCKERFILE" ]; then
+            docker buildx build --platform linux/${{ matrix.arch }} \
+              -f "$DOCKERFILE" -t "$TAG" --push "$WORKDIR"
+          else
+            docker buildx build --platform linux/${{ matrix.arch }} \
+              -t "$TAG" --push "$WORKDIR"
+          fi
+
+  manifest:
+    name: Stitch ${{ matrix.image }} multi-arch tags
```
vinothrallapalli-eGov: **[efficiency]** The manifest job fans out a matrix job per image (~19 runner VMs), each doing only a login plus one `imagetools create` — a few seconds of API calls per VM, with boot/checkout overhead dwarfing the real work and consuming org runner slots every nightly. The adjacent prune job already shows the cheaper pattern: a single job looping over `needs.enumerate.outputs.images`.

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,180 @@
...
+    if: ${{ inputs.prune_develop && !cancelled() }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Prune dated tags on Docker Hub (egovio namespace)
+        env:
+          DOCKER_USERNAME: ${{ vars.DOCKER_USERNAME }}
+          DOCKER_PASSWORD: ${{ secrets.DOCKER_ACCESS_TOKEN }}
+        run: |
+          set -euo pipefail
+          TOKEN=$(curl -s -X POST "https://hub.docker.com/v2/users/login/" \
```
vinothrallapalli-eGov, line 237: **[simplification]** This re-implements the Docker Hub JWT login + tag-listing curl/jq dance that already exists in build.yml's *Generate the Next Tag* step (same `users/login` POST, same `jq -r .token`, same `tags?page_size=100`). When the Hub auth flow changes, each copy must be found and fixed separately. A shared composite action (e.g. `.github/actions/dockerhub-api`) would give one place to fix.

dhruv-1001:
Thanks both — all 10 confirmed findings triaged. **This PR is now CI/build only**; the deploy-side half is #1582.

### Fixed here (`a3bea92b`)

| Finding | What changed |
|---|---|
| `build-images.yml` staging tags not scoped per run | Staging tags now carry `github.run_id`, so a manifest can only be stitched from images **this** run produced. Your flaky-arm64 scenario now skips that image loudly instead of publishing today's amd64 + yesterday's arm64. |
| `release-build.yml` shared `staging_prefix: release` | Resolved by the same run-id scoping — two concurrent releases can no longer collide. I did **not** derive the prefix from the tag, because release tag names aren't constrained to Docker's tag grammar and that would trade one failure for another. |
| `build.yml` `grep … \|\| true \| sort \| tail` precedence | Grouped as `{ grep … \|\| true; } \| sort \| tail`, plus the `-amd64`/`-arm64` suffixes excluded. Verified against the exact failure input: old → 3 lines (the `Invalid format` you predicted), new → 1 correct tag, no-match → empty. |
| Template injection via `inputs.ref` / `inputs.version_tag` | Both now arrive through `env:`. The resolved version tag is also validated against Docker's tag grammar before it's used or pushed. |
| Manifest job fans out ~19 runner VMs | Collapsed into a single job looping over `needs.enumerate.outputs.images`, matching the prune job's pattern. Per-image isolation is kept inside the loop. |
| *(follow-on)* run-scoped tags would accumulate 38/nightly | `prune` → `cleanup`: deletes this run's staging tags, and runs on `!cancelled()` so a **failed** run doesn't leak them either. Dated-tag pruning still gated on the caller opting in. |

### Moved to #1582

- `local-deploy.sh` `--profile` placement — confirmed against compose v5.1.4: rejected after `up`, accepted before it. Now a first-class `--profile` option.
- `local-deploy.sh` `--build` word-split/glob — `read -ra`.
- `local-deploy.sh` `--help` `sed -n '2,40p'` overshoot — stops at the first non-comment line.
- `LOCAL-DEPLOY.md` expired dated-tag example — `<YYYYMMDD>` placeholder + a command to list retained tags.

### Dropped rather than fixed

The compose repoints are **gone from this PR**. You were right that `egovio/digit-ui:nightly-develop` and `egovio/novu-bridge-endpoint:nightly-develop` are 404 — and since compose validates every image before starting any service, that fails the whole `up`, not just those two.

Two further things I found while verifying: `master` has since migrated **every** compose image to `egovio/*` independently (and pins digit-ui to an immutable `egovio/digit-ui-esbuild:master-<sha>`), so the repoints were largely redundant; and the two `FLAG(#1309)` comments were wrong — `egovio/inbox:state-tenant-fix` and `egovio/token-exchange-svc:keycloak-deploy` both exist, published 2026-08-03. They went away with the hunks.

### Deferred, with reasons

- **Docker Hub JWT/tag-listing duplication** (`build-images.yml` ↔ `build.yml`) — agreed, but a shared composite action is a bigger refactor than this PR should carry.
- **`read_targets` flatten duplication** — now spans this PR and #1582, so a partial fix would straddle the split. The `endswith` vs exact-match divergence is latent today: every `build-config.yml` entry uses the exact `build/maven/Dockerfile` path. Noted in #1582.

dhruv-1001:
Retargeted from `develop` to **`master`** (the repo default). No rebase or force-push was needed — this branch's base commit is already an ancestor of `master`, and `develop` has no commits `master` lacks, so the diff is byte-identical (6 files, +356/-5) and the review threads above are preserved.

This also resolves the first known follow-up: scheduled workflows only fire from the default branch, so the nightly cron goes live on merge rather than staying dispatch-only. #1582 was retargeted to `master` the same way.

local-setup/docker-compose.egov-digit.yaml
dhruv-1001: **Dropped, not fixed.** You were right on both counts — `egovio/digit-ui:nightly-develop` and `egovio/novu-bridge-endpoint:nightly-develop` are 404, and since compose validates every image before starting any service that fails the whole `up`, not just those two.

Rather than patch the tags, the compose repoints are **removed from this PR entirely** (`fcf037e8`); `local-setup/docker-compose.egov-digit.yaml` is back to base.

Two things I found while verifying your point: `master` has since migrated *every* compose image to `egovio/*` on its own — more completely than this PR attempted — and pins digit-ui to an immutable `egovio/digit-ui-esbuild:master-<sha>`. And the two `FLAG(#1309)` comments were wrong: `egovio/inbox:state-tenant-fix` and `egovio/token-exchange-svc:keycloak-deploy` both exist (published 2026-08-03). They went away with the hunks.

Repointing defaults onto the nightly channel is now a follow-up, gated on the nightly having actually published.

.github/workflows/build-images.yml
dhruv-1001: **Fixed** in `a3bea92b`. Per-arch staging tags now carry `github.run_id`:

```
<prefix>-staging-<run_id>-<arch>
```

so a manifest can only ever be stitched from images *this* run produced. Your flaky-arm64 scenario now fails the `imagetools inspect` for that image and skips it loudly, instead of silently stitching today's amd64 to yesterday's arm64.

I went with run-id scoping over `needs.build.result == 'success'` because the latter is all-or-nothing across the matrix — one image failing would block every other image's manifest. Scoping keeps per-image granularity.

Follow-on: run-scoped tags would accumulate ~38/nightly, so the `prune` job became `cleanup` and deletes this run's staging tags. It runs on `!cancelled()` so a *failed* run doesn't leak them either.

.github/workflows/release-build.yml
dhruv-1001: **Fixed** — by the run-id scoping in the sibling thread, rather than the prefix change you suggested. Two concurrent releases now get different `run_id`s, so their staging tags can't collide even though they share `staging_prefix: release`.

I deliberately did *not* derive the prefix from `github.event.release.tag_name`: release tag names aren't constrained to Docker's tag grammar, so that risks trading a collision for an invalid-tag failure. Added a comment in `release-build.yml` recording why the per-tag concurrency group is safe, so this doesn't look like an oversight to the next reader.

.github/workflows/build.yml
dhruv-1001: **Fixed** in `a3bea92b` — grouped exactly as you suggested, plus the arch suffixes excluded:

```bash
LATEST_TAG=$({ echo "$EXISTING_TAGS" \
  | grep -E "^${BRANCH}-${COMMIT_HASH}([._-][A-Za-z0-9._-]*)?$" \
  | grep -Ev -- '-(amd64|arm64)$' || true; } \
  | sort -V | tail -n 1)
```

Verified against your exact scenario (`dev-<hash>`, `dev-<hash>-amd64`, `dev-<hash>-arm64` all present): old form returned **3 lines** — the `Invalid format` failure you predicted — new form returns the single correct tag, and the no-match case returns empty without crashing.

local-setup/local-deploy.sh
dhruv-1001: **Moved to #1582 and fixed there.** `--profile` is now a first-class option inserted *before* the subcommand, rather than something passed after `--`.

Confirmed your finding empirically on compose v5.1.4: rejected after `up`, accepted before it. `--` is now documented as carrying `up` flags only (`--wait`, `--force-recreate`, …), and the `docs/LOCAL-DEPLOY.md` example 6 was corrected.

local-setup/local-deploy.sh
dhruv-1001: **Moved to #1582 and fixed there.** Rather than hard-code `2,33p` — which just moves the breakage to the next time the header changes length — it now stops at the first non-comment line:

```bash
awk 'NR==1 {next} /^#/ {sub(/^# ?/, ""); print; next} {exit}' "${BASH_SOURCE[0]}"
```

Verified: `--help` output no longer contains `set -euo pipefail`, `SCRIPT_DIR=`, or any other raw source.

local-setup/local-deploy.sh
dhruv-1001: **Moved to #1582**, and deferred there with the reasoning recorded in the PR body — resolving here since `local-deploy.sh` is no longer part of this PR.

Agreed on the substance. I didn't unify it because the three copies now straddle two PRs, so a partial extraction would leave the codebase in a worse intermediate state than either endpoint. Worth noting the divergence you spotted is currently **latent**: every `build-config.yml` entry uses the exact string `build/maven/Dockerfile`, so `endswith` and `==` agree on all of them today. It becomes real the first time someone adds a differently-rooted maven path.

The `build-images.yml` yq reimplementation stays tracked in the JWT-duplication thread below, which I've left open.

local-setup/local-deploy.sh
dhruv-1001: **Moved to #1582 and fixed there** — `read -ra` as you suggested:

```bash
read -ra _toks <<<"$1"
for tok in "${_toks[@]}"; do [ -n "$tok" ] && BUILD_SVCS+=("$tok"); done
```

Verified from `local-setup/`: `--build 'pgr-*'` now reports `'pgr-*' is not an image-name in build/build-config.yml` instead of blaming a globbed directory entry.

.github/workflows/build-images.yml
dhruv-1001: **Fixed** in `a3bea92b` — collapsed to a single job looping over `needs.enumerate.outputs.images`, matching the pattern you pointed at in the prune job. Per-image isolation is preserved inside the loop: a failed image is reported and skipped, the others still publish, and the job exits non-zero if any were skipped.

.github/workflows/build-images.yml
dhruv-1001: **Acknowledged — not fixed. Leaving this thread open** rather than resolving it, since the duplication is still present in this PR's code.

You're right that it's the same Hub JWT + tag-listing dance as `build.yml`'s *Generate the Next Tag*. A `.github/actions/dockerhub-api` composite is the right shape, but it touches a workflow outside this PR's scope and would be better reviewed on its own — I'd rather not bundle a shared-action refactor into a PR that's already been split once.

Note the count didn't get worse here: the `prune` job became `cleanup` and still does exactly one login, now covering both staging-tag deletion and dated-tag pruning.

docs/LOCAL-DEPLOY.md
```diff
@@ -0,0 +1,164 @@
...
+  `novu-bridge-endpoint` to `egovio/novu-bridge-endpoint:nightly-develop`).
+- **`develop-YYYYMMDD`** — an immutable daily snapshot for reproducible pins /
+  rollback. The nightly retains the **newest 5** per image and prunes older ones.
+
+A pure `./local-deploy.sh` therefore runs whatever `nightly-develop` currently
+points at. To **pin a reproducible snapshot** instead of the rolling tag, set the
+image env var the compose exposes, for example:
+
+```bash
+DIGIT_UI_IMAGE=egovio/digit-ui:develop-20260726 ./local-deploy.sh
```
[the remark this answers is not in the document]
dhruv-1001: **Moved to #1582 and fixed there** — resolving here since `docs/LOCAL-DEPLOY.md` is no longer part of this PR (it went with the deploy-side split).

You were right that the example was already stale: only the newest 5 dated tags are retained, so a hard-coded date is guaranteed to rot. It's now a `develop-<YYYYMMDD>` placeholder, stated as needing to be one of the retained tags, plus a command to list what actually exists on the Hub. I also corrected a related claim in that doc that the compose defaults track the nightly channel — they don't.

dhruv-1001:
### Second review pass — `84f88837`

CodeRabbit re-reviewed after the first fix commit and found five more, all valid. Fixed:

| Finding | Fix |
|---|---|
| `github.run_id` is **constant across re-runs** — only `run_attempt` increments | Real hole in my own fix. A **cancelled** attempt skips `cleanup` (`if: !cancelled()`), leaving staging tags that the next attempt of the *same run* would consume — exactly the mixed-manifest bug the scoping was meant to close. Tags now carry `<run_id>-<run_attempt>`, applied at all four sites. |
| Build legs re-checked out the mutable `ref` | `enumerate` resolves the ref to a SHA once; every leg checks out that SHA. With ~38 legs, a mid-run push to `develop` could otherwise publish one `develop-YYYYMMDD` tag spanning two commits. |
| `del()` returned the trailing `echo`'s status | Now inspects HTTP status: `2xx` ok, **`404` idempotent** (already gone), anything else counted → `::error::` and non-zero exit after all images. Verified the counter isn't lost to a subshell. |
| Prune read a single `page_size=100` page | `list_dated()` follows `.next` until exhausted. |
| `docker/login-action@v2` unsupported | Both references → `@v4`. |

The last two compound and are worth calling out together: silent delete failures are precisely *how* a repo accumulates >100 dated tags, at which point the single-page read quietly stopped pruning the overflow. Fixing either alone would have left the other reachable.

Validated: YAML parses, every `run:` block passes `bash -n`, and the delete-failure counter simulated end-to-end (3 failures → exit 1).

**One thread deliberately left open** — the Docker Hub JWT duplication between `build-images.yml` and `build.yml`. Still true in this PR's code, so it shouldn't be marked resolved; a `.github/actions/dockerhub-api` composite is the right fix but belongs in its own PR.

dhruv-1001:
### Nightly tag is now the commit hash, not the calendar date (`527a7b9`)

The nightly's immutable tag was `develop-$(date -u +%Y%m%d)`. A date names **when the build ran**, which is not the same fact as **what it built** — and the gap was real:

* a re-run later the same day overwrote an "immutable" `develop-20260612` with different code, so it was never actually immutable;
* two nights on an unmoved `develop` minted two tags for one commit;
* nothing tied the tag to the SHA the build legs actually checked out.

#### a) Tag format: `develop-<sha8>`

The default is now sliced from `steps.sha.outputs.sha` — the same SHA `enumerate` resolves once and every build leg checks out — rather than a second `git rev-parse`. So the tag *provably* names the built commit: `git show 84f88837` **is** `egovio/pgr-services:develop-84f88837`.

```yaml
env:
  RESOLVED_SHA: ${{ steps.sha.outputs.sha }}
```
```bash
if ! printf '%s' "$RESOLVED_SHA" | grep -qE '^[0-9a-f]{40,}$'; then
  echo "FATAL: resolved SHA is not a hex object id: $RESOLVED_SHA"; exit 1
fi
VERSION_TAG="develop-${RESOLVED_SHA:0:8}"
```

Why these choices:

* **Kept the `develop-` prefix** — the channel stays in the tag (per `NIGHTLY-BUILDS.md`), the server-side `?name=develop-` filter keeps working, and the prune can still target only this channel.
* **8 chars, not 7 or 12** — `git rev-parse --short HEAD` returns **8** in this repo today, and `build.yml` already tags `${BRANCH}-${COMMIT_HASH}` with exactly that. So the nightly and a `build.yml` dispatch of the same `develop` commit now agree on **one** name instead of inventing two for the same code.
* **Guard on the SHA** — not decoration. A non-hex value would produce a tag outside the prune pattern, and it would then leak on the Hub forever. Better to fail the run.

#### b) Retention — the part that actually breaks

Dated tags sorted **newest-first by name for free**; hashes have **no intrinsic order**, so the old `sort -r | tail -n +6` over tag names would have deleted an essentially arbitrary set. Fixed at the source: `list_dated` → `list_snapshots` now emits `<last_updated>\t<name>` and retention sorts on the *timestamp*.

```bash
| jq -r '.results[]?
        | select((.name // "") | test("^develop-[0-9a-f]{8,}$"))
        | "\(.last_updated // .tag_last_pushed // "")\t\(.name)"'
...
for T in $(list_snapshots "$IMG" | LC_ALL=C sort -r | tail -n +6 | cut -f2); do
```

`last_updated` falls back to `tag_last_pushed`, then to `""` — an empty timestamp starts the line with TAB, sorts last, and is pruned first, which is the safe direction. `LC_ALL=C` so collation can't reorder the ISO-8601 field.

#### ⚠️ The prune pattern is calibrated, not incidental

`^develop-[0-9a-f]{8,}$`:

| Tag | Matches? | Why that's required |
|---|---|---|
| `develop-84f88837` (new) | ✅ | the thing we're retaining |
| `develop-20260612` (legacy dated) | ✅ | **8 digits are 8 hex chars** — legacy tags keep ageing out instead of leaking forever now that nothing mints them |
| `develop-70916ea` (**7** chars) | ❌ | **this is the important one.** `docker-compose.yml`, `docker-compose.deploy.yaml`, `docker-compose.egov-digit.yaml`, `docker-compose.db-migrations.yml` and `devops/deploy-as-code/**/values.yaml` all pin `develop-70916ea` as the image default — in the **same `egovio` repos the prune deletes from**. A `{7,}` bound would have deleted live deployment pins. |
| `nightly-develop` | ❌ | rolling tag, untouched (unchanged behaviour) |
| `nightly-develop-staging-123-1-amd64` | ❌ | returned by the `?name=develop-` substring filter, excluded by the anchors |
| `v2.13.0` | ❌ | release tags are immutable and kept |

#### c) Docs / other date assumptions

`grep -rn` for `YYYYMMDD`, `%Y%m%d`, `[0-9]{8}`, "dated tag", "daily snapshot" across the repo. Updated in scope: the `version_tag` and `prune_develop` input descriptions, the `cleanup` job name and its comments, `nightly-build-develop.yml`'s inline comment, and `build/NIGHTLY-BUILDS.md` (tags table, driver §1, the verify-what's-running snippet, plus a new *Why the commit, not the day* section).

Deliberately **not** changed:

* `local-setup/ansible/files/nightly-build-push.sh` — the on-box VPC-registry driver still uses `develop-$(date -u +%Y%m%d)`. It's out of this PR's scope, pushes to a **different registry**, and `NIGHTLY-BUILDS.md` documents it as unchanged. The tags table now marks `develop-YYYYMMDD` as *legacy, on-box only* rather than pretending it's gone.
* `.github/workflows/build.yml`, `release-build.yml`, `build/build-config.yml` — no change needed. `build.yml` was **already** hash-based; `release-build.yml` passes an explicit `version_tag` and never reaches the default branch (verified below); `build-config.yml` has no tag logic.
* `digit-mcp/.github/workflows/ci.yml`, `performance/scripts/*` — unrelated date usage, different subsystems.

#### d/e) Unchanged behaviour

Rolling `nightly-develop` is untouched. The release path is untouched — only the *default* (empty `version_tag`) changed.

---

### Validation

**YAML + shell syntax** — all four workflow files `yaml.safe_load`, and every `run:` block extracted and checked with `bash -n` (GHA expressions substituted): 6/6 blocks in `build-images.yml`, 18/18 in `build.yml`, all OK.

**Tag derivation** — the `enumerate/list` `run:` block extracted *verbatim from the committed YAML* and executed with `yq` stubbed:

```
real HEAD sha: 84f888374b05ee747d6f16f50986d49e9924f2e9

--- case: nightly default (empty version_tag)
    exit=0  version_tag=develop-84f88837
--- case: release path (explicit version_tag)
    exit=0  version_tag=v2.13.0          <- release path unaffected
--- case: guard: non-hex SHA
    exit=1  version_tag=(none)
--- case: guard: empty SHA
    exit=1  version_tag=(none)
--- case: guard: 39-char SHA (too short)
    exit=1  version_tag=(none)
```

**Retention** — the `cleanup` `run:` block, again extracted verbatim, run against a stand-in Docker Hub over **real HTTP** (real `curl`, real `next` pagination at `page_size=4`, real jq/sort/cut, real 204/404 status handling). Input deliberately shuffled so ordering has to be earned:

```
===== cleanup job output =====
  already absent pgr-services:nightly-develop-staging-999-1-amd64
  already absent pgr-services:nightly-develop-staging-999-1-arm64
  deleted pgr-services:develop-20260728
  deleted pgr-services:develop-20260612
  deleted pgr-services:develop-cafebabe
  deleted pgr-services:develop-99887766
  deleted pgr-services:develop-11223344
  already absent digit-mcp:nightly-develop-staging-999-1-amd64
  already absent digit-mcp:nightly-develop-staging-999-1-arm64
===== cleanup exit code: 0 =====

--- pgr-services
  BEFORE (name / last_updated / tag_last_pushed):
      develop-cafebabe                       2026-06-01T13:40:00Z   2026-06-01T13:40:00Z
      nightly-develop                        2026-08-05T13:40:00Z   2026-08-05T13:40:00Z
      develop-20260729                       2026-07-29T13:40:00Z   2026-07-29T13:40:00Z
      develop-70916ea                        2026-08-05T14:00:00Z   2026-08-05T14:00:00Z
      develop-11223344                       None                   None
      develop-84f88837                       2026-08-05T13:40:00Z   2026-08-05T13:40:00Z
      nightly-develop-staging-123-1-amd64    2026-08-05T13:00:00Z   2026-08-05T13:00:00Z
      develop-20260612                       2026-06-12T13:40:00Z   2026-06-12T13:40:00Z
      develop-00ff11aa                       2026-07-30T13:40:00Z   2026-07-30T13:40:00Z
      v2.13.0                                2026-07-31T13:40:00Z   2026-07-31T13:40:00Z
      develop-abcdefgh                       2026-08-05T13:50:00Z   2026-08-05T13:50:00Z
      develop-99887766                       None                   2026-05-01T00:00:00Z
      develop-deadbeef                       2026-08-01T13:40:00Z   2026-08-01T13:40:00Z
      develop-1a2b3c4d                       2026-08-04T13:40:00Z   2026-08-04T13:40:00Z
      develop-20260728                       2026-07-28T13:40:00Z   2026-07-28T13:40:00Z
  pruned          : ['develop-11223344', 'develop-20260612', 'develop-20260728', 'develop-99887766', 'develop-cafebabe']
  expected pruned : ['develop-11223344', 'develop-20260612', 'develop-20260728', 'develop-99887766', 'develop-cafebabe']
  surviving       : ['develop-00ff11aa', 'develop-1a2b3c4d', 'develop-20260729', 'develop-70916ea',
                     'develop-84f88837', 'develop-abcdefgh', 'develop-deadbeef', 'nightly-develop',
                     'nightly-develop-staging-123-1-amd64', 'v2.13.0']
  RESULT: exact match
--- digit-mcp
  pruned          : (none)          <- repo with <=5 snapshots left completely alone
  expected pruned : (none)
  RESULT: exact match
```

Note what survived: **`develop-70916ea` is the newest tag in the whole set** (`14:00`) and is still never considered — that's the 7-char pin boundary doing its job. `develop-abcdefgh` (8 chars, but `g`/`h` aren't hex) and the staging tag both survive despite matching the server-side `?name=develop-` filter.

The intermediate ordering, showing the `tag_last_pushed` fallback and the empty-timestamp case:

```
--- LC_ALL=C sort -r  (newest first) ---
     1  2026-08-05T13:40:00Z    develop-84f88837   \
     2  2026-08-04T13:40:00Z    develop-1a2b3c4d    |
     3  2026-08-01T13:40:00Z    develop-deadbeef    |  KEEP (head -5)
     4  2026-07-30T13:40:00Z    develop-00ff11aa    |
     5  2026-07-29T13:40:00Z    develop-20260729   /   <- legacy dated tag, kept on merit
     6  2026-07-28T13:40:00Z    develop-20260728   \
     7  2026-06-12T13:40:00Z    develop-20260612    |
     8  2026-06-01T13:40:00Z    develop-cafebabe    |  DELETE (tail -n +6)
     9  2026-05-01T00:00:00Z    develop-99887766    |   <- via tag_last_pushed fallback
    10  <empty>                 develop-11223344   /    <- no timestamp -> sorts last -> pruned first
```

**Leftover date references in the changed files** — `grep -n 'YYYYMMDD\|%Y%m%d\|[0-9]{8}\|dated'` returns only prose that *deliberately* describes the legacy format (why the regex still matches it; the on-box script that still mints it). `grep -rn 'date -u\|date +' .github/workflows/` → **no matches**: no workflow derives a tag from a date any more.

dhruv-1001:
## What to review here

**Scope: CI/build only.** 6 files, all under `.github/workflows/` + `build/`. Nothing in `local-setup/` — the deploy-side half is #1582.

### 1. New nightly pipeline (`nightly-build-develop.yml` → `build-images.yml`)
Builds every entry in `build/build-config.yml` multi-arch, pushes `egovio/<name>:{develop-<sha8>, nightly-develop}`, retains the newest 5 dated tags.

**Review focus — the safety property.** Per-arch staging tags are scoped `<prefix>-staging-<run_id>-<run_attempt>-<arch>`. Both parts matter: `run_id` alone is stable across re-runs, so a **cancelled** attempt (which skips `cleanup` via `if: !cancelled()`) would leave tags the next attempt of the *same run* would consume and stitch into a mixed-attempt manifest. The `manifest` job refuses to publish unless both arch tags from *this* attempt exist.

### 2. Tags are commit hashes, not dates (`develop-<sha8>`)
Derived by slicing the SHA `enumerate` already resolved — the same SHA every build leg checks out — so the tag provably names the built commit. Matches `build.yml`'s existing `<branch>-<hash>` shape, so a nightly and a manual dispatch of the same commit produce one tag, not two.

**Review focus — the prune regex `^develop-[0-9a-f]{8,}$` is load-bearing.** 9 files (`local-setup/docker-compose.egov-digit.yaml` + 5 Helm `values.yaml`) pin **`develop-70916ea`** — a 7-char tag — in the same `egovio` repos the prune deletes from. A `{7,}` bound would delete live deployment pins. Retention also had to stop sorting by name (dates sorted chronologically for free; hashes do not) and now sorts on the Hub's `last_updated`.

### 3. Review-round fixes
- `build.yml`: `grep … || true | sort | tail` parsed as `grep … || (true | sort | tail)`, so sort/tail never saw grep's output → multi-line `LATEST_TAG` → step died with `Invalid format`. Grouped, and arch suffixes excluded.
- `inputs.ref` / `inputs.version_tag` now pass through `env:` instead of being interpolated into shell; the resolved tag is validated against Docker's tag grammar before use.
- Manifest matrix (~19 runner VMs doing one `imagetools` call each) collapsed to one job; per-image isolation kept in the loop.
- `prune` → `cleanup`: also deletes this run's staging tags, runs on `!cancelled()` so a failed run doesn't leak them, fails the job on a real delete error (404 treated as idempotent), and pages through the full tag list.

---

## Validation

The workflow was **executed locally**, not just read. `act` could not run it (`actions/checkout` insists on syncing a remote; the branch under test was local-only — and `act --bind` destructively wiped its bound directory), so each `run:` block was extracted from the committed YAML and executed verbatim with the env a runner injects, pushing to a local registry so nothing could reach Docker Hub:

```
Resolve ref to an immutable SHA  -> 527a7b95fde3b6c9...
Flatten manifest + version tag   -> version_tag=develop-527a7b95
targets (real yq enumeration)    -> ['digit-ui-esbuild', 'configurator']
build leg digit-ui-esbuild       -> OK, staging tag pushed
build leg configurator           -> OK, staging tag pushed
manifest (amd64 only)            -> SKIPPED both: "staging tags ... are incomplete"
manifest (both arches present)   -> Published 2 image(s)
```

That skip is the point: with only one arch built, the gate **refused to publish** — the exact failure the run-scoping exists to prevent. With both arch tags present it published `develop-<sha>` + `nightly-develop`, and both images were then pulled from the registry and **ran**, serving HTTP 200.

Also verified: the grep fix against its real failure input (old → 3 lines, new → 1 correct tag, no-match → empty).

**Not verified:** arm64 builds (no qemu on the test host — the arm64 leg used an amd64 stand-in purely to exercise `imagetools create`); and the `cleanup` job's Docker Hub API calls, which by design only work against Hub.

vinothrallapalli-eGov (approved):

.github/workflows/build-images.yml
dhruv-1001: Resolving this as **acknowledged, deferred** — flagging that explicitly since I said earlier I would leave it open.

The finding is correct: `build-images.yml`'s cleanup job and `build.yml`'s *Generate the Next Tag* step both do the same Docker Hub JWT login + `tags?page_size=100` dance, and when the Hub auth flow changes both copies have to be found and fixed.

Not doing it here, for two reasons:

1. **It reaches outside this PR.** A `.github/actions/dockerhub-api` composite means editing `build.yml`, which this PR only touches for a one-line namespace fix (`edevops/` → `egovio/`). Bundling a shared-action refactor into a PR that has already been split once makes it harder to review, not easier.
2. **The duplication did not grow here.** The `prune` job became `cleanup` and still performs exactly one login — it now covers both staging-tag deletion and dated-tag pruning in that single authenticated pass.

Tracked under #1309 as a follow-up. Happy to do it immediately as a standalone PR if you would rather it land before this one — say the word and I will.

Issue #1050, opened 2026-07-06
Integrate Vinoth's JBAC migration into dashboard access control


KDwevedi:
## Context
Vinoth is migrating access control to **JBAC** (JSON-Based Access Control). Once that migration lands, the dashboard should consume it as the source of truth for authorization.

## Task
When the JBAC migration is complete, integrate it with the dashboard logic and use JBAC for access control **instead of the PGR-side access-control code we added** for dashboard RBAC.

## Notes
- Blocked on / depends on Vinoth's JBAC migration completing.
- Today the dashboard RBAC lives in PGR service code (the catalog-driven RBAC work, PR #963). That bespoke logic should be retired in favor of JBAC.
- Verify existing dashboard RBAC behavior (3 layers + S2/S3, catalog-driven) is preserved after the cutover.

KDwevedi:
## Upstream ABAC work this ticket consumes

- **Digit-Core `abac-changes`** — https://github.com/egovernments/Digit-Core/tree/abac-changes
- **CCRS #1441** (draft, `abac-changes` → `develop`, +4151/-973 across 71 files) — https://github.com/egovernments/Citizen-Complaint-Resolution-System/pull/1441

### Why #1441 is the concrete integration point, not just related context

#1441 builds the ABAC engine and applies it to **`/v2/request/_search`** — a JsonLogic `condition` on MDMS action **2008** that scopes an employee to the union of their HRMS **department AND jurisdiction** assignments, plus field-level PII masking on `citizen.mobileNumber/name/userName`.

**The analytics surface does not consume any of it.** Verified at `abac-changes` @ `4432477be`:

- `PrincipalScopeResolver:212-224` **resolves `jurisdictionCodes` and fails closed** — but `AnalyticsPlanner.applyScope:315-343` **never reads it**. Net effect today: a ward-scoped GRO sees every ward in their department on the dashboard, while the same principal is correctly ward-scoped on `/v2/request/_search`. The two surfaces disagree about the same user.
- Dashboard authorization is still the bespoke PGR-side catalog RBAC (from #963) that this ticket exists to retire.
[...]

PR #1441, opened 2026-07-28
feat(access-control): masters/configurator role gating + PGR row/field ABAC hardening

### LINKED ISSUE #1086: ABAC/JBAC: pgr-services Tier-2 PDP — record-level + field-level policy enforcement
Part of #1083
Depends on #1084, #1085

## Scope
The actual runtime enforcement (Tier 2 / PDP) inside `pgr-services`, for `/pgr-services/v2/request/_search` (and `_plainsearch` for field-level masking) as the reference implementation — proving the pattern is generalizable before moving it to a shared library (per #1083's rollout step 2/4).

## What's enforced
1. **Record-level**: a citizen sees only their own complaints; an employee sees only complaints in their department **and** their HRMS-assigned jurisdiction (both required — narrows, doesn't widen); a tenant-wide role bypasses. Enforced twice, deliberately:
   - SQL-level pushdown (`PGRQueryBuilder`) for correct pagination/counts.
   - A JsonLogic post-fetch re-check (`SearchAccessPolicyService`) as defense-in-depth — logs loudly if the two ever disagree ("drift").
2. **Field-level**: individual fields (e.g. `citizen.mobileNumber`) masked per-role via `FieldVisibilityService`, applied through Spring's `BeanWrapperImpl` on the typed response graph — no JSON round-trip, no reflection code of our own.

## New components (`org.egov.pgr.policy`)
`PolicyEvaluator` (JsonLogic wrapper, `io.github.jamsesso:json-logic-java`, fails closed on any evaluation error), `AccessPolicyRegistry` (fetches + caches the Action record via the API from the "policy based on config" issue, only caching successful resolutions since the fetch is role-scoped), `PolicyInputBuilder` (the one shared `user`/`resource` input-document builder both enforcement paths bind against), `SearchAccessPolicyService`, `FieldVisibilityService`, `MaskingStrategy`.

## Supporting changes
- `PrincipalScopeResolver`/`AnalyticsScope` (existing analytics-module resolver, reused rather than duplicated): added jurisdiction resolution (previously scaffolded/disabled) alongside the existing department axis, via an additive constructor overload (zero ripple to existing callers).
- `PGRQueryBuilder`: SQL predicates for department (`additionaldetails->>'department'`) and jurisdiction (`ads.locality`).

## Design docs
`docs/design/accesscontrol-policy-conditions-design.md`, `docs/design/field-level-attribute-access-design.md`, testing plan in `docs/design/accesscontrol-policy-conditions-testing-plan.md`.

## Status
Implemented; 76 JUnit tests passing; live-verified end-to-end against a running local-setup stack (citizen self-scope, department+jurisdiction combined scope, tenant-wide bypass, fail-closed on unresolvable HRMS data, field-level mobileNumber masking) — pending commit/PR.

## Known follow-ups (explicitly out of scope here)
- Generalizing this out of `pgr-services` into a shared library/gateway concern (per #1083's rollout step 4).
- UI consuming the same conditions to pre-hide actions (per #1083's open questions).

vinothrallapalli-eGov:
## Problem statement

The PGR platform needed attribute-based access control that is consistent across three surfaces that previously had none: (1) which MDMS masters a role can see/edit inside the configurator, (2) which complaint records a principal can see when searching PGR complaints, and (3) which fields on those complaint records (citizen PII) a non-owning viewer should see. Today's baseline was ad hoc: masters had no visibility gating at all, PGR search had no record-level scoping, and there was no field-masking layer.

### 1. Configurator masters/UI access gating
`configurator/src/hooks/useMastersCapability.tsx`, `providers/bridge.ts`, and per-resource List/Show/BulkImport components: role-driven view/edit capability for masters (Departments, Designations, Complaint Hierarchy/Types, Tenants), consumed by the admin dashboard and resource pages so a role only sees the masters and actions it's entitled to, with a bulk-import capability check.

### 2. PGR `/v2/request/_search` record-level ABAC — MDMS-configurable per role, per axis

Complaint search scoping is no longer hardcoded. It's declared entirely in MDMS on the existing `ACCESSCONTROL-ACTIONS-TEST.actions-test` action record (id 2008), nested inside the already-opaque `resource` field — **zero changes to egov-accesscontrol/Digit-Core** needed to support it:

```json
"resource": {
  "complaint": {
    "scope": {
      "axes": ["department", "jurisdiction"],
      "roleScopes": {
        "GRO":        { "department": "ALL", "jurisdiction": "OWN" },
        "PGR_LME":    { "department": "OWN", "jurisdiction": "OWN" },
        "SUPERVISOR": { "department": "OWN", "jurisdiction": "ALL" }
      },
      "default": { "department": "ALL", "jurisdiction": "OWN" }
    }
  }
}
```

Each role gets an independent `OWN` (restricted to the caller's own HRMS-resolved value) or `ALL` (unrestricted) level per axis — no redeploy needed to reshape who sees what; it's a pure MDMS edit. Multiple held roles combine most-permissive-wins per axis. A generic `EMPLOYEE`/`COMMON_EMPLOYEE` marker role every HRMS token carries alongside the functional role is explicitly excluded from that union so it can never silently neutralize a functional role's explicit restriction.

New classes (`org.egov.pgr.policy`):
- **`ScopeLevel`** — `OWN` / `ALL` (named for clarity — `ALL` reads as "sees all values on this axis", not "no access").
- **`ScopePolicy`** — parses the `resource.complaint.scope` block.
- **`ScopePolicyEngine`** — pure resolver: caller's roles + HRMS-resolved values + `ScopePolicy` → per-axis allowed values.
- **`PolicyDrivenScopeResolver`** — HRMS lookup + citizen-self-scope + calls `ScopePolicyEngine`, producing a **`PgrSearchScope`** (a scope value object dedicated to PGR search — deliberately *not* reusing Dashboard/Analytics' own `AnalyticsScope`, so a change to either can never risk the other).

The same `ScopePolicy` drives **both** the Tier-1 SQL prefilter (`PolicyDrivenScopeResolver` → `PgrSearchScope` → `PGRQueryBuilder`) and the Tier-2 JsonLogic per-row re-check (`AccessPolicyRegistry.getCondition` generates the condition from the same `scope` block via `ScopePolicyEngine`) — the two are generated from one source of truth and cannot drift apart. A hand-authored `condition` field is only consulted when no `scope` block is present at all (backward-compatible fallback for actions/tenants that haven't adopted this yet).

`ServiceRequestValidator.validateSearchParam` also got a null-safe guard so a request with no `RequestInfo.userInfo` fails with a clear `INVALID_REQUESTINFO` error instead of an NPE.

Dashboard/Analytics is intentionally untouched by any of this — it keeps its own `PrincipalScopeResolver`/`AnalyticsScope`/`ScopeAxis` model, a separate resolver in a separate package.

### 3. Field-level PII masking
Same action's `resource.complaint.attributes`: `citizen.mobileNumber`, `citizen.name`, and `citizen.userName` all use `onDeny.strategy = "REDACT"` (full `null`, no partial reveal) for any non-owning, non-tenant-wide viewer — e.g. CSR/GRO/PGR_LME see none of these three fields on a complaint they didn't file, while the citizen who owns the record (or a tenant-wide caller) still sees them in full.

### 4. Design docs
- `docs/design/composable-scope-policies-design.md`, `docs/design/generic-abac-policy-framework-design.md` — the ABAC engine's design and a runbook for onboarding a second API to the same engine.
- `docs/design/accesscontrol-policy-conditions-testing-plan.md` — kept in sync with the current engine (this pass): the MDMS-configured per-role `ScopePolicy` table, the generated-condition behavior, the role-scoped 15-minute cache key, and REDACT-only masking.

### 5. Local-setup / seed data
- `local-setup/scripts/seed-tenant-city-data.py`: idempotent seed script for tenant-specific data `tenant_bootstrap` doesn't copy (`RAINMAKER-PGR.ComplaintHierarchy`, a city tenant's boundary hierarchy).
- `full-dump.sql`: action 2008's seed now carries the current `resource.complaint.scope` (GRO/PGR_LME/SUPERVISOR/default) and REDACT-only `attributes`, no dead legacy `condition` field. `EGOV_ACCESSCONTROL_HOST` wiring fix for `pgr-services`' compose service.

### Live verification (`ap` tenant, real HTTP login + search calls — not mocks)

| Employee | Role(s) | HRMS dept/ward | Expected | Actual | Result |
|---|---|---|---|---|---|
| EMP001 | PGR_LME | DEPT_1 / WARD_001 | own dept ∩ own ward = 4 | 4 | ✅ |
| EMP002 | GRO | DEPT_1 / WARD_001 | all depts, own ward = 6 | 6 | ✅ |
| EMP003 | CSR *(default policy)* | DEPT_1 / WARD_001 | all depts, own ward = 6 | 6 | ✅ |
| EMP004 | PGR_LME | DEPT_2 / WARD_002 | own dept ∩ own ward = 1 | 1 | ✅ |
| EMP007 | PGR_LME | DEPT_3 / WARD_003 | no matching data = 0 | 0 | ✅ |
| CMSADMIN | PGR_LME + CSR + GRO *(multi-role union)* | DEPT_4 / WARD_003 | dept unrestricted (GRO wins), own ward = 0 | 0 | ✅ |
| TESTSUPERVISOR | SUPERVISOR | DEPT_1 / WARD_001 | own dept, all wards = 6 | 6 | ✅ |
| TESTNODEPT | PGR_LME, **no HRMS department** | — / WARD_001 | fail-closed on unresolved required axis = 0 | 0 | ✅ |
| EMPMULTI01 | PGR_LME, multi-ward | ENV / WARD_001+002 | own dept ∩ own wards (no ENV data) = 0 | 0 | ✅ |
| ADMIN | 8 roles *(multi-role union)* | CENTER / `ap` (placeholder) | dept unrestricted (GRO wins), own jurisdiction (no match) = 0 | 0 | ✅ |

**One real bug found and fixed during this pass**: `DGRO` role was configured in `resource.complaint.scope.roleScopes` (`department: OWN, jurisdiction: OWN`) but had no `ACCESSCONTROL-ROLEACTIONS` mapping for action 2008 — egov-accesscontrol's role-scoped lookup returned "no action visible" for DGRO-only callers, silently falling back to the built-in default policy instead of DGRO's configured scope. Resolved by retiring `DGRO` from this policy entirely (role deprecated) rather than patching the mapping — confirmed via re-test that DGRO-holding callers now correctly fall back to the default (jurisdiction-only) policy.

- [x] Citizen isolation: two distinct citizens; citizen B's `_search` returns zero results for citizen A's complaint, citizen A sees her own.
- [x] Field redaction: confirmed a non-owning employee (GRO) gets `null` for `citizen.mobileNumber`/`name`/`userName` on complaints they didn't file; confirmed real values come back for the owning citizen.
- [x] Multi-jurisdiction match: employee with two ward jurisdictions, one complaint per ward (same department); `_search` returns both.
- [x] Configurator masters gating: admin UI resource List/Show/BulkImport pages respect the resolved capability per role.
- [x] `ScopeLevel.NONE → ALL` rename validated as behavior-neutral: full matrix above re-run after the rename, identical results.
- [x] Full backend suite: `mvn -f backend/pgr-services/pom.xml test` green (363 tests, 0 failures/errors) after every structural change (resolver split, `PgrSearchScope` extraction, rename).
- [x] `EGOV_ACCESSCONTROL_HOST`/action-2008 wiring verified via `/access/v1/actions/mdms/_get` — condition/scope/field rules are actually returned to the caller.

## Summary by CodeRabbit

* **New Features**
  * Added role-based access controls for complaint searches, including tenant, citizen, department, and jurisdiction-based visibility.
  * Added field-level protection that redacts or partially masks restricted complaint information.
  * Added role-based visibility and editing controls across Configurator masters, onboarding, and bulk imports.
  * Added automatic complaint hierarchy defaults based on tenant configuration.

* **Bug Fixes**
  * Improved validation for incomplete search request information.
  * Restricted previously unprotected tenant creation access to authenticated, authorized users.


backend/pgr-services/src/main/java/org/egov/pgr/policy/PgrSearchScope.java
```diff
@@ -0,0 +1,38 @@
...
+ * Fields (all "null/empty = no restriction on this axis"):
+ * - tenant scope:      always applied (LIKE prefix at state level, = at city level).
+ * - citizenUuid:       a pure CITIZEN sees only their own complaints (account_id = their uuid).
+ * - departmentCodes:   an employee is restricted to the union of their HRMS assignment departments.
+ * - jurisdictionCodes: an employee is restricted to the union of their HRMS jurisdiction (boundary)
+ *                      assignments, exact-matched against a complaint's address locality.
+ */
+public final class PgrSearchScope {
+    public final String tenantId;
+    public final boolean tenantStateLevel;
```
KDwevedi, line 25: Blocker: these server-resolved tenant fields are never enforced; SQL still trusts `criteria.tenantId`. Validate the requested tenant against the caller’s tenant/subtree and apply that authorized scope in SQL, with a cross-tenant regression test.

backend/pgr-services/src/main/java/org/egov/pgr/policy/PolicyDrivenScopeResolver.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    /**
+     * Scope for an employee whose department could not be resolved. Fail-CLOSED (deny-all sentinel)
+     * for constrained roles; unrestricted only for tenant-wide (admin/supervisor) roles.
+     */
+    private PgrSearchScope unresolvedScope(User u, String tenantId, boolean stateLevel, String reason) {
+        if (hasTenantWideRole(u)) {
+            log.debug("scope unresolved ({}) for tenant-wide role '{}' — unrestricted", reason, u.getUserName());
+            return new PgrSearchScope(tenantId, stateLevel, null, null, null);
```
KDwevedi: Do not grant tenant-wide access merely because a hard-coded role has no HRMS record. Resolve empty HRMS axes through `ScopePolicyEngine` so only an explicit policy `ALL` can remove a restriction; otherwise fail closed.

backend/pgr-services/src/main/java/org/egov/pgr/policy/AccessPolicyRegistry.java
```diff
@@ -0,0 +1,353 @@
...
+            action = getAction(actionUrl, requestInfo, tenantId);
+        } catch (AccessControlUnavailableException e) {
+            log.error("AccessPolicyRegistry: accesscontrol unavailable for url='{}' tenant='{}' — failing closed: {}",
+                    actionUrl, tenantId, e.getMessage());
+            return null;
+        }
+        if (action == null) {
+            log.info("AccessPolicyRegistry: no ACCESSCONTROL-ACTIONS-TEST entry visible for url='{}' tenant='{}' — policy not defined, allowing (backward compatible)",
+                    actionUrl, tenantId);
+            return ALWAYS_ALLOW_CONDITION;
```
KDwevedi: Missing/invisible action is indistinguishable from a broken role-action mapping—the DGRO incident in this PR already demonstrates that. ABAC-enabled tenants must fail closed; if backward compatibility is required, gate it behind an explicit rollout mode.

backend/pgr-services/src/main/java/org/egov/pgr/policy/ScopePolicy.java
```diff
@@ -0,0 +1,157 @@
...
+    @SuppressWarnings("unchecked")
+    public static Optional<ScopePolicy> parse(Object raw) {
+        if (!(raw instanceof Map))
+            return Optional.empty();
+        Map<String, Object> scopeMap = (Map<String, Object>) raw;
+
+        Object axesRaw = scopeMap.get("axes");
+        if (!(axesRaw instanceof List) || ((List<?>) axesRaw).isEmpty()) {
+            log.warn("ScopePolicy: 'axes' missing/empty — treating as not configured");
+            return Optional.empty();
```
KDwevedi: A present-but-malformed scope is not the same as no scope and must not fall into the permissive default. Preserve absent vs invalid as distinct outcomes and reject/fail closed on invalid policy.

backend/pgr-services/src/main/java/org/egov/pgr/policy/AccessPolicyRegistry.java
```diff
@@ -0,0 +1,353 @@
...
+     * {@link ScopePolicy#parse}. Returns null only if JSON serialization itself fails.
+     */
+    private String synthesizeCondition(ScopePolicy policy) {
+        List<Object> employeeClauseParts = new ArrayList<>();
+        employeeClauseParts.add(Map.of("==", List.of(Map.of("var", "user.type"), "EMPLOYEE")));
+        for (String axis : policy.getAxes()) {
+            AxisFields fields = AXIS_FIELDS.get(axis);
+            if (fields == null) {
+                log.warn("AccessPolicyRegistry: scope declares axis '{}' with no known resource/user field mapping — skipping in generated condition", axis);
+                continue;
```
KDwevedi: Unknown axes cannot be skipped: that silently removes an authored restriction from both enforcement tiers. Reject unsupported axes at parse time and fail the policy closed.

backend/pgr-services/src/main/java/org/egov/pgr/repository/rowmapper/PGRQueryBuilder.java
```diff
@@ -187,22 +214,53 @@ public String getPGRSearchQuery(RequestSearchCriteria criteria, List<Object> pre
...
+    /**
+     * Injects the RBAC scope's WHERE predicates. Mirrors the same axes/pattern as
+     * {@code AnalyticsPlanner.applyScope} in the analytics module (citizen self-scope, employee
+     * department-scope), plus PGR search's own jurisdiction axis (exact-match on the complaint's
+     * address locality — see {@code PgrSearchScope#jurisdictionCodes}, distinct from the analytics
+     * module's own hierarchical {@code boundaryPrefix}, which stays unwired here).
+     */
+    private void applyScope(PgrSearchScope scope, StringBuilder builder, List<Object> preparedStmtList) {
+        if (scope == null)
+            return;
```
KDwevedi, line 243: Null scope must not mean unrestricted inside the security-sensitive builder. Throw on scoped search/count paths and give approved internal/`plainSearch` callers a separately named explicit bypass.

backend/pgr-services/src/main/java/org/egov/pgr/service/PGRService.java
```diff
@@ -329,6 +352,14 @@ public List<ServiceWrapper> plainSearch(RequestInfo requestInfo, RequestSearchCr
         Map<String, ComplaintTemplateTypeConfig> configCache = buildConfigCache(requestInfo, tenantIdForMdms, enrichedServiceWrappers);
         applyDecryptOrMask(enrichedServiceWrappers, requestInfo, configCache);
 
+        // plainSearch stays record-level unrestricted (see PGRRepository/PGRQueryBuilder — no
+        // scope threaded into the query), but field-level PII masking (e.g. citizen.mobileNumber)
+        // must still apply here — this scope is used ONLY for that, never for row filtering.
+        PgrSearchScope fieldVisibilityScope = searchAccessPolicyService.resolveScope(
+                requestInfo, tenantIdForMdms, config.getStateLevelTenantIdLength());
+        fieldVisibilityService.apply(requestInfo, tenantIdForMdms, fieldVisibilityScope,
+                AccessPolicyRegistry.PGR_REQUEST_SEARCH_URL, "complaint", enrichedServiceWrappers);
```
KDwevedi: Remove this from this PR: `_plainsearch` is a different endpoint, remains row-unrestricted, and cannot safely reuse `_search` action 2008. If it needs ABAC, give it its own action plus row and field policy end-to-end.

backend/pgr-services/src/main/java/org/egov/pgr/service/PGRService.java
```diff
@@ -474,19 +505,15 @@ private String getDepartmentFromMDMS(ServiceRequest request, Object mdmsData) {
...
-                }
-            } catch (Exception e) {
-                log.warn("Failed to parse MDMS response for department name lookup, code: {}. Falling back to code.", departmentCode, e);
-            }
-
-            return departmentCode;
+            if (departmentCode == null)
+                return "NA";
+            String normalized = departmentCode.trim();
+            return normalized.isEmpty() || normalized.equalsIgnoreCase("NA") ? "NA" : normalized;
```
KDwevedi, line 516: This changes persisted department semantics from display name to code, so existing complaints will not match HRMS codes. Add a backfill/dual-read migration and mixed-data tests, or extract this storage migration until it is ready.

configurator/packages/data-provider/src/providers/accessPolicy.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export async function loadMastersCapability(
+  client: DigitApiClient,
+  tenantId: string,
+  roles: string[],
+): Promise<MastersCapability> {
+  if (!tenantId || roles.length === 0) return OPEN_CAPABILITY;
+
+  const [actionRecords, roleActionRecords] = await Promise.all([
+    client.mdmsSearch(tenantId, ACTIONS_TEST_SCHEMA, { limit: POLICY_FETCH_LIMIT }).catch(() => []),
```
KDwevedi: Do not convert policy-fetch failure or truncation into “view everything”; the fixed 5000 cap can create the same state. Fetch exhaustively/filter server-side and expose a loading/error capability that does not reveal restricted masters.

KDwevedi (changes requested):
Requesting changes. The latest resolver split correctly removes the Dashboard/Analytics collision; keep `AnalyticsScope`, `PrincipalScopeResolver`, `AnalyticsPlanner`, and `AnalyticsService` untouched in this PR, with the later dashboard policy cutover remaining in #1050.

Required before approval:
- Enforce the server-authorized tenant/subtree; do not trust the requested tenant.
- Remove hard-coded tenant-wide fallback and let the authored policy decide `ALL`.
- Fail closed for missing/invisible/malformed policies and unsupported axes.
- Make a null scope impossible on scoped search/count paths.
- Remove `_plainsearch` changes unless it gets its own complete action policy.
- Provide backward compatibility/migration for department name→code storage.
- Make Configurator capability loading non-leaky and exhaustive.
- Add regression coverage for these cases and get the currently red integration check green.

The legacy `/dashboard` path is outside this review and outside this effort.

backend/pgr-services/src/main/java/org/egov/pgr/policy/PgrSearchScope.java
vinothrallapalli-eGov: Fixed in 80760f464: `PolicyDrivenScopeResolver.resolve()` now validates the requested tenantId against the caller's own token-derived home tenant/subtree (`isAuthorizedTenant`) *before* any role/HRMS resolution runs, denying closed (via `denyAllScope`) on mismatch. `PGRQueryBuilder.applyScope` also now actually applies `scope.tenantId`/`scope.tenantStateLevel` in SQL (LIKE at state level, = at city level), matching this class's own doc comment which described that behavior but never wired it up. Added regression tests covering: cross-tenant denial for a regular employee, cross-tenant denial for a tenant-wide role (the exact reported hole), legitimate state->city narrowing, and a missing caller-tenant denying. Verified live against a running stack (hot-swap): a city-level tenant-wide admin widening to the parent state tenant is now denied with the new fail-closed log line, while legitimate narrowing still succeeds.

backend/pgr-services/src/main/java/org/egov/pgr/repository/rowmapper/PGRQueryBuilder.java
vinothrallapalli-eGov: Fixed in 80760f464: `PGRQueryBuilder.applyScope` now throws `IllegalStateException` on a literal null scope instead of silently treating it as unrestricted. The only sanctioned bypass is the new `PgrSearchScope.UNRESTRICTED` sentinel, which plainSearch, the escalation scheduler, and the two internal fetch-by-id callers (`PGRService.restoreMaskedPlaceholders`, `ServiceRequestValidator.validateUpdate`) now pass explicitly via `PGRRepository`'s no-scope convenience overloads, rather than relying on an implicit null default. Verified live: no unexpected `IllegalStateException` in logs across search/count/create/update/plainSearch after the swap, confirming no other caller was silently depending on the old null-means-unrestricted behavior.

backend/pgr-services/src/main/java/org/egov/pgr/policy/PolicyDrivenScopeResolver.java
vinothrallapalli-eGov: Fixed in 366b8b17c: removed the hard-coded `TENANT_WIDE_ROLES` bypass entirely. Every outcome (a populated HRMS record, an empty one, a missing employee, or an HRMS error) now routes through the same `ScopePolicyEngine.resolve()` call, passing whatever department/jurisdiction values were actually resolved (possibly empty). The engine already fails closed per-axis when a required axis has no resolvable value, and only lets a role through unrestricted when the authored policy explicitly grants that role `ALL` for that axis — so a hard-coded tenant-wide role with no HRMS record is now denied unless the policy itself says `ALL`. Added tests covering both directions: no-explicit-opinion + no HRMS data now denies (previously bypassed), and an explicit policy `ALL` still resolves unrestricted with no HRMS data.

backend/pgr-services/src/main/java/org/egov/pgr/policy/AccessPolicyRegistry.java
vinothrallapalli-eGov: Fixed in 366b8b17c: the "missing action -> allow" fallback is now gated behind a new `pgr.abac.strict-mode` property (`PGRConfiguration`), default `false` — preserving today's backward-compatible behavior for every currently-deployed tenant (verified live: the tenant this branch was hot-swap-tested against has zero MDMS action 2008 config and would otherwise lose all PGR search results). Once a tenant's ABAC policies are fully authored, setting `pgr.abac.strict-mode=true` makes a missing/invisible action fail closed instead of silently allowing — the explicit rollout mode you asked for. Added a test constructing a strict-mode registry and asserting the fail-closed path.

backend/pgr-services/src/main/java/org/egov/pgr/policy/ScopePolicy.java
vinothrallapalli-eGov: Fixed in 366b8b17c: `ScopePolicy.parse()` now distinguishes a genuinely absent `scope` key (`raw == null` -> empty Optional, stays backward compatible) from a present-but-broken one (anything else that fails to parse -> throws the new `MalformedScopePolicyException`). `AccessPolicyRegistry.getCondition` catches it and fails closed (returns null) rather than falling through to the legacy hand-authored `condition` field; `getScopePolicy` deliberately lets it propagate, same treatment as `AccessControlUnavailableException`, so `SearchAccessPolicyService` never applies its permissive default to a broken policy. Updated the existing `notAMapReturnsEmpty`/`missingOrEmptyAxesReturnsEmpty` tests (they now assert the throw) and added a `getScopePolicy`/`getCondition` coverage pair for the malformed-vs-absent distinction.

backend/pgr-services/src/main/java/org/egov/pgr/policy/AccessPolicyRegistry.java
vinothrallapalli-eGov: Fixed in 366b8b17c: `synthesizeCondition` now returns null (failing the whole generated condition closed) when a scope declares an axis with no `AXIS_FIELDS` mapping, instead of logging a warning and skipping just that axis. Skipping would have silently dropped that axis's restriction from the Tier-2 re-check while Tier-1's SQL scope (reading the exact same `scope` block) may still enforce it — i.e. the two tiers could silently disagree, exactly what generating both from one artifact is meant to prevent. Added a test with an unmapped axis asserting `getCondition` returns null.

backend/pgr-services/src/main/java/org/egov/pgr/service/PGRService.java
vinothrallapalli-eGov: Fixed in 366b8b17c: removed `_plainsearch`'s reuse of `_search`'s action/policy (id 2008) for field-visibility masking entirely — it no longer resolves a scope or calls `fieldVisibilityService.apply` for that endpoint. `_plainsearch` stays record-level unrestricted (unchanged) and, now, unrestricted at the field-visibility level too, with a comment explaining it needs its own action + policy end-to-end (row and field) if that's ever needed, rather than borrowing one authored for a different endpoint's semantics.

backend/pgr-services/src/main/java/org/egov/pgr/service/PGRService.java
vinothrallapalli-eGov: Fixed in 366b8b17c via dual-read rather than a data migration (discussed and confirmed this approach): added `MDMSUtils#getDepartmentCodeToNameMap` (fetches `common-masters.Department`, best-effort — an MDMS failure just skips the expansion, never fails the search) and wired it into `PolicyDrivenScopeResolver`'s department-axis resolution — the scope's department list now unions in each HRMS code's display name, so the SQL `IN` filter matches complaints written before this PR (display name) and after (code) with no backfill required. The deny-all sentinel is excluded from the lookup (never a real code, and skips a wasted MDMS call on the already-denied path). Added tests for both the expansion and the sentinel-skips-lookup case.

configurator/packages/data-provider/src/providers/accessPolicy.ts
vinothrallapalli-eGov: Fixed in 366b8b17c: replaced the fixed 5000-row single fetch with exhaustive pagination (`fetchAllMdmsRecords`, pages until a page comes back shorter than the page size, so truncation can no longer happen silently), and removed the `.catch(() => [])` that let a fetch failure resolve into an empty policy indistinguishable from "nothing configured, everything visible". `loadMastersCapability` now propagates a fetch failure as a rejected promise; `useMastersCapability.tsx`'s catch (the layer that actually decides the fallback capability) now sets a distinct deny-by-default `ERROR_PERMISSIONS` on that failure, separate from the transient loading-state default it previously reused. Added tests for both the pagination (a record only visible on page 2 is still picked up) and the fetch-failure propagation.

backend/pgr-services/src/main/java/org/egov/pgr/policy/ScopePolicy.java
```diff
@@ -0,0 +1,160 @@
...
+            throw new MalformedScopePolicyException("scope is present but not a JSON object: " + raw.getClass().getSimpleName());
+        Map<String, Object> scopeMap = (Map<String, Object>) raw;
+
+        Object axesRaw = scopeMap.get("axes");
+        if (!(axesRaw instanceof List) || ((List<?>) axesRaw).isEmpty())
+            throw new MalformedScopePolicyException("'axes' missing/empty");
+        List<String> axes = new java.util.ArrayList<>();
+        for (Object a : (List<?>) axesRaw)
+            if (a instanceof String && !((String) a).isBlank())
+                axes.add((String) a);
```
KDwevedi, line 128: Blocker: unsupported axes are still accepted here. Tier 2 returns null, but Tier 1 drops unknown entries when constructing `PgrSearchScope`; `_count` never runs Tier 2, so it can expose an insufficiently scoped count. Reject unsupported axes before resolution and add search/count regressions.

backend/pgr-services/src/main/java/org/egov/pgr/config/PGRConfiguration.java
```diff
@@ -127,6 +127,22 @@ public MappingJackson2HttpMessageConverter jacksonConverter(ObjectMapper objectM
...
+    @Value("${egov.accesscontrol.actions.mdms.get.path}")
+    private String accessControlActionsMdmsGetPath;
+
+    // Explicit rollout gate for org.egov.pgr.policy.AccessPolicyRegistry#getCondition: false (the
+    // default) preserves today's backward-compatible behavior — a missing/invisible
+    // ACCESSCONTROL-ACTIONS-TEST entry for an action allows, same as before ABAC existed. A tenant
+    // only gets the ABAC-strict "missing action fails closed" behavior once it explicitly opts in
+    // here, after its policies are fully authored — never as a silent default that could brick an
+    // existing deployment's PGR search on the next redeploy.
+    @Value("${pgr.abac.strict-mode:false}")
```
KDwevedi, line 143: Blocker: this gate defaults to false and this PR does not enable it in any target deployment. Missing or role-invisible action 2008 therefore takes the compatibility path; explicitly enable strict mode for ABAC-enabled deployments and test the deployed configuration.

configurator/src/hooks/useMastersCapability.tsx
```diff
@@ -0,0 +1,93 @@
...
+  // regardless of whether identityKey actually changed (e.g. the same user
+  // logging back in), which is fine — the fetch below is idempotent.
+  useEffect(() => onAuthChange(() => setIdentityKey(identityKeyOf())), []);
+
+  useEffect(() => {
+    let cancelled = false;
+    // Reset to the fail-open default immediately on identity change, rather
+    // than leaving the PREVIOUS user's (possibly more permissive) capability
+    // visible for the duration of the new fetch.
+    setPermissions(OPEN_PERMISSIONS);
```
KDwevedi: Blocker: resetting to `OPEN_PERMISSIONS` during every identity change exposes all resources while policy loading is pending and can initiate reads before the denial arrives. Use a deny/loading capability from first render and during refetch, with a delayed-promise identity-switch regression test.

configurator/packages/data-provider/src/providers/accessPolicy.ts
```diff
@@ -0,0 +1,133 @@
...
+      if (!schema) return true;
+      const rule = mastersConditions[schema];
+      if (!rule?.condition) return true; // no entry = today's behavior (§2.5)
+      try {
+        return Boolean(jsonLogic.apply(rule.condition, userDoc));
+      } catch {
+        // Fail OPEN here (not the fail-closed rule the PGR PDP uses) — this is
+        // UI-only presentation, not a security boundary (§1.1); a bad condition
+        // should never lock an admin out of the console.
+        return true;
```
KDwevedi: Blocker: a malformed authored condition is treated as view-allowed, so policy corruption exposes the resource. Fail closed here and update the malformed-JsonLogic test to require `canView(...) === false`.

KDwevedi (changes requested):
Follow-up re-review on the current head. Most of the earlier implementation requests are addressed, but approval remains blocked by the four enforcement gaps below: unsupported axes can leak counts, strict mode is not enabled by deployment, and Configurator still has two fail-open paths. Add end-to-end regressions for each. The legacy `/dashboard` path remains outside this review and this effort; the unrelated baseline Local Setup localization failure is not being counted as a blocker here.

backend/pgr-services/src/main/java/org/egov/pgr/config/PGRConfiguration.java
vinothrallapalli-eGov: Addressed in fff2b036c, but as an explicit documented decision rather than flipping the gate on: `pgr.abac.strict-mode` stays `false` in `local-setup/docker-compose.egov-digit.yaml`, now set explicitly (was previously just relying on the code default) with a comment explaining why. This stack is process-wide per pgr-services instance, not per-tenant — it serves tenants that don't yet have action 2008's `resource`/`condition` (or `resource.complaint.scope`) authored (e.g. a freshly-onboarded tenant). Flipping it on here would fail PGR search closed for every one of those, not just the ones that do have it configured. The reference tenant `pg` already has a full scope policy authored via this stack's own seed data (`db/full-dump.sql`) and would work fine with the flag on — but making that the deployment-wide default would silently break every other tenant sharing the same instance. Verified the env var (`PGR_ABAC_STRICT_MODE`) does correctly bind to the property via Spring's actual `SystemEnvironmentPropertySource` resolution, so turning it on for a specific, fully-authored deployment is a one-line, low-risk change whenever that's wanted. Open to enabling it for a specific target deployment if you have one in mind — happy to wire that up and verify live.

KDwevedi (approved):
Approved. The requested enforcement fixes are addressed, and the remaining strict-mode concern is accepted as an explicit rollout prerequisite: enable `PGR_ABAC_STRICT_MODE=true` and deploy/verify the complete action 2008 policy on Bomet. The legacy `/dashboard` path remains outside the scope of this effort. Please resolve the current base-branch conflict and rerun the relevant tests before merge.

backend/pgr-services/src/main/java/org/egov/pgr/policy/PolicyInputBuilder.java
```diff
@@ -0,0 +1,96 @@
...
+        attributes.put("jurisdictions", scope.jurisdictionCodes == null ? List.of() : scope.jurisdictionCodes);
+
+        Map<String, Object> userDoc = new LinkedHashMap<>();
+        userDoc.put("uuid", user != null ? user.getUuid() : null);
+        userDoc.put("type", user != null ? user.getType() : null);
+        userDoc.put("attributes", attributes);
+        return userDoc;
+    }
+
+    public Map<String, Object> buildResourceDoc(Service service) {
```
subhashini-egov, line 53: **MEDIUM — Tier-1/Tier-2 drift risk.** `resource.complaint.tenantId` is populated in the JsonLogic input here but `AccessPolicyRegistry.AXIS_FIELDS` doesn't reference it, so Tier-2 conditions never consult it. Today the field is dead. If a future MDMS-authored policy ever references `resource.complaint.tenantId` in a JsonLogic condition, Tier-2 will silently succeed while Tier-1 has no matching SQL predicate — undermining the "single source of truth" invariant that Tier-1 and Tier-2 are generated from the same `ScopePolicy`.

Either drop the field from `buildResourceDoc` until an axis actually maps to it, or add tenantId to `AXIS_FIELDS` and to the SQL side.

configurator/packages/data-provider/src/providers/accessPolicy.ts
```diff
@@ -0,0 +1,135 @@
...
+    offset += POLICY_FETCH_PAGE_SIZE;
+  }
+}
+
+export async function loadMastersCapability(
+  client: DigitApiClient,
+  tenantId: string,
+  roles: string[],
+): Promise<MastersCapability> {
+  if (!tenantId || roles.length === 0) return OPEN_CAPABILITY;
```
subhashini-egov: **MEDIUM — fail-open branch inconsistent with documented posture.** When `!tenantId || roles.length === 0`, this returns `OPEN_CAPABILITY` (`canView: () => true`). `authProvider.ts:49` resolves `user?.tenantId || stateTenantId`, which can be `undefined` for a partially-configured client — yielding view-everything.

This contradicts the deny-by-default posture that `useMastersCapability.tsx:22-27` documents. Not a security bug (server still enforces via egov-accesscontrol), but any operator relying on the hook's fail-closed guarantee will be surprised.

Return the deny capability here; let the "no policy configured on tenant" case be handled by the `mastersConditions` empty-object fallback that's already there (§2.5 "no entry = visible" still applies per-schema).

configurator/packages/data-provider/src/providers/accessPolicy.ts
```diff
@@ -0,0 +1,135 @@
...
+ */
+export interface MastersCapability {
+  canView(schema: string | undefined): boolean;
+  canEdit(schema: string | undefined): boolean;
+}
+
+export const ACTIONS_TEST_SCHEMA = 'ACCESSCONTROL-ACTIONS-TEST.actions-test';
+export const ROLEACTIONS_SCHEMA = 'ACCESSCONTROL-ROLEACTIONS.roleactions';
+const SEARCH_ACTION_URL = '/mdms-v2/v2/_search';
+const WRITE_URL_RE = /^\/mdms-v2\/v2\/_(?:create|update)\/(.+)$/;
```
subhashini-egov, line 37: **LOW — undocumented URL coverage.** `WRITE_URL_RE` only matches `/mdms-v2/v2/_(create|update)/<schema>`. Any masters that write through a non-mdms-v2 endpoint (legacy `/egov-mdms-create`, HRMS, access-role/action endpoints, boundary) will always report `canEdit=false`.

Today only mdms-backed masters use the hook for edit gating, so this is by design — but `access-roles` and `access-actions` now carry MDMS `schema` keys in the registry. If anyone later wires `hasEdit={canEditResource('access-roles')}` on those pages, the button will be permanently hidden for everyone with no obvious cause. Add a code comment noting the constraint.

local-setup/scripts/seed-tenant-city-data.py
```diff
@@ -0,0 +1,413 @@
...
+
+
+def create_boundary_relationship(tok, tenant, code, btype, parent):
+    body = {"RequestInfo": {"apiId": "seed-tenant-city-data", "authToken": tok},
+            "BoundaryRelationship": {"code": code, "tenantId": tenant, "hierarchyType": HIERARCHY_TYPE,
+                                     "boundaryType": btype, "parent": parent}}
+    _post("/boundary-service/boundary-relationships/_create", body).read()
+
+
+def sql_insert_boundary_relationship(tenant, code, btype, parent, path):
```
subhashini-egov, line 342: **MEDIUM — SQL string interpolation into `psql -c`.** `sql_insert_boundary_relationship` builds a SQL string via `%` formatting from env-derived values (`tenant`, `code`, `HIERARCHY_TYPE`, `btype`, `parent`, `path`) sourced from `SEED_CITY_TENANT`, `BOUNDARY_CITY_CODE`, `BOUNDARY_WARD_CODES`, `BOUNDARY_HIERARCHY_TYPE`, then shells out to `docker exec ... psql -c`.

An operator passing a value containing `'` will break the SQL and could inject. Not remotely exploitable (operator already has docker exec on the host), but a typo silently corrupts data. Use `psql --set var=value` with `:'var'` quoting, or pipe a heredoc via `psql -f -` with proper parameterization.

local-setup/scripts/seed-tenant-city-data.py
```diff
@@ -0,0 +1,413 @@
...
+import time
+import urllib.error
+import urllib.parse
+import urllib.request
+
+URL = os.environ["DIGIT_URL"].rstrip("/")
+TENANT = os.environ["SEED_TENANT"]
+CITY_TENANT = os.environ["SEED_CITY_TENANT"]
+USERNAME = os.environ.get("DIGIT_USERNAME", "ADMIN")
+PASSWORD = os.environ.get("DIGIT_PASSWORD", "eGov@123")
```
subhashini-egov, line 59: **LOW — hardcoded admin password `eGov@123`.** Fine for local-setup by convention, but flag if this script is ever repurposed to run outside dev. Consider requiring the env var explicitly (no default) when a non-`localhost` egov host is passed.

backend/pgr-services/src/main/java/org/egov/pgr/policy/PolicyDrivenScopeResolver.java
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    /**
+     * POST /egov-hrms/employees/_search with codes=[userName] + tenantId. Returns the Employees
+     * JSON array node (or null). Uses the internal gateway host from egov.hrms.host.
+     */
+    private JsonNode searchHrmsByCode(RequestInfo requestInfo, String tenantId, String userName) {
+        // Built via UriComponentsBuilder (query params encoded), not raw string concatenation: an
+        // HRMS-linked username containing '&', '=', '%', or '#' (plausible for email-based/SSO
+        // usernames) would otherwise corrupt the query string and resolve the wrong employee — the
```
subhashini-egov, line 253: **LOW — raw string concatenation in HRMS URL.** `tenantId` and `userName` are appended as query params without URL encoding. Today `userName` comes from the trusted token, so the risk is low, but `UriComponentsBuilder` is the safe idiom — a future refactor that lets a caller-controlled value through here would silently misparse on `&` or `=`.

backend/pgr-services/src/main/java/org/egov/pgr/policy/MaskingStrategy.java
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Applies the strategy named in onDeny.strategy. An unrecognized or missing strategy name
+     * fails closed to REDACT rather than leaving the field visible — a masking config typo must
+     * never result in exposure.
+     */
+    public static Object apply(Object value, Map<String, Object> onDeny) {
+        if (value == null)
+            return null;
+
+        String name = onDeny == null ? null : String.valueOf(onDeny.get("strategy"));
```
subhashini-egov: **LOW — correctness via exception path.** `String.valueOf(onDeny.get("strategy"))` returns the literal string `"null"` when the map has `strategy=null`, which then throws `IllegalArgumentException` inside `valueOf`, caught below → REDACT. Works, but relying on the exception path for correctness is fragile. An explicit `null` / empty-string guard here would document the intent and avoid a hot-path throw when policies are missing the field.

subhashini-egov (commented):
## Review — masters/configurator role gating + PGR row/field ABAC hardening

**Scope reviewed:** 88 files, +7269/-970. Backend PGR ABAC engine + configurator UI gating + MDMS/seed/infra.

### Overall

The design is genuinely defensive: fail-closed on parse errors, on HRMS misses for required axes, and on accesscontrol outages. Tier-1 SQL prefilter and Tier-2 JsonLogic re-check are correctly generated from one `ScopePolicy` source of truth so they cannot drift. Test coverage on the engine internals is solid; live-verified against a 10-employee matrix in the `ap` tenant per the PR body.

### Verified

- **Single source of truth** — `SearchAccessPolicyService.resolveScope` and `AccessPolicyRegistry.getCondition` both derive from the same parsed `ScopePolicy`.
- **Fail-closed contracts hold** — `ScopePolicy.parse` rejects unsupported axes at parse time; `getCondition` returns null (evaluator denies) on `AccessControlUnavailableException`; `PGRQueryBuilder.applyScope` throws on null scope (locked by test); `PgrSearchScope`+`applyScope` uses a `__scope_denied__` sentinel so an OWN axis with no HRMS values yields zero rows.
- **SQL is parameterized end-to-end** — every new `.append(` uses `?` placeholders via `createQuery`; scope `tenantId` also parameterized.
- **Tenant isolation is strengthened**, not weakened — scope `tenantId` is now a first-class SQL predicate (state LIKE vs city `=`) using the server-derived tenant; cross-tenant/cross-subtree requests are rejected in `isAuthorizedTenant` before scope resolution.
- **Cache key safety** — `AccessPolicyRegistry` cache keyed on `tenantId | actionUrl | uppercased-sorted-distinct-roleKey`; misses not cached; no cross-tenant/cross-role leak.
- **Marker-role bypass** — `EMPLOYEE`/`COMMON_EMPLOYEE` naturally excluded because `explicitLevelFor` returns empty for roles missing from `roleScopes`; tested in `ScopePolicyEngineTest`.
- **`validateSearchParam` null guard** — narrow, uses `INVALID_REQUESTINFO`, doesn't swallow other NPEs.
- **Client-side gating framed as UX-only** — no POST/PUT/DELETE is protected solely by `useMastersCapability`; every gated component documents the RoleAction gateway as the real boundary. `useMastersCapability` fails closed on pre-first-fetch, mid-refetch, and fetch-failure; tests cover all three.

Inline comments below cover specific bugs / risks (5 medium, 8 low). Non-anchored notes follow.

---

### Not anchored to a diff line

**Low — `configurator/packages/data-provider/src/providers/authProvider.ts:26`** — pre-existing behavior touched by this PR: `checkError` treats 401 and 403 identically (clearAuth + redirect). A 403 typically means "authenticated but forbidden," not "session expired" — clearing auth on 403 forces re-login for legitimately-authenticated users hitting a forbidden endpoint. Worth splitting.

**Informational — `AccessPolicyRegistry.getFieldVisibilityRules`** — if an operator authors `resource.complaint` but omits `attributes`, un-masked PII flows through. A warning-log on this shape would catch authoring mistakes.

**Style / minor**
- `PolicyDrivenScopeResolver.DENY_ALL_DEPARTMENT` and `ScopePolicyEngine.UNRESOLVED_SENTINEL` are duplicate `__scope_denied__` literals that must stay in sync — consolidate to one constant.
- `AccessPolicyRegistry.CachedEntry.action` stores the map by reference. Safe today; `Map.copyOf` would future-proof.
- `AccessPolicyRegistry.PGR_REQUEST_SEARCH_URL` is a public constant — extracting to `PGRConfiguration` would allow per-tenant configurability if ever needed.

---

### Test coverage gaps

- **No end-to-end integration test** combining Tier-1 SQL, Tier-2 enforce, and field-visibility apply in a single `search()` call — the PR body's matrix was live-tested but not codified.
- **`ServiceRequestValidator.validateSearchParam`** — no unit test for the new `INVALID_REQUESTINFO` guard (null requestInfo, null userInfo, blank type).
- **`count()` path** — DENY sentinel not exercised for the count SQL specifically.
- **Empty-non-null `departmentCodes`/`jurisdictionCodes`** in `PGRQueryBuilderTest` — would lock in the current "silently unrestricted" behavior (see inline comment on `PGRQueryBuilder.java`).
- **Cache TTL expiry** in `AccessPolicyRegistry` — no test refetches after 15 min.
- **Case sensitivity** — no test authoring `"gro"` lowercase against uppercase `GRO` token role, though both sides uppercase.
- **`useMastersCapability.test.tsx`** mocks `getResourceConfig: (name) => ({ schema: name })`, so never exercises resources with no `schema` (employees, complaints, boundaries, users…). Under those, edit buttons silently hide for everyone — worth locking behavior.
- **`accessPolicy.test.ts`** — no test for the `!tenantId` fallback (see inline), no test for unknown roles / malformed rows, no test for the identity-change mid-flight cancellation flag in `useMastersCapability`.
- **`authProvider.test.ts`** — no assertion that `checkError` clears auth on 403.
- **`plainSearch`** — the comment on `PGRService:355-361` claims unscoped/unmasked; no integration test locks this.

---

### Recommendation

**Approve after addressing** the seed-row duplication in `full-dump.sql`, the master vs seed shape+strategy drift in `actions-test.json`, the `accessPolicy.ts:70` fail-open branch, and the per-branch image tag in `docker-compose.egov-digit.yaml`. The rest are follow-ups; the core engine is well-designed, correctly fail-closed, and well-tested.

backend/pgr-services/src/main/java/org/egov/pgr/repository/rowmapper/PGRQueryBuilder.java
```diff
@@ -187,22 +215,80 @@ public String getPGRSearchQuery(RequestSearchCriteria criteria, List<Object> pre
...
+        }
+
+        if (scope.citizenUuid != null) {
+            addClauseIfRequired(preparedStmtList, builder);
+            builder.append(" ser.accountId = ? ");
+            preparedStmtList.add(scope.citizenUuid);
+        }
+
+        if (!CollectionUtils.isEmpty(scope.departmentCodes)) {
+            addClauseIfRequired(preparedStmtList, builder);
```
subhashini-egov, line 275: **LOW — safeguard is contractual, not enforced.** `applyScope` guards `departmentCodes`/`jurisdictionCodes` with `CollectionUtils.isEmpty(...)`, meaning an empty non-null list would silently drop the axis and return unrestricted rows (not "zero rows" as the surrounding fail-closed design implies).

Today this isn't reachable because `ScopePolicyEngine.resolve` guarantees the `__scope_denied__` sentinel for OWN axes with no values. But if that contract ever regresses, the failure mode here is silent over-exposure. Prefer emitting `WHERE FALSE` (or asserting non-empty) so the local invariant is enforced independently of upstream.

_(re-posting: my earlier inline attachment on the batch review landed at the wrong diff position — this thread supersedes it.)_

local-setup/db/full-dump.sql
```diff
@@ -3120,6 +3120,14 @@ complaint-hierarchy-data-005	pg	PGR.GarbageNeedsTobeCleared	RAINMAKER-PGR.Compla
...
... "Search PGR Request", "path": "", "enabled": false, "displayName": "Search PGR Request", "orderNumber": 0, "serviceCode": "pgr-services", "parentModule": "", "method": "POST", "resource": {"complaint": {"scope": {"axes": ["department", "jurisdiction"], "roleScopes": {"GRO": {"department": "ALL", "jurisdiction": "OWN"}, "PGR_LME": {"department": "OWN", "jurisdiction": "OWN"}, "SUPERVISOR": {"department": "OWN", "jurisdiction": "ALL"}}, "default": {"department": "ALL", "jurisdiction": "OWN"}}, "attributes": {"citizen.mobileNumber": {"condition": {"or": [{"==": [{"var": "user.attributes.tenantWide"}, true]}, {"==": [{"var": "resource.complaint.accountId"}, {"var": "user.uuid"}]}]}, "onDeny": {"strategy": "REDACT"}}, "citizen.name": {"condition": {"or": [{"==": [{"var": "user.attributes.tenantWide"}, true]}, {"==": [{"var": "resource.complaint.accountId"}, {"var": "user.uuid"}]}]}, "onDeny": {"strategy": "REDACT"}}, "citizen.userName": {"condition": {"or": [{"==": [{"var": "user.attributes.tenantWide"}, true]}, {"==": [{"var": "resource.complaint.accountId"}, {"var": "user.uuid"}]}]}, "onDeny": {"strategy": "REDACT"}}}}}}	t	system-mdms-seed	system-mdms-seed	1785200000000	1785200000000
```
subhashini-egov, line 3117: **MEDIUM — duplicate seed rows for action 2008 and its role mappings.** This new row coexists with the pre-existing row at line 2444 (different `uniqueidentifier`: `2008` vs the SHA hash `09124d06…`). The PK is `(tenantid, schemacode, uniqueidentifier)`, so both persist and egov-accesscontrol's `ActionService` deduplicates by `id` in the JSON payload — likely the newer one wins by search order, but this is a latent foot-gun.

Same duplication for the ROLEACTIONS entries: old rows at 2710/2755/2757/2758/2759 (with integer `id` field) vs the new rows at 3124-3130 (without). `ActionService` may see two mappings per (role, action).

Either `DELETE` the legacy rows first in the SQL, or overwrite the same `uniqueidentifier` so upsert semantics hold.

_(re-posting: my earlier inline attachment on the batch review landed at the wrong diff position — this thread supersedes it.)_

utilities/default-data-handler/src/main/resources/mdmsData/ACCESSCONTROL-ACTIONS-TEST/ACCESSCONTROL-ACTIONS-TEST.actions-test.json
```diff
@@ -1406,7 +1427,37 @@
...
+                { "==": [ { "var": "user.attributes.tenantWide" }, true ] },
+                { "==": [ { "var": "resource.complaint.accountId" }, { "var": "user.uuid" } ] }
+              ]
+            },
+            "onDeny": { "strategy": "MASK_SHOW_LAST_N", "n": 2, "maskChar": "X" }
+          }
+        }
+      }
+    },
+    "condition": {
```
subhashini-egov: **MEDIUM — master file is stale relative to the DB seed.** This master JSON still uses the top-level `condition` shape (JsonLogic OR across CITIZEN/EMPLOYEE) *and* `strategy: MASK_SHOW_LAST_N`, while `local-setup/db/full-dump.sql:3123` has moved to the new `resource.complaint.scope` shape with no top-level `condition` and `strategy: REDACT`.

Whichever loads last wins — likely the DB seed in practice, but the two representations of the same action disagree. Commit `380618c6 "restore action 2008 field masking, REDACT-only"` explicitly says REDACT is intended. Update the master JSON to match the seed so this can't drift back on the next MDMS-master-driven load.

_(re-posting: my earlier inline attachment on the batch review landed at the wrong diff position — this thread supersedes it.)_

backend/pgr-services/src/main/resources/application.properties
```diff
@@ -55,6 +55,10 @@ egov.localization.statelevel=true
 egov.mdms.host=http://localhost:8082
 egov.mdms.search.endpoint=/egov-mdms-service/v1/_search
 
+#accesscontrol — Tier-2 PDP condition lookup (role-scoped action + condition resolution)
```
subhashini-egov, line 58: **LOW — dev defaults for new accesscontrol properties.** The new `egov.accesscontrol.host` / `.actions.mdms.get.path` default to `http://localhost:8080` here. Per-env `application-*.properties` and helm values need to override before shipping. `pgr.abac.strict-mode=false` is correctly a backward-compat default. Worth calling out in the deployment checklist / runbook.

_(re-posting: my earlier inline attachment on the batch review landed at the wrong diff position — this thread supersedes it.)_

local-setup/docker-compose.egov-digit.yaml
```diff
@@ -1032,7 +1032,10 @@ services:
     - egov-network
 
   egov-accesscontrol:
-    image: egovio/egov-accesscontrol:maven-jdk21-9f83afb
+    # Parameterized so deploys can pin a local ABAC-testing build via
```
subhashini-egov, line 1035: **LOW — per-branch image tag pinned into shared compose.** `egovio/egov-accesscontrol:abac-changes-cb6a372` is a feature-branch build. When this PR merges and the branch is deleted, this tag may disappear from the registry and break future `docker compose pull` for anyone on master. Retag to a stable release tag before merge.

_(re-posting: my earlier inline attachment on the batch review landed at the wrong diff position — this thread supersedes it.)_

local-setup/docker-compose.egov-digit.yaml
```diff
@@ -1522,6 +1529,12 @@ services:
         condition: service_healthy
       egov-enc-service:
         condition: service_healthy
+      egov-accesscontrol:
+        # service_healthy is unavailable — see the "No healthcheck" note on egov-accesscontrol's
+        # own service block above; this still orders pgr-services' start after the container
+        # starts, which is what PGR startup actually needs (fail-closed access-control lookups
+        # already handle "not up yet" at the application layer).
+        condition: service_started
```
subhashini-egov, line 1545: **LOW — `service_started` vs `service_healthy`.** pgr-services depends on egov-accesscontrol with `condition: service_started`, which only waits for the container process to launch, not for Spring Boot to bind :8090.

If integration tests hit pgr-services `_search` before accesscontrol is truly ready, the accesscontrol lookup will 500/connection-refused. Per the design, `AccessPolicyRegistry` should catch the `ConnectException` and fail closed — worth verifying that's actually what happens (not a bubble-up that surfaces as a 5xx to the caller). Given `PGR_ABAC_STRICT_MODE=false` is the default, this is likely OK in practice.

_(re-posting: my earlier inline attachment on the batch review landed at the wrong diff position — this thread supersedes it.)_

backend/pgr-services/src/main/java/org/egov/pgr/policy/PolicyInputBuilder.java
vinothrallapalli-eGov: Fixed — dropped the dead `resource.complaint.tenantId` field from `buildResourceDoc` (no axis maps to it today). Re-adding it is left for whenever a real axis is wired into `AXIS_FIELDS` + the SQL side alongside it.

configurator/packages/data-provider/src/providers/accessPolicy.ts
vinothrallapalli-eGov: Fixed — the `!tenantId || roles.length === 0` branch now returns a deny capability instead of `OPEN_CAPABILITY` (removed entirely). Updated `accessPolicy.test.ts` and `authProvider.test.ts` to assert the new deny-by-default behavior.

configurator/packages/data-provider/src/providers/accessPolicy.ts
vinothrallapalli-eGov: Fixed — added a comment documenting `WRITE_URL_RE`'s mdms-v2-only coverage and the silent-hide risk if a non-mdms-v2 resource is ever wired to `canEditResource()`.

local-setup/scripts/seed-tenant-city-data.py
vinothrallapalli-eGov: Fixed — `sql_insert_boundary_relationship` now passes tenant/code/parent/path as `psql -v` variables and substitutes via `:'var'` (psql's quoted-literal form, which escapes embedded quotes/backslashes) instead of raw `%s` string formatting.

local-setup/scripts/seed-tenant-city-data.py
vinothrallapalli-eGov: Addressed — added a warning to stderr when `DIGIT_PASSWORD` wasn't explicitly set and `DIGIT_URL` doesn't look local. Didn't hard-require the env var since that would also break legitimate internal docker-network usage that isn't 'local' by a simple hostname check either.

backend/pgr-services/src/main/java/org/egov/pgr/policy/PolicyDrivenScopeResolver.java
vinothrallapalli-eGov: Already fixed earlier on this branch — `searchHrmsByCode` builds the URL via `UriComponentsBuilder` with `.queryParam(...).encode()` instead of string concatenation.

backend/pgr-services/src/main/java/org/egov/pgr/policy/MaskingStrategy.java
vinothrallapalli-eGov: Fixed — added an explicit null/blank `strategy` guard before calling `valueOf(...)`, so a missing/blank strategy no longer relies on the caught `IllegalArgumentException` to fail closed.

backend/pgr-services/src/main/java/org/egov/pgr/repository/rowmapper/PGRQueryBuilder.java
vinothrallapalli-eGov: Fixed — `applyScope` now distinguishes `null` (axis not restricted, skip clause) from a non-null empty list (axis restricted, resolved to zero values → emits `1 = 0`) for both `departmentCodes` and `jurisdictionCodes`, instead of collapsing both via `CollectionUtils.isEmpty`. Added two regression tests locking this in.

local-setup/db/full-dump.sql
vinothrallapalli-eGov: Fixed — deleted the 6 legacy rows (action 2008's old hash-uniqueidentifier row + its 5 old ROLEACTIONS mappings for SUPERVISOR/PGR_LME/GRO/CSR/CITIZEN) that coexisted with the new `resource.complaint.scope`-shaped rows. Verified by loading the corrected COPY block into a scratch Postgres — exactly one action row and one mapping per role now.

utilities/default-data-handler/src/main/resources/mdmsData/ACCESSCONTROL-ACTIONS-TEST/ACCESSCONTROL-ACTIONS-TEST.actions-test.json
vinothrallapalli-eGov: Fixed — updated the master JSON's action 2008 entry to the same `resource.complaint.scope` shape + REDACT-only masking as the DB seed, and dropped the old top-level `condition` + `MASK_SHOW_LAST_N`.

backend/pgr-services/src/main/resources/application.properties
vinothrallapalli-eGov: Acknowledged, no code change — these are correct as committed local-setup defaults (matching this file's existing pattern for per-env properties), and `pgr.abac.strict-mode=false` is intentionally the safe default. Deployment-checklist documentation is outside this PR's code diff; flagging for whoever owns the env-specific rollout.

local-setup/docker-compose.egov-digit.yaml
vinothrallapalli-eGov: Partially addressed, left open — added an explicit pre-merge TODO on the default value spelling out the risk (verified live the tag does carry the needed /access/v1/actions/mdms/_get endpoint). Didn't retag to a stable release build myself: that needs an actual CI-published image I have no visibility into or authority to fabricate. Leaving this thread unresolved as a to-do for whoever owns the egov-accesscontrol release pipeline before merging to master.

local-setup/docker-compose.egov-digit.yaml
vinothrallapalli-eGov: Verified, no code change needed — egov-accesscontrol's image has no shell/wget/curl (confirmed via docker exec), so service_healthy genuinely isn't usable here, matching the existing comment on that service block. Confirmed AccessPolicyRegistry's accesscontrol calls go through MDMSUtils.fetchAccessControlActions, which wraps connection failures in AccessControlUnavailableException — caught and failed closed, not bubbled up as a raw 5xx. So a not-yet-ready accesscontrol at pgr-services startup fails closed rather than erroring the caller, consistent with PGR_ABAC_STRICT_MODE's default.

local-setup/docker-compose.egov-digit.yaml
vinothrallapalli-eGov: this is temporary image, once merged into master then will update master branch build

PR #1445, opened 2026-07-28
feat(db-seed): curate default localisation + ComplaintHierarchy master seed data

vinothrallapalli-eGov:
## Summary
- Seed `RAINMAKER-PGR.ComplaintHierarchy`/`ComplaintHierarchyDefinition` schema + sample data (Garbage/StreetLights categories and sub-types) at the `pg` bootstrap tenant, plus the `common-masters.MobileNumberValidation` schema and a missing `x-unique` constraint fix on `Workflow.BusinessServiceMasterConfig`.
- Curate ~2,065 default-data-handler localisation messages (`en_IN`) actually relevant to this system: PGR citizen/employee flows, HRMS/department/designation labels used in complaint assignment, and the workbench/configurator admin console — while dropping unrelated business-service noise (FSM, Property Tax, PQM, TradeLicense, WS-Services) that ships in the same upstream source files.
- Seed the `digit-privacy-policy` module (login consent screen) with just the 5 codes it actually needs (`ES_BY_CLICKING`, `ES_PRIVACY_POLICY`, `DIGIT_I_ACCEPT`, `DIGIT_I_DO_NOT_ACCEPT`, `DIGIT_TABLE_OF_CONTENTS`), instead of duplicating the entire common message set.

## Test plan
- [x] Restored the modified `full-dump.sql` into a disposable Postgres 15 container with `ON_ERROR_STOP=1` — no errors, no `unique_message_entry` constraint violations.
- [x] Verified live against a running local-setup deployment: loaded the same curated set into a freshly-bootstrapped `ka` tenant and confirmed via the `egov-localization` API (through Kong) that citizen/employee UI strings resolve correctly, including the login consent screen.

### DIFF
--- a/local-setup/db/full-dump.sql
+++ b/local-setup/db/full-dump.sql
@@ -2148,19 +2148,19 @@ COPY public.accesscontrol_schema_version (installed_rank, version, description, 
 --
 
 COPY public.boundary (id, tenantid, code, geometry, additionaldetails, createdtime, createdby, lastmodifiedtime, lastmodifiedby) FROM stdin;
-098443ce-1c97-40d3-aec1-44a2e45d22da	pg	PG_CITYA	{"type": "Polygon", "coordinates": [[[74.8, 31.5], [74.9, 31.5], [74.9, 31.7], [74.8, 31.7], [74.8, 31.5]]]}	\N	1781073055245	system	1781073055245	system
-a1fb63a5-8305-4365-a3f3-e86337b6c809	pg	PG_CITYA_B1	{"type": "Polygon", "coordinates": [[[74.83, 31.57], [74.87, 31.57], [74.87, 31.63], [74.83, 31.63], [74.83, 31.57]]]}	\N	1781073055245	system	1781073055245	system
-215072ee-890f-4140-9451-f3c08929e614	pg	PG_CITYA_Z1	{"type": "Polygon", "coordinates": [[[74.82, 31.55], [74.88, 31.55], [74.88, 31.65], [74.82, 31.65], [74.82, 31.55]]]}	\N	1781073055245	system	1781073055245	system
-f971f0f0-f691-4b3d-bde1-a539daf5029c	pg	PG_STATE	{"type": "Polygon", "coordinates": [[[74.0, 30.0], [78.0, 30.0], [78.0, 33.0], [74.0, 33.0], [74.0, 30.0]]]}	\N	1781073055245	system	1781073055245	system
-8cf59d36-879b-498d-886c-969b799b9ed0	pg	SUN01_LOCALITY	{"type": "Point", "coordinates": [74.87155, 31.63089]}	\N	1781073055245	system	1781073055245	system
-d160ceda-65a5-4d2a-b19d-e69ef9be2c22	pg	SUN02_LOCALITY	{"type": "Point", "coordinates": [74.85, 31.61]}	\N	1781073055245	system	1781073055245	system
-78120481-4402-4284-a81c-c8f2b00e58ad	pg	SUN03_LOCALITY	{"type": "Point", "coordinates": [74.84, 31.60]}	\N	1781073055245	system	1781073055245	system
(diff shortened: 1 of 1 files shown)


singhalkarun:
Tested on a `develop` deployment — cherry-picked only `local-setup/db/full-dump.sql` from this branch, everything else unchanged. Two findings: one regression, one question.

## 1. Regression — `boundary.additionaldetails` reverts to NULL

This branches off `master`, which never received `7f2e5d1c` *("fix(db-seed): seed boundary.additionaldetails as '{}' not NULL")* — that landed on `develop` only. Same rows, both dumps:

```
develop's dump    PG_CITYA  additionaldetails=[{}]
this PR's dump    PG_CITYA  additionaldetails=[\N]
```

`boundary-service` then NPEs on `/boundary/_search` with `argument "content" is null`, which breaks **HRMS employee create** — the employee XLSX upload fails with a 400 whose message points nowhere near the cause. Worked around locally:

```sql
UPDATE boundary SET additionaldetails = '{}' WHERE additionaldetails IS NULL;  -- 13 rows
```

Rebasing onto `develop` (or cherry-picking `7f2e5d1c`) would prevent this regressing for anyone deploying against develop.

## 2. Question — localisation seeded only at the state root

The curated messages all land on `pg`:

```
message rows at pg        2,077
message rows at pg.citya    247
```

On a deployment whose SPA reads a **city** tenant (`ui_state_tenant_id: pg.citya`), the UI still renders raw keys. `tenant_bootstrap` Step 6 normally bridges this, but it's skipped when `state_tenant_id == 'pg'`. Not a defect here — our config — but it affects out-of-box usability: is the intent that deployments point the UI at the state root, or should the seed also populate the city tenant?

## Working well

The ComplaintHierarchy seed does its job — the citizen complaint-type dropdown went from `0 items` to populated cascades (8 / 4 / 2) across a 288-test integration run.

vinothrallapalli-eGov:
@singhalkarun All localization data will be loaded under the root tenant. The UI will fetch localization data by making separate API calls for both the root tenant and the sub-tenant. If this not happening then fix should be done at UI level.

For the sub-tenant, only tenant-specific localization data will be loaded, such as boundary-related localizations and other sub-tenant-specific master data localizations.

singhalkarun:
Correction to my earlier comment — you were right, and my diagnosis was wrong.

**Your dual-fetch explanation checks out.** Kong logs from a full UI exercise (a 288-test integration run):

```
160 requests   tenantId=pg.citya
155 requests   tenantId=pg
```

The SPA does query both root and sub-tenant and merge them. My "seeded at the wrong tenant" theory was incorrect — please disregard it, along with the implication that a UI-level fix might be needed.

**The actual cause is module coverage.** The UI requests 10 modules; the curated set covers 5:

| module requested | rows at `pg` |
|---|---|
| `rainmaker-pgr` | 850 |
| `rainmaker-common` | 748 |
| `rainmaker-hr` | 244 |
| `rainmaker-workbench` | 228 |
| `digit-privacy-policy` | 5 |
| **`digit-ui`** | **0** |
| **`digit-tenants`** | **0** |
| **`rainmaker-common-masters`** | **0** |
| **`rainmaker-boundary-admin`** | **0** |
| `rainmaker-pg.citya` | 0 — tenant-specific, correctly sub-tenant per your explanation |

Observed request shapes: `module=rainmaker-common,digit-ui,digit-tenants,rainmaker-pg.citya&locale=en_IN` (110 calls), plus separate calls for `rainmaker-common-masters` (21) and `rainmaker-pgr,rainmaker-boundary-admin` (22).

`digit-ui` is the most visible of these — it carries button and navigation chrome, which is what renders as raw keys on a fresh deployment. `rainmaker-common-masters` covers department/designation display names, `rainmaker-boundary-admin` the boundary labels.

These don't look like the business-service noise the curation deliberately dropped (FSM / Property Tax / PQM / TradeLicense / WS-Services) — more that they fell off the keep-list. Would it make sense to include them?

**Also confirming:** the `boundary.additionaldetails` fix in `deeb758d` resolves finding #1 — verified `[{}]` on the latest head. Thanks for the quick turnaround.

vinothrallapalli-eGov:
@singhalkarun As per my knowledge the rainmaker-common-masters and rainmaker-boundary-admin module name is not used?

vinothrallapalli-eGov:
@singhalkarun The `digit-ui` and `digit-tenants` modules are used at the core component level for sandbox. With empty localsation data of two modules does not impact the Complaints product, as they are not used by any of the currently available pages in the application.

singhalkarun (approved):

pradeepkumarcm-egov (approved):

Issue #1465, opened 2026-07-28
[Bug] KPI headings are ambigious


elzanmathew-eGov:
# Description

What is the difference between "Complaints created today" and "New Complaints Created"? Both show different values in the dashboard. In the file complaints screen, there is no provision to enter pre-dated complaints. 

---

# Steps to Reproduce

1.
2.
3.

---

# Expected Behavior

What should have happened instead?

---

# Actual Behavior

What happened? Include error messages if any.

---

# Environment

- **Deployment**: 
- **Tenant**: 
- **Browser**: 
- **User role**: 

---

# Screenshots / Logs

Attach screens

<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/c3eaa756-4230-4293-9c13-c8c524fe83f9" />

hots, browser console errors, or service logs if available.

---

# Classification

> Add these labels before submitting. See [CONTRIBUTING.md](../CONTRIBUTING.md) for the full list.

- **Feature**: 
- **Area**: 
- **Priority**: 
- **Source**:

KDwevedi:
The two tiles are meant to measure different things, but on the current build the difference doesn't come out the way the titles imply. Splitting it into the part that is MDMS wording and the part that is a real defect:

**Intended meaning**
- `cl_new_created_count` ("New complaints created") — complaints created in the **selected date range**.
- `cl_created_today_count` ("Complaints created today") — complaints created **today**, regardless of the range.

**What actually happens today**
- With a date range applied, the global range overrides the tile's own time window, so *both* tiles count the same range and show the **same** number.
- With no range applied, "created today" falls back to a **rolling 24 hours** (`last_1d` = now − 24h), not the calendar day, while "new created" uses its own default window — so they differ for a reason nobody would guess.

Neither is what the heading promises. That behaviour is tracked in #1462 and needs a service-side change (a calendar-day window + letting the tile keep its own time axis) — it cannot be expressed in MDMS today. Confirmed with @Subham: the tile should always mean today, and when the applied range excludes today the widget should stay visible with a "No data available with applied filters" message rather than a number (ties into #1456).

**The MDMS part** — the headings and subtitles are config, so we can disambiguate them straight away in `dss.KpiDefinition`:

| KPI id | field | proposed |
|---|---|---|
| `cl_new_created_count` | `viz.subtitle` | "Filed in the selected period" |
[...]

Issue #1456, opened 2026-07-28
[Bug]  Dashboard: “No data for applied filters “ to be added for tiles


rahuljose-egov:
# Description

What went wrong? Be specific.

When a filter combination is applied that returns no matching records, affected tiles shows "No data". The tiles can display the wording "No data for applied filters" to make it obvious that the tile is empty because of the current filter selection.
---

# Steps to Reproduce

1. Set the Ward filter to Kipreres.
2. Set the Complaint Type filter to All Types.
3. Set the Date Range to 28/06/26 – 28/07/26.

Observe the tiles that have no matching data for this filter combination.

---

KDwevedi:
Heads-up for whoever picks this up: **#1483** (merged into `master`, for #1462) introduces a first reason code for the empty state — an entry that cannot be answered under the current filters returns `{ rows: [], rowCount: 0, suppressed: "filter_excludes_window" }` instead of a number.

So the copy work here has something concrete to key off: a tile can distinguish "no records matched" from "this tile cannot be answered for the selected range" and word them differently, rather than both falling back to a bare "No data". Today the suppressed case renders as the existing unavailable state.

KDwevedi:
- the meaning of `Tile` is not obvious
- Update the error message 'No Data' --> 'No Data to Display'

@rahuljose-egov to clarify error message that's more graceful

rahuljose-egov:
@KDwevedi "Not enough data to populate chart." can be the error message when applied filters do not return any data for computing KPI's. 

The next fallback incase anything fails, the error state can be defined separately - " Data could not be loaded. Please refresh dashboard."

PR #1483, opened 2026-07-29
fix(analytics): pin "complaints created today" to the calendar day (#1462)

### LINKED ISSUE #1462: [Bug] Complaint created today is showing some other number
# Description

Complaints created today is supposed to show only the complaints created today irrespective of the date range selected

---

# Steps to Reproduce

1. Keep date range for a moth and see the report
2. Keep date range for a week
3. Both shows different data for complaints created today

---

# Expected Behavior
Irrespective of the date range the complaints created today should be complaints created today(system date)

Image

---

# Actual Behavior
Not sure what number is showing. It is giving different numbers for different date range.
 #[/](url)
What happened? Include error messages if any.

---

# Environment

- **Deployment**: 
- **Tenant**: 
- **Browser**: 
- **User role**: 

---

# Screenshots / Logs

Attach screenshots, browser console errors, or service logs if available.

---

# Classification

> Add these labels before submitting. See [CONTRIBUTING.md](../CONTRIBUTING.md) for the full list.

- **Feature**: 
- **Area**: 
- **Priority**: 
- **Source**:

KDwevedi:
Fixes #1462. Also removes the mechanism behind the "why do these two tiles disagree" half of #1465.

## The bug

"Complaints created today" reported the **selected date range**, not today. Two independent causes:

1. The def carries no time predicate of its own — it relied on a `window` param. `KpiQueryComposer` deletes the base window whenever `dateFrom`/`dateTo` are present, so with a range applied the tile counted the whole range. With the pilot's 28/06–28/07 filter it read **8**, exactly equal to "New complaints created"; the true figure for that day was 7.
2. Its fallback window `last_1d` is a **rolling 24 hours**, not the calendar day — so even with no range applied it never meant "today", and drifted across midnight.

## The fix

- **`dtd`** ("day-to-date") joins `wtd`/`mtd`/`qtd`/`ytd` in the window vocabulary — the calendar day in EAT.
- **`window.pinned: true`** lets a def own its time axis: the `window` param and `dateFrom`/`dateTo` no longer rewrite the predicate.
- When the selected range **cannot contain** the pinned interval the entry is **suppressed** — no SQL is run, the result is `{ rows: [], rowCount: 0, suppressed: "filter_excludes_window" }` — so the tile renders "no data for the applied filters" rather than a number for the wrong period. This is the behaviour agreed with @subhamegov on the ticket: widget stays, data does not.
- Window-start resolution is now shared between `AnalyticsPlanner` and `KpiQueryComposer`, so a window cannot mean one thing when planned and another when range-checked.

Deliberate scoping:

- **Pinning fixes time, not filters** — `ward` / `serviceCode` / `complaintPath` / `hierLevel` still narrow a pinned query.
- **`compare:"prior"`** on a pinned window means the preceding window of equal span (yesterday, for `dtd`) instead of the prior-equal-duration of the selected range — which is what the tile's "vs yesterday" label always claimed.
- **`series:"daily"` is untouched** — the sparkline is trend context over the selected range, so it stays answerable even when the headline value is suppressed.

## Blast radius

`pinned` defaults to false, so **every other tile is unchanged** — the global range keeps governing their time axis exactly as before. That is asserted explicitly, not assumed: `unpinnedDefsStillHonourTheDateRange`, `unpinnedDefsAreNeverSuppressed`, `unpinnedWindowParamOverrideStillWorks`.

The only def flipped in this PR is `cl_created_today_count` (window pinned to `dtd`; its now-meaningless `window` param dropped).

## Tests

`KpiQueryComposerPinnedWindowTest` — 16 cases: `dtd` vs `last_1d` semantics, pinning beating both the `window` param and an explicit range, suppression on ranges ending before / starting after today, the inclusive single-day boundary, the marker never reaching SQL, prior-means-yesterday, sparkline survival, narrowing params still applying, and the three unpinned regression guards.

```
Tests run: 191, Failures: 0, Errors: 0, Skipped: 0   (full pgr-services suite)
```

## Not in this PR

- The empty-state **copy** ("No data available with applied filters") is #1456 — this PR gives that work a `suppressed` reason to render against; today the tile falls back to the existing unavailable state.
- The wording changes that disambiguate the two "created" tiles are the MDMS half of #1465.

## Summary by CodeRabbit

* **New Features**
  * Added pinned KPI time windows that remain fixed despite dashboard date-range or tile-window settings.
  * Introduced `dtd` calendar-day windows, prior-period comparisons, and daily-series support.
  * Updated the daily complaint-count KPI to use a fixed day-to-date window.
* **Bug Fixes**
  * Suppressed results when selected ranges don’t fully cover pinned windows.
  * Preserved non-time filters for pinned KPIs.
* **Documentation**
  * Documented pinned-window behavior and parameter handling.
* **Tests**
  * Added coverage for pinned-window composition, suppression, comparisons, and filtering.

### DIFF
--- a/ansible/nairobi-mdms/mdms/dss/KpiDefinition.json
+++ b/ansible/nairobi-mdms/mdms/dss/KpiDefinition.json
@@ -66,7 +66,12 @@
             "name": "total",
             "agg": "count"
           }
-        ]
+        ],
+        "window": {
+          "name": "dtd",
+          "timeRole": "filed_at",
+          "pinned": true
+        }
       },
       "supportsSeries": true,
       "viz": {
@@ -87,19 +92,7 @@
         "dateKey": "created_date",
         "sparklineMeasureKey": "total"
       },
-      "params": [
-        {
-          "name": "window",
-          "default": "last_1d",
-          "allowed": [
-            "last_1d",
-            "last_7d",
-            "last_30d",
-            "wtd",
-            "mtd"
-          ]
-        }
-      ],
+      "params": [],
       "rbac": {
         "visibleTo": []
       }
--- a/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsService.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsService.java
@@ -486,6 +486,19 @@ public class AnalyticsService {
      */
     private Map<String,Object> runOne(JsonNode q, AnalyticsScope scope, QueryTelemetry tel,
                                       String entryName, String kpiId){
+        // #1462: a pinned-window def whose interval falls outside the selected date range is
+        // unanswerable, not empty-by-filter. Return no rows WITHOUT running SQL, flagged so the tile
+        // renders "no data for the applied filters" rather than a zero the user would read as fact.
+        if (q.path(KpiQueryComposer.SUPPRESSED).asBoolean(false)) {
+            Map<String,Object> r = new LinkedHashMap<>();
+            r.put("grain", q.hasNonNull("grain") ? q.get("grain").asText() : "facts");
+            r.put("columns", Collections.emptyList());
+            r.put("rows", Collections.emptyList());
+            r.put("rowCount", 0);
+            r.put("suppressed", "filter_excludes_window");
+            r.put("tookMs", 0L);
+            return r;
+        }
         AnalyticsPlanner.Planned p = planner.plan(q, scope);
         long t0 = System.currentTimeMillis();
         List<Map<String,Object>> rows = jdbc.queryForList(p.sql, p.params.toArray());
@@ -511,7 +524,7 @@ public class AnalyticsService {
         Map<String,Object> out = new LinkedHashMap<>();
         out.put("aggFns", AnalyticsCatalog.AGG_FNS);
         out.put("filterOps", Arrays.asList("eq","ne","gt","gte","lt","lte","in","isnull","starts_with","subtree"));
-        out.put("windows", Arrays.asList("all","live","last_<N>d","wtd","mtd","qtd","ytd"));
+        out.put("windows", Arrays.asList("all","live","last_<N>d","dtd","wtd","mtd","qtd","ytd"));
         out.put("timeBuckets", Arrays.asList("day","week","month","quarter","year"));
         Map<String,Object> grains = new LinkedHashMap<>();
         for (Grain g : catalog.grains()) {
--- a/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsService.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsService.java
@@ -205,7 +205,11 @@ public class AnalyticsService {
      */
     private JsonNode resolveKpiRef(JsonNode queryNode, String tenantId, Set<String> callerRoles,
                                    List<String> paramsIgnored) {
-        if (!queryNode.has("kpiId")) return queryNode;
+        // Inline path: the suppression marker is composer-internal, so a caller-supplied one is
+        // stripped rather than trusted. Replaying a logged effective query (which legitimately
+        // carries it) must still be planned and validated, not short-circuited into a clean empty
+        // result that masks whatever the planner would have rejected.
+        if (!queryNode.has("kpiId")) return stripSuppressionMarker(queryNode);
 
         String kpiId = queryNode.get("kpiId").asText();
         Optional<KpiDefinition> def = kpiCatalogService.getDef(kpiId, tenantId);
@@ -343,6 +347,20 @@ public class AnalyticsService {
             if (srcQuery == null)
                 throw new IllegalArgumentException("kpi_forbidden: compose source '" + srcId.asText() + "' not authorized");
             Map<String,Object> r = runOne(srcQuery, scope, tel, entryName, srcId.asText());
+            // A suppressed source is UNANSWERABLE, not zero. firstRow() would flatten it to {} and
+            // computeCompose would read a missing measure as 0 — netBacklogDaily would then publish a
+            // confident number derived from a period the filter excludes. Propagate instead.
+            if (r.containsKey("suppressed")) {
+                Map<String,Object> out = new LinkedHashMap<>();
+                out.put("grain", "compose");
+                out.put("columns", Collections.emptyList());
+                out.put("rows", Collections.emptyList());
+                out.put("rowCount", 0);
+                out.put("compose", type);
+                out.put("suppressed", r.get("suppressed"));
+                if (!paramsIgnored.isEmpty()) out.put("paramsIgnored", paramsIgnored);
+                return out;
+            }
             sourceRows.add(firstRow(r));
         }
 
@@ -358,6 +376,20 @@ public class AnalyticsService {
         out.put("compose", type);
         if (!paramsIgnored.isEmpty()) out.put("paramsIgnored", paramsIgnored);
         return out;
+    }
+
+    /**
+     * Remove a caller-supplied {@link KpiQueryComposer#SUPPRESSED} marker. Copy-on-write: the request
+     * body is left untouched unless the marker is actually present.
+     */
+    private JsonNode stripSuppressionMarker(JsonNode queryNode) {
+        if (queryNode == null || !queryNode.isObject() || !queryNode.has(KpiQueryComposer.SUPPRESSED))
+            return queryNode;
+        com.fasterxml.jackson.databind.node.ObjectNode copy =
+                (com.fasterxml.jackson.databind.node.ObjectNode) queryNode.deepCopy();
+        copy.remove(KpiQueryComposer.SUPPRESSED);
+        log.debug("ignoring caller-supplied '{}' on an inline query", KpiQueryComposer.SUPPRESSED);
+        return copy;
     }
 
     /** Build a synthetic {kpiId, params} ref node so a source kpiId resolves through the normal path. */
--- a/backend/pgr-services/src/main/java/org/egov/pgr/analytics/KpiQueryComposer.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/analytics/KpiQueryComposer.java
@@ -85,11 +85,16 @@ import java.util.regex.Pattern;
  * <p><b>Pinned windows (#1462).</b> A def may declare {@code window: {name, pinned: true}} to keep
  * its own time axis — "complaints created today" means today whatever range the user picked. For
  * such a def the {@code window} param and {@code dateFrom}/{@code dateTo} do NOT rewrite the time
- * predicate. If the selected range does not overlap the pinned interval the entry is SUPPRESSED
- * (no rows, {@code suppressed:"filter_excludes_window"}) so the tile can render an explicit
- * "no data for the applied filters" instead of a number for the wrong period; {@code compare:"prior"}
- * then means the preceding window of equal span (yesterday, for {@code dtd}). Non-time params
- * (ward / serviceCode / complaintPath / hierLevel) still apply — pinning fixes time, not filters.
+ * predicate — a supplied {@code window} param is reported back as {@code paramsIgnored:["window"]}
+ * rather than silently swallowed. If the selected range does not COVER the pinned interval the entry
+ * is SUPPRESSED (no rows, {@code suppressed:"filter_excludes_window"}) so the tile can render an
+ * empty state instead of a number for a period the filter excludes; {@code compare:"prior"} then
+ * means the preceding window of equal span (yesterday, for {@code dtd}), and {@code series:"daily"}
+ * gets an axis wider than the pin (the range, else the {@code window} param, else a rolling default)
+ * so the sparkline is a trend rather than a single point. Non-time params (ward / serviceCode /
+ * complaintPath / hierLevel) still apply — pinning fixes time, not filters. Pinning a BOUNDLESS
+ * window ({@code all} / {@code live}) is meaningless and ignored: there is no interval to cover and
+ * no preceding period, so such a def takes the ordinary path.
  *
  * <p>{@code compare}/{@code series} compose with {@code window}/{@code dateFrom}/{@code dateTo}/
  * {@code ward}/{@code serviceCode}: the window/range params resolve the <em>current</em> range first,
@@ -109,6 +114,11 @@ public class KpiQueryComposer {
     private static final long MS_PER_DAY = 86_400_000L;
     /** Sparkline daily-series safety cap, matching the FE {@code Math.min(366, ...)}. */
     private static final int MAX_SERIES_DAYS = 366;
+    /**
+     * Trend axis for a pinned def's sparkline when the request carries neither a date range nor a
+     * window param — a pinned window is too narrow to be a trend (dtd would be one point).
+     */
+    private static final String DEFAULT_SERIES_WINDOW = "last_30d";
     /**
      * Marker written onto the merged query when a pinned window cannot be answered under the
      * selected range (#1462). Read and stripped by {@link AnalyticsService} before planning — it is
@@ -185,20 +195,31 @@ public class KpiQueryComposer {
         // A PINNED window owns its own time axis: the def declares what period it means ("today",
         // "this week") and the dashboard's date range must not redefine it (#1462). Unlike a live-open
         // snapshot — which is timeless and simply ignores the range — a pinned window occupies a real
-        // interval, so a range that excludes that interval makes the tile unanswerable rather than
-        // merely unfiltered. That case is SUPPRESSED: no value, so the tile can say "no data for the
-        // applied filters" instead of quietly reporting a number for a different period.
-        if (isPinnedWindow(next)) {
+        // interval, so a range that does not COVER that interval makes the tile unanswerable rather
+        // than merely unfiltered. That case is SUPPRESSED: no value, so the tile can render an empty
+        // state instead of quietly reporting a number for a period the filter excludes.
+        PinnedWindow pinned = pinnedWindow(next);
+        if (pinned != null) {
             if (series && !prior) {
-                // The sparkline is trend CONTEXT for the pinned value, not the value itself: it keeps
-                // its existing range-bucketed behaviour and stays answerable even when the pinned
-                // interval sits outside the range (the tile then shows a trend but no headline number).
+                // The sparkline is trend CONTEXT for the pinned value, not the value itself, and a
+                // trend needs an axis WIDER than the pinned interval — a pinned dtd would otherwise
+                // bucket "today" into a single flat point. So the series follows the selected range,
+                // else an explicit window param, else a rolling default; and it stays answerable even
+                // when the pinned interval sits outside the range (trend shown, headline suppressed).
+                applyPinnedSeriesAxis(next, params, bounds);
                 applyDailySeries(next, g, bounds);
             } else {
-                // Overlap is decided FIRST: applyPinnedPrior consumes the window node.
-                boolean suppress = bounds != null && !pinnedWindowOverlaps(next, bounds);
-                if (prior) applyPinnedPrior(next, g);
+                // Decided BEFORE the window is consumed below, and from the SAME clock reading the
+                // bounds are materialized with, so the suppression verdict and the executed SQL can
+                // never straddle EAT midnight and disagree.
+                boolean suppress = bounds != null && !rangeCoversPinnedWindow(pinned, bounds);
+                if (prior) applyPinnedPrior(next, g, pinned);
+                else       materializePinnedWindow(next, g, pinned);
                 if (suppress) next.put(SUPPRESSED, true);
+                // The window param cannot override a pin — say so, rather than silently ignoring it
+                // (the complaintPath precedent: an unappliable param is reported, never swallowed).
+                if (params.hasNonNull("window") && !params.get("window").asText().isEmpty())
+                    reportIgnored(paramsIgnoredOut, "window");
             }
             applyNarrowingParams(next, g, params, paramsIgnoredOut);
             return next;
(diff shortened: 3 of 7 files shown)


ansible/nairobi-mdms/mdms/dss/KpiDefinition.json
```diff
@@ -66,7 +66,12 @@
             "name": "total",
             "agg": "count"
           }
-        ]
+        ],
+        "window": {
+          "name": "dtd",
+          "timeRole": "filed_at",
+          "pinned": true
```
vinothrallapalli-eGov, line 73: **[CONFIRMED] digit-mcp seed copy of `cl_created_today_count` not updated.**

The pinned `dtd` fix is applied only to this ansible Nairobi seed; the parallel copy of the same catalog in `digit-mcp/src/tools/dashboard-catalog-seed.ts` (~line 429) still ships `cl_created_today_count` with the old unpinned query and `window` param default `last_1d`.

*Failure scenario:* a tenant onboarded or reseeded via digit-mcp's dashboard catalog seed tool gets the old def — its "Complaints created today" tile keeps reporting the dashboard's selected range (identical to "New complaints created" under a month filter) and falls back to a rolling 24h with no range. Bug #1462 remains user-visible on those tenants, and behavior diverges between ansible-seeded and mcp-seeded deployments.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/KpiQueryComposer.java
```diff
@@ -289,6 +336,68 @@ private void bindRange(ObjectNode query, String col, LocalDate from, LocalDate t
...
+     * range ending "today" counts as containing today regardless of clock time.
+     */
+    private boolean pinnedWindowOverlaps(JsonNode query, Bounds bounds) {
+        long now = System.currentTimeMillis();
+        Long startMs = AnalyticsPlanner.windowStartMs(query.path("window").path("name").asText(), now);
+        if (startMs == null) return true;                       // boundless window: nothing to miss.
+        LocalDate windowStart = Instant.ofEpochMilli(startMs).atZone(EAT).toLocalDate();
+        LocalDate today = Instant.ofEpochMilli(now).atZone(EAT).toLocalDate();
+        LocalDate rangeEnd = bounds.toExclusive.minusDays(1);   // back to the inclusive dateTo
+        return !rangeEnd.isBefore(windowStart) && !bounds.fromDate.isAfter(today);
```
vinothrallapalli-eGov: **[CONFIRMED] Suppression checks overlap, not containment.**

A range that partially overlaps a multi-day pinned window (`wtd`/`mtd`/`qtd`/`ytd`) is not suppressed, so the tile silently reports the full pinned-window count for a period the filter excludes — contradicting the documented contract ("if the selected range cannot contain the pinned interval, the entry is suppressed").

*Failure scenario:* a tenant pins `{name:'wtd', pinned:true}` on a "created this week" tile (legal per the new grammar/doc). On a Friday the user filters to yesterday–today. `pinnedWindowOverlaps` returns true (`rangeEnd >= Monday`, `fromDate <= today`), so no suppression: the tile shows the full Monday–Friday count under a 2-day filter with no indicator — exactly the wrong-period-number class #1462 set out to eliminate, now doc-blessed as impossible.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsService.java
```diff
@@ -486,6 +486,19 @@ boolean projectsForbiddenPii(JsonNode queryNode, Set<String> callerRoles) {
...
+        // #1462: a pinned-window def whose interval falls outside the selected date range is
+        // unanswerable, not empty-by-filter. Return no rows WITHOUT running SQL, flagged so the tile
+        // renders "no data for the applied filters" rather than a zero the user would read as fact.
+        if (q.path(KpiQueryComposer.SUPPRESSED).asBoolean(false)) {
+            Map<String,Object> r = new LinkedHashMap<>();
+            r.put("grain", q.hasNonNull("grain") ? q.get("grain").asText() : "facts");
+            r.put("columns", Collections.emptyList());
+            r.put("rows", Collections.emptyList());
+            r.put("rowCount", 0);
+            r.put("suppressed", "filter_excludes_window");
```
vinothrallapalli-eGov, line 530: **[CONFIRMED] `suppressed: 'filter_excludes_window'` has no consumer — the tile just goes blank.**

No frontend in the repo reads `suppressed`, and `KpiTile.resolveScalar` returns null for `rows:[]`, so the promised "no data for the applied filters" rendering never happens.

*Failure scenario:* a user filters the dashboard to last month. The "Complaints created today" headline comes back `rows:[], suppressed:'filter_excludes_window'`, and `digit-ui-esbuild/products/dashboard/src/components/KpiTile.jsx` renders a bare dash with no explanation while the sparkline still draws — the user can't tell an outage/zero apart from "filter excludes today", and the behaviour ANALYTICS-QUERY-API.md documents does not exist anywhere.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsService.java
```diff
@@ -486,6 +486,19 @@ boolean projectsForbiddenPii(JsonNode queryNode, Set<String> callerRoles) {
      */
     private Map<String,Object> runOne(JsonNode q, AnalyticsScope scope, QueryTelemetry tel,
                                       String entryName, String kpiId){
+        // #1462: a pinned-window def whose interval falls outside the selected date range is
+        // unanswerable, not empty-by-filter. Return no rows WITHOUT running SQL, flagged so the tile
+        // renders "no data for the applied filters" rather than a zero the user would read as fact.
+        if (q.path(KpiQueryComposer.SUPPRESSED).asBoolean(false)) {
```
vinothrallapalli-eGov, line 524: **[CONFIRMED] `__suppressed` marker is honored on client-supplied inline queries.**

`runOne` trusts the composer-internal `__suppressed` marker on ANY query node, including caller-supplied inline queries, so an inline body containing `"__suppressed": true` short-circuits before planner validation and returns a successful empty result instead of an error.

*Failure scenario:* an authenticated client (or a developer replaying a previously-logged effective/merged query, which legitimately contains `__suppressed`) POSTs it as an inline query: instead of the planner's `invalid_param`/validation verdict, the API returns `rowCount:0, suppressed:'filter_excludes_window', tookMs:0` with no error — malformed or forbidden-shaped inline queries are masked as clean empty results, and the marker the javadoc claims "never reaches" the service from outside is honoured from outside.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/KpiQueryComposer.java
```diff
@@ -166,6 +182,28 @@ public JsonNode mergeParams(JsonNode baseQuery, JsonNode params, List<String> pa
...
+        // snapshot — which is timeless and simply ignores the range — a pinned window occupies a real
+        // interval, so a range that excludes that interval makes the tile unanswerable rather than
+        // merely unfiltered. That case is SUPPRESSED: no value, so the tile can say "no data for the
+        // applied filters" instead of quietly reporting a number for a different period.
+        if (isPinnedWindow(next)) {
+            if (series && !prior) {
+                // The sparkline is trend CONTEXT for the pinned value, not the value itself: it keeps
+                // its existing range-bucketed behaviour and stays answerable even when the pinned
+                // interval sits outside the range (the tile then shows a trend but no headline number).
+                applyDailySeries(next, g, bounds);
```
vinothrallapalli-eGov, line 210: **[CONFIRMED] Default-state sparkline collapses to a single `dtd` bucket.**

With no active date range, the pinned-def sparkline keeps the `dtd` window (the deleted params default `last_1d` no longer supplies a wider axis), so the "trend context" collapses to a single day bucket.

*Failure scenario:* in the dashboard's default state (`dateRangeActive` false, so the FE `__series` ref carries `series:"daily"` but no `dateFrom`/`dateTo`), `applyDailySeries` takes its no-bounds branch and keeps the pinned `dtd` window: the created-today tile's sparkline is bucketed over today only and renders as a single flat point, not the trend the doc promises. Pre-PR, the declared default (`last_1d`, or any allowed window like `last_7d`) gave the series a multi-day axis.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/KpiQueryComposer.java
```diff
@@ -166,6 +182,28 @@ public JsonNode mergeParams(JsonNode baseQuery, JsonNode params, List<String> pa
         // delta uses a prior-week comparison and the sparkline a rolling window, per reference.)
         boolean liveOpenSnapshot = isLiveOpenSnapshot(next, g);
 
+        // A PINNED window owns its own time axis: the def declares what period it means ("today",
+        // "this week") and the dashboard's date range must not redefine it (#1462). Unlike a live-open
+        // snapshot — which is timeless and simply ignores the range — a pinned window occupies a real
+        // interval, so a range that excludes that interval makes the tile unanswerable rather than
+        // merely unfiltered. That case is SUPPRESSED: no value, so the tile can say "no data for the
+        // applied filters" instead of quietly reporting a number for a different period.
+        if (isPinnedWindow(next)) {
```
vinothrallapalli-eGov: **[CONFIRMED] Pinned branch silently drops the `window` param with no `paramsIgnored` report.**

This drops both the previously-honored allowed window overrides and the `invalid_param` rejection that the deleted KpiDefinition params allowed-list provided.

*Failure scenario:* an API consumer or saved report calling `cl_created_today_count` with `params {"window":"last_30d"}` — a value the def's now-deleted allowed list explicitly permitted and the server honored pre-PR — now silently receives today's count instead: no `invalid_param` error, no `paramsIgnored:["window"]` entry in the envelope (the `complaintPath` precedent in this same file reports unappliable params for exactly this reason). Likewise a previously-rejected value like `window:"qtd"` now succeeds silently instead of failing with `invalid_param`.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/KpiQueryComposer.java
```diff
@@ -289,6 +336,68 @@ private void bindRange(ObjectNode query, String col, LocalDate from, LocalDate t
...
+     */
+    private void applyPinnedPrior(ObjectNode query, Grain g) {
+        String col = dateFilterColumn(query, g);
+        if (col == null || !g.filterable.contains(col)) {
+            log.debug("grain '{}' has no filterable time column for a pinned compare:prior; skipping", g.name);
+            return;
+        }
+        long now = System.currentTimeMillis();
+        Long startMs = AnalyticsPlanner.windowStartMs(query.path("window").path("name").asText(), now);
+        if (startMs == null) return;
```
vinothrallapalli-eGov: **[CONFIRMED] Pinned boundless window makes the prior delta permanently 0%.**

`applyPinnedPrior` silently returns for a pinned boundless window (`all`/`live`, both accepted by `isPinnedWindow`), leaving the original window on the query, so the `__prior` entry re-runs the CURRENT query and the delta is permanently 0%.

*Failure scenario:* an admin publishes a def with `window {name:'all', pinned:true}` (nothing rejects it — `isPinnedWindow` only requires `name` + `pinned`) and `viz.delta compare:'prior'`. The FE's `<kpiId>__prior` request goes through the pinned branch, `applyPinnedPrior` returns at `startMs==null` without altering the query, prior == current, and the tile forever shows "0% vs prior" — a wrong, confidence-inspiring delta rather than an error or an unavailable delta.

backend/pgr-services/src/test/java/org/egov/pgr/analytics/KpiQueryComposerPinnedWindowTest.java
```diff
@@ -0,0 +1,232 @@
...
+    private final AnalyticsCatalog catalog = new AnalyticsCatalog();
+    private final KpiQueryComposer composer = new KpiQueryComposer(catalog);
+    private final AnalyticsPlanner planner = new AnalyticsPlanner(catalog);
+    private final AnalyticsScope stateScope = new AnalyticsScope("ke", true, null, null, null);
+
+    private JsonNode json(String s) {
+        try { return om.readTree(s); } catch (Exception e) { throw new RuntimeException(e); }
+    }
+
+    private LocalDate today() { return Instant.now().atZone(EAT).toLocalDate(); }
```
vinothrallapalli-eGov, line 42: **[CONFIRMED] Suppression-boundary tests race the clock across EAT midnight.**

These tests capture `today()` via `Instant.now()` before `mergeParams` internally calls `System.currentTimeMillis()`, so a test run that crosses EAT midnight (21:00 UTC — a common CI window) uses two different "today"s and the suppression assertions flip.

*Failure scenario:* CI starts `rangeOfExactlyTodayIsNotSuppressed` at 20:59:59.9 UTC: `today()` returns day D, then `mergeParams` computes now on day D+1 (EAT), so the range [D,D] ends before the new `dtd` window start and the query is suppressed — `assertFalse(suppressed)` fails and the build breaks spuriously. `rangeEndingBeforeTodaySuppresses`, `dateRangeDoesNotRewriteAPinnedWindow`, and `priorOnAPinnedDayMeansYesterday` have the same race. Consider injecting a `Clock`.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/KpiQueryComposer.java
```diff
@@ -166,6 +182,28 @@ public JsonNode mergeParams(JsonNode baseQuery, JsonNode params, List<String> pa
...
+        // applied filters" instead of quietly reporting a number for a different period.
+        if (isPinnedWindow(next)) {
+            if (series && !prior) {
+                // The sparkline is trend CONTEXT for the pinned value, not the value itself: it keeps
+                // its existing range-bucketed behaviour and stays answerable even when the pinned
+                // interval sits outside the range (the tile then shows a trend but no headline number).
+                applyDailySeries(next, g, bounds);
+            } else {
+                // Overlap is decided FIRST: applyPinnedPrior consumes the window node.
+                boolean suppress = bounds != null && !pinnedWindowOverlaps(next, bounds);
```
vinothrallapalli-eGov: **[PLAUSIBLE] Two independent clock reads let a request straddle EAT midnight.**

The suppression/overlap decision uses its own `System.currentTimeMillis()` while `AnalyticsPlanner.applyWindow` later resolves the pinned `dtd` window with a fresh clock read, so a request straddling EAT midnight passes the overlap check for the old day but executes SQL for the new day.

*Failure scenario:* a dashboard request with `dateTo = today` arrives at 23:59:59.9 EAT: `pinnedWindowOverlaps` approves it against the old calendar day, then the planner computes `dtd` after midnight — the tile returns the new day's count (typically 0) labeled under a range that excludes that day. Sub-second window, but silent and unflagged; resolving the window once and passing it through (or injecting a `Clock`) closes it.

vinothrallapalli-eGov (commented):
10 findings — 9 inline below, plus one on a line outside the diff:

**[CONFIRMED] `AnalyticsService.java:346` — Composed KPIs turn suppressed sources into fabricated zeros.**
The compose-source call site of `runOne()` treats a suppressed pinned-window source as an empty row set: `firstRow()` returns an empty map and `computeCompose()`'s `orZero()` turns the unanswerable source into `0`, so composed KPIs silently compute with a fabricated zero instead of propagating suppression.

*Failure scenario:* an admin publishes a compose def allowed by the `dss.KpiDefinition` schema, e.g. `netBacklogDaily` with `sourceKpiIds [cl_created_today_count, cl_resolved_today_count]`. A dashboard user selects a date range ending yesterday: the pinned created-today source is suppressed (`rows:[]`), inflow becomes 0, and the "Net backlog today" tile displays `-<resolved count>` as a real number — exactly the wrong-period-number-presented-as-fact failure #1462 set out to eliminate, with no `suppressed` flag on the composed result.

---
Minor cleanup items (duplicated bindRange logic, grain fallback mismatch in the suppressed branch, dead code) were found but omitted here to keep the review focused.

ansible/nairobi-mdms/mdms/dss/KpiDefinition.json
KDwevedi: Confirmed and fixed in 8d598a0. `digit-mcp/src/tools/dashboard-catalog-seed.ts` now carries the same pinned `dtd` window and the emptied `params` — mcp-seeded tenants would otherwise have kept #1462 while ansible-seeded ones were fixed. Verified the seed array still parses (39 defs) with the def reading as expected.

Worth noting the duplication itself is the underlying hazard: two hand-maintained copies of the same catalog will drift again on the next KPI change. Raising that separately rather than folding a de-duplication into this PR.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/KpiQueryComposer.java
KDwevedi: Confirmed and fixed in 8d598a0. `pinnedWindowOverlaps` is now `rangeCoversPinnedWindow` — the range must cover `[windowStart, today]` in full, not merely intersect it. Your `wtd`-on-a-Friday case now suppresses.

You are right that the doc blessed it as impossible; that wording is what made this a contract bug rather than a preference. `dtd` is unaffected either way — for a single-day window coverage and overlap coincide — so the shipped def behaves identically and only the general case changes.

Test added: `partialOverlapOfAMultiDayPinIsSuppressed`, pinned `wtd` under a two-day filter (skipped early in the week, where the range genuinely does cover week-to-date).

backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsService.java
KDwevedi: Correct — no frontend reads `suppressed`, and I overstated the contract in the doc. Fixed the doc rather than the frontend, in 8d598a0.

`ANALYTICS-QUERY-API.md` now says what is actually true: `suppressed` is a machine-readable reason code on the response, the dashboard currently falls back to its existing unavailable state, and wording the empty state — plus distinguishing it from an ordinary zero — is #1456. This PR exists partly to give #1456 something to key off; I have linked it from that ticket.

I deliberately did not implement the rendering here: the copy ("No data available with applied filters") is #1456's open wording decision with @subhamegov, and picking a string in this PR would pre-empt it. Happy to fold it in if you would rather this PR carry the visible behaviour end-to-end.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsService.java
KDwevedi: Confirmed and fixed in 8d598a0. `resolveKpiRef` now strips the marker on the inline path (copy-on-write, so an untouched body is not cloned), meaning a replayed effective query is planned and validated normally instead of short-circuiting into a clean empty result.

Your point about replaying a logged merged query is the sharp end of it: that node legitimately contains `__suppressed`, so this was reachable without any malicious intent.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/KpiQueryComposer.java
KDwevedi: Confirmed and fixed in 8d598a0. The pinned branch now sets a series axis wider than the pin before delegating: selected range → explicit `window` param → rolling `last_30d`, and it clears `pinned` on that axis since a trend is not the pin.

The `window` param is honoured *here* precisely because it is meaningful for a trend axis, which also gives the param a purpose again on the series path while it stays reported-as-ignored on the scalar path. Two tests: `defaultStateSparklineGetsAWiderAxisThanThePin` and `windowParamSetsTheSparklineAxis`.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/KpiQueryComposer.java
KDwevedi: Confirmed and fixed in 8d598a0. A `window` param supplied to a pinned def on the scalar/prior path now comes back as `paramsIgnored:["window"]`, using a shared `reportIgnored` helper (the inline `complaintPath` reporting is refactored onto it).

On the lost `invalid_param`: the allow-list rejection went away with the def's `params` block, which had to go — the values it permitted (`last_7d`, `mtd`, …) are exactly the overrides a pinned tile must refuse. So an out-of-list value is now reported rather than rejected. If you would rather a pinned def reject a `window` param outright with `invalid_param`, say so and I will switch it; I chose reporting because a saved report or bookmarked URL carrying `window` should still render the tile correctly rather than error.

Test: `windowParamDoesNotOverrideAPinnedWindowAndIsReported`.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/KpiQueryComposer.java
KDwevedi: Confirmed and fixed in 8d598a0. `pinnedWindow()` now returns null for a window with no start instant, so `pinned: true` on `all`/`live` is ignored (debug-logged) and the def takes the ordinary path — prior shifts the range as it always did, instead of silently equalling current.

Test: `pinningABoundlessWindowIsIgnored` asserts the prior bounds actually move.

backend/pgr-services/src/test/java/org/egov/pgr/analytics/KpiQueryComposerPinnedWindowTest.java
KDwevedi: Fair, and fixed in 8d598a0 — though not with a `Clock`.

The five day-sensitive cases now run inside an `onAStableDay(...)` guard that re-reads the EAT date afterwards and uses a JUnit *assumption*, so a run crossing midnight aborts the case instead of failing the build. That keeps the suite honest in both directions: it never asserts against a stale day, and never reports red for one.

I considered injecting a `Clock` into the planner and composer. It is the better long-term shape, but it changes two production constructors and every call site for a test-only benefit, and the production race you flagged separately is now closed at the source (single resolution per request — see the reply on the two-clock-reads comment). Happy to do the `Clock` refactor as a follow-up if you would rather have determinism than an assumption.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/KpiQueryComposer.java
KDwevedi: Agreed, and closed at the source in 8d598a0 rather than mitigated.

A pinned window is now resolved **once** per request into a `PinnedWindow` record (start instant, now, and both EAT calendar days), and that same snapshot drives the suppression verdict, the materialized `gte`/`lt` bounds and the prior period. The window node is then removed, so `AnalyticsPlanner.applyWindow` never re-resolves it against a second clock read — your 23:59:59.9 case can no longer approve against one day and execute against the next.

Test: `pinnedWindowIsMaterializedNotLeftForThePlanner`. Note one pre-existing carve-out it documents: an entirely empty `params` object short-circuits `mergeParams` and returns the base query untouched, in which case the planner resolves the window itself — still a single clock read, so no exposure.

KDwevedi:
> **Preserve source suppression through compose computation.** (`AnalyticsService.java` 312-361)

Confirmed and fixed in 8d598a0. `maybeComposeResult` now checks each source result for `suppressed` before `firstRow()` flattens it, and short-circuits the whole compose with the source's reason code:

```json
{ "grain": "compose", "rows": [], "rowCount": 0, "compose": "<type>", "suppressed": "filter_excludes_window" }
```

Failing the compose rather than propagating a per-source flag, because the alternative — computing from the sources that *are* answerable — is the misleading-partial-result case you describe. `netBacklogDaily` can no longer read an unanswerable source as `0`.

---

Also addressed in the same commit, from @vinothrallapalli-eGov's review: suppression now requires the range to **cover** the pinned interval rather than merely overlap it; `pinned` is ignored on boundless windows; the pinned window is resolved once and materialized into explicit bounds (closing a two-clock-read midnight race); an unappliable `window` param is reported via `paramsIgnored`; the sparkline gets an axis wider than the pin; a caller-supplied `__suppressed` marker is stripped on inline queries; and the digit-mcp copy of the seed def is updated.

196 tests pass.

vinothrallapalli-eGov (approved):

PR #1511, opened 2026-07-30
fix(dashboard): remove broken header search box (#1474)

KDwevedi:
Fixes #1474

## Problem

The dashboard header search input advertised `Search complaints, wards, citizens.` but only substring-matched **KPI widget titles**, client-side. Any query that didn't match a tile title applied `.dashboard-search-dimmed` (`opacity: 0.12; pointer-events: none`) to every widget on the grid — so the whole dashboard went dim and non-interactive, with no results, no empty state, and no network request. Clearing the box was the only way out, which is exactly the "permanent dimmed/loading state" reported.

## Fix

Remove the control entirely rather than reworking it. The complaint/ward/citizen search it advertises is not implemented on this surface, and inbox search already covers that need.

Deleted:
- the `<input type="search">` and its local `SearchIcon` (`DashboardHeader.jsx`)
- the `searchQuery` / `onSearchQueryChange` prop chain (`DashboardLayout.jsx`)
- the `matchesSearch` callback, `searchQuery` state, and `dimClass` application (`AdminDashboard.jsx`)
- the associated rules in `styles/input.css`, including the `.dashboard-header-mobile-search` wrapper that was already dead (no JSX referenced it)
- `utils/dashboardSearch.js` (119 lines) — exports `widgetMatchesSearch` / `countMatchingWidgets`, documented as matching "the header search query". A richer matcher than the inline one `AdminDashboard` actually used, and never wired up: nothing in the repo imports it, and the only references to either export are inside the file itself. Already dead before this branch; unambiguously so now.

`styles/dashboard.css` is a committed build artifact; it was regenerated with `yarn build:dashboard-css` (tailwindcss 3.4.17) rather than hand-edited. **The regenerated diff there is pure deletion (90 lines, 0 additions)**, which also confirms the committed artifact was in sync with `input.css` before this change.

## Verification

- Confirmed no remaining references to `searchQuery` / `matchesSearch` / `dimClass` / `dashboard-search-dimmed` / `dashboard-header-search` anywhere in the dashboard subtree or its styles.
- Confirmed nothing else is orphaned: `ExportIcon` stays, `.dashboard-header-controls` is still needed by the Add KPI / Reset / Export buttons, and `useState` / `useCallback` remain in use in both touched components.
- All three modified `.jsx` files parse clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Changes**
  * Removed dashboard tile search from the header.
[...]

PR #1512, opened 2026-07-30
feat(dashboard): add "Powered by DIGIT" footer (#1453)

### LINKED ISSUE #1453: [Enhancement] "Powered by DIGIT" footer addition to dashboard
Add "Powered by DIGIT" footer at the bottom of the dashboard

KDwevedi:
Fixes #1453

## Problem

The citizen and employee shells both render the DIGIT attribution footer (`packages/modules/core/src/pages/{citizen,employee}/index.js`), but the dashboard never did. Root cause: `App.js` routes `/employee/dashboard` straight to `<AdminDashboard />` **instead of** `<DigitUI>`, so the dashboard inherits none of that shell's chrome.

## Fix

New `DashboardFooter` local to the dashboard subtree, mounted in `DashboardLayout`.

**Why local instead of importing `HomeFooter` from `@egovernments/digit-ui-components`** (which is already a dependency and does export it): the dashboard subtree is deliberately self-contained, and `HomeFooter`'s class names are unprefixed while this subtree runs Tailwind with a `tw-` prefix and its own palette — importing it would pull styling in from the CDN stylesheet outside the dashboard's design system.

**Config:** reads the same two globalConfigs keys as every other surface — `DIGIT_FOOTER` for the logo, `DIGIT_HOME_URL` for the link target. Both are already defined in every deployed variant (`configs/assets/globalConfigsPGR.js`, `local-setup/nginx/globalConfigs.js`, `public/globalConfigs.example.js`), so **no config or MDMS change is needed**. Fallbacks are baked into the component so the logo still renders anywhere `globalConfigs.js` hasn't been injected.

**Placement:** mounted as a sibling of `<main>`, not a child. The shell is `tw-h-screen` with an `overflow-hidden` flex column and `<main>` is the scroll container — a footer inside it would scroll away with the react-grid-layout grid instead of staying pinned at the bottom. A `position: fixed` footer (what the citizen shell does) is also wrong here for the same reason: the page itself never scrolls.

## Notes

- No localization change: the footer is an `Powered by DIGIT`, same as every other surface — the visible content is an image, not text. (The dashboard subtree also has no i18n wiring at all today.)
- `styles/dashboard.css` is a committed build artifact, regenerated with `yarn build:dashboard-css` (tailwindcss 3.4.17). The only new rule is `.dashboard-footer-img`; every `tw-` utility the component uses was already present in the generated output.
- **Not included:** the pre-auth `DashboardLogin` screen. The classic DIGIT login pages do carry the b/w footer variant, but that screen is a centered flex box and would need a layout change to host one — out of scope for this issue. Happy to fold it in if you'd like it for parity.

## Summary by CodeRabbit

- **New Features**
  - Added a dashboard footer with a configurable “Powered by DIGIT” logo.
  - Added a linked home-page destination, with a default fallback URL when none is configured.
  - Footer links open securely in a new browser tab.

- **Style**
  - Added responsive styling for the footer logo, including proportional sizing and pointer interaction.

### DIFF
--- a/frontend/micro-ui/web/src/dashboard/components/DashboardFooter.jsx
+++ b/frontend/micro-ui/web/src/dashboard/components/DashboardFooter.jsx
@@ -1,0 +1,34 @@
+import React from "react";
+
+// Link target when the deployment has not configured DIGIT_HOME_URL.
+const DEFAULT_HOME_URL = "https://egov.org.in/digit/";
+
+const getConfig = (key) => window?.globalConfigs?.getConfig?.(key);
+
+// "Powered by DIGIT" attribution, matching the citizen/employee DIGIT shells.
+// The dashboard bypasses that shell (App.js routes /employee/dashboard straight
+// to AdminDashboard), so it has to render its own.
+//
+// The logo comes from globalConfigs only — no baked-in URL — so a deployment can
+// rebrand or drop the attribution without a code change. When DIGIT_FOOTER is
+// unset we render nothing rather than a broken image, which is the same guard
+// the employee shell uses (packages/modules/core/src/pages/employee/index.js).
+const DashboardFooter = () => {
+  const logoUrl = getConfig("DIGIT_FOOTER");
+  if (!logoUrl) return null;
+
+  return (
+    <footer className="dashboard-footer tw-flex tw-flex-shrink-0 tw-items-center tw-justify-center tw-border-t tw-border-border tw-bg-surface tw-py-2">
+      <a
+        href={getConfig("DIGIT_HOME_URL") || DEFAULT_HOME_URL}
+        target="_blank"
+        rel="noopener noreferrer"
+        className="tw-inline-flex tw-items-center"
+      >
+        <img className="dashboard-footer-img" alt="Powered by DIGIT" src={logoUrl} />
+      </a>
+    </footer>
+  );
+};
+
+export default DashboardFooter;
--- a/frontend/micro-ui/web/src/dashboard/components/DashboardLayout.jsx
+++ b/frontend/micro-ui/web/src/dashboard/components/DashboardLayout.jsx
@@ -2,6 +2,7 @@ import React, { useMemo } from "react";
 import { getBrandTheme } from "../config/dashboardConfig";
 import DashboardHeader from "./DashboardHeader";
 import DashboardFilters from "./DashboardFilters";
+import DashboardFooter from "./DashboardFooter";
 import Sidebar from "./Sidebar";
 
 const DashboardLayout = ({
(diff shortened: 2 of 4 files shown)


KDwevedi:
@claude review

frontend/micro-ui/web/src/dashboard/components/DashboardFooter.jsx
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+
+// Fallbacks for the two globalConfigs keys this footer reads. Every deployed
+// globalConfigs variant defines both (configs/assets/globalConfigsPGR.js,
+// local-setup/nginx/globalConfigs.js), but the dashboard is also mounted in
+// setups where globalConfigs.js has not been injected yet, and a broken image
+// is worse than a stale-but-correct one.
+const DEFAULT_FOOTER_LOGO =
+  "https://s3.ap-south-1.amazonaws.com/egov-uat-assets/digit-footer.png";
+const DEFAULT_HOME_URL = "https://www.digit.org/";
```
vinothrallapalli-eGov: @KDwevedi Please use https://egov.org.in/digit/ as the DEFAULT_HOME_URL.

Also, avoid hardcoding the footer logo URL. Instead, read it from the global configuration so it can be configured without code changes.

vinothrallapalli-eGov (changes requested):

frontend/micro-ui/web/src/dashboard/components/DashboardFooter.jsx
KDwevedi: Both done, thanks — and you were right that the fallback defeated the point.

**Home URL** — `DEFAULT_HOME_URL` is now `https://egov.org.in/digit/`.

**Logo** — removed the hardcoded URL entirely rather than just moving it. The component now reads `DIGIT_FOOTER` from globalConfigs and **renders `null` when it is unset**, instead of falling back to a baked-in S3 path. That was the real problem with my first version: a fallback meant a deployment that deliberately left `DIGIT_FOOTER` unset would still get the eGov logo, so it was not actually configurable. Rendering nothing also matches the guard the employee shell already uses (`packages/modules/core/src/pages/employee/index.js:159` gates on the same key), so the two surfaces now behave the same way.

Net effect: no image URL anywhere in the code, and a partner can rebrand or drop the attribution purely through `globalConfigs.js`.

Also rebased onto master — #1511 merged in the meantime and had touched the same two stylesheets, which had put this PR into conflict. Now mergeable again. `dashboard.css` was regenerated with `yarn build:dashboard-css` on top of the new base rather than hand-merged, so the only rule it adds is `.dashboard-footer-img`.

vinothrallapalli-eGov (approved):

PR #1515, opened 2026-07-30
feat(dashboard): cohesive auth module with silent token refresh (#1466)

### LINKED ISSUE #1466: [Bug] Dashboard: 401 error when user session times out
# Description

What went wrong? Be specific.
When a user leaves the dashboard idle for a period, the session expires in the background, but the UI does not handle the expiry gracefully. Instead of prompting the user to re-authenticate, the dashboard surfaces a raw 401 (Unauthorized) error. The user is left with an error state and no clear guidance on what happened or how to recover. 

Reloading the page did not work, but since it was a 401 error, I have logout and logged back in for session to be continued. 

---

# Expected Behavior

Session expiry is handled gracefully. On detecting an expired session (401), the app should either silently refresh the token (if a valid refresh token exists) and retry the request, or redirect the user to the login screen with a clear message such as "Your session has expired. Please log in again." The user should not see a raw 401 error, and should land back where they were after re-authenticating where feasible.
---

KDwevedi:
Fixes #1466

## Problem

Auth had no owner. `getEmployeeToken` / `getEmployeeInfo` / `buildRequestInfo` were duplicated **verbatim** across `analyticsService`, `boundaryService` and `complaintHierarchyService`; `DashboardLogin` wrote localStorage directly; and **nothing inspected `response.status`**. Hence the three symptoms in the issue:

1. A 401 rendered as the literal string `Analytics request failed (401)` in a banner.
2. The auth gate was `useState(() => hasDashboardSession())` — evaluated **once at mount**, so a session dying mid-tab could never flip it.
3. **Reload didn't help** (the reporter had to sign out and in) because `hasDashboardSession()` only checked the token was *present*, never valid.

## What this does

`services/authService.js` owns token storage, `RequestInfo`, and the 401 cycle. All four authenticated call sites go through `authFetch`. Login persists `refresh_token` (previously discarded). An unrecoverable session drops to the login gate with an explanation.

### Server behaviour — verified against egov-user, not assumed

| | |
|---|---|
| `grant_type=refresh_token` | supported, returns a new `access_token` |
| Refresh token rotation | **not rotated** — same value returned |
| Old access token after refresh | **immediately invalidated → 401** |
| Dead refresh token | **HTTP 400 `invalid_grant`** (not 401) |
| Token format | **opaque UUIDs, not JWTs** |

Consequences that shaped the design: refresh must be **single-flight** and the request replayed (otherwise concurrent 401s invalidate each other's tokens); a failed refresh can't re-enter the retry path (it's a 400, not a 401); and refresh is necessarily **reactive**, since an opaque token has no inspectable expiry.

### The 401 contract

A 401 has four causes and they are **not** interchangeable. Treating them alike is what made the first cut destructive — every 401 meant "session dead" and wiped storage **shared with the co-hosted digit-ui app**.

| Cause | Result | Session |
|---|---|---|
| Another caller already refreshed | replay with new token | — |
| Token dead, refresh succeeded | replay | — |
| Token dead, server refused refresh | `err.sessionExpired` | storage **cleared** |
| Token dead, no refresh token held | `err.sessionExpired` | storage **kept** |
| Auth server unreachable | `err.refreshUnavailable` | **kept** |
| Fresh token, endpoint still 401s | `err.endpointUnauthorized` | **kept** |

Only the two `sessionExpired` rows dispatch `SESSION_EXPIRED_EVENT`. **The invariant: never destroy a session we have not proven is dead, because the keys are not ours alone.**

Guarantees: at most two sends and one refresh per call (cannot loop); concurrent 401s share one refresh; a refresh never re-enters the retry path.

### The dashboard is not standalone

`App.js:67` routes `/employee/dashboard` to `AdminDashboard` inside the **same SPA** that renders `DigitUI`, and `index.js:60-72` copies `Employee.token` into `Digit.SessionStorage`'s `User` at boot — the cache the rest of the employee UI authenticates from, which is never re-read. So this module shares its token *and* its localStorage keys with the surrounding app.

- `persistSession` now syncs `Digit.SessionStorage`, so a silent refresh doesn't kill the main app's token (no-op in the standalone build).
- Storage is cleared only on a **positive rejection**. This matters because the only writer of `Employee.refresh-token` is this dashboard's own login form — a session seeded by the main digit-ui login has none, and per `DashboardLogin.jsx:19` that is the *normal* path. Clearing unconditionally turned one dashboard 401 into a full-product logout.

### Predictability

- Session keys come from **one `KEYS` table** that also marks which keys the co-hosted app reads. Previously "what login writes" and "what signout clears" were two hand-maintained lists that had already drifted (the refresh token was written, never cleared). That class of bug is now structurally impossible.
- The caller error contract is documented and uniform: `analyticsService` propagates, `boundaryService` returns partial data, `complaintHierarchyService` returns null. Services that swallow still get the gate flip, since the event fires before the throw.

## Review history

Three independent passes — CodeRabbit, Codex, and Claude Fable. Every defect was in the refresh half; the 401→login half was correct from commit 1.

| Found by | Severity | Issue |
|---|---|---|
| Codex + Fable (independently) | HIGH | Late-401 race: single-flight only merged 401s *overlapping* the refresh window, so a later 401 started a second refresh that invalidated the token the first caller was retrying with — producing the exact false logout this PR prevents. Fixed with a `tokenGeneration` counter. |
| Fable | HIGH | Co-hosted digit-ui session breakage (above). |
| Codex + Fable | MED | A stalled `/user/oauth/token` left every 401'd caller pending forever. Now aborts at 15s. |
| Fable | MED | A network blip mapped to the same `false` as `invalid_grant`, deleting a valid session — a laptop waking from sleep hit this. Outcomes are now three-valued. |
| Fable | MED | Persistently-401ing endpoint became a login loop. Now `endpointUnauthorized`. |
| CodeRabbit + Fable | MED | `fetchBoundariesByCodes` could throw where it never did, asymmetric with its own sibling. |
| Fable | LOW | Refresh reverted login's role ordering. `persistSession` normalises. |

On simplicity, all three reviewers independently rejected the same alternatives: a global `fetch` interceptor is wrong here (the token lives *inside* the JSON body, so it would parse and mutate arbitrary bodies — and would intercept the co-hosted app's traffic), per-service retry re-duplicates what this PR deletes and can't share single-flight state, and react-query doesn't apply to hand-rolled promises. Dropping refresh entirely is genuinely simpler and was consciously declined in favour of fixing the co-hosting.

## Verification

Harness driving `authService` against a mocked `fetch`:

```
late-401 race        A(T1)401 REFRESH#1->T2 A(T2)200 B(T1)401 B(T2)200 -> 1 refresh, 0 expiry events
5 concurrent 401s    1 refresh, 5 successes
no refresh token     gate flips, shared keys SURVIVE
refresh rejected     gate flips, shared keys cleared
network blip         no expiry event, refresh token survives
endpoint 401s w/ fresh token   no expiry event, session intact
refresh ok           Digit.SessionStorage.User.token updated, roles non-EMPLOYEE-first
dead refresh token   1 attempt then expire
clearSession         leaves nothing behind
```

**Not covered:** a real browser session against a live tenant. This repo has no PR-time frontend build (`pgr-ui-build.yml` is `workflow_dispatch`-only). The specific thing no harness can prove is the co-hosting fix — that after a silent refresh the surrounding employee UI still works. Worth a bomet smoke test before merge.

## Summary by CodeRabbit

- **New Features**
  - Added centralized dashboard authentication with automatic session renewal and secure session handling.
  - Added clear session-expiration feedback and returned users to the login experience when their session ends.
  - Improved authentication support across dashboard data and complaint hierarchy requests.

- **Bug Fixes**
  - Prevented active dashboard sessions from appearing authenticated after expiration.
  - Improved handling of unauthorized requests and authentication failures.
  - Dashboard data loading now continues when individual boundary requests fail.

### DIFF
--- a/digit-ui-esbuild/products/dashboard/src/components/DashboardLogin.jsx
+++ b/digit-ui-esbuild/products/dashboard/src/components/DashboardLogin.jsx
@@ -5,6 +5,13 @@ import {
   getStateLabel,
   getBrandTheme,
 } from "../config/dashboardConfig";
+import {
+  clearSession,
+  getOAuthBasic,
+  hasAuth,
+  persistSession,
+  withSignificantRoleFirst,
+} from "../services/authService";
 import useDashboardT from "../i18n/useDashboardT";
 
 /**
@@ -25,8 +32,6 @@ import useDashboardT from "../i18n/useDashboardT";
  * they do not bypass authentication.
  */
 
-// Basic egov-user-client: (empty secret), base64 of "egov-user-client:"
-const OAUTH_BASIC = "Basic ZWdvdi11c2VyLWNsaWVudDo=";
 
 // label/hint are functions of t so they resolve at render time (never frozen at import).
 const DEMO_USERS = [
@@ -47,35 +52,11 @@ const DEMO_USERS = [
   },
 ];
 
-/** Order roles so the first non-EMPLOYEE role leads (drives the scoping badge). */
-function withSignificantRoleFirst(userInfo) {
-  if (!userInfo || !Array.isArray(userInfo.roles)) return userInfo;
-  const roles = [...userInfo.roles].sort(
-    (a, b) => (a.code === "EMPLOYEE" ? 1 : 0) - (b.code === "EMPLOYEE" ? 1 : 0)
-  );
-  return { ...userInfo, roles };
-}
 
-export function hasDashboardSession() {
-  try {
-    const raw = window.localStorage?.getItem("Employee.token");
-    return Boolean(raw && raw !== "undefined" && raw !== "null");
-  } catch {
-    return false;
-  }
-}
-
-export function clearDashboardSession() {
-  ["Employee.token", "Employee.user-info", "Employee.tenant-id", "user-info", "token"].forEach(
-    (k) => {
-      try {
-        window.localStorage?.removeItem(k);
-      } catch {
-        /* ignore */
-      }
-    }
-  );
-}
+// Session storage lives in authService (single source of truth); these thin
+// re-exports keep the existing AdminDashboard import sites unchanged.
+export const hasDashboardSession = hasAuth;
+export const clearDashboardSession = clearSession;
 
 const inputStyle = {
   borderRadius: "0.375rem",
@@ -87,12 +68,16 @@ const inputStyle = {
   outline: "none",
 };
 
-const DashboardLogin = ({ onLogin }) => {
+const DashboardLogin = ({ onLogin, expired = false }) => {
   const { t } = useDashboardT();
   const [username, setUsername] = useState("");
   const [password, setPassword] = useState("");
   const [submitting, setSubmitting] = useState(false);
-  const [error, setError] = useState(null);
+  const [error, setError] = useState(
+    // Distinguish "your session ended" from "you got the password wrong": the
+    // user did nothing wrong and reloading will not help them (#1466).
+    expired ? "Your session has expired. Please sign in again." : null
+  );
 
   const tenantId = getTenantId();
   const productLabel = useMemo(() => getProductLabel(), []);
@@ -127,7 +112,7 @@ const DashboardLogin = ({ onLogin }) => {
         method: "POST",
         headers: {
           "Content-Type": "application/x-www-form-urlencoded",
-          Authorization: OAUTH_BASIC,
+          Authorization: getOAuthBasic(),
         },
         body,
       });
@@ -143,11 +128,14 @@ const DashboardLogin = ({ onLogin }) => {
       }
       const data = await res.json();
       const userInfo = withSignificantRoleFirst(data.UserRequest);
-      window.localStorage.setItem("Employee.token", JSON.stringify(data.access_token));
-      window.localStorage.setItem("Employee.user-info", JSON.stringify(userInfo));
-      window.localStorage.setItem("Employee.tenant-id", JSON.stringify(tenantId));
-      window.localStorage.setItem("user-info", JSON.stringify(userInfo));
-      window.localStorage.setItem("token", JSON.stringify(data.access_token));
+      // refresh_token is persisted (owner-stamped) so authService can silently
+      // renew the access token on a 401 instead of forcing a re-login (#1466).
+      persistSession({
+        accessToken: data.access_token,
+        refreshToken: data.refresh_token,
+        userInfo,
+        tenantId,
+      });
       if (onLogin) onLogin(userInfo);
     } catch (err) {
       setError(err.message || `${t("DASHBOARD_LOGIN_SIGN_IN_FAILED", "Sign-in failed")}.`);
--- a/digit-ui-esbuild/products/dashboard/src/components/DashboardLogin.jsx
+++ b/digit-ui-esbuild/products/dashboard/src/components/DashboardLogin.jsx
@@ -76,7 +76,12 @@ const DashboardLogin = ({ onLogin, expired = false }) => {
   const [error, setError] = useState(
     // Distinguish "your session ended" from "you got the password wrong": the
     // user did nothing wrong and reloading will not help them (#1466).
-    expired ? "Your session has expired. Please sign in again." : null
+    // Localised like every sibling string on this screen — this is the one
+    // message shown to a user who did nothing wrong, so it should not be the
+    // only English text on an otherwise translated form.
+    expired
+      ? t("DASHBOARD_LOGIN_SESSION_EXPIRED", "Your session has expired. Please sign in again.")
+      : null
   );
 
   const tenantId = getTenantId();
--- a/digit-ui-esbuild/products/dashboard/src/services/analyticsService.js
+++ b/digit-ui-esbuild/products/dashboard/src/services/analyticsService.js
@@ -1,4 +1,13 @@
 import { withTraceHeaders, recordApiCall } from "./dashboardMetrics";
+import {
+  authFetch,
+  buildRequestInfo,
+  getTenantId,
+  toRequestError,
+} from "./authService";
+
+// Re-exported for existing importers; authService is the definition.
+export { getTenantId };
 
 /** Browser-facing analytics API base (relative path or absolute URL). */
 export function getAnalyticsBase() {
@@ -10,70 +19,15 @@ export function getAnalyticsBase() {
 
 const ANALYTICS_BASE = getAnalyticsBase();
 
-function parseJson(raw) {
-  try {
-    return JSON.parse(raw);
-  } catch {
-    return raw;
-  }
-}
-
-function getEmployeeToken() {
-  const raw = window.localStorage?.getItem("Employee.token");
-  return raw && raw !== "undefined" ? parseJson(raw) : null;
-}
-
-function getEmployeeInfo() {
-  const raw = window.localStorage?.getItem("Employee.user-info");
-  return raw && raw !== "undefined" ? parseJson(raw) : null;
-}
-
-export function getTenantId() {
-  return (
-    window.globalConfigs?.getConfig("STATE_LEVEL_TENANT_ID") ||
-    process.env.REACT_APP_STATE_LEVEL_TENANT_ID ||
-    "ke"
-  );
-}
-
-export function hasAuth() {
-  return Boolean(getEmployeeToken());
-}
-
-/**
- * Stable identity of the signed-in employee (user uuid), or null when no
- * session exists. Used to scope per-user client state (e.g. the dashboard
- * layout storage key) — NOT for authorization, which stays server-side.
- */
-export function getUserUuid() {
-  const info = getEmployeeInfo();
-  return info && typeof info === "object" ? info.uuid || null : null;
-}
-
-function buildRequestInfo() {
-  const authToken = getEmployeeToken();
-  const userInfo = getEmployeeInfo();
-  return {
-    apiId: "Rainmaker",
-    ver: ".01",
-    ts: Date.now(),
-    action: "_search",
-    msgId: `dashboard-${Date.now()}`,
-    ...(authToken && { authToken }),
-    ...(userInfo && { userInfo }),
-  };
-}
-
 async function postAnalytics(path, body) {
   const startedAt = typeof performance !== "undefined" ? performance.now() : Date.now();
-  const response = await fetch(`${ANALYTICS_BASE}${path}`, {
-    method: "POST",
-    // Per-load W3C traceparent + x-trace-id (dashboardMetrics) — Kong's otel
-    // plugin and the pgr javaagent continue the browser's trace id end-to-end.
-    headers: withTraceHeaders({ "Content-Type": "application/json" }),
-    credentials: "omit",
-    body: JSON.stringify({
-      RequestInfo: buildRequestInfo(),
+  // The analytics API is the dashboard's primary data path, so it is the one
+  // surface allowed to conclude the session is dead (sessionCritical defaults
+  // to true). Trace headers are preserved through the wrapper.
+  const response = await authFetch(`${ANALYTICS_BASE}${path}`, {
+    headers: withTraceHeaders({}),
+    buildBody: () => ({
+      RequestInfo: buildRequestInfo("dashboard"),
       ...body,
     }),
   });
@@ -81,14 +35,7 @@ async function postAnalytics(path, body) {
   if (!response.ok) {
     const endedAt = typeof performance !== "undefined" ? performance.now() : Date.now();
     recordApiCall(`${ANALYTICS_BASE}${path}`, endedAt - startedAt, 0, false);
-    const error = new Error(`Analytics request failed (${response.status})`);
-    error.status = response.status;
-    try {
-      error.payload = await response.json();
-    } catch {
-      error.payload = await response.text();
-    }
-    throw error;
+    throw await toRequestError(response, "Analytics request");
   }
 
   return response.json();
--- a/digit-ui-esbuild/products/dashboard/src/services/analyticsService.js
+++ b/digit-ui-esbuild/products/dashboard/src/services/analyticsService.js
@@ -24,13 +24,24 @@ async function postAnalytics(path, body) {
   // The analytics API is the dashboard's primary data path, so it is the one
   // surface allowed to conclude the session is dead (sessionCritical defaults
   // to true). Trace headers are preserved through the wrapper.
-  const response = await authFetch(`${ANALYTICS_BASE}${path}`, {
-    headers: withTraceHeaders({}),
-    buildBody: () => ({
-      RequestInfo: buildRequestInfo("dashboard"),
-      ...body,
-    }),
-  });
+  // authFetch THROWS on every terminal 401 verdict, which would skip the
+  // recordApiCall below — so during an auth incident the metrics would show zero
+  // failed analytics calls while users stared at error banners, hiding exactly
+  // the outage this module is meant to surface. Record, then rethrow.
+  let response;
+  try {
+    response = await authFetch(`${ANALYTICS_BASE}${path}`, {
+      headers: withTraceHeaders({}),
+      buildBody: () => ({
+        RequestInfo: buildRequestInfo("dashboard"),
+        ...body,
+      }),
+    });
+  } catch (error) {
+    const endedAt = typeof performance !== "undefined" ? performance.now() : Date.now();
+    recordApiCall(`${ANALYTICS_BASE}${path}`, endedAt - startedAt, 0, false);
+    throw error;
+  }
 
   if (!response.ok) {
     const endedAt = typeof performance !== "undefined" ? performance.now() : Date.now();
--- a/digit-ui-esbuild/products/dashboard/src/services/authService.js
+++ b/digit-ui-esbuild/products/dashboard/src/services/authService.js
@@ -1,0 +1,546 @@
+import { getTenantId } from "../config/dashboardConfig";
+
+/**
+ * Single source of truth for dashboard auth: token storage, RequestInfo
+ * construction, and the 401 -> refresh -> retry cycle.
+ *
+ * Before this module, getEmployeeToken / getEmployeeInfo / buildRequestInfo were
+ * duplicated verbatim in analyticsService, boundaryService and
+ * complaintHierarchyService, and no caller inspected response.status. A 401
+ * therefore surfaced as the raw string "Analytics request failed (401)" in a
+ * banner, and the auth gate (evaluated once at mount) never re-ran (#1466).
+ *
+ * Behaviour verified against egov-user on bomet, not assumed:
+ *   - grant_type=refresh_token is supported and returns a new access_token.
+ *   - The refresh token is NOT rotated: the same value comes back.
+ *   - Refreshing IMMEDIATELY invalidates the previous access token — a request
+ *     still in flight with the old token gets 401. This is why refresh must be
+ *     single-flight and why we retry the original request afterwards, rather
+ *     than letting each 401 refresh independently.
+ *   - A dead refresh token returns HTTP 400 invalid_grant (not 401), so the
+ *     refresh call's own failure must never be fed back into this retry path.
+ *   - Tokens are opaque UUIDs, not JWTs, so expiry cannot be inspected client
+ *     side. Refresh is necessarily reactive (on 401), never pre-emptive.
+ *
+ * ---------------------------------------------------------------------------
+ * CONTRACT — what authFetch() does with a 401, and what callers should expect.
+ *
+ * A 401 has four distinct causes and they are NOT interchangeable. Collapsing
+ * them is what made the first cut of this module dangerous: it treated every
+ * one as "session dead" and wiped storage shared with the co-hosted digit-ui
+ * app. Each outcome now throws a differently-flagged error:
+ *
+ *   another caller already refreshed  -> replay with the new token, no error
+ *   token dead, refresh succeeded     -> replay, no error
+ *   raced a refresh on every attempt  -> err.refreshRaced, session KEPT
+ *   auth server unreachable           -> err.refreshUnavailable, session KEPT
+ *   fresh token but endpoint 401s     -> err.endpointUnauthorized, session KEPT
+ *   non-critical call, token refused  -> err.endpointUnauthorized, session KEPT
+ *   CRITICAL call, token refused      -> err.sessionExpired, storage CLEARED
+ *
+ * Only the last row dispatches SESSION_EXPIRED_EVENT and drops the user to the
+ * login gate. Two rules govern it:
+ *
+ *   1. Never destroy a session we have not PROVEN is dead — the keys are shared
+ *      with the co-hosted digit-ui app, so a wrong guess signs the user out of
+ *      the whole product.
+ *   2. Only a SESSION-CRITICAL call may form that verdict. Pass
+ *      `sessionCritical: false` for auxiliary data (boundary geometry,
+ *      hierarchy labels): those endpoints can 401 for their own reasons (an ACL
+ *      gap, a gateway answering 401 instead of 403) and must not be able to tear
+ *      the dashboard down to a login screen that falsely blames expiry.
+ *
+ * When a critical call IS refused, storage is cleared even if we held no refresh
+ * token: the server has proven the token dead, and leaving it behind is what
+ * made a reload re-pass the auth gate and re-render a broken dashboard (#1466's
+ * "reloading did not help" symptom).
+ *
+ * Guarantees: at most three sends and at most one refresh per call (cannot
+ * loop); concurrent 401s share a single refresh; a refresh never re-enters this
+ * path; the refresh token is owner-stamped so a previous user's token on the
+ * same browser is never used.
+ *
+ * Caller error contract:
+ *   analyticsService        — critical; propagates (the dashboard shows it)
+ *   boundaryService         — non-critical; partial data, throws only if NOTHING
+ *                             loaded so the map can show its error state
+ *   complaintHierarchyService — non-critical; returns null (labels fall back)
+ * ---------------------------------------------------------------------------
+ */
+
+/**
+ * Every localStorage key this module owns, in one table.
+ *
+ * `shared: true` marks keys the co-hosted digit-ui app also reads — index.js
+ * copies them into Digit.SessionStorage at boot to seed the employee session.
+ * Writing them affects the whole product, not just the dashboard, which is why
+ * clearing is gated (see announceSessionExpired).
+ *
+ * The previous code kept "keys login writes" and "keys signout clears" as two
+ * hand-maintained lists and they had already drifted apart once — the refresh
+ * token was written but never cleared. Deriving both from this table makes that
+ * class of bug structurally impossible.
+ */
+const KEYS = {
+  token: { key: "Employee.token", shared: true },
+  refreshToken: { key: "Employee.refresh-token", shared: false },
+  userInfo: { key: "Employee.user-info", shared: true },
+  tenantId: { key: "Employee.tenant-id", shared: true },
+  // Unprefixed aliases the main UI's boot path reads for login detection.
+  tokenAlias: { key: "token", shared: true },
+  userInfoAlias: { key: "user-info", shared: true },
+};
+
+const TOKEN_KEY = KEYS.token.key;
+const REFRESH_KEY = KEYS.refreshToken.key;
+const USER_INFO_KEY = KEYS.userInfo.key;
+const TENANT_KEY = KEYS.tenantId.key;
+
+/** Every key this module writes — so signout can never fall behind login. */
+const SESSION_KEYS = Object.values(KEYS).map((k) => k.key);
+
+export const SESSION_EXPIRED_EVENT = "dashboard:session-expired";
+
+function parseJson(raw) {
+  try {
+    return JSON.parse(raw);
+  } catch {
+    return raw;
+  }
+}
+
+function readStorage(key) {
+  try {
+    const raw = window.localStorage?.getItem(key);
+    return raw && raw !== "undefined" && raw !== "null" ? parseJson(raw) : null;
+  } catch {
+    return null;
+  }
+}
+
+function writeStorage(key, value) {
+  try {
+    window.localStorage?.setItem(key, JSON.stringify(value));
+  } catch {
+    /* private mode / quota — non-fatal, the session just won't survive reload */
+  }
+}
+
+export function getEmployeeToken() {
+  return readStorage(TOKEN_KEY);
+}
+
+/**
+ * The refresh token, but only if it belongs to the user currently signed in.
+ * A token stamped for a different uuid is a leftover from a previous session on
+ * this browser; using it would swap the whole product to that user's identity.
+ */
+export function getRefreshToken() {
+  const stored = readStorage(REFRESH_KEY);
+  if (!stored) return null;
+  // Legacy shape: a bare string written before owner-stamping existed. It has no
+  // provable owner, so it is not trusted.
+  if (typeof stored === "string") return null;
+  const currentUuid = getEmployeeInfo()?.uuid ?? null;
+  if (stored.userUuid && currentUuid && stored.userUuid !== currentUuid) return null;
+  return stored.token || null;
+}
+
+export function getEmployeeInfo() {
+  return readStorage(USER_INFO_KEY);
+}
+
+// Re-exported, NOT re-implemented. A second copy here had a different fallback
+// ("ke" vs dashboardConfig's "default"); DashboardLogin uses dashboardConfig's,
+// so on a deployment with neither config set, login would authenticate as
+// "default" while refresh posted "ke" — egov-user rejects that, which this
+// module classifies as a positive rejection and answers by clearing storage
+// shared with the co-hosted app. One definition, one fallback.
+export { getTenantId };
+
+/**
+ * True when a token is PRESENT. Deliberately not a validity check — tokens are
+ * opaque, so validity is only knowable by calling the API. Session death is
+ * detected reactively via SESSION_EXPIRED_EVENT.
+ */
+export function hasAuth() {
+  return Boolean(getEmployeeToken());
+}
+
+/** Order roles so the first non-EMPLOYEE role leads (drives the scoping badge). */
+export function withSignificantRoleFirst(userInfo) {
+  if (!userInfo || !Array.isArray(userInfo.roles)) return userInfo;
+  const roles = [...userInfo.roles].sort(
+    (a, b) => (a.code === "EMPLOYEE" ? 1 : 0) - (b.code === "EMPLOYEE" ? 1 : 0)
+  );
+  return { ...userInfo, roles };
+}
+
+/**
+ * The dashboard is NOT a standalone app: App.js routes /employee/dashboard to
+ * AdminDashboard inside the same SPA that renders DigitUI, and index.js copies
+ * Employee.token into Digit.SessionStorage's "User" at boot. That cache is what
+ * the rest of the employee UI authenticates from, and it is never re-read.
+ *
+ * Since refreshing invalidates the previous access token server-side, a silent
+ * refresh would otherwise kill the token the surrounding app is still holding —
+ * every main-UI call would 401 until a full reload. Push the new token into
+ * that cache so both surfaces stay on the same session.
+ */
+function syncDigitSession(accessToken, userInfo) {
+  const session = window.Digit?.SessionStorage;
+  if (!session?.set || !session?.get) return; // standalone build — nothing to sync
+  try {
+    const existing = session.get("User") || {};
+    session.set("User", {
+      ...existing,
+      token: accessToken,
+      access_token: accessToken,
+      ...(userInfo != null && { info: userInfo }),
+    });
+  } catch {
+    /* best-effort — localStorage remains the source of truth */
+  }
+}
+
+export function persistSession({ accessToken, refreshToken, userInfo, tenantId }) {
+  // Normalise here rather than at the call site so a refresh cannot silently
+  // revert the role order that login established.
+  const normalisedUser = userInfo != null ? withSignificantRoleFirst(userInfo) : null;
+
+  if (accessToken != null) {
+    writeStorage(TOKEN_KEY, accessToken);
+    writeStorage("token", accessToken);
+  }
+  if (refreshToken != null) {
+    // Stored WITH its owner. Only this dashboard writes the refresh token, and
+    // the main digit-ui login neither knows nor clears it — so without an owner
+    // stamp, user A's refresh token can survive user B signing in on the same
+    // browser and silently re-authenticate the whole product as A.
+    writeStorage(REFRESH_KEY, {
+      token: refreshToken,
+      userUuid: normalisedUser?.uuid ?? getEmployeeInfo()?.uuid ?? null,
+    });
+  }
+  if (normalisedUser != null) {
+    writeStorage(USER_INFO_KEY, normalisedUser);
+    writeStorage("user-info", normalisedUser);
+  }
+  if (tenantId != null) writeStorage(TENANT_KEY, tenantId);
+
+  if (accessToken != null) syncDigitSession(accessToken, normalisedUser);
+}
+
+export function clearSession() {
+  SESSION_KEYS.forEach((k) => {
+    try {
+      window.localStorage?.removeItem(k);
+    } catch {
+      /* ignore */
+    }
+  });
+  // Symmetric with syncDigitSession. Clearing only localStorage would leave the
+  // co-hosted UI authenticating from a cached dead token — and index.js boots
+  // from SessionStorage FIRST, skipping the localStorage re-seed when "User"
+  // exists, so even a reload would re-enter the employee UI "logged in" on a
+  // token that no longer works.
+  const session = window.Digit?.SessionStorage;
+  if (session?.set) {
+    try {
+      session.set("User", null);
+    } catch {
+      /* ignore */
+    }
+  }
+}
+
+/**
+ * Show the login gate.
+ *
+ * `clearStorage` is deliberately opt-in. Employee.token / user-info / token are
+ * SHARED with the co-hosted digit-ui app, so wiping them signs the user out of
+ * the whole product — appropriate when the server has positively rejected our
+ * refresh token, but not when we merely lack one. Notably the only writer of
+ * Employee.refresh-token is this dashboard's own login form; a session seeded by
+ * the main digit-ui login has none, and that is the common path. Clearing
+ * unconditionally would turn one dashboard 401 into a full-product logout.
+ */
+function announceSessionExpired({ clearStorage } = { clearStorage: false }) {
+  if (clearStorage) clearSession();
+  try {
+    window.dispatchEvent(new CustomEvent(SESSION_EXPIRED_EVENT));
+  } catch {
+    /* CustomEvent unavailable — the auth gate still re-checks on next render */
+  }
+}
+
+export function buildRequestInfo(msgIdPrefix = "dashboard") {
+  const authToken = getEmployeeToken();
+  const userInfo = getEmployeeInfo();
+  return {
+    apiId: "Rainmaker",
+    ver: ".01",
+    ts: Date.now(),
+    action: "_search",
+    msgId: `${msgIdPrefix}-${Date.now()}`,
+    ...(authToken && { authToken }),
+    ...(userInfo && { userInfo }),
+  };
+}
+
+/**
+ * OAuth client credentials header. globalConfigs stores JWT_TOKEN as the bare
+ * base64 of "egov-user-client:" WITHOUT the scheme, so the "Basic " prefix is
+ * added here — matching what DashboardLogin sends on the password grant.
+ */
+export function getOAuthBasic() {
+  const configured = window.globalConfigs?.getConfig?.("JWT_TOKEN");
+  const raw = configured || "ZWdvdi11c2VyLWNsaWVudDo=";
+  return /^Basic\s/i.test(raw) ? raw : `Basic ${raw}`;
+}
+
+/**
+ * In-flight refresh, shared by every caller that 401s in the same window.
+ * Without this, N concurrent batch queries would fire N refreshes, and since
+ * each refresh invalidates the prior access token they would invalidate each
+ * other's tokens in turn.
+ */
+let refreshInFlight = null;
+
+/**
+ * Bumped on every successful refresh. Single-flight alone is NOT sufficient:
+ * it only merges 401s that overlap the refresh window. A request whose 401
+ * lands just AFTER a refresh settled would see refreshInFlight === null and
+ * start a SECOND refresh, which immediately invalidates the token the first
+ * caller is already retrying with — turning a recovered session into a false
+ * logout. The generation lets a caller tell "my token is stale because someone
+ * else already refreshed" (just retry) apart from "the current token is dead"
+ * (refresh).
+ */
+let tokenGeneration = 0;
+
+/** Refresh must not hang the whole dashboard: every 401'd caller waits on it. */
+const REFRESH_TIMEOUT_MS = 15000;
+
+/**
+ * Refresh outcomes are three-valued, not boolean. Collapsing them loses the
+ * distinction that decides whether the shared session may be destroyed:
+ *   REFRESHED   — new token in hand, replay the request.
+ *   REJECTED    — the server positively refused the refresh token (400
+ *                 invalid_grant). The session really is dead; clearing is right.
+ *   NO_TOKEN    — we never had a refresh token. The 401 is authoritative, but
+ *                 the keys belong to the co-hosted app; flip the gate, keep them.
+ *   UNAVAILABLE — the refresh call itself never got an answer (offline, DNS,
+ *                 gateway reset, timeout). We cannot conclude anything, so the
+ *                 session must survive. Waking a laptop from sleep hits this.
+ */
+export const REFRESH_REFRESHED = "refreshed";
+export const REFRESH_REJECTED = "rejected";
+export const REFRESH_NO_TOKEN = "no_token";
+export const REFRESH_UNAVAILABLE = "unavailable";
+
+async function requestRefresh() {
+  const refreshToken = getRefreshToken();
+  if (!refreshToken) return REFRESH_NO_TOKEN;
+
+  const body = new URLSearchParams({
+    grant_type: "refresh_token",
+    refresh_token: refreshToken,
+    tenantId: getTenantId(),
+  });
+
+  // Without this, a stalled /user/oauth/token leaves every waiting caller
+  // pending forever and the dashboard sits on a permanent loading state.
+  const controller = typeof AbortController !== "undefined" ? new AbortController() : null;
+  const timer = controller ? setTimeout(() => controller.abort(), REFRESH_TIMEOUT_MS) : null;
+
+  let res;
+  try {
+    res = await fetch("/user/oauth/token", {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/x-www-form-urlencoded",
+        Authorization: getOAuthBasic(),
+      },
+      body,
+      ...(controller && { signal: controller.signal }),
+    });
+  } catch {
+    // Network failure or the 15s abort — NOT a verdict on the refresh token.
+    // Treating this as rejection would delete a still-valid session over a
+    // transient blip.
+    return REFRESH_UNAVAILABLE;
+  } finally {
+    if (timer) clearTimeout(timer);
+  }
+
+  // Only a client-error answer is the server REFUSING the token (a dead one
+  // gives 400 invalid_grant). A 5xx, a 429, or a gateway error page is the auth
+  // server failing — not a verdict on our credentials — and must not be allowed
+  // to wipe storage shared with the co-hosted digit-ui app. Getting this wrong
+  // would sign every user out of the product on a deploy blip.
+  if (!res.ok) {
+    const refused = res.status === 400 || res.status === 401 || res.status === 403;
+    return refused ? REFRESH_REJECTED : REFRESH_UNAVAILABLE;
+  }
+
+  let data;
+  try {
+    data = await res.json();
+  } catch {
+    return REFRESH_UNAVAILABLE;
+  }
+  // A 2xx with no token is a malformed/intercepted response (proxy login page,
+  // truncated body), not a rejection. Same reasoning as above: do not destroy.
+  if (!data?.access_token) return REFRESH_UNAVAILABLE;
+
+  persistSession({
+    accessToken: data.access_token,
+    // Not rotated today, but honour a rotated value if the server ever sends one.
+    refreshToken: data.refresh_token || refreshToken,
+    userInfo: data.UserRequest || undefined,
+  });
+  tokenGeneration += 1;
+  return REFRESH_REFRESHED;
+}
+
+/** Single-flight wrapper around requestRefresh. Resolves to a REFRESH_* value. */
+function refreshSession() {
+  if (!refreshInFlight) {
+    refreshInFlight = requestRefresh()
+      .catch(() => REFRESH_UNAVAILABLE)
+      .finally(() => {
+        refreshInFlight = null;
+      });
+  }
+  return refreshInFlight;
+}
+
+/**
+ * fetch() for authenticated dashboard APIs.
+ *
+ * `buildBody` is a function, not a value: on retry the RequestInfo must be
+ * rebuilt so the request carries the NEW token. Passing a pre-serialised body
+ * would replay the dead one and 401 again.
+ *
+ * On a 401 it either refreshes (shared across callers) or, when another caller
+ * has already refreshed since this request was sent, simply replays with the
+ * newer token. At most two sends happen, so this cannot loop. If refresh fails,
+ * or the replay 401s again, the session is announced dead and the error is
+ * thrown with `status` and `sessionExpired` set for callers.
+ */
+export async function authFetch(
+  url,
+  { buildBody, method = "POST", headers, sessionCritical = true } = {}
+) {
+  const send = () =>
+    fetch(url, {
+      method,
+      headers: { "Content-Type": "application/json", ...headers },
+      credentials: "omit",
+      body: buildBody ? JSON.stringify(buildBody()) : undefined,
+    });
+
+  // Three sends max, still at most ONE refresh per call. Three because there are
+  // two independent reasons to replay and a caller can need both: a
+  // generation-mismatch replay (someone else refreshed mid-flight) and the
+  // guaranteed retry after this caller's OWN refresh. With a cap of two, a
+  // mismatch replay could consume the budget and the loop would exit having
+  // never sent the token it just minted — reporting failure on a session it had
+  // successfully repaired. The refresh itself stays capped, so this cannot loop.
+  let didRefresh = false;
+  let outcome = null;
+
+  for (let attempt = 0; attempt < 3; attempt += 1) {
+    const generationAtSend = tokenGeneration;
+    const response = await send();
+    if (response.status !== 401) return response;
+
+    // If the generation has already moved, another caller refreshed while this
+    // request was in flight — our token was merely stale, so replay with the
+    // new one instead of refreshing again (a second refresh would invalidate
+    // the token that caller is using).
+    if (tokenGeneration !== generationAtSend) continue;
+
+    if (didRefresh) break;
+    didRefresh = true;
+    outcome = await refreshSession();
+    if (outcome !== REFRESH_REFRESHED) break;
+  }
+
+  // Could not reach the auth server: say nothing about the session. Leaving it
+  // intact lets the next attempt succeed once connectivity returns, instead of
+  // signing the user out of the whole product over a dropped packet.
+  if (outcome === REFRESH_UNAVAILABLE) {
+    const error = new Error("Could not reach the sign-in service. Please retry.");
+    error.status = 401;
+    error.refreshUnavailable = true;
+    throw error;
+  }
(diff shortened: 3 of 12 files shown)


KDwevedi:
Two independent review passes (Codex + Fable) found real defects. All fixed; pushed as commits 2 and 3.

### Both reviewers independently found the same HIGH bug

**Single-flight did not prevent the cascade it was written for.** It only merges 401s that *overlap* the refresh window. Given the verified invalidate-the-predecessor semantics: A and B both send with `T1`; A 401s, refreshes to `T2`, retries; `refreshInFlight` clears; B's slower 401 sees no refresh in flight and starts a **second** refresh to `T3`, invalidating the `T2` that A is mid-retry with. A's retry 401s and A tears down a live session — the exact false logout this PR exists to prevent.

My original harness could not have caught it: all its concurrent calls 401'd *before* the refresh settled, which is precisely the case single-flight already handled.

Fixed with a `tokenGeneration` counter captured before each send, so a caller can tell "my token is stale because someone else refreshed" (replay) from "the current token is dead" (refresh). Also capped at one refresh per call — refreshing after a replay still 401s would mint a token only to invalidate one another caller may be using.

### The finding that changed the design

**The dashboard is not standalone.** `App.js:67` routes `/employee/dashboard` to `AdminDashboard` inside the same SPA that renders `DigitUI`, and `index.js:60-72` copies `Employee.token` into `Digit.SessionStorage`'s `User` at boot — the cache the rest of the employee UI authenticates from, never re-read. The module shared its token *and* its localStorage keys with the surrounding app and did not know it.

Two consequences, both fixed:

- **Refresh killed the main app's session.** Refreshing invalidates the predecessor, so the first refresh on the dashboard route left every other digit-ui call 401ing until a full reload. `persistSession` now syncs `Digit.SessionStorage` (no-op in the standalone build).
- **A dashboard 401 could sign the user out of the whole product.** The only writer of `Employee.refresh-token` is this dashboard's own login form — a session seeded by the main digit-ui login has none, and per `DashboardLogin.jsx:19` that is the *normal* path. So the common case was: no refresh token → refresh "fails" → `clearSession()` wipes `Employee.token` / `token` / `user-info` → logged out everywhere, where pre-PR it was a banner on one page. Storage is now cleared **only** on a positive rejection from the server.

### Also fixed

- **A network blip destroyed a recoverable session** — fetch rejections were mapped to the same `false` as `400 invalid_grant`, so a laptop waking from sleep (the most common expiry scenario) deleted a still-valid refresh token. Outcomes are now three-valued; `UNAVAILABLE` concludes nothing, throws a retryable error, fires no expiry event.
- **A stalled `/user/oauth/token` hung the dashboard forever** — every 401'd caller awaits the shared promise. Now aborts after 15s.
- **Refresh reverted the role order** login establishes; `persistSession` applies `withSignificantRoleFirst` itself and the helper is no longer duplicated.

### On simplicity

Both reviewers independently reached the same verdict: the global `fetch` interceptor is wrong here (the token lives *inside* the JSON body, so it would have to parse and mutate arbitrary bodies — and it would intercept the co-hosted app's traffic), per-service retry re-duplicates what this PR deleted and cannot share single-flight state, and react-query does not apply since these are hand-rolled promises.

The one genuinely simpler option is dropping refresh entirely (401 → login), which removes every HIGH finding at once. That was considered and consciously rejected in favour of keeping silent renew and fixing the co-hosting properly.

### Verification

Left unverified as noted before: a real browser session against a live tenant. Everything below is a harness against a mocked fetch.

```
late-401 race    A(T1)401 REFRESH#1->T2 A(T2)200 B(T1)401 B(T2)200  -> 1 refresh, 0 expiry events
5 concurrent 401s                                                   -> 1 refresh, 5 successes
no refresh token         gate flips, shared keys SURVIVE
refresh rejected (400)   gate flips, shared keys cleared
network blip             no expiry event, refresh token survives
refresh ok               Digit.SessionStorage.User.token = T2, roles non-EMPLOYEE-first
dead refresh token       1 attempt then expire
API 401s on fresh token  exactly 1 refresh (bounded, no loop)
```

KDwevedi:
## Reviewer notes

Out of draft. Two commits landed since the previous comment, and the PR description has been rewritten to describe the final state rather than the original design — **read the description, not the first commit.**

### Suggested read order

The diff is 6 files but only one carries risk.

1. **`services/authService.js`** — the whole design. Start with the header block: it states the 401 contract in a table. Everything else follows from it.
2. **`AdminDashboard.jsx`** (~20 lines) — the auth gate, now event-driven instead of mount-once.
3. The three services + `DashboardLogin` — **mechanical**. They delete duplicated helpers and swap `fetch` → `authFetch`. No logic worth arguing about.

### What to actually scrutinise

- **The `tokenGeneration` guard** (`authFetch`). This is the subtle part. It exists because single-flight alone is *not* enough given that refreshing invalidates the previous token server-side: a 401 arriving just after a refresh settles would otherwise start a second refresh and invalidate the token the first caller is mid-retry with. If you break one thing, break this — a staggered-401 test is the way in.
- **The clear-vs-keep decision** in `announceSessionExpired`. Clearing signs the user out of the **entire employee app**, not just the dashboard. Only a positive server rejection should do it. This is the judgement call most worth a second opinion.
- **`syncDigitSession`.** Correct only if `index.js`'s boot-time copy into `Digit.SessionStorage` really is the main app's auth source and really is never re-read. I believe both; happy to be wrong.

### Deliberate decisions, so they don't read as oversights

- **Refresh is reactive, never pre-emptive.** Tokens are opaque UUIDs — there is no expiry to inspect client-side. Not fixable without a server change.
- **`buildBody` is a function, not a value.** The replay must rebuild `RequestInfo` with the new token; a pre-serialised body would replay the dead one. This also rules out a generic `fetch` interceptor, since the token lives *inside* the JSON body.
- **The two swallowing services stay swallowing.** Boundary geometry and hierarchy labels are best-effort; they still get the gate flip because the expiry event fires before the throw.
- **Refresh was kept rather than dropped.** Dropping it (401 → login) is genuinely simpler and removes every HIGH finding — it was considered and declined in favour of fixing the co-hosting. Reopening that is a legitimate review outcome.

### How to exercise it

There is no PR-time frontend build here (`pgr-ui-build.yml` is `workflow_dispatch`-only), so nothing compiles this automatically.

Manual smoke test, which is also **the one thing no harness can prove**:

1. Sign in to the employee UI normally, open `/employee/dashboard`.
2. Invalidate the access token server-side (sign in again elsewhere, or refresh the token out of band).
3. Dashboard should silently renew and keep rendering — no banner, no login screen.
4. **Then navigate back to `/employee/...`** and confirm the main UI still works. That is the co-hosting fix; steps 1–3 passing while step 4 fails is the exact regression this PR is guarding against.
5. Separately: clear `Employee.refresh-token`, force a 401, and confirm you land on the dashboard login **without** being signed out of the rest of the app.

### Known gaps

- **No browser verification.** Everything below is a harness against a mocked `fetch`.
- **Multi-tab is not handled.** `refreshInFlight` and `tokenGeneration` are per-tab, so two dashboard tabs can still invalidate each other. The blast radius is bounded — the loser refreshes and recovers, rather than being logged out — but it is not coordinated. Cross-tab coordination (BroadcastChannel or a storage-event listener) is a follow-up, not in scope here.
- **`@claude review` has never run on this PR** — the fork restriction; the trigger comment was not honoured.

### Risk / rollback

Confined to `src/dashboard/`. The one change that reaches outside it is `syncDigitSession` writing `Digit.SessionStorage`, which is a no-op wherever `window.Digit` is absent. Reverting the whole PR restores the previous behaviour exactly — the raw `(401)` banner — since no server, schema, or config change is involved.

frontend/micro-ui/web/src/dashboard/services/authService.js
```diff
@@ -0,0 +1,464 @@
+/**
```
vinothrallapalli-eGov, line 3: **[CONFIRMED / correctness] Fix applied only to the legacy UI copy — the deployed esbuild dashboard is unfixed**

The entire auth-module fix lands only in `frontend/micro-ui` (which the repo docs describe as the superseded legacy UI), while the byte-for-byte-parallel dashboard vendored in `digit-ui-esbuild/products/dashboard` — the UI actually built by Ansible (`local-setup/ansible/files/digit-ui-build.sh`) and served in local-setup deploys — still has the mount-time-only auth gate (`AdminDashboard.jsx:148`: `useState(() => embedded || hasDashboardSession())`) and the duplicated `getEmployeeToken`/`buildRequestInfo` helpers in all three service files.

**Failure scenario:** An employee on a deployed environment whose session dies mid-tab still sees the raw "Analytics request failed (401)" banner, the gate never flips to the login screen, and reload re-passes the gate on the stale token — every user-visible symptom of #1466 persists in the UI users actually run.

frontend/micro-ui/web/src/dashboard/services/authService.js
```diff
@@ -0,0 +1,464 @@
...
+    // Treating this as rejection would delete a still-valid session over a
+    // transient blip.
+    return REFRESH_UNAVAILABLE;
+  } finally {
+    if (timer) clearTimeout(timer);
+  }
+
+  // Only a client-error answer is the server REFUSING the token (a dead one
+  // gives 400 invalid_grant). A 5xx, a 429, or a gateway error page is the auth
+  // server failing — not a verdict on our credentials — and must not be allowed
```
vinothrallapalli-eGov, line 440: **[CONFIRMED / correctness] Any non-2xx refresh response is treated as a positive rejection, wiping shared storage**

`requestRefresh` treats ANY non-2xx (502/503/504/429 from the gateway) as `REFRESH_REJECTED`, so a transient auth-server outage positively "proves" the session dead and `authFetch` clears the localStorage keys shared with the co-hosted digit-ui app.

**Failure scenario:** egov-user pod restarts while an employee works in the main digit-ui; a dashboard KPI call 401s, the silent refresh hits Kong and gets a 502; `if (!res.ok) return REFRESH_REJECTED` classifies this as a rejected refresh token, `authFetch` runs `announceSessionExpired({clearStorage:true})` which wipes `Employee.token`/`user-info`/`token` — the employee is signed out of the entire product over a dropped-service blip. This violates the module's own stated rule that only a positive 400 `invalid_grant` justifies destroying shared storage; a 5xx should map to `REFRESH_UNAVAILABLE`.

frontend/micro-ui/web/src/dashboard/services/authService.js
```diff
@@ -0,0 +1,464 @@
...
+    error.status = 401;
+    error.refreshUnavailable = true;
+    throw error;
+  }
+
+  // We refreshed successfully and the replay STILL 401'd. The token is provably
+  // fresh, so this endpoint is rejecting us for its own reason (an auth gap, a
+  // misrouted service, a role check answering 401 instead of 403) — not because
+  // the session died. Reporting expiry here would sign the user out on every
+  // load and re-login would change nothing: a loop. Surface it as an ordinary
```
vinothrallapalli-eGov: **[CONFIRMED / correctness] Endpoint-level 401 tears down the whole dashboard when no refresh token exists (the common co-hosted path)**

The old code guaranteed a dashboard-API 401 only produced a widget-level error banner (session and dashboard UI untouched). That is removed: in the common co-hosted path the session is seeded by the main digit-ui login and `Employee.refresh-token` does not exist (only the dashboard's own login form writes it), so `requestRefresh` returns `NO_TOKEN` and any single endpoint 401 dispatches `SESSION_EXPIRED_EVENT`, unmounting the entire dashboard to a login screen claiming "Your session has expired". The `endpointUnauthorized` escape hatch that distinguishes an endpoint auth gap from session death is unreachable without a refresh token.

**Failure scenario:** Employee logs in via the main employee portal (no refresh token written), opens `/employee/dashboard`; one backing endpoint (e.g. boundary-service behind a role/ACL gap, or a gateway answering 401 instead of 403) returns 401 while the token is valid for everything else. Before: a warn/banner on that one widget. After: the whole dashboard is torn down and replaced by a login form falsely telling the user their session expired, while the surrounding employee UI in the same tab still works.

frontend/micro-ui/web/src/dashboard/services/authService.js
```diff
@@ -0,0 +1,464 @@
...
+  // generation-mismatch replay (someone else refreshed mid-flight) and the
+  // guaranteed retry after this caller's OWN refresh. With a cap of two, a
+  // mismatch replay could consume the budget and the loop would exit having
+  // never sent the token it just minted — reporting failure on a session it had
+  // successfully repaired. The refresh itself stays capped, so this cannot loop.
+  let didRefresh = false;
+  let outcome = null;
+
+  for (let attempt = 0; attempt < 3; attempt += 1) {
+    const generationAtSend = tokenGeneration;
```
vinothrallapalli-eGov, line 516: **[CONFIRMED / correctness] A refresh completed on the final loop iteration is never replayed**

When the first 401 exits via the generation-moved `continue` and the replay 401s with generation unchanged, `refreshSession()` runs on the last loop iteration and a `REFRESHED` outcome exits the loop with no attempt left — the request is never replayed with the newly minted token and the caller gets a thrown error instead. (Same root cause interacts with lines 388 and 397.)

**Failure scenario:** Attempt 0 401s because another caller refreshed mid-flight (`continue`); attempt 1 replays and 401s with generation unchanged; the loop refreshes successfully (bumping generation and invalidating the token every concurrent caller holds) but the loop ends — `authFetch` throws `endpointUnauthorized` ("Request failed (401) with a valid session") without ever sending the request with the token it just minted, surfacing an error banner even though one more replay would have succeeded.

frontend/micro-ui/web/src/dashboard/services/authService.js
```diff
@@ -0,0 +1,464 @@
...
+    throw error;
+  }
+
+  // We refreshed successfully and the replay STILL 401'd. The token is provably
+  // fresh, so this endpoint is rejecting us for its own reason (an auth gap, a
+  // misrouted service, a role check answering 401 instead of 403) — not because
+  // the session died. Reporting expiry here would sign the user out on every
+  // load and re-login would change nothing: a loop. Surface it as an ordinary
+  // request failure and leave the session alone.
+  if (outcome === REFRESH_REFRESHED) {
```
vinothrallapalli-eGov, line 567: **[CONFIRMED / correctness] Dead token kept on NO_TOKEN — reload re-passes the auth gate (the #1466 reload symptom remains)**

In the `REFRESH_NO_TOKEN` case (the common main-UI-seeded session), the dead `Employee.token` is deliberately kept, so `hasDashboardSession()` still returns true and a page reload re-passes the auth gate on the dead token.

**Failure scenario:** An employee whose session was seeded by the main digit-ui login (no refresh token exists) hits token expiry: the gate flips with "Your session has expired", but on reload (the natural reaction, and what `AdminDashboard`'s own `onLogin` handler does) `hasAuth()` finds the stale token, the dashboard mounts, every KPI/boundary call 401s and renders error states before the NO_TOKEN expiry event flips the gate again — a flash-of-broken-dashboard loop on every reload instead of landing directly on the login gate.

frontend/micro-ui/web/src/dashboard/services/authService.js
```diff
@@ -0,0 +1,464 @@
...
+ *                 gateway reset, timeout). We cannot conclude anything, so the
+ *                 session must survive. Waking a laptop from sleep hits this.
+ */
+export const REFRESH_REFRESHED = "refreshed";
+export const REFRESH_REJECTED = "rejected";
+export const REFRESH_NO_TOKEN = "no_token";
+export const REFRESH_UNAVAILABLE = "unavailable";
+
+async function requestRefresh() {
+  const refreshToken = getRefreshToken();
```
vinothrallapalli-eGov, line 404: **[PLAUSIBLE / correctness] Stale refresh token can silently swap the session to a previous user**

`requestRefresh` blindly uses whatever `Employee.refresh-token` is in localStorage, but that key is written only by the dashboard's own login form and is never cleared or overwritten when the co-hosted main digit-ui login establishes a new session (the main UI does not know the key exists) — a stale refresh token belonging to a previous user can silently re-authenticate the whole product as that user.

**Failure scenario:** User A signs in via the dashboard login form (A's refresh token persisted). Later, user B signs in through the main digit-ui login on the same browser without a full logout (main-UI login overwrites `Employee.token`/`user-info` but leaves `Employee.refresh-token` untouched). When B's access token expires, a dashboard 401 triggers `requestRefresh` with A's refresh token; `persistSession` then writes A's `access_token` and A's `UserRequest` into `Employee.token`/`user-info` AND syncs them into `Digit.SessionStorage` — B's entire employee session silently becomes user A: wrong KPI scoping, and every subsequent action across the co-hosted app runs under A's identity and roles.

vinothrallapalli-eGov (commented):
10 findings below as inline comments — the two most severe are the fix landing only in the legacy `frontend/micro-ui` dashboard copy (the deployed `digit-ui-esbuild` copy is untouched), and `requestRefresh` treating any non-2xx as a positive rejection that wipes localStorage shared with the co-hosted employee UI.

frontend/micro-ui/web/src/dashboard/services/authService.js
```diff
@@ -0,0 +1,464 @@
...
+    });
+
+  // Three sends max, still at most ONE refresh per call. Three because there are
+  // two independent reasons to replay and a caller can need both: a
+  // generation-mismatch replay (someone else refreshed mid-flight) and the
+  // guaranteed retry after this caller's OWN refresh. With a cap of two, a
+  // mismatch replay could consume the budget and the loop would exit having
+  // never sent the token it just minted — reporting failure on a session it had
+  // successfully repaired. The refresh itself stays capped, so this cannot loop.
+  let didRefresh = false;
```
vinothrallapalli-eGov, line 512: **[PLAUSIBLE / correctness] Two stale-generation `continue`s exhaust the loop and falsely announce session expiry**

If both loop iterations exit via the generation-moved `continue` (a concurrent refresh lands during each send), the loop exhausts with `outcome === null` and control falls through to the sessionExpired branch, falsely announcing session expiry for a session that was just successfully refreshed.

**Failure scenario:** Request A is sent with token T0; another caller's refresh completes mid-flight so A gets 401 and `tokenGeneration !== generationAtSend` → `continue`; A replays with T1, a second refresh completes during that replay so A 401s again with generation moved → `continue` exits the 2-iteration loop; `outcome` is still `null` so it skips the UNAVAILABLE and REFRESHED branches and executes `announceSessionExpired` + throws "Your session has expired" — the user is dropped to the login gate even though the session is alive and freshly refreshed.

frontend/micro-ui/web/src/dashboard/services/boundaryService.js
```diff
@@ -53,21 +27,28 @@ export async function fetchBoundariesByCodes(codes = []) {
...
+        continue;
+      }
 
-    if (!response.ok) {
-      console.warn(`boundary/_search failed (${response.status})`);
-      continue;
+      const payload = await response.json();
+      const boundaries = payload?.Boundary || [];
+      all.push(...boundaries);
+    } catch (error) {
```
vinothrallapalli-eGov, line 51: **[CONFIRMED / correctness] Boundary fetch errors are now swallowed — the map's error state can never fire**

`fetchBoundariesByCodes` no longer propagates fetch/parse failures: the old code let a rejected fetch or failed `response.json()` reject the whole call, which `GeographyChoroplethMap.jsx` caught to set `boundariesError` ("Could not load ward boundaries"); the new per-chunk try/catch swallows every error and resolves with a partial or empty array, so the map's only error indicator can never fire for these failures.

**Failure scenario:** Boundary-service is briefly unreachable while an employee opens the geography widget: before, the map showed "Could not load ward boundaries"; after, every chunk throws, is swallowed with only a `console.warn`, the call resolves to `[]`, and the user sees a silently blank choropleth with no error message.

frontend/micro-ui/web/src/dashboard/services/authService.js
```diff
@@ -0,0 +1,464 @@
...
+
+export function getRefreshToken() {
+  return readStorage(REFRESH_KEY);
+}
+
+export function getEmployeeInfo() {
+  return readStorage(USER_INFO_KEY);
+}
+
+export function getTenantId() {
```
vinothrallapalli-eGov: **[CONFIRMED / cleanup] Duplicate `getTenantId` with a divergent fallback (`"ke"` here vs `"default"` in dashboardConfig)**

`getTenantId` is re-implemented here, duplicating the existing `dashboard/config/dashboardConfig.js` `getTenantId` with a divergent fallback (`"ke"` here vs `"default"` there, used by `DashboardLogin`'s password grant).

**Failure scenario:** On a deployment where neither the `STATE_LEVEL_TENANT_ID` globalConfig nor `REACT_APP_STATE_LEVEL_TENANT_ID` is set, login creates the session with tenantId `default` but the refresh grant posts tenantId `ke` — egov-user can reject the refresh, which `authFetch` classifies as `REFRESH_REJECTED` and answers by clearing the shared storage, signing the user out of the whole product. Import the existing `dashboardConfig.getTenantId` (or move it wholesale) instead of keeping a second copy — the duplication also means any future tenant-resolution change must be made twice.

frontend/micro-ui/web/src/dashboard/services/authService.js
```diff
@@ -0,0 +1,464 @@
...
+  if (normalisedUser != null) {
+    writeStorage(USER_INFO_KEY, normalisedUser);
+    writeStorage("user-info", normalisedUser);
+  }
+  if (tenantId != null) writeStorage(TENANT_KEY, tenantId);
+
+  if (accessToken != null) syncDigitSession(accessToken, normalisedUser);
+}
+
+export function clearSession() {
```
vinothrallapalli-eGov, line 282: **[CONFIRMED / cleanup] `clearSession` leaves the `Digit.SessionStorage` "User" cache on the dead token**

`clearSession` wipes the localStorage keys but never clears or invalidates the `Digit.SessionStorage` "User" cache that `persistSession` takes care to sync, leaving the co-hosted UI on the dead token.

**Failure scenario:** After a positive refresh rejection (`REFRESH_REJECTED`), `clearSession` removes `Employee.token`/`user-info`/`token` from localStorage, but `window.Digit.SessionStorage` "User" still holds the invalidated access token. The surrounding employee UI keeps authenticating from that cache, so every main-UI call 401s with no recovery; worse, `src/index.js:48` boots from SessionStorage first and skips the localStorage re-seed when "User" exists, so even a full reload in the same tab re-enters the employee UI as "logged in" with the dead token while the dashboard shows the login gate. The sync should be symmetric: `clearSession` should clear/patch the SessionStorage "User" entry the same way `syncDigitSession` writes it.

KDwevedi:
@vinothrallapalli-eGov thank you — this was a genuinely valuable review, and the first finding was the important one. All seven live findings are addressed; two were already fixed in the commit the review was anchored to.

### 1. Wrong copy — confirmed, and fixed

You're right, and it was the most consequential thing here. I verified `digit-ui-esbuild/products/dashboard` still had the mount-time-only gate (`AdminDashboard.jsx:148`), `getEmployeeToken`/`buildRequestInfo` duplicated across all three services, and **zero** occurrences of `401`. `local-setup/ansible/files/digit-ui-build.sh` builds `digit-ui-esbuild`, so every symptom of #1466 persisted in the UI users actually run. My original exploration concluded the esbuild copy "wasn't it" — that was simply wrong.

Ported, and it is a port rather than a copy: the esbuild dashboard has diverged and is **ahead** (localisation, number-format masks, trace headers, `embedded` mode), so the module was carried over and each call site adapted. That copy has **seven** authenticated call sites vs micro-ui's five — a second `ComplaintHierarchy` search and the boundary-relationships call. Trace headers are preserved through a new `headers` option.

Two deliberate calls in the port, worth your eye:
- **`embedded` mode**: the gate is event-driven except when embedded, where the DigitUI chrome owns the session and its own expiry handling — rendering our standalone login form inside the host would be wrong. `authService` has already cleared the dead token, so the host sees the same state on its next call.
- **`dashboardMetrics.js` keeps its raw `fetch`**: it's a fire-and-forget telemetry beacon to `/otel` with its own backoff, and a 401 there must never be able to expire a session.

### 2. Non-2xx → rejection, and 3. the never-replayed refresh

Both already fixed in `0bdea65ed`, the commit your review is anchored to — I think the analysis was drafted against the prior state. `requestRefresh` now returns `REJECTED` only for 400/401/403 (a 5xx/429 is `UNAVAILABLE`), and the send cap is 3, not 2, precisely so a refresh completed on the last iteration still gets replayed.

### 4. Endpoint 401 tearing down the dashboard — fixed, and you identified the right principle

`authFetch` now takes `sessionCritical` (default `true`). Boundary geometry and hierarchy labels pass `false`, so an ACL gap or a gateway answering 401 instead of 403 on an auxiliary endpoint can no longer end the session. Only the analytics API — the dashboard's primary data path — may form that verdict.

### 5. Dead token kept on NO_TOKEN — fixed, and your framing changed my answer

I'd kept it deliberately to avoid signing the user out of the co-hosted app, but you're right that it reintroduces the original "reload didn't help" symptom. Now that only *critical* calls reach that branch, a refusal there is **proof** the token is dead — so clearing is justified, and the reload lands on the login screen.

### 6. Stale refresh token → user swap — fixed

Serious catch. The refresh token is now **owner-stamped** with the uuid it was issued for and ignored if it doesn't match the current user. The legacy bare-string shape has no provable owner and is likewise not trusted.

### 7. `getTenantId` divergence — fixed

Removed the duplicate rather than aligning it; it is re-exported from `dashboardConfig`, so login and refresh cannot disagree again.

### 8. `clearSession` leaving the SessionStorage cache — fixed

Now symmetric with `syncDigitSession`. Your point about `index.js` booting from SessionStorage first is what makes this matter: without it, a reload re-entered the employee UI "logged in" on a dead token.

### 9. Boundary errors swallowed — fixed

Fair, and it was a regression I introduced answering the *previous* round. Per-chunk tolerance is kept (partial geometry is usable) but it rethrows when **nothing** loaded, so the map's error state can fire again.

### 10. Generation-mismatch exhaustion — fixed

Distinct `refreshRaced` error rather than falling through to the expiry branch on a healthy session.

---

**Also caught while doing this:** the branch was based on pre-#1511 master, so it would have **reverted the search removal** on merge. Rebased; the diff no longer touches `dashboardSearch.js` or the search CSS.

**Verification.** The scattered ad-hoc harnesses are consolidated into one 13-check suite — single-flight, the late-401 generation guard, 5xx/429 vs 400 classification, network blips, critical vs auxiliary 401s, dead-token clearing, foreign refresh tokens, endpoint-401 with a fresh token, tenant resolution. It passes against **both** copies from the same source, so the two now behave identically.

Still not covered: a real browser session. There's no PR-time frontend build here, so the co-hosting behaviour — that after a silent refresh the surrounding employee UI still works — needs a smoke test on a deployed environment before merge.

**Known straggler, not addressed:** `i18n/localeRuntime.js:89` still reads `Employee.token` directly rather than going through `authService`. Happy to fold it in if you'd prefer it consistent now.

vinothrallapalli-eGov (commented):
**Resolution status of the 10 prior findings:**

| # | Prior finding | Status |
|---|---|---|
| 1 | Fix only in legacy micro-ui copy; esbuild copy unfixed | ✅ Fixed — esbuild dashboard now updated |
| 2 | Any non-2xx refresh treated as REFRESH_REJECTED, wipes shared storage | ✅ Fixed |
| 3 | Endpoint 401 with no refresh token tears down the dashboard | ⚠️ **Partially fixed** — still tears down, and now also wipes shared storage (see inline) |
| 4 | Refresh on final loop iteration never replayed | ⚠️ **Partially fixed** — loop is now 3 attempts but the same flaw persists (see inline) |
| 5 | Dead token kept on NO_TOKEN; reload re-passes auth gate | ✅ Fixed |
| 6 | Stale refresh token can swap session identity | ✅ Mostly fixed — owner stamp added; a fail-open residual remains (see inline) |
| 7 | Double stale-generation continue falsely expires session | ✅ Fixed |
| 8 | Boundary fetch errors swallowed | ✅ Fixed |
| 9 | Duplicate getTenantId with divergent fallback | ✅ Fixed — but the unification changed the effective fallback 'ke'→'default' (see inline) |
| 10 | clearSession leaves Digit.SessionStorage 'User' | ✅ Fixed — but the unconditional clear introduces a new citizen-session risk (see inline) |

10 findings below as inline comments — 2 residuals from the prior review plus new issues introduced by the fixes themselves.

digit-ui-esbuild/products/dashboard/src/services/authService.js
```diff
@@ -0,0 +1,546 @@
...
+    error.endpointUnauthorized = true;
+    throw error;
+  }
+
+  // The primary API refused this token, so it IS dead — regardless of whether we
+  // held a refresh token. Clearing is therefore proven-correct, and it is what
+  // stops a reload from re-passing the auth gate on the same dead token and
+  // re-rendering a broken dashboard before flipping back to login (#1466's
+  // original "reload didn't help" symptom).
+  announceSessionExpired({ clearStorage: true });
```
vinothrallapalli-eGov: **[CONFIRMED / correctness] Prior finding #3 only partially fixed: an endpoint 401 with no refresh token still destroys the session — and now also wipes shared storage**

A session-critical (analytics) 401 with `REFRESH_NO_TOKEN` still calls `announceSessionExpired({clearStorage:true})`, wiping `Employee.token`/`user-info` + the unprefixed aliases AND nulling `Digit.SessionStorage` "User" on the strength of a single endpoint 401, with no auth-server verdict on the token (identical code in the frontend/micro-ui copy).

**Failure scenario:** An employee signs in through the main digit-ui (such sessions never have a dashboard refresh token) and opens the dashboard. If pgr-analytics answers 401 for its own reason (role/ACL gap, gateway answering 401 instead of 403, misrouted service) while the egov-user session is still perfectly valid, `authFetch` gets NO_TOKEN, clears all shared localStorage keys and the Digit session cache, and the user is signed out of the ENTIRE co-hosted employee product — in embedded mode with no visible explanation, since embedded `AdminDashboard` deliberately does not listen for `SESSION_EXPIRED_EVENT`. The module's own rule 1 ("never destroy a session we have not PROVEN is dead") is violated: with no refresh token there was no server verdict, only one endpoint's 401, and re-login then re-visiting the dashboard reproduces the logout in a loop.

digit-ui-esbuild/products/dashboard/src/services/authService.js
```diff
@@ -0,0 +1,546 @@
...
+  if (normalisedUser != null) {
+    writeStorage(USER_INFO_KEY, normalisedUser);
+    writeStorage("user-info", normalisedUser);
+  }
+  if (tenantId != null) writeStorage(TENANT_KEY, tenantId);
+
+  if (accessToken != null) syncDigitSession(accessToken, normalisedUser);
+}
+
+export function clearSession() {
```
vinothrallapalli-eGov, line 282: **[CONFIRMED / correctness] `clearSession` can destroy an active citizen session via the unprefixed aliases**

`clearSession` — now invoked automatically on any critical 401 verdict rather than only on explicit sign-out — removes the unprefixed `token`/`user-info` aliases and nulls the single `Digit.SessionStorage` "User" slot without checking whose session they hold.

**Failure scenario:** On a shared browser a citizen is logged into the co-hosted digit-ui (localStorage `token` === Citizen.token, SessionStorage "User" holds the citizen session) while a dead `Employee.token` lingers from earlier employee use. The standalone dashboard gate (`hasAuth`) passes on the dead Employee.token, analytics 401s, `REFRESH_NO_TOKEN` + sessionCritical triggers `clearSession`, which deletes `token`, `user-info` and sets SessionStorage "User" to null — the logged-in citizen is silently signed out of the citizen portal and loses any in-progress complaint draft, from a 401 on a dashboard they never opened as that user.

digit-ui-esbuild/products/dashboard/src/services/authService.js
```diff
@@ -0,0 +1,546 @@
...
+    error.refreshRaced = true;
+    throw error;
+  }
+
+  // We refreshed successfully and the replay STILL 401'd. The token is provably
+  // fresh, so this endpoint is rejecting us for its own reason (an auth gap, a
+  // misrouted service, a role check answering 401 instead of 403) — not because
+  // the session died. Reporting expiry here would sign the user out on every
+  // load and re-login would change nothing: a loop.
+  if (outcome === REFRESH_REFRESHED) {
```
vinothrallapalli-eGov, line 567: **[CONFIRMED / correctness] Prior finding #4 only partially fixed: a refresh succeeding on the final (3rd) attempt is still never replayed**

In the 3-attempt `authFetch` loop, an own refresh that succeeds on the FINAL iteration (attempt 2, after two generation-mismatch continues on attempts 0 and 1) is never replayed — the loop exits with `outcome === REFRESH_REFRESHED` and throws `err.endpointUnauthorized` ("Request failed (401) with a valid session") even though the freshly minted token was never sent (same code in the frontend/micro-ui copy).

**Failure scenario:** Under concurrent widget loads where other callers' refreshes land during this request's first two sends (each refresh immediately invalidates the in-flight token, per the module's own documented server behaviour), the third send 401s with a matching generation, `refreshSession()` succeeds, and the loop terminates without the guaranteed post-refresh replay. For the session-critical analytics path the dashboard then shows the error banner instead of rendering data, falsely blaming the endpoint, when one more send with the new token would have succeeded.

digit-ui-esbuild/products/dashboard/src/services/authService.js
```diff
@@ -0,0 +1,546 @@
...
+ * this browser; using it would swap the whole product to that user's identity.
+ */
+export function getRefreshToken() {
+  const stored = readStorage(REFRESH_KEY);
+  if (!stored) return null;
+  // Legacy shape: a bare string written before owner-stamping existed. It has no
+  // provable owner, so it is not trusted.
+  if (typeof stored === "string") return null;
+  const currentUuid = getEmployeeInfo()?.uuid ?? null;
+  if (stored.userUuid && currentUuid && stored.userUuid !== currentUuid) return null;
```
vinothrallapalli-eGov: **[PLAUSIBLE / correctness] Prior finding #6 residual: the refresh-token owner check fails open when either uuid is null**

`getRefreshToken`'s owner check `if (stored.userUuid && currentUuid && ...)` trusts the stored refresh token whenever EITHER uuid is null, so a stamped token belonging to user A is used when `Employee.user-info` is missing/cleared, or when the stamp itself was written as null (`persistSession` falls back to `getEmployeeInfo()?.uuid ?? null` when a refresh response carries no `UserRequest`).

**Failure scenario:** User A uses the dashboard login (refresh token stamped A). Later `Employee.user-info` is absent while `Employee.token` exists (partial clear by another surface, storage eviction, or a main-UI flow that clears user-info without the dashboard-owned refresh token). The next dashboard 401 finds `currentUuid === null`, the guard passes, A's refresh token is posted, and `persistSession` writes A's access token and A's `UserRequest` into shared storage and `Digit.SessionStorage` — the whole co-hosted product silently swaps to user A's identity, the exact identity-swap the owner stamp was added to prevent. Consider failing closed (treat a missing uuid on either side as a mismatch).

digit-ui-esbuild/products/dashboard/src/services/analyticsService.js
```diff
@@ -10,85 +19,23 @@ export function getAnalyticsBase() {
...
-    // Per-load W3C traceparent + x-trace-id (dashboardMetrics) — Kong's otel
-    // plugin and the pgr javaagent continue the browser's trace id end-to-end.
-    headers: withTraceHeaders({ "Content-Type": "application/json" }),
-    credentials: "omit",
-    body: JSON.stringify({
-      RequestInfo: buildRequestInfo(),
+  // The analytics API is the dashboard's primary data path, so it is the one
+  // surface allowed to conclude the session is dead (sessionCritical defaults
+  // to true). Trace headers are preserved through the wrapper.
+  const response = await authFetch(`${ANALYTICS_BASE}${path}`, {
```
vinothrallapalli-eGov: **[CONFIRMED / correctness] `postAnalytics` no longer records failed API calls for 401 outcomes**

On master a 401 response reached the `!response.ok` branch and called `recordApiCall(url, duration, 0, false)`, but now `authFetch` THROWS for all terminal 401 verdicts (sessionExpired / endpointUnauthorized / refreshRaced / refreshUnavailable) before that line can run, so the failure is never recorded in dashboardMetrics.

**Failure scenario:** During a session-expiry or auth-outage incident every failing analytics request bypasses `recordApiCall`, so the dashboard's API-call metrics show zero failed analytics calls — operators inspecting the metrics see a healthy API while users see error banners, hiding exactly the incidents this PR is about (esbuild copy only; the legacy copy never had `recordApiCall`). Consider a try/catch around the `authFetch` call that records the failure before rethrowing.

digit-ui-esbuild/products/dashboard/src/services/analyticsService.js
```diff
@@ -1,4 +1,13 @@
 import { withTraceHeaders, recordApiCall } from "./dashboardMetrics";
+import {
+  authFetch,
+  buildRequestInfo,
+  getTenantId,
+  toRequestError,
+} from "./authService";
+
+// Re-exported for existing importers; authService is the definition.
+export { getTenantId };
```
vinothrallapalli-eGov, line 10: **[PLAUSIBLE / correctness] The getTenantId unification silently changes the effective tenant fallback from `"ke"` to `"default"`**

The deleted analyticsService `getTenantId` carried a `"ke"` fallback that every dashboard data call (analytics, boundary, hierarchy RequestInfo/tenantId) previously resolved to when neither the `STATE_LEVEL_TENANT_ID` globalConfig nor `REACT_APP_STATE_LEVEL_TENANT_ID` is set; the unification (the fix for prior finding #9 — the right direction!) re-points all of them at dashboardConfig's `"default"` fallback.

**Failure scenario:** On an existing deployment (e.g. a Kenya/bomet-style env) that never set `STATE_LEVEL_TENANT_ID` and relied on the hardcoded `"ke"` fallback, every `/_query`, `/packs`, boundary and MDMS call now posts `tenantId="default"`. The analytics service has no data for tenant `default`, so all KPIs render zero/empty and the choropleth loads no boundaries — a fully blank dashboard after upgrade, with no error surfaced because the requests succeed with empty result sets. Worth calling out in the PR description / release notes at minimum.

digit-ui-esbuild/products/dashboard/src/services/authService.js
```diff
@@ -0,0 +1,546 @@
...
+
+/** Shared non-401 error shape so every service reports failures identically. */
+export async function toRequestError(response, label) {
+  const error = new Error(`${label} failed (${response.status})`);
+  error.status = response.status;
+  try {
+    error.payload = await response.json();
+  } catch {
+    try {
+      error.payload = await response.text();
```
vinothrallapalli-eGov: **[CONFIRMED / cleanup] `toRequestError`'s `text()` fallback can never succeed — the body stream is already consumed**

`response.json()` failing to parse has already consumed the body stream, so the subsequent `response.text()` throws "body already read" and `payload` is always null for non-JSON error bodies.

**Failure scenario:** Analytics returns a 502 with an HTML gateway error page; `error.payload` ends up null instead of the page text, so the on-screen banner and console logs lose the only diagnostic content. Simpler form that works: read `await response.text()` once, then `try { JSON.parse(text) } catch { text }`. Duplicated in the frontend/micro-ui copy.

frontend/micro-ui/web/src/dashboard/services/authService.js
```diff
@@ -0,0 +1,546 @@
+import { getTenantId } from "../config/dashboardConfig";
```
vinothrallapalli-eGov, line 1: **[CONFIRMED / cleanup] A byte-identical 546-line security-sensitive auth module is now maintained in two trees**

One of the two trees is the one CLAUDE.md calls superseded ("frontend/micro-ui — legacy… Appears superseded; only referenced in a stale Dockerfile comment"); the module's own header decries exactly this hand-maintained-duplicate drift pattern.

**Failure scenario:** The next auth fix lands in only one copy (this is literally prior finding #1 of this same PR) and the trees drift on session-teardown behavior — a bug class the PR's KEYS-table comment says it exists to make "structurally impossible", yet reintroduces at file granularity. Cheaper: keep the module only in the deployed digit-ui-esbuild tree, or extract one shared source both builds vendor.

digit-ui-esbuild/products/dashboard/src/services/authService.js
```diff
@@ -0,0 +1,546 @@
...
+ *
+ * `clearStorage` is deliberately opt-in. Employee.token / user-info / token are
+ * SHARED with the co-hosted digit-ui app, so wiping them signs the user out of
+ * the whole product — appropriate when the server has positively rejected our
+ * refresh token, but not when we merely lack one. Notably the only writer of
+ * Employee.refresh-token is this dashboard's own login form; a session seeded by
+ * the main digit-ui login has none, and that is the common path. Clearing
+ * unconditionally would turn one dashboard 401 into a full-product logout.
+ */
+function announceSessionExpired({ clearStorage } = { clearStorage: false }) {
```
vinothrallapalli-eGov, line 329: **[CONFIRMED / cleanup] `announceSessionExpired`'s doc comment contradicts its actual use**

The comment says clearing shared storage is "appropriate when the server has positively rejected our refresh token, but not when we merely lack one", yet `authFetch`'s only call site passes `clearStorage:true` for both `REJECTED` and `NO_TOKEN` outcomes (line 525).

**Failure scenario:** A maintainer reading the function doc believes NO_TOKEN sessions are preserved and debugs the co-hosted digit-ui logout-on-analytics-401 symptom in the wrong place; or "fixes" the call site to match the comment, silently reintroducing the dead-token-survives-reload bug from the prior review. Same contradiction in the frontend/micro-ui copy. Whichever behavior is intended, make the comment and the call site agree.

digit-ui-esbuild/products/dashboard/src/components/DashboardLogin.jsx
```diff
@@ -87,12 +68,16 @@ const inputStyle = {
...
+const DashboardLogin = ({ onLogin, expired = false }) => {
   const { t } = useDashboardT();
   const [username, setUsername] = useState("");
   const [password, setPassword] = useState("");
   const [submitting, setSubmitting] = useState(false);
-  const [error, setError] = useState(null);
+  const [error, setError] = useState(
+    // Distinguish "your session ended" from "you got the password wrong": the
+    // user did nothing wrong and reloading will not help them (#1466).
+    expired ? "Your session has expired. Please sign in again." : null
```
vinothrallapalli-eGov: **[CONFIRMED / cleanup] Hardcoded English session-expired message bypasses `t()`**

"Your session has expired. Please sign in again." bypasses the `t()` localization wrapper that every other user-facing string in this file uses (e.g. `DASHBOARD_LOGIN_SIGN_IN_FAILED`), and duplicates the same English literal baked into authService's sessionExpired error.

**Failure scenario:** On a non-English locale (the dashboard ships `useDashboardT` and localized keys for all sibling strings), the session-expiry explanation — the one message shown to a user who did nothing wrong — renders in English while the rest of the login screen is localized; the string also exists in two places that can drift.

KDwevedi:
@vinothrallapalli-eGov — you were right that three of my fixes only moved the problem. All eight actionable findings are addressed; one is deliberately not changed and flagged below.

### The three residuals — you were right on each

**#3 NO_TOKEN still destroying the session, now with storage.** This one I got backwards last round. I reasoned that a critical-path refusal was *proof* the token was dead, so clearing was justified. It isn't proof: with no refresh token there is no auth-server verdict at all — just one endpoint's 401 — and your embedded-mode point makes it worse, since there is no visible explanation there. Storage is now destroyed only on a positive `REJECTED`. The reload symptom that made me clear in the first place is handled instead by marking the refused token dead in **sessionStorage**: tab-scoped, keeps the gate closed across a reload, touches nobody else's session.

**#4 the replay.** Raising the cap 2→3 moved the defect to the final iteration rather than fixing it — exactly as you say. The replay now happens **immediately after our own refresh** instead of relying on a remaining attempt, so no budget can starve it.

**#6 owner check failing open.** Now fails closed: a missing uuid on either side is treated as a mismatch. You're right that the null case is precisely when the risk is highest.

### New findings

**`clearSession` killing a citizen session** — a genuinely nasty one I would not have found. The aliases and the single `Digit.SessionStorage` slot are now dropped only when they actually hold the employee token being torn down.

**`toRequestError` body already read** — correct, the fallback could never fire. Reads once, then attempts `JSON.parse`.

**`recordApiCall` skipped on 401** — correct and self-defeating, since it hides exactly the incidents this PR is about. Wrapped so the failure is recorded before rethrowing.

**Localised session-expired string** — fixed, `DASHBOARD_LOGIN_SESSION_EXPIRED` with an English default. Only in the esbuild copy; the micro-ui copy has no i18n wiring at all.

**Doc comment vs call site** — now agree.

### Not changed — needs a decision, not a patch

**The `"ke"` → `"default"` tenant fallback.** You're right this is a real behaviour change for any deployment that set no `STATE_LEVEL_TENANT_ID`. I've left the unification because the two were already inconsistent — `DashboardLogin` resolved `"default"` while data calls resolved `"ke"`, so a standalone-login deployment relying on `"ke"` could never have logged in. Embedded deployments are the exposed case. Reverting to `"ke"` just moves the inconsistency back, so this wants a call on the intended default rather than a code tweak. Flagged for release notes; happy to pin either value if you have a preference.

### On the duplicated module

Fair, and it is the same bug class as your finding #1. The plan is to delete the micro-ui dashboard copy outright in a separate PR rather than maintain two — it is the tree CLAUDE.md already calls superseded. Doing it here would bury a large deletion inside an auth change, and its blast radius (what else still imports from that tree) hasn't been mapped. Keeping both copies byte-identical in the interim is the weaker option, and I'd rather it be explicit than implied.

### Verification

Suite now **21 checks, passing against both copies from one source** — including the new cases: NO_TOKEN keeps shared storage but closes the gate on reload; `clearSession` spares a foreign citizen alias while still clearing our own; a refresh on the final attempt is replayed; the owner check fails closed on a missing uuid; a non-JSON error body is captured.

One check from the previous round had to be inverted rather than kept — it asserted the storage-clearing behaviour your #3 argues against. Worth naming, since it means the earlier "13 passed" included a check encoding the wrong intent.

Still not covered: a real browser session, which is where the co-hosting and embedded-mode behaviour actually needs proving.

digit-ui-esbuild/products/dashboard/src/services/authService.js
```diff
@@ -0,0 +1,622 @@
...
+}
+
+export function persistSession({ accessToken, refreshToken, userInfo, tenantId }) {
+  // Normalise here rather than at the call site so a refresh cannot silently
+  // revert the role order that login established.
+  const normalisedUser = userInfo != null ? withSignificantRoleFirst(userInfo) : null;
+
+  if (accessToken != null) {
+    writeStorage(TOKEN_KEY, accessToken);
+    writeStorage("token", accessToken);
```
vinothrallapalli-eGov, line 253: **[CONFIRMED / correctness] `persistSession` unconditionally stomps the shared unprefixed aliases — the write-side counterpart of the clearSession bug just fixed**

`persistSession` (invoked by the silent background refresh) overwrites the shared unprefixed `token` and `user-info` aliases with the employee identity, with no ownership check — while `clearSession` now carefully checks `ownsAliases` before touching the same keys.

**Failure scenario:** A citizen is actively signed into the co-hosted digit-ui portal (the unprefixed aliases hold the citizen session — the exact shared-browser scenario `clearSession`'s own comment describes). A stale dashboard tab in the background gets a 401, silently refreshes, and `requestRefresh` → `persistSession` executes `writeStorage("token", accessToken)` / `writeStorage("user-info", normalisedUser)`, clobbering the citizen's aliases with the employee token and user-info. The main UI's boot path reads these aliases for login detection, so on next reload the citizen is silently switched to (or broken by) the employee identity — an identity swap performed without any user action. The same `ownsAliases`-style check used in `clearSession` is needed on the write side. Same code in the frontend/micro-ui copy.

digit-ui-esbuild/products/dashboard/src/components/DashboardLogin.jsx
```diff
@@ -47,35 +52,11 @@ const DEMO_USERS = [
...
-      } catch {
-        /* ignore */
-      }
-    }
-  );
-}
+// Session storage lives in authService (single source of truth); these thin
+// re-exports keep the existing AdminDashboard import sites unchanged.
+export const hasDashboardSession = hasAuth;
+export const clearDashboardSession = clearSession;
```
vinothrallapalli-eGov, line 59: **[CONFIRMED / correctness] Explicit sign-out lost its unconditional-clear guarantee**

`clearDashboardSession` is now aliased to the ownership-gated `clearSession`, so the deleted code's invariant — "clicking Sign out always removes the token/user-info aliases" — only holds when the unprefixed `token` alias strictly equals `Employee.token`, and the `Digit.SessionStorage` "User" cache is likewise only nulled when `ownsAliases` is true.

**Failure scenario:** Employee on a shared kiosk clicks Sign out (`AdminDashboard handleSignOut` → `clearDashboardSession` → `clearSession` → `window.location.reload`). If the unprefixed `token` alias no longer strictly equals `Employee.token` (the co-hosted digit-ui or an earlier login rewrote one of the two keys, or the alias was stored unstringified so `parseJson` yields a different value), `ownsAliases` is false: `token` and `user-info` stay in localStorage and SessionStorage "User" keeps the employee token. After the reload, index.js seeds the session from those leftovers and the next kiosk user lands in the employee UI still authenticated as the employee who just signed out. Explicit sign-out should clear the Employee.* keys unconditionally and use a stronger rule for the shared aliases (e.g. clear them whenever they hold the employee's token OR are unparseable).

digit-ui-esbuild/products/dashboard/src/services/authService.js
```diff
@@ -0,0 +1,622 @@
...
+      /* ignore */
+    }
+  });
+  // Symmetric with syncDigitSession. Clearing only localStorage would leave the
+  // co-hosted UI authenticating from a cached dead token — and index.js boots
+  // from SessionStorage FIRST, skipping the localStorage re-seed when "User"
+  // exists, so even a reload would re-enter the employee UI "logged in" on a
+  // token that no longer works.
+  const session = window.Digit?.SessionStorage;
+  if (session?.set && ownsAliases) {
```
vinothrallapalli-eGov, line 306: **[CONFIRMED / correctness] The `ownsAliases` gate leaves a proven-dead employee token cached in `Digit.SessionStorage`**

`clearSession` gates the SessionStorage "User" teardown on `ownsAliases` (alias-key ownership) instead of on the cached token itself, so a dead employee token stays in the cache whenever the unprefixed `token` alias holds a different session's value or is absent.

**Failure scenario:** Employee is signed into the dashboard (index.js seeded SessionStorage "User" with the employee token); a citizen later logs into the co-hosted portal on the same browser, overwriting the unprefixed `token` alias. The employee's refresh token is then REJECTED (400 invalid_grant) → `clearSession` runs with `ownsAliases=false` → the `if (session?.set && ownsAliases)` guard skips nulling "User" even though the cached token IS the dead employee token. index.js boots from SessionStorage FIRST and skips the localStorage re-seed when "User" exists, so even after reload the co-hosted employee UI re-enters "logged in" on the dead token and every main-UI call 401s until the user manually clears storage — exactly the failure mode the surrounding comment says this block exists to prevent. The guard should compare the CACHED token against the employee token being cleared, not rely on alias ownership.

digit-ui-esbuild/products/dashboard/src/services/authService.js
```diff
@@ -0,0 +1,622 @@
...
+  return readStorage(USER_INFO_KEY);
+}
+
+// Re-exported, NOT re-implemented. A second copy here had a different fallback
+// ("ke" vs dashboardConfig's "default"); DashboardLogin uses dashboardConfig's,
+// so on a deployment with neither config set, login would authenticate as
+// "default" while refresh posted "ke" — egov-user rejects that, which this
+// module classifies as a positive rejection and answers by clearing storage
+// shared with the co-hosted app. One definition, one fallback.
+export { getTenantId };
```
vinothrallapalli-eGov, line 170: **[CONFIRMED / correctness] Prior finding not fixed: the tenant fallback change `"ke"` → `"default"` stands unmitigated**

The unconfigured-deployment tenant fallback for analytics consumers remains `"default"` (via the dashboardConfig re-export) where the pre-PR analyticsService fell back to `"ke"`. The change is documented as deliberate, but the behavior change ships without mitigation.

**Failure scenario:** On a deployment where neither the `STATE_LEVEL_TENANT_ID` globalConfig nor `REACT_APP_STATE_LEVEL_TENANT_ID` is set and data is provisioned under tenant `ke` (the old fallback), every `/_query`, `/packs`, `/catalog/_search` and MDMS/boundary call posts `tenantId="default"` after this PR ships — all KPI tiles, the hierarchy, and the map return empty results and the dashboard renders blank with no error, on an environment that worked before the upgrade. At minimum this needs a release-note callout and ideally a startup warning when the fallback is used.

digit-ui-esbuild/products/dashboard/src/components/DashboardLogin.jsx
```diff
@@ -87,12 +68,21 @@ const inputStyle = {
...
   const [submitting, setSubmitting] = useState(false);
-  const [error, setError] = useState(null);
+  const [error, setError] = useState(
+    // Distinguish "your session ended" from "you got the password wrong": the
+    // user did nothing wrong and reloading will not help them (#1466).
+    // Localised like every sibling string on this screen — this is the one
+    // message shown to a user who did nothing wrong, so it should not be the
+    // only English text on an otherwise translated form.
+    expired
+      ? t("DASHBOARD_LOGIN_SESSION_EXPIRED", "Your session has expired. Please sign in again.")
```
vinothrallapalli-eGov, line 83: **[CONFIRMED / cleanup] `DASHBOARD_LOGIN_SESSION_EXPIRED` is unseeded — the raw key renders on session expiry**

`localeRuntime.translate(key, seedEnglish)` documents "seedEnglish is NEVER rendered" and returns the key when unseeded; `digit-mcp/src/tools/dashboard-l10n-seed.ts` contains 0 occurrences of `DASHBOARD_LOGIN_SESSION_EXPIRED` (the pack was not regenerated in this PR — 34 `DASHBOARD_LOGIN` entries at both base and head).

**Failure scenario:** A user whose session expires sees the literal string `DASHBOARD_LOGIN_SESSION_EXPIRED` on the login screen instead of "Your session has expired. Please sign in again." — strictly worse UX than the hardcoded English this fix replaced. The seed pack needs the new key added and regenerated.

digit-ui-esbuild/products/dashboard/src/services/authService.js
```diff
@@ -0,0 +1,622 @@
+import { getTenantId } from "../config/dashboardConfig";
```
vinothrallapalli-eGov, line 1: **[CONFIRMED / correctness] Prior finding not fixed: authService is still byte-identical in two trees — and grew from 546 to 622 duplicated lines**

`authService.js` remains a byte-identical (md5-verified) duplicate between `digit-ui-esbuild/products/dashboard` and `frontend/micro-ui/web/src/dashboard`.

**Failure scenario:** The next fix to this security-sensitive session/refresh logic (e.g. a patch to the `ownsAliases` gating flagged in this round) lands in the production-facing esbuild copy but not the micro-ui copy. The sibling files already diverge — the micro-ui analyticsService lacks the record-then-rethrow metrics path and its DashboardLogin lacks i18n — proving the copies are hand-synced and drifting. Anyone building the classic micro-ui tree ships the unpatched auth behavior. Cheaper: keep the module only in the deployed esbuild tree, or extract one shared source both builds vendor.

frontend/micro-ui/web/src/dashboard/components/DashboardLogin.jsx
```diff
@@ -72,11 +51,15 @@ const inputStyle = {
...
-const DashboardLogin = ({ onLogin }) => {
+const DashboardLogin = ({ onLogin, expired = false }) => {
   const [username, setUsername] = useState("");
   const [password, setPassword] = useState("");
   const [submitting, setSubmitting] = useState(false);
-  const [error, setError] = useState(null);
+  const [error, setError] = useState(
+    // Distinguish "your session ended" from "you got the password wrong": the
+    // user did nothing wrong and reloading will not help them (#1466).
+    expired ? "Your session has expired. Please sign in again." : null
```
vinothrallapalli-eGov, line 61: **[CONFIRMED / correctness] Prior finding partially fixed: this micro-ui copy still hardcodes the English session-expired message**

The esbuild copy now localises via `t("DASHBOARD_LOGIN_SESSION_EXPIRED", ...)`, but this copy still hardcodes "Your session has expired. Please sign in again." — and the same literal remains duplicated in both authService.js copies' `sessionExpired` error (~line 614).

**Failure scenario:** A non-English-locale user of the classic micro-ui dashboard whose session expires is dropped to the login gate with an untranslatable English-only message; and if the wording is ever changed in authService's thrown error without updating the login copies (or vice versa), users see two different expiry messages for the same event.

digit-ui-esbuild/products/dashboard/src/services/authService.js
```diff
@@ -0,0 +1,622 @@
...
+ * token there is no verdict — only one endpoint's 401 — so the refused token is
+ * instead marked dead for THIS TAB (sessionStorage). That keeps the gate closed
+ * across a reload, which was #1466's "reloading did not help" symptom, without
+ * deleting keys the co-hosted app may still be using.
+ *
+ * Likewise clearSession only drops the unprefixed aliases and the
+ * Digit.SessionStorage cache when they actually hold the employee token being
+ * torn down — on a shared browser they may belong to a live citizen session.
+ *
+ * Guarantees: at most three sends and at most one refresh per call (cannot
```
vinothrallapalli-eGov, line 65: **[CONFIRMED / correctness] Send-count contract comments not updated for the guaranteed-replay change**

Line 65 promises "at most three sends" and the authFetch doc block (~line 489) still says "At most two sends happen", but the loop can now issue up to four sends (three loop attempts plus the inline post-refresh replay).

**Failure scenario:** A developer relying on the stated send-count contract (e.g. sizing server-side rate limits or reasoning about idempotency of the analytics POST) under-budgets by up to two requests per call; under a 401 storm each dashboard query legitimately hits the backend four times, tripping rate limiting/alerts the documented contract says cannot happen. Update both comments to match the actual worst case.

digit-ui-esbuild/products/dashboard/src/services/authService.js
```diff
@@ -0,0 +1,622 @@
...
+    const response = await send();
+    if (response.status !== 401) return response;
+
+    // If the generation has already moved, another caller refreshed while this
+    // request was in flight — our token was merely stale, so replay with the
+    // new one instead of refreshing again (a second refresh would invalidate
+    // the token that caller is using).
+    if (tokenGeneration !== generationAtSend) continue;
+
+    if (didRefresh) break;
```
vinothrallapalli-eGov, line 526: **[CONFIRMED / cleanup] `if (didRefresh) break;` is unreachable dead code**

`didRefresh` is set to true only later in the same iteration, after which every path returns or breaks (outcome !== REFRESHED → break; inline replay → return or break). The loop is only re-entered via the generation-mismatch `continue`, where `didRefresh` is still false.

**Failure scenario:** The guard can never fire, and its presence implies a multi-refresh re-entry path that does not exist, misleading the next maintainer of this security-sensitive loop. Delete it (or restructure so the single-refresh invariant is structural).

vinothrallapalli-eGov (commented):
**Resolution status of the 10 second-round findings:**

| # | Second-round finding | Status |
|---|---|---|
| 1 | NO_TOKEN 401 tore down session and wiped shared storage | ✅ Fixed |
| 2 | clearSession could wipe an active citizen session | ✅ Fixed via ownership gating — but the gating introduced two new issues (see inline) |
| 3 | Refresh on final loop attempt never replayed | ✅ Fixed — guaranteed replay added; stale contract comments + dead code remain (see inline) |
| 4 | Refresh-token owner check failed open on null uuid | ✅ Fixed |
| 5 | postAnalytics lost failed-401 metrics | ✅ Fixed (record-then-rethrow) |
| 6 | Tenant fallback silently changed 'ke'→'default' | ❌ **Not fixed** — documented as deliberate but unmitigated (see inline) |
| 7 | toRequestError text() fallback read a consumed body | ✅ Fixed |
| 8 | Auth module duplicated byte-identical across two trees | ❌ **Not fixed** — still byte-identical, grew 546→622 lines (see inline) |
| 9 | announceSessionExpired doc contradicted call site | ✅ Fixed |
| 10 | Hardcoded English session-expired string | ⚠️ **Partially fixed** — esbuild copy localized but the key is unseeded and renders raw; micro-ui copy still hardcoded (see inline) |

10 findings below as inline comments. The most serious new one: `persistSession` unconditionally stomps the shared unprefixed aliases — the exact write-side counterpart of the clearSession bug this round fixed.

digit-ui-esbuild/products/dashboard/src/components/DashboardLogin.jsx
```diff
@@ -47,35 +52,11 @@ const DEMO_USERS = [
...
-        window.localStorage?.removeItem(k);
-      } catch {
-        /* ignore */
-      }
-    }
-  );
-}
+// Session storage lives in authService (single source of truth); these thin
+// re-exports keep the existing AdminDashboard import sites unchanged.
+export const hasDashboardSession = hasAuth;
```
vinothrallapalli-eGov, line 58: **[CONFIRMED / cleanup] Auth helpers re-exported through the login component**

`hasDashboardSession`/`clearDashboardSession` re-export `hasAuth`/`clearSession` under aliases solely so `AdminDashboard`'s import sites stay unchanged, while `AdminDashboard` already imports `SESSION_EXPIRED_EVENT` directly from authService in this same PR.

**Failure scenario:** Two import paths to the same auth functions mean a future caller can pick either, and the component module becomes a load-bearing dependency of non-UI code. Importing `hasAuth`/`clearSession` directly from authService in `AdminDashboard` (a 2-line change in a file this diff already touches) removes the indirection.

vinothrallapalli-eGov (approved):

PR #1518, opened 2026-07-31
Windows (WSL2) setup: end-to-end Ansible localhost deploy with templates and quickstart

singhalkarun:
## What

Everything needed to bring the full Ansible stack up on a fresh localhost box — including Windows via WSL2 Ubuntu — plus the documentation to do it without help. Validated end-to-end on a 16 GB Windows machine (2026-07-29: `ok=98 failed=0`, all INFRA VALIDATION green; configurator, citizen SPA and MCP verified after re-enable).

## New host_vars templates

| Template | Fits | Notes |
|----------|------|-------|
| `localhost-slim.yml.example` | 16 GB host (12 GB WSL VM) | Full stack minus the Novu notifications stack (~2 GB resident) — the only functional delta |
| `localhost-full.yml.example` | 32 GB host | Everything, including Novu |

Both follow the maputo/bomet `.example` convention: operators `cp` to `<name>.yml` (gitignored) and `./deploy.sh <name>`. Because the inventory generator only globs `*.yml`, templates never enter the generated inventory — a manual `ansible-playbook` run without `--limit` cannot accidentally target the operator's own machine.

## Docs

**`WINDOWS-QUICKSTART.md`** (linked from `local-setup/README.md`'s setup-path chooser): the one-session WSL2 bring-up — install, ansible-core pin, WSL-filesystem clone, template choice, the intentional first-run stop for `.wslconfig` application, live progress tailing, verification URLs, and a troubleshooting table covering every failure mode hit during validation. Companion to `MAC-QUICKSTART.md`.

## Fresh-box playbook fixes (bite any new Ubuntu box, not just WSL)

- **Node 20 guard for `build_digit_ui` / `build_configurator`**: both build scripts hard-fail without node/npm; unlike the static/v2/integration-tests paths, nothing installed Node first. Wires the existing `tasks/ensure-node20.yml` in front of both. *Reviewer note: a Debian box with an older system node gets NodeSource Node 20 installed when those flags are on — same behavior the other three build paths already had.*
- **`user-seed` now depends on `egov-user-proxy`**, the service it actually calls — on a cold start it could exhaust its 150 s retry loop resolving a proxy whose container (and DNS name) didn't exist yet, dying with curl exit 6.

## WSL2 self-heals (skipped everywhere else)

- **Mount propagation**: WSL leaves `/` private, breaking node-exporter's `/:/rootfs:ro,rslave` bind ("path / is mounted on / but it is not a shared or slave mount"). Installs a boot-time `make-rshared-root.service` ordered before docker.service. Gated on the WSL kernel string ONLY — non-WSL hosts with a deliberately private `/` are never touched.
- **`.wslconfig` management**: writes the `[wsl2]` memory/swap/processors caps into the Windows-side `.wslconfig` via `/mnt/c` (`ini_file`; preserves other settings) and fails fast with instructions when the running VM doesn't reflect them yet (Ansible cannot restart the VM it runs inside). Adds `community.general` to requirements.

## Operator feedback

The image pull and the ~40-container `up` — the deploy's long silent stretch — stream live compose output to a per-tenant progress file, announced by a task and by a platform-aware `deploy.sh` banner (Linux runs no longer shown the macOS hint). `pipefail` keeps compose's exit code authoritative; Ansible still captures full output.

**Security (from CodeRabbit review, addressed in `bf63e0df`)**: the progress file lives under root-owned `{{ digit_dir }}` (`/opt/digit/digit-stack-up.<tenant>.progress`), not world-writable `/tmp`, closing a symlink-attack vector (root `tee` to a predictable `/tmp` path). Defense-in-depth: the path is recreated as a fresh regular file before the first write. Mode `0644` (not the suggested `0600`) so a non-root operator can still `tail -f` it.

## Known-issue workaround (remove when resolved)

The templates ship `insecure_registries: ["registry.preview.egov.theflywheel.in"]` because that registry's Let's Encrypt wildcard **expired 2026-07-25** and image pulls fail with x509. Clearly marked as a workaround to revert once the cert is renewed — worth escalating separately since it blocks every fresh deploy.

## Testing

- Full blank-box bring-up on WSL2 Ubuntu 24.04 (16 GB host) with the slim template: all containers healthy, UI/Gatus 200, ADMIN token minted, MDMS non-empty, OpenBao unsealed.
- Idempotent re-runs verified repeatedly, including the incremental re-run that added configurator + citizen SPA + MCP.
- Merges clean into current develop.
- CI: all five ansible/config/gatus/flyway checks pass. The two red checks (`test`, `tilt-test`) are the telemetry-sidecar verification, failing identically on develop since 2026-07-27 (see PR comment with run links) — unrelated to this change.

### DIFF
--- a/WINDOWS-QUICKSTART.md
+++ b/WINDOWS-QUICKSTART.md
@@ -1,0 +1,167 @@
+# Windows Quickstart — DIGIT via WSL2 in one session
+
+Validated 2026-07-29 on a 16 GB Windows machine / WSL2 Ubuntu 24.04. Brings up
+the full DIGIT stack with `./deploy.sh <name>` — the playbook self-heals every
+WSL-specific quirk (memory caps, mount propagation, Node toolchain), so the
+happy path is short. The same steps work on any Linux machine or VM; only the
+WSL2 sections are Windows-specific.
+
+## What you get
+
+~40 containers: all DIGIT core services, PGR, the employee UI, citizen SPA,
+configurator (DIGIT Studio), Kong, host nginx on port 80, Gatus health board,
+Grafana/Prometheus/Loki/Tempo observability, Jupyter DataLoader, OpenBao —
+on the dump-seeded `pg` / `pg.citya` tenants. Login works immediately.
+
+Two sizing profiles (see step 4):
+
+| Template | Fits | Difference |
+|----------|------|------------|
+| `localhost-slim.yml.example` | **16 GB machine** (12 GB WSL VM) | No Novu notifications stack (~2 GB). Everything else identical. |
+| `localhost-full.yml.example` | 32 GB machine | Adds Novu (SMS/WhatsApp delivery pipeline). |
+
+## Prerequisites
+
+- Windows 10/11 with hardware virtualization enabled (Intel VT-x / AMD SVM —
+  usually on by default; enable in BIOS if step 1 errors with `0x80370102`).
+- ≥ 16 GB RAM and ~60 GB free disk.
+- **Do NOT install Docker Desktop** (or disable its WSL integration for this
+  distro). The playbook installs Docker Engine natively inside WSL and manages
+  its daemon; Docker Desktop's injected `docker` conflicts with it.
+
+## 1. Install WSL2 + Ubuntu (PowerShell as Administrator)
+
+```powershell
+wsl --update
+wsl --install -d Ubuntu-24.04
+```
+
+Reboot if Windows asks, run the install command again if the distro isn't
+there yet, then create your Linux user on first launch. Confirm you're in:
+
+```bash
+uname -a          # must contain "microsoft ... WSL2"
+systemctl is-system-running   # "running" or "degraded" (Ubuntu 24.04 default)
+```
+
+If systemd reports `offline`, add to `/etc/wsl.conf`:
+
+```ini
+[boot]
+systemd=true
+```
+
+then `wsl --shutdown` in PowerShell and relaunch Ubuntu.
+
+## 2. Install the deploy tooling (inside WSL)
+
+```bash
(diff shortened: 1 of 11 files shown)


singhalkarun:
Review disposition (all CodeRabbit comments addressed in bf63e0df):

**Symlink attack on the `/tmp` progress file — valid, fixed.** Both the primary fix and the defense-in-depth were applied:
- `compose_progress_file` moved out of world-writable `/tmp` into `{{ digit_dir }}` (root-owned on Linux), so a non-root user cannot pre-plant a symlink at the path.
- The playbook additionally recreates the path as a fresh regular file (`state: absent` → `touch`) before the first `tee`, protecting operators who override the var back into a world-writable directory.
- One deliberate deviation from the suggested patch: mode `0644` instead of `0600` — the file's entire purpose is to be `tail -f`'d by a possibly non-root operator, and the directory permissions already prevent non-root writes.
- `deploy.sh` banner and `WINDOWS-QUICKSTART.md` updated to the new path.

**`user-seed` `service_started` vs `service_healthy` nitpick — agree with the "no change required" conclusion.** The seed script has its own 150-second retry loop against the proxy's `/health`, so connection-refused during nginx startup is already handled; `service_started` addresses the actual failure mode (DNS name not yet registered) and lets the seed start sooner.

singhalkarun:
CI note: the two failing checks (test, tilt-test) both die in the "Verify telemetry events captured" step with zero events and empty sidecar logs. This is pre-existing on develop — its own Tilt CI runs from 2026-07-27 and 2026-07-29 (e.g. run 30458992845, before this branch's changes) fail at the identical step. Nothing in this PR touches the telemetry sidecar or docker-compose.yml; all five ansible/config/gatus/flyway checks pass. The sidecar breakage is worth a separate issue.

vinothrallapalli-eGov (approved):

PR #1554, opened 2026-08-03
fix(dashboard): ownership-check alias writes; make sign-out unconditional (#1535, #1536)

### LINKED ISSUE #1535: fix(dashboard): persistSession overwrites shared token/user-info aliases without an ownership check
Follow-up from the #1515 review (merged before this was addressed): https://github.com/egovernments/Citizen-Complaint-Resolution-System/pull/1515#pullrequestreview-4828326583

cc @KDwevedi

## Problem

`persistSession` in `digit-ui-esbuild/products/dashboard/src/services/authService.js` (~line 253, byte-identical copy in `frontend/micro-ui/web/src/dashboard/services/authService.js`) unconditionally writes the shared unprefixed `token` and `user-info` aliases (`writeStorage("token", accessToken)` / `writeStorage("user-info", normalisedUser)`) with the employee identity on every silent background refresh — with no ownership check. This is the write-side counterpart of the `clearSession` bug fixed in #1515: `clearSession` now checks `ownsAliases` before touching these keys, but the write path does not.

## Failure scenario

A citizen is actively signed into the co-hosted digit-ui portal (the unprefixed aliases hold the citizen session — the exact shared-browser scenario `clearSession`'s own comment describes). A stale dashboard tab in the background gets a 401, silently refreshes, and `requestRefresh` → `persistSession` clobbers the citizen's aliases with the employee token and user-info. The main UI's boot path reads these aliases for login detection, so on next reload the citizen is silently switched to (or broken by) the employee identity — an identity swap performed without any user action.

## Suggested fix

Mirror the ownership check on the write side: only rewrite the unprefixed aliases (and the `Digit.SessionStorage` "User" sync) when they currently hold the employee's session (or are absent), analogous to `ownsAliases` in `clearSession`. Apply to both copies of `authService.js`.

### LINKED ISSUE #1536: fix(dashboard): explicit sign-out no longer clears session unconditionally; ownsAliases gate leaves dead token in Digit.SessionStorage
Follow-up from the #1515 review (merged before this was addressed): https://github.com/egovernments/Citizen-Complaint-Resolution-System/pull/1515#pullrequestreview-4828326583

cc @KDwevedi

## Problem

Two related consequences of the `ownsAliases` gating introduced in #1515 (`digit-ui-esbuild/products/dashboard/src/services/authService.js` ~line 306 and `components/DashboardLogin.jsx` ~line 59; byte-identical copies in `frontend/micro-ui/web/src/dashboard/`):

1. **Sign-out lost its unconditional-clear guarantee.** `clearDashboardSession` is now an alias for the ownership-gated `clearSession`, so the old invariant — "clicking Sign out always removes the token/user-info aliases" — only holds when the unprefixed `token` alias strictly equals `Employee.token`. On a shared kiosk, if the alias was rewritten by the co-hosted UI or stored in a different encoding (`parseJson` mismatch), Sign out + reload leaves `token`, `user-info`, and the `Digit.SessionStorage` "User" cache holding the employee's live token — the next kiosk user lands in the employee UI still authenticated as the employee who just signed out.

2. **The SessionStorage teardown is gated on the wrong condition.** `clearSession` gates the `Digit.SessionStorage` "User" clear on `ownsAliases` (alias-key ownership) instead of on the cached token itself. If a citizen later overwrote the unprefixed alias, a REJECTED (400 invalid_grant) employee refresh runs `clearSession` with `ownsAliases=false` and skips nulling "User" even though the cached token IS the dead employee token. Since index.js boots from SessionStorage first (and skips the localStorage re-seed when "User" exists), the co-hosted employee UI re-enters "logged in" on the dead token and every call 401s until the user manually clears storage — exactly the failure mode the surrounding comment says this block exists to prevent.

## Suggested fix

- Explicit sign-out should clear all `Employee.*` keys unconditionally, and clear the shared aliases whenever they hold the employee's token OR are unparseable.
- Gate the `Digit.SessionStorage` "User" clear on comparing the **cached** token against the employee token being cleared, not on alias ownership.
- Apply to both copies of `authService.js`.

KDwevedi:
Fixes #1535. Fixes #1536. The two review comments raised on #1515 that were merged past. Same root cause, one file (×2 copies), so they land together.

## Context

The dashboard and the citizen portal are co-hosted: they share the unprefixed `token` / `user-info` localStorage aliases and the single `Digit.SessionStorage` "User" slot. #1515 correctly recognised this on the **clear** side (`ownsAliases`) and stamped an owner onto the refresh token. The **write** side never got the same treatment, and the new gate had two side effects on sign-out.

## #1535 — persistSession seized aliases it did not own

`persistSession` wrote `token` and `user-info` with the employee identity on *every* silent refresh, unconditionally.

**Failure:** a citizen is signed into the portal. A stale dashboard tab in the background 401s, silently refreshes, and overwrites the citizen's aliases with the employee token and user-info. The main UI's boot path reads those aliases for login detection, so on next reload the citizen is silently switched to the employee identity — no user action anywhere.

**Fix:** alias writes are gated three ways —
- an explicit sign-in **claims** them (`claimAliases: true`, set only at the login call site — that is the one place where taking them over is the intent);
- a background refresh writes them only if they still hold the token being replaced;
- an absent alias is free to take.

`syncDigitSession` was overwriting the shared "User" cache unconditionally for the same reason and is now gated the same way. The two aliases move as a pair — a `token` from one identity beside a `user-info` from another is worse than either alone.

## #1536 — two consequences of the ownsAliases gate

**1. Sign-out lost its unconditional-clear guarantee.** `clearDashboardSession` was aliased straight to the gated `clearSession`, so "Sign out always removes the aliases" only held while the alias byte-matched `Employee.token`. On a shared kiosk, if the alias held a value other than `Employee.token` — rewritten by the co-hosted UI, or a leftover from an earlier session — Sign out + reload left `token`, `user-info` and the "User" cache holding a live employee token — the next person landed in the employee UI authenticated as the employee who had just signed out.

(An earlier draft of this description blamed an encoding mismatch. That was wrong and is corrected here: `readStorage` → `parseJson` falls back to the raw string when `JSON.parse` throws, and an opaque UUID is never valid JSON, so the comparison holds across both writers' encodings. A mismatch always means a *different value*.)

`clearSession` now takes `{ force }`. Set for user-initiated sign-out only; the automatic teardown (server-rejected refresh) keeps the ownership gate, where silently signing a citizen out would be the greater harm.

**2. The SessionStorage teardown was gated on the wrong condition** — alias ownership rather than what the cache holds. If a citizen later overwrote the alias, `ownsAliases` is false while the cache still holds the employee's dead token, so the gate stranded exactly the token the teardown exists to remove. It now keys off the cached token directly.

## Tests

New `authSessionOwnership.test.js`, **11 cases**, following the existing `dashboardMetrics.test.js` idiom (esbuild-to-CJS + mocked storage). Run from `digit-ui-esbuild/`:

```
node --test products/dashboard/src/services/authSessionOwnership.test.js
```

Every defect case was verified against the module *without* its fix, so the suite guards behaviour rather than asserting the new API. 7 cases cover #1535/#1536 (3 fail pre-fix); 4 more were added in the review round below (all 4 fail against the first commit of this PR, while the original 7 keep passing).

## Review round 2 — `bcec0b92e`

Six findings from the review of the first commit. Each is a case where token equality was standing in for ownership and cannot carry that weight.

1. **`syncDigitSession` stranded our own stale cache.** The "User" slot is `sessionStorage` (per-tab) while `previousToken` is `localStorage` (shared), so a sibling tab that refreshed first leaves this tab's cache one vintage behind — by token alone, indistinguishable from a stranger's. Skipping the write left the surrounding employee UI authenticating with a token the earlier refresh had already invalidated, and a reload did not heal it because `index.js` boots from that cache and skips the localStorage re-seed. Ownership now also accepts a matching `info.uuid` (falling back to stored identity, since a refresh usually returns no `UserRequest`).
2. **The cache write was gated on emptiness, not ownership.** Digit storage expires entries after 24h, so a kiosk left overnight has a null cache beside a citizen's still-live aliases — and the employee identity was being written into it. `syncDigitSession` now receives `mayWriteAliases`, not `claimAliases`.
3. **A refresh in flight across a sign-out resurrected the session.** Nothing broadcasts sign-out and localStorage is shared, so a refresh started before the wipe could land after it, rewriting `Employee.*` and — the aliases now being vacant, hence "claimable" — the shared aliases too. `requestRefresh` re-checks that the refresh token still exists before persisting; `clearSession` always drops it, so its absence is the marker.
4. **Forced sign-out could delete a citizen's live aliases.** It now spares aliases whose `user-info` carries a *different* uuid — those are positively identified as someone else's session, and deleting them signs a citizen out mid-flow, breaking the module's rule 1 to fix a leftover. A missing uuid on either side proves nothing and still clears, so the kiosk case of #1536 is unchanged.
5. **Token-less cache** is dropped on a non-force teardown only when it carries no identity payload either.
6. **The ownership predicates were hand-copied** across two functions × two mirrored files — the same drift that produced #1535. Extracted as `readAliasToken` / `ownsSharedAliases` / `getCachedSessionToken`; the alias is now read once in `persistSession` rather than twice.

One finding was **not** actioned: nulling the `Digit.SessionStorage` "User" cache on a forced sign-out cannot sign out a citizen in another tab, because `Digit.SessionStorage` is backed by `window.sessionStorage` (`packages/libraries/src/services/atoms/Utils/Storage.js:64`) and is therefore per-tab. The only cache reachable from that line is the one in the tab where Sign out was clicked, where an empty cache is the intended end state.

## Notes for review

- Both copies (`digit-ui-esbuild/products/dashboard/` and `frontend/micro-ui/web/src/dashboard/`) are updated; `authService.js` stays byte-identical between them.
- All four call sites audited: sign-out → forced clear; login → claims aliases; `announceSessionExpired(clearStorage)` → unforced clear (unchanged semantics); `requestRefresh` → unclaimed persist.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dashboard sign-out to reliably clear stale session data.
  * Prevented background session refreshes from overwriting another active user’s session.
  * Improved handling of shared authentication data across dashboard and citizen portal sessions.
  * Prevented in-progress refresh requests from restoring sessions after sign-out.
  * Ensured explicit sign-ins correctly establish session ownership.

* **Tests**
  * Added coverage for session ownership, refresh behavior, sign-out, cache expiry, and shared browser storage scenarios.

### DIFF
--- a/digit-ui-esbuild/products/dashboard/src/services/authService.js
+++ b/digit-ui-esbuild/products/dashboard/src/services/authService.js
@@ -60,7 +60,12 @@ import { getTenantId } from "../config/dashboardConfig";
  *
  * Likewise clearSession only drops the unprefixed aliases and the
  * Digit.SessionStorage cache when they actually hold the employee token being
- * torn down — on a shared browser they may belong to a live citizen session.
+ * torn down — on a shared browser they may belong to a live citizen session. An
+ * explicit sign-out overrides that (it must leave no employee session behind on
+ * a kiosk) with one exception: aliases carrying a DIFFERENT identity are proven
+ * to be someone else's and are still spared. The mirror-image rule governs
+ * writes: persistSession seizes the aliases only on a deliberate sign-in, never
+ * on a silent background refresh (#1535, #1536).
  *
  * Guarantees: at most three sends and at most one refresh per call (cannot
  * loop); concurrent 401s share a single refresh; a refresh never re-enters this
@@ -135,6 +140,38 @@ function writeStorage(key, value) {
 
 export function getEmployeeToken() {
   return readStorage(TOKEN_KEY);
+}
+
+/**
+ * The ownership predicates, defined ONCE.
+ *
+ * These rules were hand-copied into persistSession and clearSession, and the
+ * copies drifting apart is precisely what produced #1535 (the write side never
+ * got the clear side's ownership check) and #1536. Two mirrored files times two
+ * functions gave eight sites that had to change in lockstep; now there is one
+ * definition per rule.
+ */
+function readAliasToken() {
+  return readStorage(KEYS.tokenAlias.key);
+}
+
+/**
+ * True when the shared unprefixed aliases still hold `token`.
+ *
+ * Encoding is NOT a hazard here despite the two writers disagreeing about it:
+ * the main digit-ui login stores the token raw while this module JSON-stringifies
+ * it, but readStorage -> parseJson falls back to the raw string when JSON.parse
+ * throws, and an opaque UUID is never valid JSON. So the comparison holds across
+ * both encodings, and a mismatch always means a DIFFERENT VALUE — someone else's
+ * token, or a leftover.
+ */
+function ownsSharedAliases(token, aliasToken = readAliasToken()) {
+  return token != null && aliasToken === token;
+}
+
+/** The token a Digit.SessionStorage "User" entry authenticates with, if any. */
+function getCachedSessionToken(cached) {
+  return cached?.token ?? cached?.access_token ?? null;
 }
 
 /**
@@ -236,9 +273,24 @@ function syncDigitSession(accessToken, userInfo, { claim = false, previousToken 
     // authenticates from it too. Overwriting it on a background refresh would
     // swap an actively-signed-in citizen to the employee identity without any
     // user action (#1535). Only write when this is a deliberate sign-in, or
-    // when the cache is empty / already holds the token we are replacing.
-    const cachedToken = existing?.token ?? existing?.access_token ?? null;
-    if (!claim && cachedToken != null && cachedToken !== previousToken) return;
+    // when the cache is provably ours.
+    //
+    // Token equality alone does NOT prove that: the cache is sessionStorage
+    // (per-tab) while previousToken comes from localStorage (shared), so a
+    // sibling tab that refreshed first leaves THIS tab's cache one vintage
+    // behind — indistinguishable, by token, from a stranger's. Refusing to
+    // write there strands the tab on a token the earlier refresh already
+    // invalidated, and a reload does not heal it because index.js boots from
+    // this cache and skips the localStorage re-seed while it looks populated.
+    // Identity is the tab-independent proof; a citizen's cache still fails it.
+    const cachedToken = getCachedSessionToken(existing);
+    const cachedUuid = existing?.info?.uuid ?? null;
+    // A refresh usually carries no UserRequest, so fall back to stored identity.
+    const ourUuid = userInfo?.uuid ?? getEmployeeInfo()?.uuid ?? null;
+    const ownsCache =
+      cachedToken === previousToken ||
+      (cachedUuid != null && ourUuid != null && cachedUuid === ourUuid);
+    if (!claim && !ownsCache) return;
     session.set("User", {
       ...existing,
       token: accessToken,
@@ -267,9 +319,12 @@ export function persistSession({ accessToken, refreshToken, userInfo, tenantId, 
   // refreshes clobbers a citizen's live session on the co-hosted portal with
   // the employee identity, an identity swap with no user action (#1535).
   const previousToken = getEmployeeToken();
-  const ownsAliases =
-    previousToken != null && readStorage(KEYS.tokenAlias.key) === previousToken;
-  const mayWriteAliases = claimAliases || ownsAliases || readStorage(KEYS.tokenAlias.key) == null;
+  // Read the alias ONCE: two reads of a value that cannot change between them
+  // are a wasted parse, and they let a later edit make the two expressions
+  // disagree — silently splitting one ownership decision into two.
+  const aliasToken = readAliasToken();
+  const mayWriteAliases =
+    claimAliases || ownsSharedAliases(previousToken, aliasToken) || aliasToken == null;
 
   if (accessToken != null) {
     writeStorage(TOKEN_KEY, accessToken);
@@ -300,7 +355,13 @@ export function persistSession({ accessToken, refreshToken, userInfo, tenantId, 
     } catch {
       /* ignore */
     }
-    syncDigitSession(accessToken, normalisedUser, { claim: claimAliases, previousToken });
+    // Pass mayWriteAliases, not claimAliases: an EMPTY "User" cache is not ours
+    // to take while the aliases belong to someone else. Digit's storage expires
+    // entries after 24h, so a kiosk left overnight nulls a citizen's cache while
+    // their aliases stay live — and writing the employee identity into it would
+    // re-enter #1535 through the cache door, since index.js boots from the cache
+    // first and would then serve the employee session under the citizen's aliases.
+    syncDigitSession(accessToken, normalisedUser, { claim: mayWriteAliases, previousToken });
   }
 }
 
@@ -319,11 +380,24 @@ export function clearSession({ force = false } = {}) {
   // tearing down, or we would sign an unrelated user out of a session they are
   // actively using.
   const employeeToken = getEmployeeToken();
-  const ownsAliases =
-    employeeToken != null && readStorage(KEYS.tokenAlias.key) === employeeToken;
+  const ownsAliases = ownsSharedAliases(employeeToken);
+
+  // A forced sign-out must leave no employee session behind, so it drops the
+  // aliases even when they no longer match our token — that mismatch is usually
+  // a leftover, and leaving it stranded is what #1536 was about. But "not
+  // provably ours" is not "provably someone else's": when the `user-info` alias
+  // carries a DIFFERENT uuid, the aliases are positively identified as another
+  // live session, and deleting them would sign a citizen out of the co-hosted
+  // portal mid-flow — breaking rule 1 above to fix a leftover. A missing uuid on
+  // either side proves nothing, so those still clear.
+  const employeeUuid = getEmployeeInfo()?.uuid ?? null;
+  const aliasUuid = readStorage(KEYS.userInfoAlias.key)?.uuid ?? null;
+  const aliasesAreForeign =
+    !ownsAliases && aliasUuid != null && employeeUuid != null && aliasUuid !== employeeUuid;
+  const mayDropAliases = ownsAliases || (force && !aliasesAreForeign);
 
   SESSION_KEYS.forEach((k) => {
-    if (!force && !ownsAliases && (k === KEYS.tokenAlias.key || k === KEYS.userInfoAlias.key)) return;
+    if (!mayDropAliases && (k === KEYS.tokenAlias.key || k === KEYS.userInfoAlias.key)) return;
     try {
       window.localStorage?.removeItem(k);
     } catch {
@@ -344,10 +418,13 @@ export function clearSession({ force = false } = {}) {
   if (session?.set) {
     try {
       const cached = session.get?.("User");
-      const cachedToken = cached?.token ?? cached?.access_token ?? null;
+      const cachedToken = getCachedSessionToken(cached);
       // Never null a cache holding a DIFFERENT live session — but on an
-      // explicit sign-out an empty cache is the required end state.
-      if (force || cachedToken == null || cachedToken === employeeToken) session.set("User", null);
+      // explicit sign-out an empty cache is the required end state. A cache with
+      // no token authenticates nobody, so it is safe to drop UNLESS it still
+      // carries a foreign identity payload the co-hosted UI is rendering from.
+      if (force || cachedToken === employeeToken || (cachedToken == null && !cached?.info))
+        session.set("User", null);
     } catch {
       /* ignore */
     }
@@ -493,6 +570,16 @@ async function requestRefresh() {
   // A 2xx with no token is a malformed/intercepted response (proxy login page,
   // truncated body), not a rejection. Same reasoning as above: do not destroy.
   if (!data?.access_token) return REFRESH_UNAVAILABLE;
+
+  // The refresh token was read up to 15s ago. A sign-out — in this tab or, since
+  // localStorage is shared and nothing broadcasts the event, in another one —
+  // may have wiped the session in the meantime. Persisting now would write a
+  // freshly issued employee token back into keys the user just cleared, and
+  // because the wipe leaves the aliases vacant, mayWriteAliases would let it
+  // take those too: the next person at a shared kiosk boots straight into the
+  // signed-out employee's session. clearSession always drops REFRESH_KEY, so its
+  // absence is the marker. Nothing to clean up — we wrote nothing.
+  if (readStorage(REFRESH_KEY) == null) return REFRESH_UNAVAILABLE;
 
   persistSession({
     accessToken: data.access_token,
--- a/digit-ui-esbuild/products/dashboard/src/services/authService.js
+++ b/digit-ui-esbuild/products/dashboard/src/services/authService.js
@@ -227,11 +227,18 @@ export function withSignificantRoleFirst(userInfo) {
  * every main-UI call would 401 until a full reload. Push the new token into
  * that cache so both surfaces stay on the same session.
  */
-function syncDigitSession(accessToken, userInfo) {
+function syncDigitSession(accessToken, userInfo, { claim = false, previousToken = null } = {}) {
   const session = window.Digit?.SessionStorage;
   if (!session?.set || !session?.get) return; // standalone build — nothing to sync
   try {
     const existing = session.get("User") || {};
+    // The "User" slot is a SINGLE shared cache — the co-hosted citizen portal
+    // authenticates from it too. Overwriting it on a background refresh would
+    // swap an actively-signed-in citizen to the employee identity without any
+    // user action (#1535). Only write when this is a deliberate sign-in, or
+    // when the cache is empty / already holds the token we are replacing.
+    const cachedToken = existing?.token ?? existing?.access_token ?? null;
+    if (!claim && cachedToken != null && cachedToken !== previousToken) return;
     session.set("User", {
       ...existing,
       token: accessToken,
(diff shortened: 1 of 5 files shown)


vinothrallapalli-eGov (commented):
Automated high-effort review (Claude Code workflow: 4 independent finders + adversarial verification per finding). 7 findings survived verification — 2 confirmed correctness regressions, 3 plausible variants around the same ownership logic, 2 cleanup items. All findings apply byte-identically to the mirrored copy at `frontend/micro-ui/web/src/dashboard/services/authService.js`.

digit-ui-esbuild/products/dashboard/src/services/authService.js
```diff
@@ -227,11 +227,18 @@ export function withSignificantRoleFirst(userInfo) {
...
   if (!session?.set || !session?.get) return; // standalone build — nothing to sync
   try {
     const existing = session.get("User") || {};
+    // The "User" slot is a SINGLE shared cache — the co-hosted citizen portal
+    // authenticates from it too. Overwriting it on a background refresh would
+    // swap an actively-signed-in citizen to the employee identity without any
+    // user action (#1535). Only write when this is a deliberate sign-in, or
+    // when the cache is empty / already holds the token we are replacing.
+    const cachedToken = existing?.token ?? existing?.access_token ?? null;
+    if (!claim && cachedToken != null && cachedToken !== previousToken) return;
```
vinothrallapalli-eGov: **[CONFIRMED — correctness]** The new gate (`cachedToken !== previousToken` ⇒ skip) can't distinguish a citizen-owned cache from the employee's **own older token**, so a tab whose per-tab `Digit.SessionStorage` "User" cache holds a stale employee token is never re-synced after a successful silent refresh — a regression from the previous unconditional write.

**Failure scenario:** Employee has the SPA open in two tabs. Tab A silently refreshes (T0→T1, server invalidates T0; localStorage `Employee.token` becomes T1, but tab B's sessionStorage-backed "User" cache still holds T0). A request in tab B that was in flight with T0 401s and triggers tab B's own refresh (T1→T2); `persistSession` computes `previousToken=T1`, tab B's `cachedToken=T0 !== T1` and `claim=false`, so the sync returns without writing. Tab B's co-hosted employee UI keeps authenticating from dead T0 and every main-UI call 401s; even a full reload does not recover because `index.js` boots from the SessionStorage "User" slot first and skips the localStorage re-seed when it exists — the exact trap this module's own comments describe. Before this PR, tab B's refresh repaired the cache.

*(Mirrored in `frontend/micro-ui/web/src/dashboard/services/authService.js:241`.)*

vinothrallapalli-eGov: **[PLAUSIBLE — correctness]** The cache-empty branch of this gate (`cachedToken == null` ⇒ write) claims the shared "User" cache on a silent background refresh even when the unprefixed aliases are owned by a live citizen session — re-introducing the #1535 identity swap through the cache path.

**Failure scenario:** Digit Storage entries expire after `ttl=86400s`, so on a kiosk left ~24h the citizen's "User" cache entry expires (`session.get` returns null) while their `token`/`user-info` localStorage aliases remain live. A stale dashboard surface then 401s and silently refreshes: `mayWriteAliases` is false so the aliases are correctly spared, but `syncDigitSession` sees `cachedToken == null` and writes the employee token/identity into the "User" cache. The citizen's next action in the co-hosted SPA authenticates as the employee — a no-user-action identity swap. The write gate should mirror the alias-ownership check, not just cache emptiness.

digit-ui-esbuild/products/dashboard/src/services/authService.js
```diff
@@ -243,14 +250,30 @@ function syncDigitSession(accessToken, userInfo) {
...
 
+  // Read BEFORE writing: the token the aliases must currently hold for us to
+  // still own them. This is the write-side counterpart of clearSession's
+  // ownsAliases check — without it, a stale tab that 401s and silently
+  // refreshes clobbers a citizen's live session on the co-hosted portal with
+  // the employee identity, an identity swap with no user action (#1535).
+  const previousToken = getEmployeeToken();
+  const ownsAliases =
+    previousToken != null && readStorage(KEYS.tokenAlias.key) === previousToken;
+  const mayWriteAliases = claimAliases || ownsAliases || readStorage(KEYS.tokenAlias.key) == null;
```
vinothrallapalli-eGov: **[PLAUSIBLE — correctness]** `mayWriteAliases` treats a vacant `token` alias (`readStorage == null`) as claimable by a **silent** background refresh, so a refresh in flight (or firing later from another employee-dashboard tab) after `clearDashboardSession`'s force wipe re-populates `Employee.token` AND the shared aliases — resurrecting the employee session that #1536's sign-out guarantee just destroyed.

**Failure scenario:** Shared kiosk, dashboard open in two tabs; in tab A the employee clicks Sign out — `clearSession({force:true})` wipes all keys and the page reloads to the login gate. Tab B's dashboard had already 401'd and its `requestRefresh` (started before the wipe, refresh token read pre-clear) resolves afterwards: `persistSession` runs with `claimAliases=false`, sees the token alias vacant, and writes the fresh employee token + user-info into both `Employee.*` keys and the unprefixed aliases. The next person at the kiosk boots straight into the signed-out employee's authenticated session — the exact #1536 symptom the force flag was added to prevent.

digit-ui-esbuild/products/dashboard/src/services/authService.js
```diff
@@ -275,22 +300,30 @@ export function persistSession({ accessToken, refreshToken, userInfo, tenantId }
...
+  // sign-out, only drop them when they actually hold the employee token we are
+  // tearing down, or we would sign an unrelated user out of a session they are
+  // actively using.
   const employeeToken = getEmployeeToken();
   const ownsAliases =
     employeeToken != null && readStorage(KEYS.tokenAlias.key) === employeeToken;
 
   SESSION_KEYS.forEach((k) => {
-    if (!ownsAliases && (k === KEYS.tokenAlias.key || k === KEYS.userInfoAlias.key)) return;
+    if (!force && !ownsAliases && (k === KEYS.tokenAlias.key || k === KEYS.userInfoAlias.key)) return;
```
vinothrallapalli-eGov: **[CONFIRMED — correctness]** `clearSession({force:true})` unconditionally deletes the shared aliases here (and nulls the "User" cache at line 350) even when they **provably hold a DIFFERENT user's live session**, contradicting both the module's rule 1 ('never destroy a session we have not proven is dead') and the adjacent comment 'Never null a cache holding a DIFFERENT live session'.

**Failure scenario:** A citizen is actively signed into the co-hosted portal (unprefixed `token`/`user-info` aliases hold the citizen's live token; `Employee.token` holds a distinguishable stale employee token). The employee returns to an old dashboard tab and clicks Sign out: `clearDashboardSession()` forces removal of the citizen's `token`/`user-info` aliases from shared localStorage, silently signing the citizen out mid-flow (e.g. losing a half-filed complaint) — even though `ownsAliases` was false and the code could tell the aliases were not the employee's. The force flag needed to cover only the same-token-different-encoding kiosk case, but it also destroys sessions the ownership check positively identifies as foreign.

digit-ui-esbuild/products/dashboard/src/services/authService.js
```diff
@@ -302,13 +335,19 @@ export function clearSession() {
...
-  if (session?.set && ownsAliases) {
+  if (session?.set) {
     try {
       const cached = session.get?.("User");
       const cachedToken = cached?.token ?? cached?.access_token ?? null;
-      // Same ownership test: never null a cache holding a different session.
-      if (cachedToken == null || cachedToken === employeeToken) session.set("User", null);
+      // Never null a cache holding a DIFFERENT live session — but on an
+      // explicit sign-out an empty cache is the required end state.
+      if (force || cachedToken == null || cachedToken === employeeToken) session.set("User", null);
```
vinothrallapalli-eGov: **[PLAUSIBLE — correctness]** The automatic (non-force) teardown dropped the `ownsAliases` guard on the Digit "User" cache: a foreign cache object whose token is stored under neither `token` nor `access_token` (`cachedToken == null`) is now nulled during an employee session teardown, where the old code left non-owned caches untouched.

**Failure scenario:** A citizen is signed into the co-hosted portal whose "User" cache entry carries their identity/info but no `token`/`access_token` field (e.g. info-only seed or a different token key). A stale employee tab's refresh is positively rejected (400 invalid_grant) → `clearSession()` with `force=false`: `ownsAliases` is false so aliases are spared, but the cache branch now runs unconditionally and `cachedToken == null` triggers `session.set('User', null)`, deleting the citizen's cached session object. Since the main UI boots from SessionStorage "User" first, the citizen's next in-tab navigation loses their identity/session state.

digit-ui-esbuild/products/dashboard/src/services/authService.js
```diff
@@ -243,14 +250,30 @@ function syncDigitSession(accessToken, userInfo) {
...
   const normalisedUser = userInfo != null ? withSignificantRoleFirst(userInfo) : null;
 
+  // Read BEFORE writing: the token the aliases must currently hold for us to
+  // still own them. This is the write-side counterpart of clearSession's
+  // ownsAliases check — without it, a stale tab that 401s and silently
+  // refreshes clobbers a citizen's live session on the co-hosted portal with
+  // the employee identity, an identity swap with no user action (#1535).
+  const previousToken = getEmployeeToken();
+  const ownsAliases =
+    previousToken != null && readStorage(KEYS.tokenAlias.key) === previousToken;
```
vinothrallapalli-eGov: **[CONFIRMED — cleanup]** The diff grows the number of hand-rolled copies of the same ownership predicates: alias-ownership (`x != null && readStorage(KEYS.tokenAlias.key) === x`) is now duplicated in `persistSession` (lines 270-272) and `clearSession` (lines 322-323), and the cached-token extraction (`?.token ?? ?.access_token ?? null`) is duplicated in `syncDigitSession` (line 240) and `clearSession` (line 347).

Four copies across two functions, times two byte-identical mirrored files, gives eight sites that must change in lockstep whenever the ownership rule evolves — and #1536 itself was caused by exactly this kind of drift. Extracting helpers such as `readAliasToken()`, `ownsSharedAliases(token)` and `getCachedSessionToken(session)` in this module would leave one place per copy to fix.

digit-ui-esbuild/products/dashboard/src/services/authService.js
vinothrallapalli-eGov: **[CONFIRMED — efficiency]** `persistSession` calls `readStorage(KEYS.tokenAlias.key)` twice back-to-back (once inside `ownsAliases` on line 271, again for the `== null` check here), doing a second `localStorage.getItem` + `JSON.parse` for a value that cannot have changed between the two expressions.

Beyond the wasted read, two separate reads let a future edit accidentally make them non-equivalent (e.g. changing the key in one expression but not the other), silently splitting the ownership decision. Reading once into `const aliasToken = readStorage(KEYS.tokenAlias.key)` and using it in both `ownsAliases` and `mayWriteAliases` is cheaper and safer.

KDwevedi:
Thanks — all seven addressed in `bcec0b92e`. Six actioned, one I'd push back on.

**`:241` (own older token, CONFIRMED)** — agreed, and the mechanism is worse than "can't distinguish": the cache is `sessionStorage` (per-tab) while `previousToken` is `localStorage` (shared), so tabs of the *same* employee legitimately diverge. Ownership now also accepts a matching `info.uuid`, falling back to `getEmployeeInfo()` since a refresh usually returns no `UserRequest`. A citizen's cache still fails it.

**`:241` (cache-empty branch, PLAUSIBLE)** — confirmed, including the 24h TTL premise (`Storage.js:29` `set(key, value, ttl = 86400)`, `:18-21` returns null and evicts past expiry). Fixed by passing `mayWriteAliases` instead of `claimAliases` into `syncDigitSession`, so the cache gate mirrors the alias gate exactly as you suggested.

**`:272` (vacant alias claimable after a force wipe, PLAUSIBLE)** — confirmed, and there was no guard: `tokenGeneration` only de-duplicates concurrent refreshes, and nothing broadcasts sign-out across tabs. `requestRefresh` now re-checks that the refresh token still exists immediately before persisting; `clearSession` drops it unconditionally, so its absence is the marker. Covered by a new test that signs out from inside the mocked refresh response.

**`:326` (force deletes a foreign live session, CONFIRMED)** — agreed on the aliases. The rule is now "wipe unless *provably* someone else's": a `user-info` alias carrying a different uuid is positive evidence and is spared; a missing uuid on either side proves nothing and still clears, so the kiosk case of #1536 is untouched (test 7 still passes unchanged).

**`:350` (force nulls the "User" cache) — not actioned.** `Digit.SessionStorage` is `getStorage(window.sessionStorage)` (`packages/libraries/src/services/atoms/Utils/Storage.js:64`; `PersistantStorage` on the next line is the localStorage one, and the dashboard doesn't use it). It's per-tab, so the only cache that line can reach is the one in the tab where Sign out was just clicked — a citizen in another tab is unreachable, and an empty cache in *this* tab is the intended end state of a deliberate sign-out. Happy to be shown otherwise if you have a path where those two tabs share the slot.

**`:271` / `:272` (cleanup + double read)** — done: `readAliasToken()`, `ownsSharedAliases(token, aliasToken?)` and `getCachedSessionToken(cached)` extracted, alias read once in `persistSession`, both mirrored copies still byte-identical.

One correction to my own PR description while here: the #1536 rationale I wrote blamed an encoding mismatch (`parseJson`). That was wrong. `readStorage` → `parseJson` falls back to the raw string when `JSON.parse` throws, and an opaque UUID is never valid JSON, so `readStorage("token") === readStorage("Employee.token")` holds across both writers' encodings. The only reachable case is the alias holding a *different value*. Description and the `DashboardLogin.jsx` comment are corrected.

Suite is now 11 cases. The 4 new ones all fail against `c821d9c1d` and the original 7 pass on both, so the round is guarded the same way the first one was:

```
not ok 5 - a refresh re-syncs our OWN cache even when it is a vintage behind
not ok 6 - a silent refresh does not claim an EXPIRED cache while aliases are foreign
not ok 8 - forced sign-out spares aliases that are provably a DIFFERENT user's
not ok 10 - a refresh in flight when the user signs out does not resurrect the session
```

vinothrallapalli-eGov (approved):

PR #1561, opened 2026-08-04
fix(pgr): don't restrict inbox-v2 complaint search to open states

### LINKED ISSUE #1367: [Bug] Employee: CSR user is unable to search a complaint in inbox
# Description

When a complaint is rejected by GRO, it should be visible in CSR inbox/search screen so that the user should be able to open complaint details page and reopen the complaint 

---

# Steps to Reproduce

1. Create a complaint from CSR
2. Go to GRO inbox and reject the complaint (search complaint works in GRO inbox)
3. Go to CSR employee inbox 

---

# Expected Behavior

CSR shall be able to search a specific complaint with id

---

# Actual Behavior

CSR unable to search complaint with id 

---

https://github.com/user-attachments/assets/a48456db-c782-4827-a93b-bfc3ca212501

Lokendra-egov:
CSR searching for a specific complaint number in inbox-v2 got zero results once GRO rejected it, because PGRInboxConfig.preProcess always defaulted applicationStatus to OPEN_STATES/allActionableStates when no status checkbox was checked — even for an explicit complaintNumber/ mobileNumber/date-range search. REJECTED is a terminal state (excluded from that default) even though CSR can still REOPEN it.

Skip the default status restriction whenever the search form has an explicit complaintNumber, mobileNumber, or date range, so a targeted search is never narrowed by application status. The _count request reuses the same params, so it's fixed too.

Fixes #1367

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved PGR inbox search behavior for complaint numbers, mobile numbers, and complete request-date ranges.
  - Targeted searches no longer apply a default application-status filter when no status is selected.
  - Existing open-status defaults remain unchanged for broader searches, and explicit status selections continue to work as expected.

### DIFF
--- a/digit-ui-esbuild/products/pgr/src/configs/UICustomizations.js
+++ b/digit-ui-esbuild/products/pgr/src/configs/UICustomizations.js
@@ -1,4 +1,5 @@
 import _ from "lodash";
+import Urls from "../utils/urls";
 import { complaintLabel } from "../utils/complaintLabel";
 import { useLocation, useHistory, Link, useParams } from "react-router-dom";
 import React, { useState, Fragment } from "react";
@@ -1652,22 +1653,14 @@ export const UICustomizations = {
       ];
       const rawStatuses = filterForm.status || {};
       const statuses = Object.keys(rawStatuses).filter((key) => rawStatuses[key] === true);
-      // A targeted lookup (complaint number / mobile number / date range —
-      // the "search" section, as opposed to the "filter" section) must not be
-      // narrowed to open states: CCRS#1367 — once GRO rejects a complaint it
-      // moves to the terminal REJECTED state, which isn't in OPEN_STATES/
-      // allStates, so a CSR searching that exact complaint number got zero
-      // results back even though the complaint still needs to be reopened.
-      const hasExplicitSearch = Boolean(
-        searchForm.complaintNumber || searchForm.mobileNumber || (requestDate?.startDate && requestDate?.endDate)
-      );
-      // Status scope: an explicit filter wins; a targeted search leaves the
-      // status scope open; otherwise both tabs default to every open/
-      // actionable state. The tabs differ on the ASSIGNEE axis, not the
-      // status axis (PO decision 2026-07-15).
+      // CCRS#1367: an exact complaint-number lookup (not mobile number or
+      // date range — those are broad, multi-result searches) must ignore
+      // every inbox default, not just status: it needs to find that one
+      // complaint no matter who it's assigned to or what state it's in.
+      const isComplaintLookup = Boolean(params.serviceRequestId);
       if (statuses.length > 0) {
         params.applicationStatus = statuses;
-      } else if (!hasExplicitSearch) {
+      } else if (!isComplaintLookup) {
         params.applicationStatus = allStates.length > 0 ? allStates : OPEN_STATES;
       }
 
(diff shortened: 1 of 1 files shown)


digit-ui-esbuild/products/pgr/src/configs/UICustomizations.js
```diff
@@ -1652,12 +1652,22 @@ export const UICustomizations = {
...
+        searchForm.complaintNumber || searchForm.mobileNumber || (requestDate?.startDate && requestDate?.endDate)
+      );
+      // Status scope: an explicit filter wins; a targeted search leaves the
+      // status scope open; otherwise both tabs default to every open/
+      // actionable state. The tabs differ on the ASSIGNEE axis, not the
+      // status axis (PO decision 2026-07-15).
       if (statuses.length > 0) {
         params.applicationStatus = statuses;
-      } else {
+      } else if (!hasExplicitSearch) {
```
vinothrallapalli-eGov: **[CONFIRMED — correctness]** Explicit search opens the status scope but not the **assignee** scope: on the default "MY" tab (client mode), the unchanged block at lines 1686-1691 still sets `params.assignee = [me]`, so an exact-complaint-number search for a rejected complaint still returns zero results — CCRS#1367 remains unfixed on the tab the user lands on.

**Failure scenario:** `PGRInbox.js` initializes `activeTab` to "MY" (`useState("MY")`, line 77). A CSR opens the inbox and types the exact complaint number of a complaint the GRO rejected. `hasExplicitSearch` correctly skips the `applicationStatus` default, but `params.assignee = [CSR uuid]` is still added. The rejected complaint is terminal with no active workflow assignment to the CSR, so pgr-services returns zero rows — the exact symptom of #1367 persists unless the user knows to switch to the "All" tab first.

vinothrallapalli-eGov: **[CONFIRMED — correctness]** In server-side visibility mode, an explicit search on the **All** tab still sends `scope=TEAM` (line 1685), whose backend predicate (`serviceRequestId IN` team-assigned IDs `OR applicationStatus IN` unassigned states — `PGRQueryBuilder.java:143-157`) excludes terminal REJECTED complaints, so the #1367 fix is ineffective for supervisors with projected reportees.

**Failure scenario:** On a `serverSide=true` tenant, a supervisor with reportees in the HRMS projection searches the exact complaint number of a rejected complaint from the All tab: the frontend now omits `applicationStatus`, but `VisibilityService.resolve` sets `visibilityIds` (team assignments) + `visibilityUnassignedStates` (open queue states); the rejected complaint has no active assignee and REJECTED is not an unassigned-queue state, so the search still returns zero rows and the complaint remains unfindable/un-reopenable from the inbox.

digit-ui-esbuild/products/pgr/src/configs/UICustomizations.js
```diff
@@ -1652,12 +1652,22 @@ export const UICustomizations = {
...
-      // every open/actionable state. The tabs differ on the ASSIGNEE axis, not
-      // the status axis (PO decision 2026-07-15).
+      // A targeted lookup (complaint number / mobile number / date range —
+      // the "search" section, as opposed to the "filter" section) must not be
+      // narrowed to open states: CCRS#1367 — once GRO rejects a complaint it
+      // moves to the terminal REJECTED state, which isn't in OPEN_STATES/
+      // allStates, so a CSR searching that exact complaint number got zero
+      // results back even though the complaint still needs to be reopened.
+      const hasExplicitSearch = Boolean(
+        searchForm.complaintNumber || searchForm.mobileNumber || (requestDate?.startDate && requestDate?.endDate)
```
vinothrallapalli-eGov: **[CONFIRMED — correctness]** Counting a bare **date range** as an "explicit search" drops the open/actionable-states default for a broad filtering flow, flooding the inbox with closed/resolved/rejected/cancelled rows.

**Failure scenario:** An operator picks the date-range preset "This Month" (a routine workload-review action — no complaint number or mobile number) with no status checkboxes. `hasExplicitSearch` becomes true, so `params.applicationStatus` is omitted entirely and pgr-services returns every state. The inbox — which per the in-code rationale defaults to open states because it is otherwise "dominated by closed/resolved/rejected/cancelled rows that nobody can act on" — now shows exactly those non-actionable rows in both MY and All tabs, and pagination/counts are dominated by dead complaints; the operator must manually check every open-state checkbox to restore the previous behaviour.

A date range refines a *browse*; unlike a complaint number it doesn't identify a specific complaint. Suggest limiting the status bypass to `complaintNumber` (and possibly `mobileNumber` — see separate comment).

vinothrallapalli-eGov: **[CONFIRMED — correctness]** Mobile-number-only searches lost the open-state scoping: a mobile number identifies a **citizen**, not a complaint, so dropping the status default returns that citizen's entire multi-year complaint history instead of their currently-actionable complaints.

**Failure scenario:** A CSR takes a call and enters the caller's mobile number to pull up their pending complaint. A repeat complainant with e.g. 40 historical closed/resolved complaints and 1 open one now gets all 41 back, and with the default sla-ascending sort the long-expired closed rows fill page 1 — the CSR must paginate or manually re-check status filters to find the one open complaint they could previously see immediately.

digit-ui-esbuild/products/pgr/src/configs/UICustomizations.js
```diff
@@ -1652,12 +1652,22 @@ export const UICustomizations = {
...
-      // Status scope: an explicit filter wins; otherwise both tabs default to
-      // every open/actionable state. The tabs differ on the ASSIGNEE axis, not
-      // the status axis (PO decision 2026-07-15).
+      // A targeted lookup (complaint number / mobile number / date range —
+      // the "search" section, as opposed to the "filter" section) must not be
+      // narrowed to open states: CCRS#1367 — once GRO rejects a complaint it
+      // moves to the terminal REJECTED state, which isn't in OPEN_STATES/
+      // allStates, so a CSR searching that exact complaint number got zero
+      // results back even though the complaint still needs to be reopened.
+      const hasExplicitSearch = Boolean(
```
vinothrallapalli-eGov: **[CONFIRMED — cleanup]** `hasExplicitSearch` re-derives conditions already materialized into `params` just above; testing the params themselves (`Boolean(params.serviceRequestId || params.mobileNumber || params.fromDate)`) would be simpler and drift-proof, instead of duplicating the `searchForm.complaintNumber` / `searchForm.mobileNumber` / `requestDate?.startDate && requestDate?.endDate` predicates (the date-range predicate is copy-pasted verbatim from line 1618).

**Risk:** The set of 'targeted search' fields is now encoded in two places that must stay in sync. When a future search-section field (e.g. landline or email) is added, the developer sets its param in the search-form block but misses the `hasExplicitSearch` duplicate; that field's searches silently keep the OPEN_STATES scoping and return zero results for closed/REJECTED complaints — the exact CCRS#1367 bug recurs for the new field.

vinothrallapalli-eGov (commented):
Automated high-effort review (Claude Code workflow: 4 independent finders + adversarial verification per finding; 11 candidates verified, 0 refuted). 5 findings survived — 4 correctness, 1 cleanup. Headline: the CCRS#1367 fix is incomplete — `hasExplicitSearch` only bypasses the status default, while the assignee narrowing on the default MY tab and the server-side TEAM visibility predicate still AND onto explicit complaint-number searches; and classifying bare date-range / mobile-number searches as explicit lookups drops the open-states default for routine triage flows.

digit-ui-esbuild/products/pgr/src/configs/UICustomizations.js
Lokendra-egov: fixed

digit-ui-esbuild/products/pgr/src/configs/UICustomizations.js
Lokendra-egov: fixed

vinothrallapalli-eGov (approved):

PR #1564, opened 2026-08-04
chore(ansible): configure Docker data-root relocation and add Bomet e…

priyanshu-egov:
…xample settings

- Add docker_data_root variable to group_vars/all.yml with detailed documentation for relocating Docker/containerd storage to dedicated partitions
- Introduce docker_data_root configuration to host_vars/_example.yml with migration guidance for SELinux-aware relabeling
- Set docker_data_root: "" in bomet.yml.example with local-box context and reference to _example.yml documentation
- Add employee_module_denylist with IM module disabled (search_stack dependency)
- Add core_postal_configs for Kenya postal-code validation (5-digit pattern)
- Add footer branding URL variables (footer_logo_url, footer_bw_logo_url)
- Add db_fast_path_ack_data_wipe explicit acknowledgment flag for destructive fast-path deployments
- Add build_novu_bridge, build_pgr_services, and requires_nairobi_mdms feature flags
- Add build_novu_dashboard flag for subpath-rebased dashboard serving
- Add optional services toggles: enable_otp_services, enable_turbopass for local development flexibility
- Update ensure-node20.yml task with improved Node.js version detection and caching
- Update playbook-deploy.yml with Bomet-specific task sequencing and error handling
- Relocate containerd configuration to sibling directory when docker_data_root is set, with automatic SELinux relabeling on RHEL targets
- Prevents mid-deploy "no space left on device" failures on boxes with small root filesystems pulling ~30 container images

## Summary by CodeRabbit

* **New Features**
  * Added support for deploying on Debian, RHEL, and macOS systems.
  * Added optional relocation of Docker and container storage to a dedicated partition.
  * Added RHEL-specific Docker, firewall, SELinux, nginx, and static-content configuration.
  * Added OS-specific setup for Node.js, Python, Chromium, and integration testing tools.
  * Added deployment options for branding, postal-code validation, UI modules, optional services, and CI workflows.

* **Documentation**
  * Documented storage relocation requirements and migration considerations.

### DIFF
--- a/local-setup/ansible/inventory/group_vars/all.yml
+++ b/local-setup/ansible/inventory/group_vars/all.yml
@@ -19,3 +19,16 @@ configurator_www_dir: "{{ '/var/www/configurator' if (ansible_system | default('
 # Host port the macOS nginx container publishes (maps to its :80). Linux is
 # unaffected (host nginx binds 80/443 directly).
 nginx_host_port: 80
+
+# Optional: move Docker's data-root (images + named volumes — the bulk of
+# disk usage) off the root filesystem onto a dedicated partition, e.g.
+# "/opt/docker". Empty = Docker's default (/var/lib/docker). Set this on any
+# box where `/` is small and a larger partition (commonly /opt) is available
+# — a fresh RHEL box pulling the ~30 images in this stack onto a small `/`
+# reliably fills it mid-deploy. Setting this ALSO relocates containerd's own
+# root (a separate config file Docker's data-root does not control — the
+# image layers actually live there) to a sibling "containerd" directory, e.g.
+# "/opt/containerd". Only takes effect on a box where Docker/containerd have
+# never been started before — see local-setup/ansible/README.md for the
+# migration steps if they're already populated on the root filesystem.
+docker_data_root: ""
--- a/local-setup/ansible/playbook-deploy.yml
+++ b/local-setup/ansible/playbook-deploy.yml
@@ -665,7 +665,21 @@
                 - ansible_facts['os_family'] == 'RedHat'
                 - _selinux_enforced
 
-            - name: "SELinux — restorecon the relocated storage dirs (RHEL family)"
+            # Expensive: walks every file individually (read + possibly
+            # rewrite its security.selinux xattr). Only needed the FIRST time
+            # the relocation happens — the semanage equivalence above makes
+            # new files Docker/containerd create afterward inherit the
+            # correct label automatically (default SELinux file-creation
+            # behavior: a new file/dir inherits its parent's context absent
+            # an explicit type-transition rule, and the parent here is
+            # already correctly labeled after this one pass). Re-running
+            # restorecon on every subsequent deploy against an
+            # already-relocated, already-populated data-root just re-walks
+            # a now-large Docker/containerd tree for nothing — on a
+            # multi-GB overlay2 store this has been observed to take
+            # 15-20+ minutes on IOPS-limited cloud block storage, blocking
+            # every redeploy for no benefit.
+            - name: "SELinux — restorecon the relocated storage dirs (RHEL family, first relocation only)"
               ansible.builtin.command: "restorecon -RF {{ item }}"
               loop:
                 - "{{ _docker_data_root }}"
@@ -674,6 +688,7 @@
               when:
                 - ansible_facts['os_family'] == 'RedHat'
                 - _selinux_enforced
+                - not (_relocation_already_applied | bool)
 
             - name: "Restart containerd if its config changed"
               ansible.builtin.service:
--- a/local-setup/ansible/playbook-deploy.yml
+++ b/local-setup/ansible/playbook-deploy.yml
@@ -185,25 +185,23 @@
 
   tasks:
     # ==================== Preflight: target OS ====================
-    # The playbook installs Docker + nginx + node + python via apt and
-    # manages docker.service via systemd. Both assume an Ubuntu-family
-    # target. Fail fast on Mac/RHEL/etc with a clear message rather than
-    # halfway through with `apt: command not found` or systemd errors.
+    # The playbook installs Docker + nginx + node + python and manages
+    # docker.service via systemd. Every apt/dnf-package task below is
+    # branched on ansible_facts['os_family'] — Debian (apt) or RedHat
+    # (dnf: Rocky/Alma/CentOS Stream/RHEL). Fail fast on anything else
+    # (Alpine, etc) with a clear message rather than halfway through with
+    # a "module not found" or systemd error.
     #
-    # Mac operators: spin up an Ubuntu 24.04 VM (UTM, OrbStack, Multipass,
-    # Lima, or a remote VPS) and point `ansible_host` in your host_vars
-    # at that VM. The Mac itself runs only ansible-playbook + ssh.
-    # Supported targets: Debian-family (Ubuntu) for full host provisioning,
-    # and macOS (Darwin) via the thin Mac path — Docker Desktop/OrbStack
+    # Mac operators: see the thin Mac path below — Docker Desktop/OrbStack
     # already provides Docker, nginx runs as a container, and the
-    # apt/systemd/host-nginx tasks are skipped (when: ansible_system != 'Darwin').
-    # Anything else (RHEL, Alpine, …) is genuinely unsupported.
-    - name: "preflight — target OS must be Debian-family or macOS"
+    # apt/dnf/systemd/host-nginx tasks are skipped (when: ansible_system != 'Darwin').
+    - name: "preflight — target OS must be Debian-family, RHEL-family, or macOS"
       ansible.builtin.fail:
         msg: |
           Target host {{ inventory_hostname }} is {{ ansible_facts['os_family'] }}
           ({{ ansible_facts['distribution'] | default('?') }} {{ ansible_facts['distribution_version'] | default('?') }}).
-          This playbook supports Ubuntu/Debian (full host provisioning) and
+          This playbook supports Ubuntu/Debian (apt), RHEL-family (dnf: Rocky
+          Linux/AlmaLinux/CentOS Stream/RHEL) for full host provisioning, and
           macOS (thin Mac path — containerized nginx, Docker Desktop/OrbStack).
 
           On macOS, in inventory/host_vars/<tenant>.yml set:
@@ -212,7 +210,7 @@
           and ensure Docker Desktop or OrbStack is running, then run
           `./deploy.sh <tenant>` from the Mac terminal.
       when:
-        - ansible_facts['os_family'] != 'Debian'
+        - ansible_facts['os_family'] not in ['Debian', 'RedHat']
         - ansible_system != 'Darwin'
 
     # common-masters.MobileNumberValidation (utilities/default-data-handler)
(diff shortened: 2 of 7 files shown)


local-setup/ansible/playbook-deploy.yml
```diff
@@ -567,7 +774,9 @@
                   {{ {'log-driver': 'json-file',
                       'log-opts': {'max-size': docker_log_max_size,
                                    'max-file': docker_log_max_file_effective | string},
-                      'insecure-registries': insecure_registries | default([])} }}
+                      'insecure-registries': insecure_registries | default([])}
+                     | combine({'data-root': docker_data_root}
```
vinothrallapalli-eGov: **[CONFIRMED — correctness, most severe]** The data-root merge into `/etc/docker/daemon.json` (followed by the existing "Restart docker if daemon.json changed" task) has **no guard that `/var/lib/docker` is empty**, despite the docs claiming the feature "only takes effect on a never-started box" — nothing enforces that precondition.

**Failure scenario:** Operator adds `docker_data_root: /opt/docker` to the host_vars of an already-live tenant and re-runs `deploy.sh`: daemon.json changes, docker restarts pointed at an empty `/opt/docker`, all existing images and named volumes (**including the Postgres complaint database**) become invisible, and the stack re-initializes from scratch — effective production data loss with no warning.

Suggest a precondition task: fail (or skip the merge) when `docker_data_root` is set, differs from the current effective data-root, and `/var/lib/docker` is non-empty.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -390,15 +393,219 @@
...
+              ansible.builtin.lineinfile:
+                path: /etc/containerd/config.toml
+                regexp: '^\s*root\s*='
+                line: 'root = "{{ _containerd_root }}"'
+              register: containerd_root_config
+
+            # semanage errors if the exact equivalence already exists (e.g. a
+            # re-run) — that's not a real failure, just already-done.
+            - name: "SELinux — label the relocated storage dirs like the originals (RHEL family)"
+              ansible.builtin.command: "semanage fcontext -a -e {{ item.orig }} {{ item.new }}"
```
vinothrallapalli-eGov: **[CONFIRMED — correctness]** This `semanage fcontext` command runs before `policycoreutils-python-utils` (which provides `semanage`) is installed — that package is only installed by the later "SELinux — allow nginx to proxy" block, and it is not preinstalled on minimal RHEL-family cloud images.

**Failure scenario:** Fresh minimal Rocky/Alma/RHEL box with SELinux enforcing and `docker_data_root` set in host_vars: deploy.sh reaches "SELinux — label the relocated storage dirs", the shell cannot find `semanage` (rc 127, stderr lacks 'already defined'), `failed_when` evaluates true, and the whole deploy aborts before Docker storage is ever relocated.

vinothrallapalli-eGov (commented):
Automated high-effort review (Claude Code workflow: 4 independent finders + adversarial verification per finding; 26 candidates verified, 3 refuted). 10 findings survived — all CONFIRMED. The severe cluster: the docker data-root relocation can silently orphan a live tenant's Docker data (effective data loss); fresh RHEL deploys abort because `semanage` runs before its package is installed and nginx is never started/enabled; and redeploys abort because the `failed_when` tolerance string doesn't match the actual re-run error. Plus a trailing-slash bug in the containerd-root derivation, Playwright RHEL deps missing fonts and the integration-tests runner path, triplicated NodeSource logic, a dead `changed_when`, and a dangling README reference.

(Three candidates about the RHEL stock-default-site removal were refuted during verification.)

local-setup/ansible/playbook-deploy.yml
```diff
@@ -390,15 +393,219 @@
...
+              loop:
+                - { orig: /var/lib/docker, new: "{{ docker_data_root }}" }
+                - { orig: /var/lib/containerd, new: "{{ _containerd_root }}" }
+              register: _semanage_fcontext
+              failed_when: >-
+                _semanage_fcontext.rc != 0
+                and 'already defined' not in (_semanage_fcontext.stderr | default(''))
```
vinothrallapalli-eGov: **[CONFIRMED — correctness]** This `failed_when` only tolerates stderr containing `'already defined'`, but re-adding an fcontext **equivalence** (`semanage fcontext -a -e`) fails with "Equivalence class for <path> already exists" — a different string — so the task's own re-run tolerance never matches.

**Failure scenario:** Second `./deploy.sh <tenant>` run on a RHEL box with `docker_data_root` set: semanage exits non-zero with "Equivalence class for /opt/docker already exists", the 'already defined' check does not match, `failed_when` trips, and an otherwise-idempotent redeploy hard-fails at the SELinux labeling task. Add `'already exists' in ...stderr` to the tolerance (or check both strings).

local-setup/ansible/playbook-deploy.yml
```diff
@@ -2370,15 +2599,43 @@
...
           when: not (nginx_preserve_vhost | default(false))
 
-        - name: "Host nginx — flush handlers so reload happens before validation"
-          ansible.builtin.meta: flush_handlers
+    - name: "Host nginx — Linux, RHEL family"
+      when:
+        - ansible_system != "Darwin"
+        - ansible_facts['os_family'] == 'RedHat'
+      block:
+        - name: "Host nginx — install if missing (RHEL family)"
```
vinothrallapalli-eGov, line 2891: **[CONFIRMED — correctness]** The new RHEL host-nginx block installs nginx via dnf but never **starts or enables** the service; unlike apt on Debian, RPM installs leave `nginx.service` dead/disabled, and the only service interaction is the "Reload nginx" handler which uses `state: reloaded`.

**Failure scenario:** First deploy on any RHEL-family box: after rendering the vhost, `flush_handlers` runs `systemctl reload nginx` against an inactive unit, which fails ("nginx.service is not active, cannot reload") and aborts the play; even if it were ignored, nginx would stay stopped and disabled, so the site never serves and dies again on reboot. Add a `systemd: name=nginx state=started enabled=true` task after the dnf install.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -390,15 +393,219 @@
...
+        # Without this, a fresh box with a small `/` and a larger `/opt`
+        # fills `/` mid-deploy pulling the ~30 images in this stack and dies
+        # with "no space left on device". Must run before containerd/docker
+        # are ever started — relocating an already-populated /var/lib/docker
+        # or /var/lib/containerd needs a stop+rsync+relabel migration instead
+        # (not handled here; do that by hand first if Docker's already run).
+        - name: "Docker/containerd storage relocation (when docker_data_root is set)"
+          when: (docker_data_root | default('') | length) > 0
+          vars:
+            _containerd_root: "{{ docker_data_root | dirname }}/containerd"
```
vinothrallapalli-eGov: **[CONFIRMED — correctness]** `_containerd_root` is derived as `docker_data_root | dirname ~ '/containerd'`, which breaks when `docker_data_root` has a trailing slash: `dirname('/opt/docker/')` is `/opt/docker`, nesting containerd's root **inside** Docker's data-root instead of the documented sibling directory (and `dirname('/docker')` yields `/`, producing `//containerd`).

**Failure scenario:** host_vars sets `docker_data_root: "/opt/docker/"`: containerd's root becomes `/opt/docker/containerd` inside Docker's own data-root, the SELinux equivalence is registered for that nested path instead of the documented `/opt/containerd` sibling, and Docker manages a data-root containing a foreign containerd tree — divergence from the documented layout with mislabeled/misplaced image-layer storage. Normalize first, e.g. `docker_data_root | regex_replace('/+$', '')`.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -4014,8 +4314,40 @@
       changed_when: false
 
       when: run_ci_tests | default(false) and (run_playwright_tests | default(false))
+    # Playwright's `--with-deps` OS-dependency installer only knows how to
+    # drive apt (Ubuntu/Debian) — on RHEL family it no-ops with a warning
+    # and leaves Chromium's shared libs missing, so the browser fails to
+    # launch later with a cryptic "error while loading shared libraries".
+    # Install the equivalent RPM packages ourselves first, then run
+    # `playwright install chromium` WITHOUT --with-deps on that family.
+    - name: "Install Chromium runtime dependencies (RHEL family, Playwright has no dnf support)"
```
vinothrallapalli-eGov, line 4602: **[CONFIRMED — correctness]** The replaced `npx playwright install --with-deps` also installed **font packages** (fonts-liberation, fonts-noto-color-emoji, fonts-unifont) on Debian; this RHEL replacement dnf list re-establishes only shared libraries and installs no font package at all.

**Failure scenario:** On a minimal Rocky/RHEL server with no font packages installed, `run_ci_tests` + `run_playwright_tests` deploys launch Chromium with zero system fonts: every rendered page draws text as blank/tofu glyphs, so Playwright failure screenshots, trace artifacts, and any screenshot/visual assertions produce unreadable or mismatching output. Add e.g. `liberation-fonts` and `google-noto-emoji-color-fonts` (or the Rocky equivalents) to the dnf list.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -4014,8 +4314,40 @@
       changed_when: false
 
       when: run_ci_tests | default(false) and (run_playwright_tests | default(false))
+    # Playwright's `--with-deps` OS-dependency installer only knows how to
```
vinothrallapalli-eGov: **[CONFIRMED — correctness]** This RHEL Playwright-deps fix only covers the e2e-tests path; the **integration-tests runner** (task at line 2526, outside this diff) still runs `playwright install --with-deps chromium` unconditionally on Linux, which — per this very comment — no-ops on RHEL and leaves Chromium's shared libraries missing.

**Failure scenario:** On a RHEL-family box with `enable_integration_tests_runner: true`, Chromium libs are never installed and every dashboard-triggered test run dies with "error while loading shared libraries" — the exact failure this PR fixed for the CI path but not there. Suggest moving the RPM dep list into a shared include (e.g. `tasks/ensure-chromium-deps.yml`) used by both call sites.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -3591,7 +3868,22 @@
       changed_when: "'nodejs is already' not in nodejs_install.stdout"
       when:
         - install_claude_code | default(false)
-        - ansible_system != "Darwin"
+        - ansible_facts['os_family'] == 'Debian'
+
+    - name: Install Node.js 20 LTS (needed for CC and tests, RHEL family)
```
vinothrallapalli-eGov: **[CONFIRMED — cleanup]** This inline "Install Node.js 20 LTS (RHEL family)" shell task re-implements `tasks/ensure-node20.yml`, which this same PR already extended with the identical rpm.nodesource.com/dnf logic.

The NodeSource-install logic now exists in **three copies** (ensure-node20.yml Debian+RHEL tasks, plus the two inline Claude-Code tasks at lines 3858/3873) with already-divergent version regexes (`^v2[02]` inline vs `^v(2[0-9]|[3-9][0-9])` in the helper) — a future Node bump will miss one copy. The inline pair could be a single `include_tasks: tasks/ensure-node20.yml` with `node_ensure_label`, like the 5 existing call sites (lines 1318, 1655, 1811, 2201, 2390).

local-setup/ansible/playbook-deploy.yml
```diff
@@ -3591,7 +3868,22 @@
...
+      shell: |
+        set -o pipefail
+        if ! node --version 2>/dev/null | grep -qE "^v2[02]"; then
+          curl -fsSL https://rpm.nodesource.com/setup_20.x | bash -
+          dnf install -y nodejs
+        fi
+      args:
+        executable: /bin/bash
+      register: nodejs_install_rhel
+      changed_when: "'nodejs is already' not in nodejs_install_rhel.stdout"
```
vinothrallapalli-eGov: **[CONFIRMED — cleanup]** `changed_when: "'nodejs is already' not in stdout"` was copy-pasted from the apt task, but dnf never prints that phrase (it says "Nothing to do" / "already installed"), and when node>=20 the script prints nothing at all.

**Effect:** On every RHEL re-deploy with `install_claude_code: true` the task reports `changed` even when it did nothing (empty stdout ⇒ condition true), polluting the play recap and defeating changed-based tooling. The version-probe gating in `ensure-node20.yml` (skip, rather than changed_when string-matching) is the correct mechanism.

local-setup/ansible/inventory/group_vars/all.yml
```diff
@@ -19,3 +19,16 @@ configurator_www_dir: "{{ '/var/www/configurator' if (ansible_system | default('
...
+# Optional: move Docker's data-root (images + named volumes — the bulk of
+# disk usage) off the root filesystem onto a dedicated partition, e.g.
+# "/opt/docker". Empty = Docker's default (/var/lib/docker). Set this on any
+# box where `/` is small and a larger partition (commonly /opt) is available
+# — a fresh RHEL box pulling the ~30 images in this stack onto a small `/`
+# reliably fills it mid-deploy. Setting this ALSO relocates containerd's own
+# root (a separate config file Docker's data-root does not control — the
+# image layers actually live there) to a sibling "containerd" directory, e.g.
+# "/opt/containerd". Only takes effect on a box where Docker/containerd have
+# never been started before — see local-setup/ansible/README.md for the
```
vinothrallapalli-eGov, line 32: **[CONFIRMED — documentation]** This comment tells operators to "see local-setup/ansible/README.md for the migration steps", but README.md contains no docker_data_root / data-root migration section (same dangling reference in `_example.yml` line 49).

**Effect:** An operator with an already-populated `/var/lib/docker` follows the comment, finds nothing in the README, and either sets `docker_data_root` anyway — orphaning existing images/volumes (see the data-loss finding on playbook-deploy.yml:778) — or has to reverse-engineer the stop+rsync+semanage+restorecon procedure the comment promises is documented. Either add the migration section or drop the reference.

PR #1568, opened 2026-08-04
Ansible docker data root rhel support

priyanshu-egov:
## Summary by CodeRabbit

* **New Features**
  * Local deployment setup now supports Red Hat-based Linux systems in addition to Debian and macOS.
  * Docker and containerd data can be relocated to a dedicated storage partition.
  * Deployment configuration now supports optional integration tests, service toggles, UI module visibility, branding, and regional validation settings.
  * Nginx, Chromium, Playwright, and Node.js setup is now handled across supported operating systems.

* **Documentation**
  * Added guidance for relocating existing container storage and configuring the new deployment options.

### DIFF
--- a/local-setup/ansible/inventory/host_vars/bomet.yml.example
+++ b/local-setup/ansible/inventory/host_vars/bomet.yml.example
@@ -41,6 +41,12 @@ docker_default_platform: ''
 domain: localhost
 tls_enabled: false
 insecure_registries: []
+# Empty = Docker's default (/var/lib/docker on the root filesystem) — fine
+# here since this is a single-partition local box. Set to e.g. "/opt/docker"
+# on a target where `/` is small and a larger partition (commonly /opt) is
+# available; see the docker_data_root comment in inventory/host_vars/_example.yml
+# for what it does (also relocates containerd's root, relabels for SELinux).
+docker_data_root: ""
 
 # ── identity / tenancy ──
 # Kit shape was bare `ke` everywhere, but the configurator wizard created the
@@ -55,6 +61,11 @@ tenant_id: ke
 ui_state_tenant_id: ke.bomet
 nginx_preserve_vhost: false        # kit: true — local box has no hand-managed vhost to preserve
 login_tenant_allowlist: []
+
+# Hide these modules in the employee UI. IM (inbox-v2) needs the search
+# stack, which is off on this box (enable_search_stack: false below).
+employee_module_denylist:
+  - IM
 
 # ── geography / boundary model (Bomet County) ──
 map_center: { lat: -0.7817, lng: 35.3428 }
@@ -74,6 +85,20 @@ core_mobile_configs:
   countryCode: '+254'
   mobileNumberRegex: ^[17][0-9]{8}$
 
+# Postal-code validation (SPA globalConfigs CORE_POSTAL_CONFIGS). Kenya: 5
+# digits. Empty/absent falls back to this same pattern anyway, but set it
+# explicitly so it's not relying on the hardcoded legacy default.
+core_postal_configs:
+  postalCodePattern: "^[0-9]{5}$"
+  postalCodeLength: 5
+  postalCodeErrorMessage: "CS_COMPLAINT_POSTALCODE_INVALID_ERROR"
+
+# ── Branding ──
+# Footer logos shown in citizen + employee UI. Use absolute URLs; copied to
+# /var/www/brand/ when nginx_features.brand_assets: true (off below).
+footer_logo_url: ""
+footer_bw_logo_url: ""
+
 # ── feature toggles (exact Bomet shape, except MCP — see header) ──
 asset_s3_bucket: pg-egov-assets
 digit_ui_mode: container
@@ -82,11 +107,26 @@ enable_mcp: true                   # kit: false — needed locally to bootstrap 
 build_mcp: true                    # kit: false — build digit-mcp:local like the mz bring-up did
 enable_novu: true                  # config-driven notifications target: Novu
 db_fast_path: true                 # REQUIRED — see kit NON-PERSISTED-STATE.md / DB section
+# New in _example.yml: scripts/preflight.py refuses db_fast_path: true without this
+# explicit ack (deploy.sh would fail preflight otherwise). Set only after confirming
+# this box has no anonymous-volume postgres data worth keeping.
+db_fast_path_ack_data_wipe: true
 build_otp_publisher: true          # not in kit: real Bomet box already has otp-publisher:local;
                                    # the notifications profile needs it, so build it here
 build_digit_ui: true
 build_configurator: true
 tolerate_bootstrap_failures: true  # bootstrap is partial-success tolerant (see kit GAPS)
+
+# Build the pass-through novu-bridge / config-driven pgr-services from source
+# on the box instead of pulling a published/preview image. Off by default,
+# same as _example.yml — this tenant currently uses the published images
+# (pgr_services_image / novu_bridge_image below).
+build_novu_bridge: false
+build_pgr_services: false
+
+# Init the nairobi-mdms git submodule on the controller. Only needed if this
+# tenant's onboarding uses those JSONs — not the case here.
+requires_nairobi_mdms: false
 
 # ── Config-driven notifications (design D1/D4/D6/D7/D8) ──
 # Bomet is the FIRST tenant cut over. PGR reads PGR.NotificationRouting +
@@ -95,6 +135,11 @@ tolerate_bootstrap_failures: true  # bootstrap is partial-success tolerant (see 
 # docker-compose.bomet.yml ALSO hard-sets PGR_NOTIFICATION_CONFIG_DRIVEN=true.
 pgr_notification_config_driven: true
 novu_bridge_channel: "sms"         # SMS-first; add WHATSAPP/EMAIL as config-only routing edits
+
+# Build a subpath-rebased Novu dashboard image so the SPA serves cleanly
+# under /novu. Not required here (enable_keycloak is false below, so no
+# domain-sharing conflict yet) — matches _example.yml's off-by-default.
+build_novu_dashboard: false
 
 # Custom image pins (built+pushed to the VPC registry by CI). Replace <sha>
 # with the actual CI tags once the Phase 1/2 images are built. Until then the
@@ -109,6 +154,27 @@ twilio_account_sid: ""
 twilio_auth_token: ""
 twilio_whatsapp_from: "whatsapp:+14155238886"
 
+# ── Optional services (opt-in via flags; all off, matching _example.yml) ──
+# Real OTP stack (egov-otp/user-otp/egov-notification-sms). Off keeps the
+# local kong.yml mock (fixed OTP 123456) — fine for this local box.
+enable_otp_services: false
+# Turbopass (OSM location autocomplete) — standalone on-box container.
+enable_turbopass: false
+# Self-hosted Overpass (boundary-polygon fetch) — falls back to public
+# overpass-api.de when off.
+enable_overpass: false
+
+# ── digit-integration-tests dashboards (off — not used on this box) ──
+enable_integration_tests: false
+integration_tests_dashboard_base: "/tests-v2/"
+integration_tests_basic_auth_user: "digit-tests"
+enable_integration_tests_runner: false
+
+# ── CI / lifecycle ──
+run_ci_tests: false                # post-deploy XLSX dataloader + Playwright suite
+force_clean: false                 # tear down + redeploy from scratch (NEVER on live data)
+install_claude_code: false         # installs Claude Code CLI on the target
+
 nginx_features:
   brand_assets: false
   configurator: true
@@ -118,6 +184,8 @@ nginx_features:
   mcp: true                        # kit: false — paired with enable_mcp above
   keycloak: true
   digit_ui_v2: true
+  integration_tests: false         # pair with enable_integration_tests above
+  integration_tests_runner: false  # pair with enable_integration_tests_runner above
 
 # ── auth (Keycloak) ──
 enable_keycloak: false
@@ -129,6 +197,7 @@ auth_provider: ""
 keycloak_client_id: digit-ui
 enable_digit_ui_v2: true
 digit_ui_esbuild_branch: feat/keycloak-auth-adapter
+digit_ui_v2_repo: "https://github.com/ChakshuGautam/Citizen-Complaint-Resolution-System.git"
 digit_ui_v2_branch: feat/digit-ui-v2-keycloak
 keycloak_google_client_id: ''      # kit: CHANGE_ME — blank skips Google IdP wiring
 
--- a/local-setup/ansible/inventory/host_vars/bomet.yml.example
+++ b/local-setup/ansible/inventory/host_vars/bomet.yml.example
@@ -86,12 +86,13 @@ core_mobile_configs:
   mobileNumberRegex: ^[17][0-9]{8}$
 
 # Postal-code validation (SPA globalConfigs CORE_POSTAL_CONFIGS). Kenya: 5
-# digits. Empty/absent falls back to this same pattern anyway, but set it
-# explicitly so it's not relying on the hardcoded legacy default.
+# digits. The pattern is the single knob — length and the localized error
+# message are derived from it (postalCodeLength / postalCodeErrorMessage
+# were removed in v2.12). Empty/absent falls back to this same pattern
+# anyway, but set it explicitly so it's not relying on the hardcoded
+# legacy default.
 core_postal_configs:
   postalCodePattern: "^[0-9]{5}$"
-  postalCodeLength: 5
-  postalCodeErrorMessage: "CS_COMPLAINT_POSTALCODE_INVALID_ERROR"
 
 # ── Branding ──
 # Footer logos shown in citizen + employee UI. Use absolute URLs; copied to
--- a/local-setup/ansible/playbook-deploy.yml
+++ b/local-setup/ansible/playbook-deploy.yml
@@ -665,7 +665,21 @@
                 - ansible_facts['os_family'] == 'RedHat'
                 - _selinux_enforced
 
-            - name: "SELinux — restorecon the relocated storage dirs (RHEL family)"
+            # Expensive: walks every file individually (read + possibly
+            # rewrite its security.selinux xattr). Only needed the FIRST time
+            # the relocation happens — the semanage equivalence above makes
+            # new files Docker/containerd create afterward inherit the
+            # correct label automatically (default SELinux file-creation
+            # behavior: a new file/dir inherits its parent's context absent
+            # an explicit type-transition rule, and the parent here is
+            # already correctly labeled after this one pass). Re-running
+            # restorecon on every subsequent deploy against an
+            # already-relocated, already-populated data-root just re-walks
+            # a now-large Docker/containerd tree for nothing — on a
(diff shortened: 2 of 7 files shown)


local-setup/ansible/inventory/host_vars/bomet.yml.example
```diff
@@ -74,6 +85,20 @@ core_mobile_configs:
   countryCode: '+254'
   mobileNumberRegex: ^[17][0-9]{8}$
 
+# Postal-code validation (SPA globalConfigs CORE_POSTAL_CONFIGS). Kenya: 5
+# digits. Empty/absent falls back to this same pattern anyway, but set it
+# explicitly so it's not relying on the hardcoded legacy default.
+core_postal_configs:
+  postalCodePattern: "^[0-9]{5}$"
+  postalCodeLength: 5
```
vinothrallapalli-eGov: @priyanshu-egov remove the postal code length and message variables

local-setup/ansible/playbook-deploy.yml
```diff
@@ -390,15 +406,284 @@
...
+            - name: "Guard — check /var/lib/docker and /var/lib/containerd are still empty"
+              ansible.builtin.command: "find {{ item }} -mindepth 1 -maxdepth 1"
+              loop:
+                - /var/lib/docker
+                - /var/lib/containerd
+              register: _live_dataroot_check
+              changed_when: false
+              failed_when: false
+
+            - name: "Guard — refuse to relocate an already-populated data-root"
```
vinothrallapalli-eGov: **[CONFIRMED — correctness, most severe]** First deploy with `docker_data_root` starts dockerd on the default `/var/lib/docker` **before** daemon.json is written, so every re-run trips this already-populated guard and hard-fails.

**Failure scenario:** Fresh Rocky 9 box, host_vars sets `docker_data_root: /opt/docker`. Run 1 passes the guard (dirs empty), but 'Ensure docker daemon is enabled + running' (line 614) starts dockerd with the default data-root — daemon.json with `data-root` is only written ~250 tasks later (line 833) — so dockerd's first start populates `/var/lib/docker` (overlay2/, image/, volumes/, engine-id, ...). Run 1 finishes and the stack works from `/opt/docker`. Any subsequent `./deploy.sh <tenant>` (routine re-deploy, config change) now fails at 'Guard — refuse to relocate an already-populated data-root', telling the operator to stop docker and rsync by hand — even though `/var/lib/docker` only holds the throwaway first-start skeleton. **Deploys of that tenant are permanently broken** until someone manually empties `/var/lib/docker`.

Fix options: write daemon.json (or at least the data-root key) before the first docker start, or skip the guard when daemon.json already points at `docker_data_root` and it is populated.

vinothrallapalli-eGov (commented):
Automated high-effort review (Claude Code workflow: 4 independent finders + adversarial verification per finding; 19 candidates verified, 1 refuted). 10 findings survived — 9 CONFIRMED, 1 PLAUSIBLE. Headline: the new data-root relocation guard cross-fires with task ordering — on RHEL every re-deploy after the first trips it, and on Debian/Ubuntu even the first deploy fails because the deb postinst auto-starts docker before the guard runs. Two more RHEL bring-up breaks: a font package name that doesn't exist on EL10, and preflight now admitting RHEL boxes without rsync/git/tmux ever being installed. Plus the surviving //containerd edge case, a plausible Chromium dependency gap vs --with-deps, and four duplication/dead-code cleanups.

(Refuted during verification: NodeSource does publish a working EL10 path, so the RHEL Node branch is fine.)

local-setup/ansible/playbook-deploy.yml
```diff
@@ -390,15 +406,284 @@
...
+          block:
+            # Relocating a data-root that Docker/containerd have already
+            # written to would point daemon.json/config.toml at an empty
+            # directory: every existing image and named volume (including the
+            # Postgres complaint database) becomes invisible and the stack
+            # silently re-initializes from scratch. Only a fresh/never-started
+            # box may set docker_data_root; an already-populated /var/lib/docker
+            # or /var/lib/containerd needs a stop+rsync+relabel migration done
+            # by hand first.
+            - name: "Guard — check /var/lib/docker and /var/lib/containerd are still empty"
```
vinothrallapalli-eGov: **[CONFIRMED — correctness]** On Debian/Ubuntu the docker-ce/containerd.io **deb postinst auto-starts docker and containerd during package install**, which runs before this relocation guard — so `docker_data_root` fails even on a genuinely fresh Ubuntu box.

**Failure scenario:** Fresh Ubuntu 24.04 box, host_vars sets `docker_data_root: /opt/docker` (the var is documented in group_vars/all.yml and _example.yml as applying to any box, not RHEL-only). 'Install Docker Engine (Debian/Ubuntu)' (line 399) runs first; Debian packaging starts containerd.service and docker.service in postinst, populating `/var/lib/containerd` and `/var/lib/docker` before the 'Guard — check ... still empty' task ever executes. The very first deploy then aborts at the guard, telling the operator to do a manual stop+rsync migration on a box Docker was never used on — the feature is unusable on the Debian family the playbook primarily targets.

Common mitigation: write daemon.json before installing docker-ce, or mask the postinst auto-start (`systemd mask` / `policy-rc.d`) until daemon.json is in place.

local-setup/ansible/tasks/ensure-chromium-deps.yml
```diff
@@ -0,0 +1,32 @@
...
+      - libXcomposite
+      - libXdamage
+      - libXfixes
+      - libXrandr
+      - mesa-libgbm
+      - alsa-lib
+      - pango
+      - cairo
+      - liberation-fonts
+      - google-noto-emoji-color-fonts
```
vinothrallapalli-eGov: **[CONFIRMED — correctness]** `google-noto-emoji-color-fonts` does not exist on EL10 (renamed `google-noto-color-emoji-fonts`), so this dnf task fails and aborts the deploy on Rocky/Alma 10.

**Failure scenario:** Deploy to Rocky Linux 10 (a platform this PR explicitly adds support for — the docker-ce.repo EL9-releasever fallback exists solely for EL10) with `run_ci_tests`+`run_playwright_tests` or `enable_integration_tests_runner` set. The dnf task fails with 'No match for argument: google-noto-emoji-color-fonts' — verified against Rocky 10 AppStream repo metadata, which ships only `google-noto-color-emoji-fonts` (Rocky 9 still has the old name) — and `state: present` treats a missing package as a hard error, so the whole run dies at the Chromium-deps step. Select the name by major version, or list both with `failed_when` tolerance / use a dnf `--skip-broken` strategy for the font.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -185,25 +185,23 @@
...
-    # at that VM. The Mac itself runs only ansible-playbook + ssh.
-    # Supported targets: Debian-family (Ubuntu) for full host provisioning,
-    # and macOS (Darwin) via the thin Mac path — Docker Desktop/OrbStack
+    # Mac operators: see the thin Mac path below — Docker Desktop/OrbStack
     # already provides Docker, nginx runs as a container, and the
-    # apt/systemd/host-nginx tasks are skipped (when: ansible_system != 'Darwin').
-    # Anything else (RHEL, Alpine, …) is genuinely unsupported.
-    - name: "preflight — target OS must be Debian-family or macOS"
+    # apt/dnf/systemd/host-nginx tasks are skipped (when: ansible_system != 'Darwin').
+    - name: "preflight — target OS must be Debian-family, RHEL-family, or macOS"
```
vinothrallapalli-eGov, line 210: **[CONFIRMED — correctness]** The removed preflight rejection of RHEL guaranteed no host with missing baseline CLI tooling ever proceeded; the new RHEL support re-establishes package parity for the apt-installed set (docker, nginx, node, python3, httpd-tools) but never installs **rsync, git, or tmux** on RedHat — which Ubuntu cloud images ship but Rocky/Alma *minimal* cloud images do not.

**Failure scenario:** `./deploy.sh <tenant>` against a fresh Rocky 9 minimal cloud image (now accepted by preflight): the always-on 'Synchronize config directories' task (`ansible.posix.synchronize`, which requires rsync on the target) fails with "rsync: command not found" — or with `digit_ui_mode` static/hmr, the pre_tasks git clone of /opt/digit-ui-esbuild fails with "Failed to find required executable git", and hmr's tmux task fails too — aborting the deploy halfway through, after Docker/firewalld/SELinux provisioning has already mutated the box. That's exactly the mid-run failure mode the deleted guard's comment ("Fail fast ... rather than halfway through") existed to prevent. Add a baseline-tools dnf task (rsync, git, tmux) to the RHEL path.

local-setup/ansible/tasks/ensure-chromium-deps.yml
```diff
@@ -0,0 +1,32 @@
...
+# to launch later with a cryptic "error while loading shared libraries",
+# or launches with every page rendering blank/tofu glyphs. Install the
+# equivalent RPM packages ourselves, then have callers run
+# `playwright install chromium` WITHOUT --with-deps on this family.
+#
+# Shared by both Playwright call sites (e2e-tests CI stage and the
+# integration-tests runner) so a future dependency bump only needs to
+# happen once.
+- name: "Install Chromium runtime dependencies (RHEL family, Playwright has no dnf support)"
+  ansible.builtin.dnf:
```
vinothrallapalli-eGov, line 19: **[PLAUSIBLE — correctness]** Dropping Playwright's `--with-deps` on RedHat replaces its complete, version-matched Chromium dependency resolution with a hand-curated 17-package RPM list that omits several libraries from Playwright's canonical Chromium dependency set (e.g. libX11/libXext/libXrender, libxshmfence, at-spi2-core, libXtst, dbus-libs), relying on undocumented transitive RPM pulls (e.g. cairo→libX11) to cover the gaps.

**Failure scenario:** On a minimal Rocky/Alma 9 box, `playwright install chromium` succeeds (no --with-deps validation either), but at test time Chromium fails to launch with "error while loading shared libraries: libxshmfence.so.1 (or libX11.so.6)" if the omitted lib isn't pulled transitively on that image — the integration-tests runner daemon crash-loops or the CI Playwright stage fails every spec. The risk also silently reappears whenever a Playwright bump adds a new native dependency, since only apt targets get the auto-resolved list. Consider listing Playwright's full published RPM set explicitly, and/or adding a post-install `ldd`-based launch smoke check.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -390,15 +406,284 @@
...
+        # (not handled here; do that by hand first if Docker's already run).
+        - name: "Docker/containerd storage relocation (when docker_data_root is set)"
+          when: (docker_data_root | default('') | length) > 0
+          vars:
+            # Strip any trailing slash before deriving the sibling containerd
+            # dir — dirname('/opt/docker/') is '/opt/docker' (nests containerd
+            # INSIDE Docker's own data-root instead of beside it), and
+            # dirname('/docker') is '/' (yields the bogus '//containerd').
+            _docker_data_root: "{{ docker_data_root | regex_replace('/+$', '') }}"
+            _containerd_root: "{{ _docker_data_root | dirname }}/containerd"
```
vinothrallapalli-eGov: **[CONFIRMED — correctness edge case]** The `_containerd_root` derivation still produces the bogus `//containerd` its own comment warns about: stripping trailing slashes fixes `/opt/docker/` but not a top-level value like `/docker`, where `dirname` is `/`.

**Failure scenario:** Operator sets `docker_data_root: "/docker"` (a plausible choice for a dedicated /docker mount). `dirname('/docker')` is `/`, so config.toml gets `root = "//containerd"` and semanage/restorecon are invoked on `//containerd` — the double-slash path lands in persistent config and SELinux equivalence rules, diverging from the `/containerd` path containerd actually normalizes to, so the relabel targets a string that never matches and image-layer writes land mislabeled (AVC denials at pull time). Guard against `dirname == '/'`, or document that docker_data_root must be at least two levels deep.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -390,15 +406,284 @@
...
+                service: "{{ item }}"
+                permanent: true
+                immediate: true
+                state: enabled
+              loop: [http, https]
+
+            # Only relevant when nginx is configured to listen on a
+            # non-standard port (host_vars `nginx_host_port`); http/https
+            # above cover the default 80/443.
+            - name: "Open nginx_host_port if it isn't 80/443"
```
vinothrallapalli-eGov: **[CONFIRMED — cleanup]** This firewalld task opens `nginx_host_port` on RHEL, but `nginx_host_port` is a macOS-container-only knob (group_vars: 'Linux is unaffected (host nginx binds 80/443 directly)') and `templates/nginx-site.conf.j2` only listens on 80/443 — the task is dead code or opens an unused port.

With the default `nginx_host_port: 80` the task is permanently skipped. If an operator copies host_vars from a Mac tenant that sets `nginx_host_port: 8080`, the RHEL deploy permanently opens 8080/tcp in firewalld even though host nginx listens only on 80/443 — an unneeded permanent firewall hole. Drop the task, or gate it on the path where the port is actually bound (the Darwin container path, which never reaches firewalld).

local-setup/ansible/playbook-deploy.yml
```diff
@@ -4014,8 +4412,17 @@
...
+      ansible.builtin.include_tasks: "{{ playbook_dir }}/tasks/ensure-chromium-deps.yml"
+      when:
+        - run_ci_tests | default(false)
+        - run_playwright_tests | default(false)
+
     - name: Install Playwright Chromium with system dependencies
-      shell: 'npx playwright install{{ " --with-deps" if ansible_system != "Darwin" else "" }} chromium'
+      shell: >-
+        npx playwright install
+        {{ " --with-deps" if (ansible_system != "Darwin" and ansible_facts['os_family'] != 'RedHat') else "" }}
```
vinothrallapalli-eGov: **[CONFIRMED — cleanup]** The Playwright `--with-deps` suppression is implemented twice with two different Jinja expressions (line 2615: `"" if os_family == 'RedHat' else "--with-deps"`; here: `" --with-deps" if (ansible_system != "Darwin" and os_family != 'RedHat') else ""`), and the two ensure-chromium-deps include call sites carry inconsistent gates (2607 excludes Darwin, 4417 does not).

When a third case appears (e.g. SUSE, or Playwright gains dnf support), one call site gets updated and the other silently keeps installing without system deps — Chromium then fails at test time on only one of the two test paths. A single early fact (e.g. `_playwright_with_deps_flag`) referenced by both commands keeps the policy in one place.

local-setup/ansible/tasks/ensure-node20.yml
```diff
@@ -1,30 +1,52 @@
...
     executable: /bin/bash
   when: >-
-    _ensure_node_ver.rc != 0
-    or (_ensure_node_ver.stdout | default('') is not match('^v(2[0-9]|[3-9][0-9])\.'))
+    ansible_facts['os_family'] == 'Debian'
+    and (_ensure_node_ver.rc != 0
+    or _ensure_npm_ver.rc != 0
+    or (_ensure_node_ver.stdout | default('') is not match('^v(2[0-9]|[3-9][0-9])\.')))
+
+- name: "{{ node_ensure_label }} — install NodeSource Node 20 (RHEL family; node <20 or npm absent)"
```
vinothrallapalli-eGov: **[CONFIRMED — cleanup]** The Debian and RHEL NodeSource tasks are copy-paste twins: identical shell body except the URL host (deb. vs rpm.) and installer (apt-get vs dnf), with the three-clause version-probe `when:` duplicated verbatim in both.

This file's own history shows the cost: the playbook previously carried 'an already-divergent version regexp' from exactly this kind of duplication (per the PR's comment at playbook-deploy.yml:4033). Next time the version floor moves to Node 22, the regexp/conditions must be edited in two places; missing one silently leaves one OS family installing the wrong Node. Compute a single `_node_needs_install` set_fact from the probes and use one shell task with templated URL/package-manager (`{{ 'deb' if os_family == 'Debian' else 'rpm' }}.nodesource.com`).

local-setup/ansible/playbook-deploy.yml
```diff
@@ -390,15 +406,284 @@
...
+        # MCP, Grafana, Gatus, Novu, …) needs httpd_can_network_connect —
+        # without it every proxy_pass 502s with an AVC denial
+        # (avc: denied { name_connect } for ... scontext=...httpd_t). This
+        # is the single most common "deploy finished green but the site
+        # 502s" gap on a fresh RHEL-family box.
+        - name: "SELinux — allow nginx to proxy to backend services (RHEL family)"
+          block:
+            - name: Ensure libselinux-python bindings + policy tools are present
+              dnf:
+                name: [policycoreutils-python-utils]
```
vinothrallapalli-eGov: **[CONFIRMED — cleanup]** `policycoreutils-python-utils` is installed by two separate new tasks (relocation block line 563 and this SELinux nginx block) instead of one early RedHat+enforcing-gated task.

Duplicate dnf transactions on every RHEL deploy (the second is a metadata-check no-op but still a dnf invocation), and two task definitions to keep in sync — the line-563 comment even exists solely to explain why the duplicate is needed. A single 'SELinux tooling' task placed before the relocation block (`when: RedHat and _selinux_enforced`) removes both the duplication and the explanatory comment.

vinothrallapalli-eGov:
Addressed all review findings in 98c03b681:

| Finding | Fix |
|---|---|
| Guard cross-fires on re-runs (dockerd first start precedes daemon.json) | `data-root` is now seeded into `daemon.json` inside the relocation block, before `docker.service` ever starts; a running dockerd is restarted only if the seed changed it. The guard also skips when `daemon.json` already points at `docker_data_root` (leftovers are a first-start skeleton, not live data) — so existing boxes bitten by the old ordering recover on their next run. |
| Debian postinst auto-start populates default roots before the guard | A temporary `policy-rc.d` (exit 101) holds the docker-ce/containerd.io postinst auto-start during install (only when `docker_data_root` is set; `force: false` so an operator-managed policy-rc.d is never touched), released right after the relocation is configured. |
| `google-noto-emoji-color-fonts` missing on EL10 | Package name now selected by EL major version (`google-noto-color-emoji-fonts` on ≥10). |
| rsync/git/tmux missing on minimal Rocky/Alma images | New "Baseline CLI tooling" dnf task in pre_tasks, before the git clone and sync tasks. |
| Hand-curated RPM list omits Playwright Chromium deps | Added `libX11`, `libXext`, `libXrender`, `libXtst`, `libxshmfence`, `at-spi2-core`, `dbus-libs`; comment notes the list mirrors Playwright's published set and must be revisited on version bumps. |
| `//containerd` for top-level `docker_data_root` like `/docker` | `dirname` result of `/` is collapsed before appending `/containerd`. |
| Dead firewalld `nginx_host_port` opening | Task dropped; comment explains why the macOS-only knob must not be opened on Linux. |
| `--with-deps` policy duplicated with divergent expressions/gates | Single `_playwright_with_deps_flag` fact resolved once after the OS preflight and used at both call sites; the second ensure-chromium-deps include now carries the same Darwin gate as the first. |
| NodeSource Debian/RHEL copy-paste twins | One decision fact + one install task with the repo host and package manager templated. |
| `policycoreutils-python-utils` installed twice | Single early "SELinux — ensure policy tooling is present" task before the relocation block; both former install sites removed. |
| bomet.yml.example postal-code vars | `postalCodeLength` / `postalCodeErrorMessage` removed; `postalCodePattern` is the single knob (v2.12 behaviour). |

`ansible-playbook --syntax-check` passes on the updated playbook.

vinothrallapalli-eGov (approved):

PR #1582, opened 2026-08-05
feat(configurator): runtime config so one image serves every deployment (#1309)

dhruv-1001:
## What & why

Part of #1309. Makes **one pre-built `configurator` image serve every deployment**, so the image the nightly publishes is actually usable.

This PR previously carried a `local-deploy.sh` selective-build script. That is **removed** — `local-setup/ansible/deploy.sh` is the only supported deploy path.

## The bug

`configurator` read four values from `import.meta.env` — `VITE_STATE_TENANT_ID`, `VITE_OVERPASS_URL`, `VITE_TURBOPASS_URL`, `VITE_BOUNDARY_SEARCH_LIMIT` — i.e. **baked in at build time**. Its Dockerfile declared no `ARG`/`ENV` for any of them, while asserting:

> *"Tenant-neutral: configurator reads no VITE_\* build-env"*

That was false. A pre-built image baked all four **empty**, and **degraded silently rather than failing**:

- the login form lost its pre-filled tenant — and that field's value is the form's actual initial value, not a placeholder, so a blank one means operators retype the tenant on every login;
- the Phase-2 boundary-polygon fetch fell back to the public `overpass-api.de`, which the source itself calls rate-limited and 504-prone.

Proven by building the image twice with a sentinel value: shipped Dockerfile → **0** occurrences in the bundle; with `ARG`/`ENV` added → **1**.

This matters right now because `docker-compose.egov-digit.yaml` on `master` already ships a `configurator` service pulling `egovio/configurator`, with a comment making the same tenant-neutral claim. This PR is what makes that claim true.

## The fix — mirror digit-ui's `globalConfigs.js`

The four values now resolve at **runtime** from `<base>/config.js` (`window.__CONFIGURATOR_CONFIG__`), loaded by `index.html` before the bundle.

**Precedence: runtime → build-time `VITE_*` → in-code default**, with blank treated as unset *at every layer*. That last part is the subtle bit: a `config.js` rendered with an empty field (the normal state for an unconfigured deployment) must fall through rather than override a good value with `""`. 12 vitest cases cover this, plus the numeric rules below.

Build-time is kept only as a fallback, so `npm run dev` and any build-with-env workflow are unchanged. A neutral `public/config.js` ships in the image, so an unconfigured run behaves exactly as before this PR.

**Deliberately not** adding `VITE_*` build args — that keeps the image deployment-specific, which is the defect.

`BOUNDARY_SEARCH_LIMIT` goes through `resolvePositiveNumber` rather than `Number(x) || 300`. The old form rejected on *truthiness*, so it swallowed a configured `0` **and passed `-1` / `Infinity` straight through** to `/boundary/_search?limit=`. Non-positive values are now rejected explicitly, on a stated rule, because this is a page size.

## Validation

- **vitest 12/12** on the precedence contract and the numeric rules.
- **Built by the actual CI pipeline** — `build-images.yml`'s `run:` blocks executed against this branch produced a `configurator:develop-<sha>` image, pushed to a registry, then **pulled fresh and run**: `/configurator/config.js` returns `content-type: application/javascript` with the neutral defaults, and overwriting that one file changes the served config with **no rebuild and no restart**.
- **Real deploy** (`./deploy.sh maputo`, `failed=0`) served `/configurator/config.js` with `STATE_TENANT_ID: "mz"` rendered from `host_vars`.
- **Browser** — the login form's Tenant Code field is pre-filled `mz`; DOM confirms `input[name=tenantCode].value === "mz"` (the value, not the placeholder), `window.__CONFIGURATOR_CONFIG__` present, no console errors.

> **Reviewer trap:** `curl`-ing a nonexistent path under `/configurator/` returns **200 with `index.html`** because of nginx `try_files`. An early check of mine passed for that reason against an image with no `config.js` at all. Verify `content-type: application/javascript`, not just the status code.

## How config.js reaches the SPA (both serving paths)

Host nginx serves `/configurator/` two ways, and the rendered `config.js` reaches both — so flipping `build_configurator` cannot change what the SPA sees:

| `build_configurator` | nginx | `config.js` arrives via |
|---|---|---|
| `false` (default) | proxies to the `configurator` container | **bind mount** over the image's neutral placeholder |
| `true` | aliases the host-built dist | **copied in after the sync**, which would otherwise restore the placeholder |

Two ordering details worth a reviewer's eye:

- **`config.js` is rendered before `up -d`**, not with the other configurator tasks that run afterwards. Docker auto-creates a missing bind-mount source as a *directory*, which would silently shadow the real file.
- **The Overpass extract check is hoisted** to that same early point and drives an effective-URL fact. The playbook skips the Overpass container when the flag is on but the operator-prepared extract is absent; publishing `/overpass/api/interpreter` regardless would point the SPA at a location that 502s rather than at the documented public fallback. The build-time `VITE_OVERPASS_URL` reads the same effective value — otherwise a correctly-blank runtime value would fall through to a baked-in dead URL.

This PR adds no new configurator deploy machinery: it uses master's existing `configurator` compose service and `CONFIGURATOR_IMAGE`, rather than duplicating them.

## Note on the red CI

The three failing checks are **pre-existing on `master`**, not introduced here:

- `Ansible static validation` and `gatus-coverage` — both fixed by #1589 (they pass there). Merging #1589 first turns them green here.
- `Config combination + infra contracts` — the frozen `:latest` debt list in `static/infra-contracts.test.ts` still names pre-migration images (`curlimages/curl`, `twinproduction/gatus`, …) while compose now uses their `egovio/*` equivalents. Registry-migration bookkeeping, unrelated to this PR.

Verified locally that this branch adds **no new** `ansible-lint` violations: it reports exactly the same 2 as plain `master`.

## Summary by CodeRabbit

- **New Features**
  - Added runtime deployment configuration for tenant settings, service endpoints, and boundary search limits.
  - Deployment-specific values can now be updated without rebuilding the configurator.
  - Configurator deployments automatically use the appropriate Overpass endpoint when available.
  - Added validation and fallback handling for invalid or missing configuration values.

- **Bug Fixes**
  - Improved consistency across source-built and container-served deployments.
  - Prevented stale runtime configuration from being served.

- **Tests**
  - Added coverage for configuration precedence, fallbacks, and numeric validation.


dhruv-1001:
## What to review here

**Scope: configurator + the deploy that consumes it.** 10 files. The `local-deploy.sh` path this PR originally carried is **removed** — `local-setup/ansible/deploy.sh` is the only supported deploy.

### The bug being fixed

`configurator` read four values from `import.meta.env` — `VITE_STATE_TENANT_ID`, `VITE_OVERPASS_URL`, `VITE_TURBOPASS_URL`, `VITE_BOUNDARY_SEARCH_LIMIT` — i.e. **baked in at build time**. Its Dockerfile declared no `ARG`/`ENV` for any of them and asserted *"Tenant-neutral: configurator reads no VITE_\* build-env"*, which was false.

So a pre-built image baked all four **empty** and **degraded silently rather than failing**: the login form lost its pre-filled tenant (it is the field's initial *value*, not a placeholder — a placeholder-only prefill submits an empty `tenantId` and no login succeeds), and the Phase-2 polygon fetch fell back to the public `overpass-api.de`, which the source itself calls rate-limited and 504-prone.

Proven by building the image twice with a sentinel: shipped Dockerfile → **0** occurrences in the bundle; with `ARG`/`ENV` added → **1**.

### The fix — mirror digit-ui's `globalConfigs.js`

The four values now resolve at **runtime** from `<base>/config.js` (`window.__CONFIGURATOR_CONFIG__`), loaded by `index.html` before the bundle.

**Review focus — precedence:** runtime → build-time `VITE_*` → in-code default, with blank treated as unset *at every layer*. That last part matters: a `config.js` rendered with an empty field (the normal state for an unconfigured deployment) must fall through, not override a good value with `""`. 7 vitest cases cover exactly this.

Build-time is retained only as a fallback, so `npm run dev` and any build-with-env workflow are unchanged. A neutral `public/config.js` ships in the image so an unconfigured run behaves exactly as before.

**Deliberately not** adding `VITE_*` build args — that keeps the image deployment-specific, which is the defect.

### Ansible side
`build_configurator` now defaults false: the deploy pulls `configurator_image` (default `egovio/configurator:nightly-develop`), lifts the dist out with `docker create`/`cp`, and syncs it. Host nginx still serves `configurator_www_dir` from disk — the image is only a carrier, so **no new compose service and no nginx change**, and no Node/Vite on the deploy box.

Two ordering details worth checking:
- `config.js` renders **after** the sync (the dist ships a neutral placeholder the sync would otherwise restore).
- It renders on **both** paths (from-source and pulled), so switching `build_configurator` cannot change what the SPA sees.

The pull accepts an already-present local image (same contract as `digit-mcp:local`) so a locally-built configurator can be validated before publication; a tag that is neither pullable nor local is a hard error.

⚠️ **`egovio/configurator:nightly-develop` returns 404 today.** It exists only once #1422 merges to `master` and its first nightly runs. Until then a tenant must override `configurator_image`.

---

## Validation

Beyond unit tests, this was driven end-to-end on a real deploy:

1. **`./deploy.sh maputo`** — `ok=133 changed=33 failed=0`. Pull-image path exercised: `ensure Node 20` SKIPPED, `pull image + extract dist` CHANGED, `render runtime config.js` CHANGED.
2. **Served result** — `/configurator/config.js` returns `STATE_TENANT_ID: "mz"`, rendered from `host_vars`, with `index.html` loading it.
3. **Browser** — the login form's Tenant Code field is **pre-filled `mz`**; DOM confirms `input[name=tenantCode].value === "mz"` (the value, not the placeholder) and `window.__CONFIGURATOR_CONFIG__` present. No console errors.
4. **Built by the actual CI workflow** — the `build-images.yml` `run:` blocks were executed against this branch, producing `configurator:develop-cfd5ec8e`, pushed to a registry, **pulled fresh and run**: `/configurator/config.js` returns `content-type: application/javascript` with the neutral defaults, and overwriting that one file changes the served config with **no rebuild and no restart**.

Point 4 is the whole premise: a CI-built image, configured by a file the deploy renders.

One trap worth knowing when reviewing: `curl` a nonexistent path under `/configurator/` and nginx's `try_files` returns **200 with `index.html`**. An early check of mine passed for that reason on an image that had no `config.js` at all — verify `content-type: application/javascript`, not just the status code.

**Also validated on this stack** (not caused by this PR, but it exercises the same deploy): a full PGR lifecycle — create → assign → resolve, ending `RESOLVED` in `eg_pgr_service_v2`, dashboard `recordCount` 0 → 2.

configurator/src/api/config.ts
```diff
@@ -114,9 +116,10 @@ export const OAUTH_CONFIG = {
 //    needs a representative set, and city/county hierarchies are well under).
 //
 // Configurable so a deployment whose boundary-service raises/lowers the cap
-// can match it without a code change: set VITE_BOUNDARY_SEARCH_LIMIT.
+// can match it without a code change: set BOUNDARY_SEARCH_LIMIT in config.js
+// (or the build-time VITE_BOUNDARY_SEARCH_LIMIT).
 export const BOUNDARY_SEARCH_LIMIT: number =
-  Number(import.meta.env.VITE_BOUNDARY_SEARCH_LIMIT) || 300;
+  Number(resolveConfig('BOUNDARY_SEARCH_LIMIT', import.meta.env.VITE_BOUNDARY_SEARCH_LIMIT)) || 300;
```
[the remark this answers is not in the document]
dhruv-1001: Fixed in 0dbf921 — **but deliberately not the way the suggestion has it**, so flagging the deviation.

Agreed on the diagnosis: `|| 300` rejects on truthiness, which is the wrong mechanism. Digging into it turned up that the truthiness form was doing more damage than the `0` case:

| configured | old `Number(x) \|\| 300` | new |
|---|---|---|
| `"0"` | 300 | 300 |
| `"-1"` | **-1** → `?limit=-1` | 300 |
| `"Infinity"` | **Infinity** → `?limit=Infinity` | 300 |
| `"abc"` | 300 | 300 |
| `"500"` | 500 | 500 |

`-1` and `Infinity` were reaching `/boundary/_search` as-is — those are the values truthiness actually *fails* to catch, and they were the real functional bug here.

**Where I diverged:** the suggested diff preserves an explicit `0`. I reject it instead, because `0` is not a meaningful value for this setting — `BOUNDARY_SEARCH_LIMIT` is a page size, and `?limit=0` makes boundary-service return nothing (or fall back to its own ~50 default), so honouring it renders an *empty* overview map. That is the silent degradation this whole runtime-config PR exists to remove, and no operator writing "limit" means "fetch none". The point of the finding still stands and is addressed: the rejection is now **explicit and stated** rather than an accident of `||`.

`configurator/src/api/runtimeConfig.ts` gains a small documented helper:

```ts
export function resolvePositiveNumber(raw: string, fallback: number): number {
  if (!raw) return fallback;
  const parsed = Number(raw);
  if (!Number.isFinite(parsed) || parsed <= 0) return fallback;
  return parsed;
}
```

and `config.ts` composes it with `resolveConfig` (`DEFAULT_BOUNDARY_SEARCH_LIMIT = 300`). Blank still falls through the precedence chain exactly as before — the fallback is applied at the caller, not inside `resolveConfig`, so nothing about runtime > build-time > default changes.

**Tests:** five new cases in `runtimeConfig.test.ts` covering `0`, `-1`, `0.5` (positive, so honoured — we reject non-positive, we don't round), `"abc"`, `"300px"`, `"Infinity"`, and a `0` arriving through `config.js` as a real JS number rather than a string. The two pre-existing limit cases now exercise the real helper instead of an inline `|| 300`, so they can't drift from production. Mutation-checked: restoring `Number(raw) || fallback` turns the new cases red (`expected -1 to be 300`, `expected Infinity to be 300`) rather than passing vacuously.

`vitest run src/api/` → 18 passed (3 files); `eslint` and `tsc --noEmit -p tsconfig.app.json` clean on the touched files.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -2003,23 +2003,72 @@
...
+
+    - name: "Configurator SPA — use extracted dist path"
+      ansible.builtin.set_fact:
+        configurator_build: "{{ configurator_extract_dir }}"
+      when:
+        - nginx_features.configurator | default(false)
+        - not (build_configurator | default(false))
+        - configurator_extract is defined
+        - configurator_extract is not skipped
+
```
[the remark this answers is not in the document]
dhruv-1001: Confirmed and fixed in 0dbf921. You're right: `configurator_extract_dir` is created by the `docker create`/`docker cp` shell task **on the target**, and `synchronize` is a controller → target push, so on a real remote deploy it would look for that directory on the controller — missing (task fails) or, worse, an unrelated same-named directory silently published as the SPA. It only looked fine in testing because every committed host_vars example sets `ansible_connection: local`, i.e. controller == target.

The single task is now two, split on `build_configurator`:

- **`build_configurator: true`** — unchanged `ansible.builtin.synchronize`, controller → target. This path's dist comes from the controller's checkout (`{{ playbook_dir }}/../../configurator`), so the push direction is correct for it.
- **`build_configurator: false`** — new target-side install from the registered extract dir:

```yaml
    - name: "Configurator SPA — install extracted dist (target-side copy)"
      ansible.builtin.shell: |
        set -euo pipefail
        src="{{ configurator_build }}"
        dest="{{ configurator_www_dir }}"
        test -f "$src/index.html"
        mkdir -p "$dest"
        if command -v rsync >/dev/null 2>&1; then
          rsync -a --omit-dir-times --itemize-changes "$src/" "$dest/"
        else
          cp -a "$src/." "$dest/"
          echo "cp: copied $src -> $dest (rsync unavailable)"
        fi
      register: configurator_install
      changed_when: configurator_install.stdout | trim | length > 0
```

On idempotence + `changed_when`: `--itemize-changes` prints one line per file actually transferred, so an unchanged re-deploy prints nothing and the task reports **ok**, not changed. `--omit-dir-times` stops directory mtimes from faking a change. No `--delete` — same contract as the `synchronize` task above it, and a hashed-asset SPA wants old assets to survive for already-cached `index.html`. The `cp -a` branch (box without rsync) announces itself, because `cp` can't distinguish a no-op from a real update and claiming "ok" there would be a lie.

Exercised the block directly against real directories: first run itemises 4 entries → changed; second run prints nothing → not changed; after the tenant `config.js` is rendered on top, the next run itemises exactly that one file → changed, which is correct. Missing `index.html` exits 1. The no-rsync branch was run with a stripped `PATH` and copies correctly.

**One thing I did not change, worth flagging separately:** the `build_configurator: true` task runs `bash {{ playbook_dir }}/files/configurator-build.sh` via `ansible.builtin.command` with **no `delegate_to: localhost`**, so it executes on the *target* against a *controller* path. The comment at L1581 describes this as "the configurator pattern (controller-side build)", and the `synchronize` direction assumes the same, so intent and implementation disagree — on a genuinely remote target that task would fail before the sync ever mattered. Fixing it means either delegating the build or syncing the source to the target first, and it is part of the broader configurator-deploy rework this PR's ansible half already needs (master now serves `/configurator/` by proxying to a compose container rather than from `/var/www/configurator`), so I've left it rather than half-change it here.

`ansible-playbook --syntax-check` passes, `ansible-lint playbook-deploy.yml` → 0 failures / 0 warnings, `bash -n` clean on the added shell.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -2040,6 +2089,21 @@
...
+    # runtime config first either way, so the two paths stay behaviourally
+    # identical and switching build_configurator can't change what the SPA sees.
+    - name: "Configurator SPA — render runtime config.js"
+      ansible.builtin.template:
+        src: configurator-config.js.j2
+        dest: "{{ configurator_www_dir }}/config.js"
+        mode: "0644"
+      when:
+        - nginx_features.configurator | default(false)
+
```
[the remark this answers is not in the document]
dhruv-1001: Confirmed and fixed in 0dbf921. The render was ~150 lines ahead of the Overpass block, and that block deliberately skips `docker run` when `enable_overpass` is true but the operator-prepared extract is absent (it only prints a warning) — while nginx's `/overpass/` location is gated on `enable_overpass` alone, so the SPA was pointed at a proxy to nothing (502) instead of falling through to `https://overpass-api.de/api/interpreter`.

Of the two options offered I took the second (**effective value**) rather than moving the render, and integrated with the existing check rather than adding one:

- The existing `Overpass — check prepared extract exists` **stat task is hoisted** to just above the configurator block. Same task, same `overpass_extract` registration — the warn and `docker run` tasks further down still consume it, so there is still exactly one check, just earlier than everything that publishes the URL.
- A new `set_fact` derives the effective value from it:

```yaml
    - name: "Overpass — effective endpoint for the configurator SPA"
      ansible.builtin.set_fact:
        configurator_overpass_url: >-
          {{ '/overpass/api/interpreter'
             if (((enable_overpass | default(false)) | bool)
                 and ((overpass_extract.stat.exists | default(false)) | bool))
             else '' }}
```

- `configurator-config.js.j2` renders `OVERPASS_URL` from that value, so it is blank exactly when the local service was not provisioned, and the app's documented public fallback applies.

Why hoist rather than move the render: moving the template task alone would have left a second copy of the bug one layer down. The `build_configurator: true` path also exports `VITE_OVERPASS_URL: "{{ '/overpass/api/interpreter' if enable_overpass ... }}"` as the **build-time fallback**, and precedence is runtime > build-time > default — so a correctly-blank `config.js` would have fallen straight through to a baked-in dead URL. That env var now reads the same `configurator_overpass_url`, so both layers agree. The render also has to stay after the dist install (the dist ships a neutral placeholder `config.js` that would otherwise overwrite the tenant-rendered one), and hoisting the check keeps that ordering intact without shuffling tasks across the Turbopass block.

On "confirmed available": I gate on the extract, not on an HTTP probe. That is the exact condition the `docker run` task itself gates on, so it is a faithful predictor of whether the container gets started — whereas probing would be misleading anyway, since the container comes up in `OVERPASS_MODE=init` and doesn't answer queries until the import finishes, long after the play ends. Flagging the distinction explicitly: this publishes the local URL when the endpoint *was provisioned*, not when the import has completed.

Verified by rendering the template both ways (`OVERPASS_URL: "/overpass/api/interpreter"` when provisioned, `OVERPASS_URL: ""` when the extract is missing; both pass `node --check`), and by evaluating the fact expression across the flag/extract matrix including `enable_overpass: "true"` as a string and an undefined `overpass_extract` — all yield the expected value with no string-truthiness trap. `--syntax-check` passes and `ansible-lint` reports 0 failures.

Also corrected the now-stale comment in the Overpass block that still described the configurator as baking `VITE_OVERPASS_URL` whenever the flag is on.

local-setup/docker-compose.egov-digit.yaml
```diff
@@ -1611,6 +1611,15 @@ services:
...
     container_name: configurator
+    # Per-tenant runtime config, rendered by the ansible deploy from host_vars
+    # (templates/configurator-config.js.j2) and mounted over the neutral
+    # placeholder baked into the image. This is what lets ONE published image
+    # serve every tenant: the deployment values the SPA needs are read at boot
+    # from this file instead of being compiled in by `vite build`.
+    # The deploy renders it BEFORE `up -d` — Docker auto-creates a missing
+    # bind-mount source as a DIRECTORY, which would shadow the real file.
+    volumes:
+      - ./configurator/config.js:/usr/share/nginx/html/configurator/config.js:ro
```
vinothrallapalli-eGov: **Most severe: breaks the documented plain `docker compose up -d` quickstart.**

This unconditional single-file bind mount assumes the ansible deploy has already rendered `configurator/config.js` first. But CLAUDE.md's documented local workflow is just `cd local-setup && docker compose up -d` — no ansible run. On a fresh checkout, `./configurator/config.js` doesn't exist, so Docker auto-creates the bind-mount source as a root-owned **directory** to satisfy the mount, shadowing the image's neutral placeholder config.js inside the container (nginx can no longer serve it as a static file).

The stray directory is root-owned in the working tree and can't be removed without `sudo`, and a later ansible deploy targeting this same path as `digit_dir` will fail when the `ansible.builtin.template` task tries to write a regular file over what is now a directory.

Suggest either: seed a placeholder `configurator/config.js` file in the repo (checked in) so the mount source always pre-exists, or guard the mount so it only applies when the file is present (e.g. render a default in a pre-step).

local-setup/ansible/playbook-deploy.yml
```diff
@@ -2535,6 +2580,45 @@
...
+      failed_when:
+        - configurator_restart.rc != 0
+        - '"No such container" not in (configurator_restart.stderr | default(""))'
+      when:
+        - nginx_features.configurator | default(false)
+        - not (build_configurator | default(false))
+        - configurator_config_render is defined
+        - configurator_config_render is changed
+
+    - name: "Configurator SPA — place runtime config.js beside the host-served dist"
```
vinothrallapalli-eGov, line 2630: **macOS/Darwin path collision clobbers the tenant-rendered config.js.**

On the documented macOS/OrbStack quickstart path (`build_configurator: true`), `configurator_www_dir` resolves to the exact same path as the new tenant-config render destination (`{{ digit_dir }}/configurator`). The earlier "sync build" rsync step (which copies the dist's neutral placeholder `config.js` into `configurator_www_dir`) runs *before* this copy and would clobber the tenant-rendered file — but since both paths are identical here, this "place runtime config.js beside the host-served dist" step becomes a same-file no-op copy that can't restore anything once clobbered.

Result: on every macOS deploy, the served `configurator/config.js` ends up with blank placeholder values instead of the tenant-rendered ones — defeating the exact anti-clobber fix this PR is adding. Currently masked for `STATE_TENANT_ID`/`OVERPASS_URL` only because the build-time `VITE_*` env fallback happens to bake in the same values at build time; it will not mask `TURBOPASS_URL`/`BOUNDARY_SEARCH_LIMIT` once those are wired up, or any value that needs to change without a full rebuild.

Worth double-checking `configurator_www_dir` vs `digit_dir}}/configurator` don't alias when `build_configurator: true`, or reordering so this copy runs strictly after the rsync with a distinct destination.

local-setup/ansible/templates/configurator-config.js.j2
```diff
@@ -0,0 +1,44 @@
...
+  // and pointing the SPA at an nginx location that proxies to nothing (502) is
+  // worse than the public fallback. `configurator_overpass_url` is set by the
+  // playbook from the same extract check those tasks gate on — it is blank
+  // whenever the local service was not actually provisioned.
+  OVERPASS_URL: {{ (configurator_overpass_url | default('')) | to_json }},
+
+  // Turbopass suggestions base (OSM place-name autocomplete). Same-origin
+  // '/turbopass' when enable_turbopass is on; blank falls through to that
+  // default anyway, so this is here for explicitness and overrides.
+  TURBOPASS_URL: {{ (configurator_turbopass_url | default('')) | to_json }},
```
vinothrallapalli-eGov, line 39: `configurator_turbopass_url` is read here (and `configurator_boundary_search_limit` on line 43) but neither fact is ever set anywhere in the playbook or inventory — only `configurator_overpass_url` gets a corresponding `set_fact` task above. Both currently always fall back to `default('')` and render blank.

An operator following the documented pattern for `state_tenant_id`/`enable_overpass` to override the Turbopass URL or boundary search page cap will find it silently has no effect — `TURBOPASS_URL`/`BOUNDARY_SEARCH_LIMIT` stay `''` every deploy, falling back to the in-code defaults (`/turbopass`, 300) with no warning that the override never took effect.

Needs a `set_fact` (or inventory var) wiring these two the same way `configurator_overpass_url` is wired, or the fields should be dropped until that plumbing exists.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -2535,6 +2580,45 @@
...
+    # container keeps serving the file it started with — compose will not
+    # recreate it, because the compose config itself has not changed. Verified on
+    # this stack: after a re-render the host file and the container's copy
+    # diverged, and `docker restart` re-resolved the mount.
+    #
+    # So: restart the container ONLY when the render actually changed. On a first
+    # deploy the container is created after the render and already has the right
+    # content, and `changed` is false on an unchanged re-run, so this is a no-op
+    # in the common case rather than a restart on every deploy.
+    - name: "Configurator SPA — restart container so a re-rendered config.js is picked up"
```
vinothrallapalli-eGov, line 2617: The comment above says restarting is "a no-op in the common case rather than a restart on every deploy" because `changed_when: configurator_restart.rc == 0` combined with the `configurator_config_render is changed` guard is assumed false on first deploy — but `ansible.builtin.template` reports `changed` on a brand-new render too (the destination file didn't exist before), so this guard is also true on first deploy, and `docker restart configurator` fires there as well, contradicting the stated intent.

Harmless today since the configurator container is already healthy at that point, but it's worth either fixing the comment to reflect reality, or (if the no-restart-on-first-deploy behavior is actually desired) adding an explicit first-deploy guard, e.g. checking whether the container was just created in this same play.

local-setup/ansible/templates/configurator-config.js.j2
```diff
@@ -0,0 +1,44 @@
...
+
+  // boundary-service page cap for the overview map. Must be a positive number;
+  // blank / non-numeric / <= 0 => the app's 300 (a page size of 0 would fetch
+  // nothing, so the app rejects it rather than render an empty map).
+  BOUNDARY_SEARCH_LIMIT: {{ (configurator_boundary_search_limit | default('')) | string | to_json }},
```
vinothrallapalli-eGov, line 44: Same undefined-fact issue as `configurator_turbopass_url` on line 38 — `configurator_boundary_search_limit` has no `set_fact`/inventory wiring anywhere, so this always renders `''` and silently falls back to the in-code default of 300.

vinothrallapalli-eGov (commented):
Automated review (high-effort, multi-agent pass) of the runtime config.js change. 5 findings below, most severe first. One additional finding (restart/validate race, PLAUSIBLE) couldn't be anchored inline since the affected `validate — configurator returns 200` task is unchanged by this PR — noted here:

- **Restart races unretried health check** (`playbook-deploy.yml`, the pre-existing `validate — configurator returns 200` task, further down from the new restart task at line 2599): unlike its sibling validate tasks (kong, public UI), this check has no retries/delay. The new "restart container so a re-rendered config.js is picked up" step bounces the configurator container; if a tenant's feature set makes the gap between the restart and this validate task unusually short, the single-attempt check can hit the container while nginx is still coming back up and fail the whole deploy even though configurator is healthy moments later. Worth adding a retry/delay to that check, or ordering the restart with more buffer.

local-setup/docker-compose.egov-digit.yaml
dhruv-1001: **Fixed** in `5c570bfc` — and you were right that this was the most severe: it broke the documented quickstart outright.

Two changes, because the cleanest fix for this also fixes the macOS aliasing in the sibling thread:

1. The mount source is now a **checked-in placeholder**, so `docker compose up -d` on a fresh checkout has a real file to bind and Docker never auto-creates a directory. It configures nothing (all fields blank), so the SPA falls through to build-time then in-code defaults — i.e. exactly pre-runtime-config behaviour.
2. It moved to `local-setup/configurator-runtime/config.js` rather than `configurator/`, so the path cannot alias `configurator_www_dir` on macOS.

Verified by starting the container with only that mount, no ansible: `/configurator/` → 200, `/configurator/config.js` → 200 `application/javascript` with blank defaults, and the mount source stayed a normal user-owned file.

Worth owning: the previous commit's own comment warned about Docker auto-creating a missing bind-mount source as a directory — I guarded the ansible path and never considered the plain-compose path the docs actually tell people to use.

local-setup/ansible/playbook-deploy.yml
dhruv-1001: **Fixed** in `5c570bfc`. Confirmed the alias by simulating both platforms:

```
Linux    www_dir=/var/www/configurator      render=/opt/digit/configurator   ALIAS=no
Darwin   www_dir=/opt/digit/configurator    render=/opt/digit/configurator   ALIAS=YES
```

So on macOS the dist rsync clobbered the tenant-rendered file with the image's blank placeholder, and the "place beside the host-served dist" copy became a same-file no-op that could not restore it — defeating the exact anti-clobber fix this PR adds.

Fixed by rendering to `{{ digit_dir }}/configurator-runtime/` instead, which cannot collide with `configurator_www_dir` on either platform. Chose a distinct path over reordering because reordering would still leave two code paths writing the same file and relying on order to stay correct.

local-setup/ansible/templates/configurator-config.js.j2
dhruv-1001: **Fixed** in `5c570bfc` — both facts are now set, alongside the Overpass one:

```yaml
- name: "Configurator SPA — effective Turbopass + boundary-limit settings"
  ansible.builtin.set_fact:
    configurator_turbopass_url: >-
      {{ configurator_turbopass_url
         | default('/turbopass' if (enable_turbopass | default(false)) else '') }}
    configurator_boundary_search_limit: >-
      {{ configurator_boundary_search_limit | default('') }}
```

You were right that this was a silent no-op: the template read facts nothing defined, so both rendered `''` every deploy and an operator's override vanished without a warning. Turbopass now follows the same shape as Overpass (same-origin only when the box actually runs it); the boundary limit passes through from inventory.

Verified on a live deploy — the rendered `config.js` now emits both fields, and the browser shows `window.__CONFIGURATOR_CONFIG__` carrying all four keys.

local-setup/ansible/playbook-deploy.yml
dhruv-1001: **Fixed** in `5c570bfc` — the comment was wrong and is now corrected rather than the behaviour changed.

You're right that `template:` reports `changed` on a brand-new render, so the restart does fire on a first deploy. I kept that behaviour (harmless — the container was created after the render and already holds the right content, it just restarts once) and rewrote the comment to say what actually happens:

> Note that `template:` reports changed on a brand-new render too, so this DOES fire on a first deploy … What it avoids is a restart on every SUBSEQUENT deploy: an unchanged re-render reports ok, and the task is skipped.

Confirmed on a redeploy with no config change: render reports `ok`, restart is skipped.

local-setup/ansible/templates/configurator-config.js.j2
dhruv-1001: **Fixed** in `5c570bfc` — same fix as the `TURBOPASS_URL` thread above: `configurator_boundary_search_limit` is now set by a `set_fact`, so an inventory override actually reaches the template instead of always rendering `''`.

The in-code side is also stricter than it was: `BOUNDARY_SEARCH_LIMIT` goes through `resolvePositiveNumber` rather than `Number(x) || 300`, so a configured `0` is rejected on a stated rule rather than by truthiness — and `-1`/`Infinity` no longer pass through to `/boundary/_search?limit=`, which the old form allowed.

configurator/src/api/runtimeConfig.test.ts
```diff
@@ -0,0 +1,116 @@
...
+  it("falls back to 'pg' when nothing is configured at any layer", () => {
+    window.__CONFIGURATOR_CONFIG__ = { STATE_TENANT_ID: '' };
+    expect(resolveConfig('STATE_TENANT_ID', undefined) || 'pg').toBe('pg');
+  });
+
+  it('a blank runtime value still defers to a build-time one before the default', () => {
+    window.__CONFIGURATOR_CONFIG__ = { STATE_TENANT_ID: '' };
+    expect(resolveConfig('STATE_TENANT_ID', 'ke') || 'pg').toBe('ke');
+  });
+});
```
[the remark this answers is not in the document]
dhruv-1001: Fixed in `a91303a3` — and the finding was sharper than "minor". I checked it by mutation before changing anything: deleting the `'pg' default from `config.ts` left **all 15 tests green**. They were asserting a test-side copy of `|| 'pg'`, so they could not have caught the regression they exist to prevent.

Took the `vi.resetModules()` + dynamic-import route (`config.ts` resolves `STATE_TENANT_ID` at module scope, so each case resets the registry and re-imports). Build-time precedence now uses `vi.stubEnv` instead of a hand-passed argument, so that layer is exercised the way the app sees it too.

Re-verified by mutation — both now go RED where they were previously green:

| mutation to `config.ts` | before | after |
|---|---|---|
| remove the `'pg'` default | 15 passed | **1 failed** |
| point the consumer at the wrong `config.js` key | 15 passed | **2 failed** |
| restored | — | 16 passed |

Also added a `getConfiguredRootTenant` case (city code must collapse to its root segment) — nothing covered that.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -2998,6 +3058,47 @@
...
+      failed_when:
+        - configurator_restart.rc != 0
+        - '"No such container" not in (configurator_restart.stderr | default(""))'
+      when:
+        - nginx_features.configurator | default(false)
+        - not (build_configurator | default(false))
+        - configurator_config_render is defined
+        - configurator_config_render is changed
+
+    - name: "Configurator SPA — place runtime config.js beside the host-served dist"
```
KDwevedi, line 3106: **Host-built deployments can cache stale runtime config.**

This copy is served directly by host nginx when `build_configurator: true`, so it bypasses the `no-store` rule added to the configurator container's nginx. With the repository default `nginx_features.configurator_caching: false`, the host nginx `/configurator/` location adds no `Cache-Control` header at all. Browsers may therefore continue using an old tenant or endpoint after this file is replaced, which defeats the runtime-config mechanism.

Please add an exact host-nginx location for `/configurator/config.js` with `Cache-Control: no-store, must-revalidate` in both host-built caching configurations.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -2998,6 +3058,47 @@
...
+    - name: "Configurator SPA — restart container so a re-rendered config.js is picked up"
+      ansible.builtin.command: docker restart configurator
+      register: configurator_restart
+      changed_when: configurator_restart.rc == 0
+      failed_when:
+        - configurator_restart.rc != 0
+        - '"No such container" not in (configurator_restart.stderr | default(""))'
+      when:
+        - nginx_features.configurator | default(false)
+        - not (build_configurator | default(false))
```
KDwevedi: **Switching from host-built to container mode can expose an old `config.js`.**

When the config changes during a `build_configurator: true` deploy, this condition skips the restart, so the still-running container remains bound to the old inode. If a later deploy flips `build_configurator` to `false` without changing the config values again, the render reports unchanged and this restart is skipped again; host nginx then proxies to a container serving the stale tenant config.

Please keep the container mount synchronized whenever `configurator_config_render` changes, regardless of which serving path is currently active (or explicitly refresh it when switching back to container mode).

local-setup/ansible/playbook-deploy.yml
dhruv-1001: Fixed in `c2fef933`. Confirmed the mechanism before changing anything, and the default is the bad case:

| host-built variant | `Cache-Control` on config.js |
|---|---|
| `configurator_caching: true` | `no-cache, must-revalidate` — revalidates, tolerable |
| `configurator_caching: false` (**repo default**) | **none** — heuristic caching off `Last-Modified` |

Added an exact-match `location = /configurator/config.js` with `no-store, must-revalidate` to both host-built branches. Exact match beats the `/configurator/` prefix regardless of order, so `index.html` keeps its existing behaviour.

Left the proxy branch alone deliberately — it has no `add_header` of its own, so the container image's `no-store` already passes through. Confirmed rather than assumed.

Verified by rendering the template and serving it through real nginx:

```
--- caching_off (repo default) ---
config.js  200  Cache-Control: no-store, must-revalidate
body:      STATE_TENANT_ID:"mz"
index      200  Cache-Control: (none — unchanged)

--- caching_on ---
config.js  200  Cache-Control: no-store, must-revalidate
index      200  Cache-Control: no-cache, must-revalidate (unchanged)
```

And rendered the **unpatched** template the same way to confirm the bug was real rather than theoretical — `config.js` came back with no `Cache-Control` header at all. `nginx -t` passes on all three rendered variants.

local-setup/ansible/playbook-deploy.yml
dhruv-1001: Fixed in `c2fef933` — and it was worse than the comment assumes. I checked whether the container even runs in host-built mode: the `configurator` compose service has **no `profiles:` key**, so it is up the whole time, sitting pinned to the old inode while host nginx serves the dist. So the stale container is not a leftover, it is always there waiting.

Your sequence is exactly right:

1. `build_configurator: true` + a tenant change → render `CHANGED`, restart skipped, container still holds the previous tenant
2. later deploy flips to `false` with the same values → render reports `ok`, nothing triggers a restart
3. host nginx proxies to a container serving the **old** tenant, no error anywhere

Not hypothetical either: `true → false` is precisely the migration this PR exists to enable.

Took the simpler of your two suggestions — dropped the `not (build_configurator | default(false))` condition so the mount is kept in sync in both modes, rather than adding mode-switch detection that would need its own state. Cost is one `docker restart` of an idle container on the host-built path, and `failed_when` already tolerates the container being absent for slim stacks.

My original condition was the wrong optimisation: I reasoned "the container is not serving traffic in host mode, so do not bother restarting it" — which is true right up to the moment it *starts* serving. Rewrote the comment above the task, since it argued for the condition I just removed.

`ansible-lint` passes the production profile; syntax-check passes.

PR #1585, opened 2026-08-05
feat(analytics): MDMS-driven analytics adapter registry (moz) — schema + portal shim + Configurator screen

Hari-egov:
Moz counterpart of **#1584** (base `develop`). Same commits cherry-picked onto
`release-v2.12-moz`, plus one moz-only commit. **Read #1584 for the full design rationale** —
this description covers what differs, plus the documentation index.

## What this does

Analytics destinations for the citizen/employee portal become **MDMS rows edited in the
Configurator** instead of deploy-time config. With no rows — the state of every environment the
moment this merges — the portal issues at most two anonymous MDMS reads per page load, caches
the empty result for 90s, and does nothing else. **No environment changes behaviour until an
admin enables a row.**

Adapters: `MATOMO`, `GA4`, `POSTHOG`, `SENTRY`, plus a data-described `CUSTOM` type.

---

## Documentation

`docs/analytics-guide/` — plain-language operator guide:

| Doc | Covers |
|---|---|
| [README](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/e6f377d2/docs/analytics-guide/README.md) | what it is, runtime flow, admin flow, the two caveats |
| [10-setup](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/e6f377d2/docs/analytics-guide/10-setup.md) | **installation** — per-environment prerequisites, seeding the schema, first destination |
| [20-configuration](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/e6f377d2/docs/analytics-guide/20-configuration.md) | **configuration** — every field, state-vs-city precedence, residency acknowledgement, ops knobs, page titles, what is and is not sent |
| [30-customization](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/e6f377d2/docs/analytics-guide/30-customization.md) | `CUSTOM` records and code adapters |
| [40-operations](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/e6f377d2/docs/analytics-guide/40-operations.md) | **deployment/ops** — 4-tier rollback, troubleshooting, propagation, verification record |
| [50-self-hosted-matomo](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/e6f377d2/docs/analytics-guide/50-self-hosted-matomo.md) | **full install → deploy → configure walkthrough** with the five traps we hit |

---

## The moz-only commit

`ccrs-migrate.cjs` (which does not exist on develop/master) admitted schemas by the
`RAINMAKER-PGR.` prefix only, so an operator running the unified runner would have got the
whole feature deployed and silently inert: no schema, every read empty, nothing saying why.
`common-masters.json` holds dozens of upstream schemas that must not be touched by this runner,
so it is admitted through an explicit one-entry allowlist rather than by loosening the prefix
test — verified: 1 of the 11 schemas in that file is admitted.

`local-setup/scripts/seed-analytics-schema.sh` stays the primitive and the only carrier that
works on develop/master, where default-data-handler is no longer in the compose stack. It now
ensures the **ACCESSCONTROL rows** too (actions 30/31 → `SUPERUSER`/`MDMS_ADMIN`), not just the
schema — nothing else carries those to a *running* environment, since `MdmsBulkLoader` skips a
file whose tenant already has rows. They are inert while Kong runs `ENFORCE_RBAC=false`; they
exist so a later flip does not silently break analytics writes.

So there are two entry points doing the same work, and `10-setup.md` now leads with a table
saying which applies where:

| Release line | Command |
|---|---|
| develop / master | `TENANT=<stateRoot> ./local-setup/scripts/seed-analytics-schema.sh` |
| **moz (this PR)** | `node docs/migration/ccrs-migrate.cjs --host … --tenant <stateRoot>` — wraps the same source data in auth, `--dry-run`, per-phase reporting, a readiness check, and the Configurator localisation keys |

Both are add-if-missing, so running both is harmless. Neither creates a destination record:
absent records *is* the default-off state.

Verified on the live local stack, both paths — not just the happy one:

```
present:                     actions 0 created/2 present · grants 0 created/4 present
after deleting SUPERUSER.30:  grants 1 created/3 present → independent MDMS read shows all 4 back
```

The first cut had a real bug, caught by running it rather than reading it: the `already present`
branch exited before the ACCESSCONTROL section, so on any environment that already had the
schema — i.e. every environment this matters on — the rows were never ensured.

---

## Conflict resolutions — please sanity-check these

All were expected keep-both cases; moz is a superset in three of them.

| File | Resolution |
|---|---|
| `application.properties` | kept **moz's** `commonMDMSConfig.PrivacyPolicy` spelling (develop has `commonUiConfig`) and appended `common-masters.AnalyticsProvider` |
| `schemaDescriptors/types.ts` | took **moz wholesale** — it already has `'json'` plus its own `'object-table'` and `FieldSpec.columns` |
| `widgets/index.tsx` | took **moz wholesale** — `JsonInput` and `ObjectTableInput` already wired |
| `schemaDescriptors/index.ts` | kept **both** — moz's landing/ui-homepage/data-security descriptors plus the analytics one |
| `themeEditor/index.ts` | kept **both** — moz's `landing-builder` plus `analytics-provider` |
| `ACCESSCONTROL-ROLEACTIONS.roleactions.json` | kept **both** sets of grants; ids 1792–1795 are free on moz (its own additions use 1800+ and 2000+), verified no collisions and every grant resolves to an existing action |

`JsonInput.tsx` did **not** conflict: it was ported byte-identical from this work, so both sides
matched exactly. That was deliberate — develop gains the widget without the branches diverging.

---

## Verification on this branch

```
digit-ui-esbuild:  npm test                              → 50 tests pass
configurator:      npm test                              → 125 tests / 13 files pass
                   npx tsc -p tsconfig.app.json --noEmit  → 7 errors, ALL pre-existing in
                     moz-only src/admin/landingBuilder/* which this branch does not touch.
                     Pristine origin/release-v2.12-moz measured 11 by the same method, so the
                     count does not increase.
data-provider:     dist rebuilt; registry entry present
ccrs-migrate:      node --check clean; allowlist admits exactly 1 of 11 schemas
```

Vendor verification is unchanged from #1584 (same commits): self-hosted Matomo end to end with
the real vendor tracker — 4 route changes → 4 collector hits, read back from MariaDB with
derived page titles — PostHog Cloud against a real project, the proof that removing the ops host
declaration takes an enabled destination from 4 hits to 0, and PII checked **at rest** in
Matomo's database (0 rows for both the mobile number and the complaint id).

---

## Deviations and known limits

Identical to #1584: no localhost guard (documented, deliberate); ACCESSCONTROL rows reach
freshly-seeded tenants only; the residency acknowledgement is a save-time control; `MATOMO` does
not trigger it (correct for self-hosted, wrong for Matomo Cloud); the screen's route is reachable
read-only by any logged-in user with credentials masked; docs at `docs/analytics-guide/`; pilot
verification deferred at the time; cms-pilot/mctd provenance unconfirmed because their
`host_vars` are gitignored.

## Merge order

Either PR can go first — they share no files with each other's base, and the branches are
independent heads.


Hari-egov:
Ported the CodeRabbit review fixes from #1584 (cherry-pick of `cac7a1d7`) so both PRs are in the same state. All three findings applied to this branch:

1. **`configurator/index.html`** — `telemetry-config.js` was requested from the origin root while `vite.config.ts` sets `base: '/configurator/'`. Deployed under that prefix the request 404s, the global is missing, and `isTelemetryKilled()` returns `false` — the kill switch failed **open**. Now loaded base-relative.

2. **`AnalyticsProvidersEditor.tsx`** — the textarea key used `editingCode ?? 'new'`, but `startCreate` sets `editingCode` to `''`, so `??` never fell back and every create session shared the key `-settings`. React reused the textarea, leaving a cancelled session's JSON on screen over an empty draft. Now `||` plus a per-session `createSeq` counter.

3. **`analytics.js` (CWE-201)** — the host allowlist gated only the loaded script. `matomoEndpoint()` returned `rec.endpointUrl` unchanged into `_paq.push(["setTrackerUrl", …])`, and PostHog's `api_host` took the same host, so a record with an allowlisted `scriptUrl` and an arbitrary `endpointUrl` posted telemetry to an undeclared host. New `validateEndpointUrl()` applies the same rule as the script URL (same-origin path OK; otherwise https + allowlisted host), enforced for MATOMO and POSTHOG.

Tests: regression test pinning the exfiltration shape (allowlisted script + off-allowlist endpoint), the http-endpoint case, the same-origin case, and both PostHog paths. Suite **51/51 green** on this branch.

PR #1587, opened 2026-08-05
fix(dashboard): add a breadcrumb trail and restore the page gutter when embedded

Hari-egov:
## Problem

Two issues on `/digit-ui/employee/dashboard`, both reported from cms-pilot.

**1. No way back.** The embedded dashboard is a leaf route with no in-page navigation. The only exit is the left rail's Home item, which is easy to miss on the collapsed 3rem rail — and is absent entirely for roles whose access-control grants don't include it. Users get stuck on the page.

**2. Missing gutter.** `DashboardLayout` gives `<main>` `tw-p-4 lg:tw-p-6` in the standalone branch but nothing in the embedded branch, substituting a flat `--dashboard-embedded-inset: 12px`. Cards sat almost flush against the left rail, the filter row against the header's bottom border, and the bottom card row against the viewport edge.

## Changes

**Breadcrumb** — new `DashboardBreadcrumb.jsx`, rendered by `DashboardLayout` **only when `embedded`**:
- Standalone is untouched — it already has its own dark `Sidebar` carrying this navigation, and gating on `embedded` also guarantees `<Link>` never mounts outside the host's Router.
- Restores the platform convention (`Home / <page>`) that every other employee screen has, rather than inventing a bespoke back button.
- Labels are `exists()`-guarded with English fallbacks. The dashboard's `t()` deliberately echoes the KEY when unseeded so gaps surface in QA — a breadcrumb reading `DASHBOARD_BREADCRUMB_CURRENT` on a live env would be worse than useless. Same pattern `DashboardHeader` already uses for `DASHBOARD_HEADER_TITLE`.
- `DASHBOARD_BREADCRUMB_HOME` falls back to the platform's existing `ACTION_TEST_HOME`, so a tenant that already localized its nav gets a consistent word for free.

**Padding** — inset now matches the standalone shell: **16px, 24px from 1024px up**. Vertical rhythm moves to its own `--dashboard-embedded-inset-y` so the grid's floor gap isn't dragged around by the horizontal value at wide viewports. Adds `padding-bottom` on the grid wrap and `padding-top` on the filters bar.

## Verification

Built the bundle and measured the compiled CSS in a browser at both breakpoints. Breadcrumb, title, filter row and KPI cards all land on the same gutter with no straggler:

| viewport | inset | breadcrumb | title | filters | cards | grid pad-bottom |
|---|---|---|---|---|---|---|
| 900px | 16px | 16 | 16 | 16 | 16 | 16px |
| 1440px | 24px | 24 | 24 | 24 | 24 | 20px |

Breadcrumb renders `Home / Dashboard`, current crumb at weight 600, Home links to `/${contextPath}/employee`.

## Notes

- Both stylesheets are edited identically — `dashboard.css` is the generated artifact, and this block sits outside the tailwind layers, so regeneration passes it through verbatim.
- **Localization follow-up:** `DASHBOARD_BREADCRUMB_HOME` / `DASHBOARD_BREADCRUMB_CURRENT` / `DASHBOARD_BREADCRUMB_ARIA_LABEL` need seeding in `en_IN` + `pt_PT` on local / cms-pilot / mctd. Until then the English fallbacks render, so this is safe to ship first.

### DIFF
--- a/digit-ui-esbuild/products/dashboard/src/components/DashboardBreadcrumb.jsx
+++ b/digit-ui-esbuild/products/dashboard/src/components/DashboardBreadcrumb.jsx
@@ -1,0 +1,70 @@
+import React from "react";
+import { Link } from "react-router-dom";
+import useDashboardT from "../i18n/useDashboardT";
+
+/**
+ * Breadcrumb trail for the EMBEDDED dashboard (mounted inside the DigitUI
+ * employee chrome by products/dashboard/Module.js).
+ *
+ * Why this exists: the embedded dashboard is a leaf route with no in-page way
+ * back — the only exit was the left rail's Home item, which is easy to miss on
+ * the collapsed 3rem rail and disappears entirely for roles whose access-control
+ * grants don't include it. Every other employee screen (see the PGR module's
+ * BreadCrumbs usage in products/pgr/src/pages/employee/index.js) carries a
+ * "Home / <page>" trail, so this restores the platform convention rather than
+ * inventing a bespoke back button.
+ *
+ * Standalone mode is unaffected: DashboardLayout only renders this when
+ * `embedded`, so <Link> is never mounted outside the host's Router.
+ *
+ * Labels resolve through the dashboard's own t(), which deliberately echoes the
+ * KEY when a message is unseeded so gaps surface in QA. A breadcrumb showing
+ * "DASHBOARD_BREADCRUMB_CURRENT" would be worse than useless on a live env, so
+ * every label goes through `label()`, which falls back to English until the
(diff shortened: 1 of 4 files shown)


Hari-egov:
Superseded by #1588, which targets `release-v2.12-moz` — the branch this fix is actually needed on (reported from cms-pilot).

Not retargeted, because all four touched files differ between `master` and `release-v2.12-moz` and moz is 132 commits behind `master`; changing the base here would have pulled all of that into the diff. #1588 is a clean cherry-pick cut off moz, re-built and re-verified against moz's own compiled CSS.

Issue #1604, opened 2026-08-05
Remove unauthenticated Kong observability routes; gate nginx /grafana/ behind a flag


singhalkarun:
Part of #1601 (monitoring stack umbrella). **P0-security · effort: S**

## Problem

Two Kong routes proxy to the observability stack with no auth plugin:

- `tempo-service` / `tempo-route` → `http://tempo:3200`, path `/tempo`, `strip_path: true` (`local-setup/kong/kong.yml:709-717`). Exposes the entire trace API. It has no legitimate consumer — Grafana queries Tempo server-side over the container network, and the browser never touches it.
- `grafana-service` / `grafana-route` → `http://grafana:3000`, path `/grafana`, `preserve_host: true` (`local-setup/kong/kong.yml:673-684`).

The Grafana route creates a fall-through problem. In static UI mode the nginx `location /` catch-all proxies to Kong, so a tenant that removes or disables the nginx `/grafana/` location is **still** served Grafana through Kong. Any nginx-level gating is cosmetic while that Kong route exists.

Separately, the nginx `/grafana/` location (`local-setup/ansible/templates/nginx-site.conf.j2:127-137`) is rendered unconditionally — unlike the Gatus `/status/` block directly below it, which is correctly gated behind `nginx_features.status` (default `false`).

## Fix

- Delete both Kong route/service pairs, including the `request-transformer` plugin attached to `grafana-service`.
- Wrap the nginx `/grafana/` location in a new feature flag:

```jinja
{% if nginx_features.grafana %}
  location /grafana/ {
    ...existing block unchanged...
  }
{% endif %}
```

- Add `grafana: true` to `nginx_features` in `ansible/inventory/group_vars/digit.yml`. Default **on**: after #1602 it carries a real login, and it is the team's dashboard entry point. Hardened tenants set it to `false` and Grafana becomes genuinely unreachable from outside, with no fall-through path.
- Local development reaches Grafana at `localhost:13000` directly — the loopback bind from #1603 is still reachable from one's own machine.

## Acceptance

- [ ] `https://<domain>/tempo/api/search` → 404
- [ ] With `nginx_features.grafana: false`, `https://<domain>/grafana/` → 404 (confirms no Kong fall-through)
- [ ] With the flag on, Grafana loads and presents its login page
- [ ] Kong config still validates (`kong config parse` / deck lint in CI) after the route removal
- [ ] No other service references the removed routes (grep `tags: [observability]` consumers)

[...]

PR #1605, opened 2026-08-05
fix(grafana): require a login instead of anonymous Admin

### LINKED ISSUE #1602: Grafana deploys with no authentication (anonymous Admin)

## Problem

`local-setup/docker-compose.egov-digit.yaml:46-48` deploys Grafana with anonymous access enabled at org role **Admin**, the login form disabled, and no admin password set anywhere:

```yaml
GF_AUTH_ANONYMOUS_ENABLED: "true"
GF_AUTH_ANONYMOUS_ORG_ROLE: Admin
GF_AUTH_DISABLE_LOGIN_FORM: "true"
```

This is the shipped configuration for every tenant deployed from this repo. Anyone who reaches the UI is an Admin: they can edit datasources, alter or delete dashboards, and run arbitrary queries against the Loki, Prometheus and Tempo datasources.

The Loki datasource is the sharpest edge — Kong access logs are shipped into Loki and contain bearer tokens in query strings, so unauthenticated Grafana access is effectively unauthenticated access to credentials in the log store.

## Fix

- Anonymous access off by default, exposed as a per-tenant knob (`GF_ANONYMOUS_ENABLED`, default `false`). If a tenant ever enables it, the role is `Viewer` — never `Admin`.
- Re-enable the login form (`GF_AUTH_DISABLE_LOGIN_FORM: "false"`).
- Render `GF_SECURITY_ADMIN_PASSWORD` into the per-tenant `.env` from OpenBao, following the existing Keycloak-secrets pattern (`digit.env.j2`, blockinfile append after the secret fetch). Add `grafana_admin_password` to `host_vars/_example.yml`.
- Add an Ansible task after the Grafana health gate:

```yaml
- name: Grafana — enforce admin password (idempotent reset)
  ansible.builtin.command: >
    docker exec digit-grafana
    grafana cli admin reset-admin-password {{ grafana_admin_password }}
  changed_when: false
  no_log: true
```

This last step is not optional: `GF_SECURITY_ADMIN_PASSWORD` only seeds the admin user on the **first** boot of a fresh `grafana_data` volume. Boxes already deployed have an admin row written with the old default and would silently keep it.

## Acceptance

- [ ] Unauthenticated `GET /grafana/` returns the login page
- [ ] Anonymous `GET /grafana/api/datasources` returns 401
- [ ] The OpenBao-sourced password logs in on a **fresh** box
- [ ] The OpenBao-sourced password logs in on an **already-deployed** box (proves the reset task works)
- [ ] Provisioned dashboards and datasources still load after the auth change

## Notes

Ships together with #1604 (ingress consolidation) — the login is what replaces network-level obscurity, so the route removal and the auth change belong in one PR.

singhalkarun:
Fixes #1602

## What

Grafana shipped with anonymous access at org role **Admin**, the login form disabled, and no admin password set anywhere — on every tenant deployed from this repo. Anyone reaching the UI could edit datasources and query the Loki datasource, which holds Kong access logs containing bearer tokens in query strings.

## Changes

**`docker-compose.egov-digit.yaml`**
- `GF_AUTH_ANONYMOUS_ENABLED` → `${GF_ANONYMOUS_ENABLED:-false}` (opt-in per tenant)
- `GF_AUTH_ANONYMOUS_ORG_ROLE` → `Viewer` (never Admin, even when enabled)
- `GF_AUTH_DISABLE_LOGIN_FORM` → `"false"`
- new `GF_SECURITY_ADMIN_USER` / `GF_SECURITY_ADMIN_PASSWORD`, fed from `.env`

**`playbook-deploy.yml`**
- resolve `grafana_admin_password` from OpenBao; generate a 24-char one when absent and persist it back (merged into the existing secret map, never replacing it)
- write `GF_ADMIN_PASSWORD` into the existing `TENANT SECRETS` block in `.env`
- after the stack is up: wait for Grafana health, then `grafana cli admin reset-admin-password`

**`host_vars/_example.yml`** — documents the optional `grafana_admin_password` bootstrap secret.

## Why the generate-and-persist step

The `bootstrap_secrets` seed writes with `cas: 0` — it only fires when no version of the tenant secret exists. Tenants seeded before this key existed would therefore never receive one. Generating on first deploy and storing it back means no operator action, no shipped default password, and a value that stays stable across deploys.

Operators read it with:
```
bao kv get -field=grafana_admin_password <secrets_path>
```

## Why the reset task

`GF_SECURITY_ADMIN_PASSWORD` only seeds the admin user on the **first** boot of a fresh `grafana_data` volume. Every already-deployed box has an admin row from the era when the login form was disabled, so without the reset they would keep unknown legacy credentials for a UI this PR just closed off. The task is idempotent (`changed_when: false`) and gated on the health check.

## Testing

Not yet deployed. Needs a run on the localhost/WSL2 template box, then a scratch tenant, before any live box:

- [ ] fresh deploy: `/grafana/` shows a login page; OpenBao password works
- [ ] redeploy of an **existing** box: same, proving the reset path
- [ ] anonymous `GET /grafana/api/datasources` → 401
- [ ] provisioned dashboards + datasources still load after login
- [ ] `GF_ANONYMOUS_ENABLED=true` yields a Viewer-role anonymous board

## Notes

Pairs with #1604 (removes the unauthenticated Kong `/grafana` route and gates the nginx location). Either can merge first; together they make the login the only way in.

## Summary by CodeRabbit

* **New Features**
  * Added persistent Grafana administrator password management, with automatic secure generation when no password is configured.
  * Added optional configuration for providing a Grafana administrator password.
  * Grafana credentials can now be configured through environment variables.

* **Security**
  * Disabled anonymous Grafana access by default.
  * Anonymous access, when enabled, is limited to viewer permissions.
  * Restored Grafana login form access.

### DIFF
--- a/local-setup/ansible/playbook-deploy.yml
+++ b/local-setup/ansible/playbook-deploy.yml
@@ -100,6 +100,14 @@
       args:
         chdir: "{{ digit_dir }}"
         executable: /bin/bash
+      # GF_SECURITY_ADMIN_PASSWORD in the compose file is `${GF_ADMIN_PASSWORD:?}`
+      # (no default, on purpose — see the grafana service there). Compose
+      # interpolates for `down` too, so a box whose previous deploy aborted
+      # before .env got the real value would fail this teardown on an
+      # unresolvable variable. The value is irrelevant to `down`; feed it a
+      # placeholder rather than let a missing secret block a teardown.
+      environment:
+        GF_ADMIN_PASSWORD: unused-by-down
       register: clean_teardown
       changed_when: true
       when:
@@ -1577,6 +1585,30 @@
       when:
         - build_digit_ui | default(false)
 
+    # ---- Grafana admin password BEFORE the first compose-up ----
+    # docker-compose.egov-digit.yaml resolves GF_SECURITY_ADMIN_PASSWORD from
+    # ${GF_ADMIN_PASSWORD:?} — no default, so an unset variable aborts the whole
+    # compose invocation (that is the point: a `:-admin` default would ship
+    # admin/admin to anyone running the file standalone). But the authoritative
+    # value only exists further down, after the OpenBao bootstrap — which itself
+    # needs the stack up. Same ordering trap the KEYCLOAK/novu prep blocks below
+    # solve, and the same fix: put a value in .env before anything runs compose.
+    #
+    # It is random per run, not a fixed placeholder, because it IS the password
+    # Grafana seeds its admin row with on a first boot (GF_SECURITY_ADMIN_PASSWORD
+    # is only read when grafana_data is empty). If a deploy dies between here and
+    # the OpenBao block, the box is left with an unguessable admin password
+    # instead of admin/admin. The cost is that Grafana gets recreated once more
+    # per deploy, when the OpenBao value replaces this line further down.
+    - name: "Grafana prep — write a GF_ADMIN_PASSWORD into .env BEFORE compose-up"
+      ansible.builtin.lineinfile:
+        path: "{{ digit_dir }}/.env"
+        regexp: '^GF_ADMIN_PASSWORD='
+        line: "GF_ADMIN_PASSWORD={{ lookup('password', '/dev/null length=24 chars=ascii_letters,digits') }}"
+        create: true
+        mode: '0600'
+      no_log: true
+
     # ---- Keycloak readiness BEFORE the dependent compose-up (converge-abort fix) ----
     # The compose-up below waits on keycloak: service_healthy, but the keycloak DB
     # secret is otherwise written to .env only AFTER it (the OpenBao block further
@@ -2279,13 +2311,19 @@
     # first run the stored value is reused, which keeps the password stable
     # across deploys. Operators read it back with:
     #   bao kv get -field=grafana_admin_password {{ secrets_path }}
+    #
+    # `or`, not `ternary`. Jinja evaluates every filter argument before calling
+    # the filter, so `… | ternary(stored, lookup('password', …))` runs the
+    # generator on every deploy and throws the result away whenever OpenBao
+    # already has a value. `or` compiles to Python's `or` and short-circuits, so
+    # the lookup only runs when there is nothing stored — which is also what the
+    # task name claims. (default('', true) turns both "undefined" and "" into the
+    # falsey '' that `or` needs.)
     - name: "OpenBao — resolve Grafana admin password (generate if absent)"
       ansible.builtin.set_fact:
         grafana_admin_pw: >-
-          {{ bao_secrets.json.data.data.grafana_admin_password
-             | default('', true)
-             | ternary(bao_secrets.json.data.data.grafana_admin_password | default(''),
-                       lookup('password', '/dev/null length=24 chars=ascii_letters,digits')) }}
+          {{ bao_secrets.json.data.data.grafana_admin_password | default('', true)
+             or lookup('password', '/dev/null length=24 chars=ascii_letters,digits') }}
       no_log: true
 
     - name: "OpenBao — persist generated Grafana admin password"
@@ -2334,10 +2372,24 @@
           AWS_SECRETKEY={{ bao_secrets.json.data.data.minio_root_password }}
           ELASTICSEARCH_MASTER_PASSWORD={{ bao_secrets.json.data.data.elasticsearch_master_password }}
           EGOV_HRMS_DEFAULT_PASSWORD={{ bao_secrets.json.data.data.egov_hrms_default_password }}
-          GF_ADMIN_PASSWORD={{ grafana_admin_pw }}
         marker: "# {mark} TENANT SECRETS"
         create: false           # .env was already written by the digit.env.j2 template earlier
         mode: '0600'            # tighten — was 0644 from the template; secrets bump to 0600
+      no_log: true
+
+    # GF_ADMIN_PASSWORD is the one OpenBao-backed value that is NOT in the block
+    # above: "Grafana prep" already wrote a pre-boot line for it before the stack
+    # came up (the compose file has no default for it), and appending a second
+    # line inside the block would leave two GF_ADMIN_PASSWORD entries in .env
+    # with the winner decided by file order. Replace the existing line in place
+    # instead, so .env always has exactly one — now the authoritative value.
+    - name: "OpenBao — put the stored Grafana admin password into .env"
+      ansible.builtin.lineinfile:
+        path: "{{ digit_dir }}/.env"
+        regexp: '^GF_ADMIN_PASSWORD='
+        line: "GF_ADMIN_PASSWORD={{ grafana_admin_pw }}"
+        create: false
+        mode: '0600'
       no_log: true
 
     # Keycloak-tier secrets — appended as a separate marked block so the
@@ -4212,12 +4264,30 @@
     # the failure mode it leaves behind is silent: the UI keeps whatever password
     # the volume was seeded with, nobody can log in with the OpenBao value, and a
     # `failed_when: false` would bury that in a 4000-line deploy log.
+    #
+    # --password-from-stdin, NOT the positional argument. As an argument the
+    # password lands in the host's process table and the container's argv, and
+    # /proc/<pid>/cmdline is world-readable — so every `no_log` on the Ansible
+    # side is undone for the lifetime of the exec by any local user running
+    # `ps aux`. The CLI reads one line off stdin instead (bufio.Scanner in
+    # pkg/cmd/grafana-cli/commands/reset_password_command.go; present in the
+    # pinned grafana 11.4.0), and `docker exec -i` is what forwards it.
+    #
+    # failed_when is explicit because a non-zero rc is not the only failure:
+    # resetPasswordCommand returns nil when the admin user cannot be found, so
+    # the CLI exits 0 after printing "Admin user cannot be found". Gate on the
+    # success line so that case is loud too.
     - name: Grafana — align admin password with OpenBao (idempotent)
       ansible.builtin.command: >-
-        docker exec digit-grafana
-        grafana cli admin reset-admin-password {{ grafana_admin_pw }}
+        docker exec -i digit-grafana
+        grafana cli admin reset-admin-password --password-from-stdin
+      args:
+        stdin: "{{ grafana_admin_pw }}"
       register: grafana_pw_reset
       changed_when: false
+      failed_when: >-
+        grafana_pw_reset.rc != 0
+        or 'Admin password changed successfully' not in grafana_pw_reset.stdout
       no_log: true
       when: grafana_health.rc == 0
 
@@ -4229,9 +4299,11 @@
         msg: >-
           Grafana: {{ 'admin password aligned with OpenBao'
                       if grafana_health.rc == 0
-                      else 'UNHEALTHY — admin password NOT applied. Once it is up, rerun the deploy or:
-                            docker exec digit-grafana grafana cli admin reset-admin-password "$(bao kv get -field=grafana_admin_password '
-                           ~ secrets_path ~ ')"' }}
+                      else 'UNHEALTHY — admin password NOT applied. Once it is up, rerun the deploy or (piped, so the
+                            password stays out of ps aux):
+                            bao kv get -field=grafana_admin_password '
+                           ~ secrets_path ~
+                           ' | docker exec -i digit-grafana grafana cli admin reset-admin-password --password-from-stdin' }}
 
     # ==================== Claude Code Setup (opt-in) ====================
     # Installs Claude Code CLI on the target and syncs the controller's
--- a/local-setup/ansible/playbook-deploy.yml
+++ b/local-setup/ansible/playbook-deploy.yml
@@ -2295,10 +2295,26 @@
         headers:
           X-Vault-Token: "{{ bao_root_token }}"
         body_format: json
-        body:
-          # Merge, never replace: a bare {grafana_admin_password: …} write would
-          # drop every other key in this tenant's secret.
-          data: "{{ bao_secrets.json.data.data | combine({'grafana_admin_password': grafana_admin_pw}) }}"
+        # The whole body is ONE template expression on purpose: Ansible preserves
+        # native types that way, so `cas` reaches OpenBao as a JSON number. Split
+        # across YAML keys it would render as the string "5" and the CAS check
+        # would depend on server-side weak decoding.
+        #
+        # cas = the version we just read. Without it this is a blind overwrite of
+        # the entire secret: anything written between our read and this write (a
+        # concurrent deploy, an operator running `bao kv put`) is silently lost,
+        # because `data` is built from the snapshot we read. With cas, a racing
+        # write makes OpenBao reject ours (400) and the deploy stops rather than
+        # clobbering. No retry loop: deploys are serialised per tenant, so a
+        # conflict means something genuinely unexpected touched the secret and a
+        # human should look before we overwrite it.
+        #
+        # `data` merges, never replaces — a bare {grafana_admin_password: …}
+        # write would drop every other key in this tenant's secret.
+        body: >-
+          {{ {'options': {'cas': bao_secrets.json.data.metadata.version | int},
+              'data': bao_secrets.json.data.data
+                      | combine({'grafana_admin_password': grafana_admin_pw})} }}
         status_code: [200]
       when: (bao_secrets.json.data.data.grafana_admin_password | default('', true)) | length == 0
       no_log: true
@@ -4191,22 +4207,31 @@
       changed_when: false
       ignore_errors: yes
 
+    # Fatal when it runs. Grafana is healthy at this point, so a failing reset is
+    # a genuine anomaly (wrong CLI path after an image bump, unwritable DB), and
+    # the failure mode it leaves behind is silent: the UI keeps whatever password
+    # the volume was seeded with, nobody can log in with the OpenBao value, and a
+    # `failed_when: false` would bury that in a 4000-line deploy log.
     - name: Grafana — align admin password with OpenBao (idempotent)
       ansible.builtin.command: >-
         docker exec digit-grafana
         grafana cli admin reset-admin-password {{ grafana_admin_pw }}
       register: grafana_pw_reset
       changed_when: false
-      failed_when: false
       no_log: true
       when: grafana_health.rc == 0
 
+    # Grafana never came up. Non-fatal, matching the Loki gate above — the
+    # dashboards are not a serving dependency — but say so loudly, because the
+    # stored password has NOT been applied to the running instance.
     - name: Show Grafana admin-password status
       debug:
         msg: >-
           Grafana: {{ 'admin password aligned with OpenBao'
-                      if (grafana_pw_reset.rc | default(1)) == 0
-                      else 'admin password NOT set — log in with the previous credentials and check `bao kv get ' ~ secrets_path ~ '`' }}
+                      if grafana_health.rc == 0
+                      else 'UNHEALTHY — admin password NOT applied. Once it is up, rerun the deploy or:
+                            docker exec digit-grafana grafana cli admin reset-admin-password "$(bao kv get -field=grafana_admin_password '
+                           ~ secrets_path ~ ')"' }}
 
     # ==================== Claude Code Setup (opt-in) ====================
     # Installs Claude Code CLI on the target and syncs the controller's
--- a/local-setup/ansible/playbook-deploy.yml
+++ b/local-setup/ansible/playbook-deploy.yml
@@ -2273,6 +2273,36 @@
       register: bao_secrets
       no_log: true
 
+    # Grafana admin password. Tenants seeded BEFORE this key existed can't get
+    # it from the bootstrap seed above (that write is cas:0 — it only fires when
+    # no version exists yet), so generate one here and persist it. After the
+    # first run the stored value is reused, which keeps the password stable
+    # across deploys. Operators read it back with:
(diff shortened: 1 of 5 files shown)


singhalkarun:

```
```

singhalkarun:
Thanks — worked through all three. Two applied in b0f5cf1, one declined with reasoning.

### 1. CAS on the OpenBao write — **applied**

Correct and worth fixing. `data` is built from a snapshot read earlier in the play, so any key written between the read and the write was silently lost. Now sends `options.cas` pinned to the version we read.

One implementation detail worth flagging for reviewers: the body is deliberately a **single template expression** rather than nested YAML keys. Ansible only preserves native types for whole-string templates, so `cas: "{{ … | int }}"` as its own key would serialise as the JSON *string* `"5"` and the CAS check would depend on OpenBao's weak decoding rather than being unambiguous.

I did **not** add the retry-on-conflict loop you suggested. Deploys are serialised per tenant, so a CAS conflict means something genuinely unexpected touched the secret mid-deploy — silently re-reading and overwriting is exactly the behaviour the CAS is there to prevent. Failing loudly is the safer default; a retry can be added if a real concurrent-deploy workflow ever appears.

### 2. `failed_when: false` on the password reset — **applied**

Agreed, and the failure mode was worse than it looks: the reset silently not running leaves the instance on whatever password its volume was seeded with, so nobody can log in with the OpenBao value — and this PR is what removes anonymous access, so that's a lockout. Now fatal when it runs (already gated on Grafana being healthy).

The unhealthy-Grafana branch stays non-fatal, matching the existing Loki gate directly above it (dashboards aren't a serving dependency), but it now prints the exact recovery command instead of a vague warning.

### 3. Removing the `admin` fallback in compose — **declined, with reasoning**

I don't think either variant is safe here, because of an ordering constraint in the playbook:

- `Start DIGIT stack` is at `playbook-deploy.yml:1741`
- the block that writes `GF_ADMIN_PASSWORD` into `.env` is at `:2306`

So on a first deploy the stack starts **before** that variable exists. Making it mandatory (`${GF_ADMIN_PASSWORD:?…}`) would hard-fail every fresh deploy at compose-up.

Dropping the default outright (`${GF_ADMIN_PASSWORD}`) is worse than it appears: compose passes an empty string, and Grafana falls back to its own built-in `admin` default anyway. Same outcome, less visible.

What actually makes this safe is the reset task at the end of the play: whatever password Grafana boots with, the final state is the OpenBao value. The exposure window is one deploy, on a port that is loopback-only after #1606. The `:-admin` fallback only persists for a bare `docker compose -f docker-compose.egov-digit.yaml up` outside Ansible, which isn't a supported path for this file — CI composes `docker-compose.yml` / `docker-compose.deploy.yaml`, and the other consumers parse the YAML as text.

Closing that window properly means moving password resolution before compose-up, which needs OpenBao reachable at that point — it isn't guaranteed on a fresh non-Keycloak tenant, since `openbao` is only started early inside the Keycloak-prep block (`:1592`). That's a real improvement but a separate change with its own ordering risk, not something to fold into this PR.

singhalkarun:
**Self-review addition.** Going over the diff again, `docker-compose.registry.yml` had an identical Grafana block — `GF_AUTH_ANONYMOUS_ENABLED: "true"`, `GF_AUTH_ANONYMOUS_ORG_ROLE: Admin`, login form disabled — that the original commit missed.

It is not the Ansible path (`local-setup/README.md:324` says the playbook never references it), but it is a documented alternative stack, so it would have remained a one-command route to an anonymous-Admin Grafana while this PR claimed to close exactly that.

Fixed in the same PR. There is no OpenBao on that path, so the password comes from the environment with an `admin/admin` fallback and a comment saying so plainly — the meaningful change there is that anonymous Admin is gone and the login form is back.

vinothrallapalli-eGov (approved):

local-setup/docker-compose.registry.yml
```diff
@@ -43,9 +43,15 @@ services:
...
+      # Same posture as docker-compose.egov-digit.yaml. This file is NOT the
+      # Ansible path (no .env from OpenBao), so set GF_ADMIN_PASSWORD in your
+      # shell or a local .env before bringing the stack up — otherwise the
+      # admin login is admin/admin.
+      GF_AUTH_ANONYMOUS_ENABLED: ${GF_ANONYMOUS_ENABLED:-false}
+      GF_AUTH_ANONYMOUS_ORG_ROLE: Viewer
+      GF_AUTH_DISABLE_LOGIN_FORM: "false"
+      GF_SECURITY_ADMIN_USER: ${GF_ADMIN_USER:-admin}
+      GF_SECURITY_ADMIN_PASSWORD: ${GF_ADMIN_PASSWORD:-admin}
       GF_SERVER_DOMAIN: api.egov.theflywheel.in
```
vinothrallapalli-eGov: In the docker compose files we should not hard code the domain url, it should get overrided from <tenant.yaml>?

local-setup/docker-compose.egov-digit.yaml
```diff
@@ -43,9 +43,19 @@ services:
...
     environment:
-      GF_AUTH_ANONYMOUS_ENABLED: "true"
-      GF_AUTH_ANONYMOUS_ORG_ROLE: Admin
-      GF_AUTH_DISABLE_LOGIN_FORM: "true"
+      # Auth. Anonymous access is OFF by default: Grafana proxies the Loki
+      # datasource, and Kong access logs in Loki carry bearer tokens in query
+      # strings — so "read-only" anonymous is still credential disclosure.
+      # A tenant that wants a public board sets GF_ANONYMOUS_ENABLED=true in
+      # .env; the role stays Viewer, never Admin.
+      GF_AUTH_ANONYMOUS_ENABLED: ${GF_ANONYMOUS_ENABLED:-false}
```
vinothrallapalli-eGov, line 67: The comment above (lines 46-50) explains that read-only anonymous access is still credential disclosure because Grafana proxies Loki, and Kong access logs in Loki carry bearer tokens in query strings. But the `GF_ANONYMOUS_ENABLED=true` escape hatch this comment describes grants exactly that read-only (Viewer) access — it doesn't re-introduce Admin, but it does re-introduce the disclosure path the same comment just warned about, for any dashboard/panel backed by the Loki datasource. Worth an explicit callout here (or in `_example.yml`) that enabling this flag exposes Loki-backed dashboards' bearer tokens to anonymous users, so operators don't reach for it thinking Viewer-role == safe.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -4146,6 +4193,46 @@
...
+
+    # Fatal when it runs. Grafana is healthy at this point, so a failing reset is
+    # a genuine anomaly (wrong CLI path after an image bump, unwritable DB), and
+    # the failure mode it leaves behind is silent: the UI keeps whatever password
+    # the volume was seeded with, nobody can log in with the OpenBao value, and a
+    # `failed_when: false` would bury that in a 4000-line deploy log.
+    - name: Grafana — align admin password with OpenBao (idempotent)
+      ansible.builtin.command: >-
+        docker exec digit-grafana
+        grafana cli admin reset-admin-password {{ grafana_admin_pw }}
```
vinothrallapalli-eGov: `grafana_admin_pw` is passed as a literal `docker exec` / `grafana cli` argument. That puts the password in the host's process table and the container's argv (readable via `ps aux`/`/proc/<pid>/cmdline` by any local user who can list processes), even though the Ansible side is careful to `no_log` every step that touches it. Consider piping it via stdin/env file into the container instead (e.g. `grafana cli admin reset-admin-password` supports reading from an env var in newer Grafana, or use `docker exec -e`/a short-lived file under a 0600 perm removed right after) to avoid the argv leak.

local-setup/docker-compose.egov-digit.yaml
```diff
@@ -43,9 +43,19 @@ services:
...
+      GF_AUTH_ANONYMOUS_ORG_ROLE: Viewer
+      GF_AUTH_DISABLE_LOGIN_FORM: "false"
+      # Admin password comes from OpenBao via .env (key: grafana_admin_password,
+      # auto-generated on first deploy). The compose fallback only applies to a
+      # bare `docker compose up` outside Ansible.
+      GF_SECURITY_ADMIN_USER: ${GF_ADMIN_USER:-admin}
+      GF_SECURITY_ADMIN_PASSWORD: ${GF_ADMIN_PASSWORD:-admin}
```
vinothrallapalli-eGov: `GF_SECURITY_ADMIN_PASSWORD: ${GF_ADMIN_PASSWORD:-admin}` falls back to the classic `admin/admin` when `GF_ADMIN_PASSWORD` isn't set. This compose file is also the one exercised by the plain `docker compose up -d` quickstart documented for this repo (not just the Ansible/OpenBao path), so anyone following that quickstart without manually setting `GF_ADMIN_PASSWORD` gets a weak, well-known default credential behind the now-enabled login form — narrower than the previous no-password Admin-anonymous hole, but still worth a startup-time warning (or failing closed) rather than a silent fallback, since the whole point of this PR is to stop shipping guessable/default Grafana admin access.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -2273,6 +2273,52 @@
...
+    # first run the stored value is reused, which keeps the password stable
+    # across deploys. Operators read it back with:
+    #   bao kv get -field=grafana_admin_password {{ secrets_path }}
+    - name: "OpenBao — resolve Grafana admin password (generate if absent)"
+      ansible.builtin.set_fact:
+        grafana_admin_pw: >-
+          {{ bao_secrets.json.data.data.grafana_admin_password
+             | default('', true)
+             | ternary(bao_secrets.json.data.data.grafana_admin_password | default(''),
+                       lookup('password', '/dev/null length=24 chars=ascii_letters,digits')) }}
```
vinothrallapalli-eGov: Nit: `lookup('password', '/dev/null length=24 ...')` is an argument to the `ternary` filter, and Jinja evaluates all filter arguments before the filter runs — it isn't short-circuited like a Python `if/else`. So this lookup (and its file-I/O against `/dev/null`) actually runs on *every* deploy, even when `grafana_admin_password` already exists in OpenBao and the generated value is immediately discarded by `ternary`. Harmless today (no_log hides it, the discarded value is never used), but worth a comment so a future reader doesn't assume 'generate if absent' means the lookup itself is conditional.

vinothrallapalli-eGov (commented):
Reviewed the diff against the prior devbox-verification claims. The core claims check out structurally: `GF_AUTH_ANONYMOUS_ENABLED` now defaults to `false` (anon → 401 by default), `GF_SECURITY_ADMIN_PASSWORD` is fed from the OpenBao-generated `grafana_admin_pw`, and the opt-in anonymous escape hatch hardcodes `GF_AUTH_ANONYMOUS_ORG_ROLE: Viewer` (not templated, so it can't be escalated to Admin via env var) — consistent with 'POST as that user returns 403'. The OpenBao CAS write and the 'fail loudly on reset failure' issues that CodeRabbit flagged on an earlier revision are already fixed in this diff (cas is present, and the reset task no longer swallows failures). Four remaining points below, none blocking: one is a real residual-risk gap in the new opt-in anonymous path, two are hardening-completeness nits, one is a no-op efficiency nit. No duplicate of the still-open `docker-compose.registry.yml` hardcoded-domain thread from the other human reviewer.

local-setup/docker-compose.egov-digit.yaml
singhalkarun: Valid, and this was the one real hole left — fixed in 381b365e.

**Correcting one detail first:** the plain `docker compose up -d` quickstart (README "Option A") loads `docker-compose.yml`, which has no `grafana` service at all — `grep -cE '^  grafana:'` is 1 only for `docker-compose.egov-digit.yaml` and `docker-compose.registry.yml`. So the quickstart never saw the fallback. But your point stands for the *other* documented path: both files' own headers say `Usage: docker compose -f <file> up -d`, and that path did get admin/admin forever.

**Why it couldn't just be `:?` on its own.** The Ansible path boots the whole stack (`Start DIGIT stack`, playbook-deploy.yml:1744) *before* OpenBao is even initialised (the bootstrap block starts at :2152). So on a fresh box there is no `GF_ADMIN_PASSWORD` in `.env` when compose first runs, and compose interpolates for every subcommand — I confirmed `down` fails too:

```
$ docker compose -f docker-compose.egov-digit.yaml down --remove-orphans --timeout 30
error while interpolating services.grafana.environment.GF_SECURITY_ADMIN_PASSWORD:
required variable GF_ADMIN_PASSWORD is missing a value: ...
```

**What I did instead** — fail closed *and* make the Ansible path supply the value first, which is the pattern this playbook already uses twice (`Keycloak prep — write KEYCLOAK SECRETS into .env BEFORE compose-up`, `novu prep — write pinned NOVU_API_KEY into .env BEFORE compose-up`):

1. `${GF_ADMIN_PASSWORD:?...}` in both compose files — no default, with the tradeoff spelled out in the comment.
2. New `Grafana prep — write a GF_ADMIN_PASSWORD into .env BEFORE compose-up`, placed ahead of every compose invocation in the play. It writes a **random** 24-char value, not a fixed placeholder, on purpose: `GF_SECURITY_ADMIN_PASSWORD` is what seeds the admin row on a first boot, so a deploy that dies between there and the OpenBao block now leaves an unguessable password instead of admin/admin. That window was previously real — the play can abort at `Start DIGIT stack`.
3. `GF_ADMIN_PASSWORD` moved out of the `TENANT SECRETS` blockinfile into a `lineinfile` with the same regexp, so `.env` holds exactly one entry that gets replaced in place by the OpenBao value rather than two lines whose winner depends on file order. Simulated the full sequence (template → prep → blockinfile → OpenBao lineinfile): 1 line, `0600`, final value = OpenBao's.
4. `force-clean — tear down existing DIGIT stack` gets `environment: GF_ADMIN_PASSWORD: unused-by-down`, so a teardown after a half-finished deploy can't be blocked by a missing secret. Verified the play-level `DOCKER_DEFAULT_PLATFORM` still merges through.

Verified statically (no devbox run — the box is serialised right now):

* `docker compose -f docker-compose.egov-digit.yaml config` → exit 1 with the required-variable message when unset; renders fine with the var set or supplied via `--env-file`. Same for `docker-compose.registry.yml`.
* `ansible-lint playbook-deploy.yml` → 0 failures (production profile), `yamllint` unchanged at 16 pre-existing line-length warnings, `npx jest --runInBand static` 35/35, `check-gatus-coverage.py` clean.

**Still wants a runtime check on a real deploy:** the first-boot ordering (prep line present in `.env` before `up -d` on a genuinely fresh box).

local-setup/docker-compose.egov-digit.yaml
singhalkarun: Valid, and worse than a documentation gap — fixed in 381b365e.

**The disclosure path is real.** I traced it end to end rather than trusting my own comment:

* `otel/grafana/provisioning/datasources/tempo.yaml` provisions Loki with `access: proxy`, so Grafana's *backend* runs any Loki query on behalf of the logged-in user — Viewer included. There's a `loki-logs.json` dashboard shipped alongside it.
* `otel/promtail-config.yaml` tails **every** container's stdout (`docker_sd_configs` over all containers) into Loki.
* `egov-user-proxy` is `egovio/nginx:alpine` (access log → stdout), and `kong/kong.yml:152` calls it as
  `http://egov-user-proxy:8107/user/_details?access_token=" .. token`.

So live session tokens sit in Loki as plain log text, and enabling the flag hands an anonymous visitor a token-harvesting console. Admin *write* access stays closed; the credential *disclosure* comes straight back.

**The hatch also didn't work.** The comment told operators to "set `GF_ANONYMOUS_ENABLED=true` in `.env`", but the playbook re-renders `/opt/digit/.env` from `templates/digit.env.j2` on every deploy (playbook-deploy.yml:1258), so that hand-edit is silently discarded on the next run. `grep -rn GF_ANONYMOUS_ENABLED` found it in the compose files only — nothing wired it.

Rather than narrow a knob that didn't function, I made it real *and* honest:

* `grafana_anonymous_enabled` is now a top-level host_vars knob (`_example.yml`), rendered into `.env` by `digit.env.j2`. Confirmed it renders `false`/`true` correctly for undefined / `false` / `true` / `"true"`.
* The compose comment now states the mechanism (Loki proxy + promtail + the `access_token` query string) instead of the vague "Kong access logs", and says outright that Viewer is not a safe role here and not to enable it on an internet-reachable box until `access_token` is scrubbed from the log pipeline.
* `_example.yml` carries the same warning at the point of decision, plus the note that editing `/opt/digit/.env` reverts.

The scrubbing itself (dropping `access_token` from the log pipeline, or stopping Kong putting the token in a query string at all) is the actual fix for the underlying leak — that's a separate change and I've left it out of this PR deliberately.

local-setup/ansible/playbook-deploy.yml
singhalkarun: Valid — fixed in 381b365e.

Confirmed the leak: `docker exec digit-grafana grafana cli admin reset-admin-password <pw>` puts the password in the host `docker` process's argv and the container's, and `/proc/<pid>/cmdline` is world-readable, so every `no_log` around it is undone for the lifetime of the exec by any local user running `ps aux`.

Went with stdin over an env file — the CLI supports it directly. Checked the pinned image's source rather than the docs (the v11.4 CLI docs page doesn't mention the flag, but it is there):

```go
// pkg/cmd/grafana-cli/commands/reset_password_command.go @ v11.4.0
if c.Bool("password-from-stdin") {
    logger.Infof("New Password: ")
    scanner := bufio.NewScanner(os.Stdin)
    ...
```

So the task is now:

```yaml
ansible.builtin.command: >-
  docker exec -i digit-grafana
  grafana cli admin reset-admin-password --password-from-stdin
args:
  stdin: "{{ grafana_admin_pw }}"
```

`-i` is what forwards it. Verified Ansible's `command` module honours `args.stdin` (round-tripped a value through `cat`). The value never appears in any argv on either side of the exec now.

Two extras while I was in there:

* **The "fail loudly" gate wasn't actually loud.** `resetPasswordCommand` `return nil`s after printing `Admin user cannot be found`, so the CLI exits 0 on that path and `rc` alone would never catch it. Added
  `failed_when: rc != 0 or 'Admin password changed successfully' not in stdout`.
  Checked that the success line goes to **stdout** — grafana-cli's logger is plain `fmt.Printf` (`pkg/cmd/grafana-cli/logger/logger.go`), not stderr.
* The manual-recovery hint in the `Show Grafana admin-password status` debug still showed the argv form; it's now the piped `bao kv get … | docker exec -i … --password-from-stdin` form.

**Needs runtime confirmation on a real box:** that `grafana cli admin reset-admin-password --password-from-stdin` behaves as read on the actual `egovio/grafana:11.4.0` image — source-verified only, not executed, since the devbox is serialised.

local-setup/ansible/playbook-deploy.yml
singhalkarun: Valid — you're right about the mechanism, and it's fixable rather than just comment-worthy, so I changed the expression in 381b365e.

Reproduced it with an instrumented lookup:

```
ternary, value present -> FROMBAO   | lookup calls: 1   <-- generated and discarded
or,      value present -> FROMBAO   | lookup calls: 0
or,      value absent  -> GENERATED | lookup calls: 1
or,      value empty   -> GENERATED | lookup calls: 1
```

Jinja compiles `or` to Python's `or`, which short-circuits, so it does what the task name claims while `ternary` can't:

```yaml
grafana_admin_pw: >-
  {{ bao_secrets.json.data.data.grafana_admin_password | default('', true)
     or lookup('password', '/dev/null length=24 chars=ascii_letters,digits') }}
```

`default('', true)` still collapses both "undefined" and "" into the falsey `''` that `or` needs, so the absent/empty/stored cases behave identically to before. Confirmed with a real `ansible-playbook` run against both shapes (stored value passes through untouched; absent generates a fresh 24-char value).

On "does it matter": you're right that it was harmless — `/dev/null` is special-cased by the `password` lookup so there's no file I/O against it and no state is written, and the value was genuinely never used. It's the readability that was off, and `or` fixes that without needing a comment to apologise for the construct. Added a short note anyway explaining why it's `or` and not `ternary`, so nobody "tidies" it back.

local-setup/docker-compose.registry.yml
singhalkarun: Agreed on the principle, with one correction to the mechanism — changed in 381b365e.

`docker-compose.registry.yml` can't be overridden from `<tenant>.yml`, because the playbook never reads this file. It deploys `docker-compose.egov-digit.yaml` plus overlays; `docker-compose.registry.yml` isn't in the `compose_files` fact and appears nowhere in playbook-deploy.yml. `local-setup/README.md` says so explicitly ("**not** `docker-compose.registry.yml`, which it never references"), and the project-structure listing repeats it. So this file is the standalone/manual path only — there's no tenant yaml in play to override anything.

What was genuinely wrong is that it was a *literal* where the sibling file uses a variable, so a standalone user had no way to point Grafana at their own host without editing the compose file. Both now use the same shape as `docker-compose.egov-digit.yaml`:

```yaml
GF_SERVER_DOMAIN: ${GF_SERVER_DOMAIN:-api.egov.theflywheel.in}
GF_SERVER_ROOT_URL: ${GF_SERVER_ROOT_URL:-https://api.egov.theflywheel.in/grafana/}
```

Verified: `docker compose --env-file <f> -f docker-compose.registry.yml config` renders `GF_SERVER_DOMAIN: grafana.example.org` from the env file, and the committed literal when unset.

Also fixed the header of this file while I was there — line 1 said `# docker-compose.egov-digit.yaml` and the `Usage:` line named that file too, copy-pasted from the sibling. Both now name `docker-compose.registry.yml`, which is probably part of why this file reads as if it should be the Ansible one.

These were the only three `theflywheel.in` occurrences in the file; the remaining one is the registry hostname in the header comment, which is accurate.

vinothrallapalli-eGov (approved):

PR #1606, opened 2026-08-05
fix(compose): bind observability ports to loopback

### LINKED ISSUE #1603: All monitoring ports bind 0.0.0.0, bypassing nginx/Kong

## Problem

Every observability service publishes its port on all interfaces in `local-setup/docker-compose.egov-digit.yaml`, so each is reachable directly from outside, bypassing both host nginx and Kong:

| Service | Published | Exposure if reachable |
|---|---|---|
| Grafana | `13000:3000` | the UI directly (see #1602) |
| Loki | `13100:3100` | `auth_enabled: false` — unauthenticated log **read and write**, including Kong access logs containing bearer tokens |
| Prometheus | `19090:9090` | all metrics, plus the lifecycle endpoints (see the follow-up item on `--web.enable-lifecycle`) |
| Tempo | `13200:3200` | the full trace API, unauthenticated |
| otel-collector | `14317`, `14318`, `13133` | OTLP ingest — arbitrary telemetry injection into the pipeline |
| Gatus | `18889:8080` | complete service-health map of the host |

None of these needs to be public. Host nginx proxies to them over `127.0.0.1` (`nginx_upstream_host` defaults to loopback), and containers reach each other over `egov-network` by service name. `openbao` already demonstrates the correct pattern by binding `127.0.0.1:1871`.

## Fix

Prefix all six mappings with `127.0.0.1:`:

```yaml
# before → after
"13000:3000"   → "127.0.0.1:13000:3000"     # grafana
"19090:9090"   → "127.0.0.1:19090:9090"     # prometheus
"13100:3100"   → "127.0.0.1:13100:3100"     # loki
"13200:3200"   → "127.0.0.1:13200:3200"     # tempo
"14317:4317"   → "127.0.0.1:14317:4317"     # otel-collector OTLP gRPC
"14318:4318"   → "127.0.0.1:14318:4318"     # otel-collector OTLP HTTP
"13133:13133"  → "127.0.0.1:13133:13133"    # otel-collector health
"18889:8080"   → "127.0.0.1:18889:8080"     # gatus
```

Public entry points are unchanged: nginx `/grafana/` and `/status/`, and Kong's rate-limited `/otel/v1/metrics` + `/otel/v1/logs` ingest routes for browser telemetry (these travel over the container network and never touch the host port).

## Acceptance

- [ ] The eight mappings time out when probed from off-box
- [ ] Ansible validation tasks (already loopback-based) stay green
- [ ] Browser dashboard telemetry still lands in Prometheus and Loki
- [ ] Grafana still loads through the domain
- [ ] `digit-mcp` tracing tools still work — they default to `localhost:13200`/`13133`/`13000`, which is a host-side path. If the MCP **container** ever consumes them, set `TEMPO_URL=http://tempo:3200` on the service (currently unset).

## Before merging

Grep tenant runbooks and team notes for direct `:13000` / `:19090` / `:13100` / `:18889` usage from off-box. Any such habit breaks and should move to an SSH tunnel (`ssh -L 13000:127.0.0.1:13000 <box>`).

singhalkarun:
Fixes #1603

## What

Every observability service published its port on `0.0.0.0`, reachable directly from off-box and bypassing both host nginx and Kong.

| Service | Was | Now |
|---|---|---|
| otel-collector | `14317`, `14318`, `13133` | `127.0.0.1:` on all three |
| tempo | `13200:3200` | `127.0.0.1:13200:3200` |
| grafana | `13000:3000` | `127.0.0.1:13000:3000` |
| prometheus | `19090:9090` | `127.0.0.1:19090:9090` |
| loki | `13100:3100` | `127.0.0.1:13100:3100` |
| gatus | `18889:8080` | `127.0.0.1:18889:8080` |

Loki is the sharpest case: `auth_enabled: false`, and it ingests Kong access logs which carry bearer tokens in query strings — so the port offered unauthenticated read *and* write over those logs.

## Why this is safe

Host nginx proxies to these over `127.0.0.1` (`nginx_upstream_host` defaults to loopback) and containers reach each other by service name on `egov-network`. `openbao` already binds `127.0.0.1:18200`; this applies the same pattern to the rest.

Public entry points are unchanged: nginx `/grafana/` and `/status/`, and Kong's rate-limited `/otel/v1/metrics` + `/otel/v1/logs` ingest routes, which travel over the container network and never touch a host port.

## Testing

Not yet deployed. Checklist for the template box → scratch tenant run:

- [ ] the eight mappings time out when probed from off-box over **IPv4** (`nc -4 -vz <public-v4> 13000 13100 13133 13200 14317 14318 18889 19090`)
- [ ] the eight mappings time out when probed from off-box over **IPv6** (`nc -6 -vz <public-v6> …`) — called out separately because the pre-fix listeners were on `0.0.0.0` *and* `[::]`. Naming an IPv4 host address in the mapping should leave the v6 socket unbound entirely rather than move it to `::1`, but that is the one part of this change that is daemon-version dependent, so probe it rather than assume it
- [ ] on-box, `ss -lntp | grep -E ':(13000|13100|13133|13200|14317|14318|18889|19090)\b'` shows **only** `127.0.0.1:` rows and no `[::]:` / `*:` rows
- [ ] Ansible validate tasks (already loopback-based) stay green
- [ ] browser dashboard telemetry still lands in Prometheus and Loki
- [ ] Grafana still loads through the domain
- [ ] Gatus `/status/` still renders where enabled

## Operator impact

Anyone curling these ports directly from their laptop now needs a tunnel:

```
ssh -L 13000:127.0.0.1:13000 <box>
```

Worth grepping tenant runbooks for `:13000` / `:19090` / `:13100` / `:18889` before this merges.

Note the tunnel target is `127.0.0.1`, not `localhost` — on a dual-stack box `localhost` can resolve to `::1` first, which is no longer bound.

`digit-mcp`'s tracing tools default to `localhost:13200`/`13133`/`13000`, which is a host-side path and unaffected. If the MCP **container** ever consumes them, set `TEMPO_URL=http://tempo:3200` on the service (currently unset).

### DIFF
--- a/local-setup/docker-compose.egov-digit.yaml
+++ b/local-setup/docker-compose.egov-digit.yaml
@@ -14,9 +14,9 @@ services:
     volumes:
       - ./otel/otel-collector-config.yaml:/etc/otelcol-contrib/config.yaml:ro
     ports:
-      - "14317:4317"
-      - "14318:4318"
-      - "13133:13133"
+      - "127.0.0.1:14317:4317"
+      - "127.0.0.1:14318:4318"
+      - "127.0.0.1:13133:13133"
     networks:
       - egov-network
 
@@ -29,7 +29,7 @@ services:
       - ./otel/tempo-config.yaml:/etc/tempo/config.yaml:ro
       - tempo_data:/var/tempo
     ports:
-      - "13200:3200"
+      - "127.0.0.1:13200:3200"
     healthcheck:
       test: ["CMD", "wget", "--spider", "-q", "http://localhost:3200/ready"]
       interval: 10s
@@ -55,7 +55,7 @@ services:
       - ./otel/grafana/provisioning:/etc/grafana/provisioning:ro
       - grafana_data:/var/lib/grafana
     ports:
-      - "13000:3000"
+      - "127.0.0.1:13000:3000"
     healthcheck:
       test: ["CMD", "curl", "-sf", "http://localhost:3000/api/health"]
       interval: 10s
@@ -86,7 +86,7 @@ services:
       - ./otel/prometheus.yml:/etc/prometheus/prometheus.yml:ro
       - prometheus_data:/prometheus
     ports:
-      - "19090:9090"
+      - "127.0.0.1:19090:9090"
     networks:
       - egov-network
 
@@ -1645,7 +1645,7 @@ services:
     volumes:
       - ./gatus/config.yaml:/config/config.yaml:ro
     ports:
-      - "18889:8080"
+      - "127.0.0.1:18889:8080"
     environment:
       GATUS_CONFIG_PATH: /config/config.yaml
       # Profile-gated stacks: keep these beside COMPOSE_PROFILES in .env.
@@ -1672,7 +1672,7 @@ services:
       - ./otel/loki-config.yaml:/etc/loki/config.yaml:ro
       - loki_data:/loki
     ports:
-      - "13100:3100"
+      - "127.0.0.1:13100:3100"
     healthcheck:
       test: ["CMD", "wget", "--spider", "-q", "http://localhost:3100/ready"]
       interval: 10s
--- a/docker-compose.egov-digit.yaml
+++ b/docker-compose.egov-digit.yaml
@@ -14,9 +14,9 @@ services:
     volumes:
       - ./otel/otel-collector-config.yaml:/etc/otelcol-contrib/config.yaml:ro
     ports:
-      - "14317:4317"
-      - "14318:4318"
-      - "13133:13133"
+      - "127.0.0.1:14317:4317"
+      - "127.0.0.1:14318:4318"
+      - "127.0.0.1:13133:13133"
     networks:
       - egov-network
 
@@ -29,7 +29,7 @@ services:
       - ./otel/tempo-config.yaml:/etc/tempo/config.yaml:ro
       - tempo_data:/var/tempo
     ports:
-      - "13200:3200"
+      - "127.0.0.1:13200:3200"
     healthcheck:
       test: ["CMD", "wget", "--spider", "-q", "http://localhost:3200/ready"]
       interval: 10s
@@ -55,7 +55,7 @@ services:
       - ./otel/grafana/provisioning:/etc/grafana/provisioning:ro
       - grafana_data:/var/lib/grafana
     ports:
-      - "13000:3000"
+      - "127.0.0.1:13000:3000"
     healthcheck:
       test: ["CMD", "curl", "-sf", "http://localhost:3000/api/health"]
       interval: 10s
@@ -86,7 +86,7 @@ services:
       - ./otel/prometheus.yml:/etc/prometheus/prometheus.yml:ro
       - prometheus_data:/prometheus
     ports:
-      - "19090:9090"
+      - "127.0.0.1:19090:9090"
     networks:
       - egov-network
 
@@ -1521,7 +1521,7 @@ services:
     volumes:
       - ./gatus/config.yaml:/config/config.yaml:ro
     ports:
-      - "18889:8080"
+      - "127.0.0.1:18889:8080"
     environment:
       GATUS_CONFIG_PATH: /config/config.yaml
     networks:
@@ -1537,7 +1537,7 @@ services:
       - ./otel/loki-config.yaml:/etc/loki/config.yaml:ro
       - loki_data:/loki
     ports:
-      - "13100:3100"
+      - "127.0.0.1:13100:3100"
     healthcheck:
       test: ["CMD", "wget", "--spider", "-q", "http://localhost:3100/ready"]
       interval: 10s
--- a/local-setup/docker-compose.db-migrations.yml
+++ b/local-setup/docker-compose.db-migrations.yml
@@ -1295,7 +1295,7 @@ services:
     volumes:
       - ./gatus/config.yaml:/config/config.yaml:ro
     ports:
-      - "18889:8080"
+      - "127.0.0.1:18889:8080"
     environment:
       GATUS_CONFIG_PATH: /config/config.yaml
       # These services are ungated in this file, so the toggles are fixed:
--- a/local-setup/docker-compose.deploy.yaml
+++ b/local-setup/docker-compose.deploy.yaml
@@ -908,7 +908,7 @@ services:
     volumes:
       - ./gatus/config.yaml:/config/config.yaml:ro
     ports:
-      - "18889:8080"
+      - "127.0.0.1:18889:8080"
     environment:
       GATUS_CONFIG_PATH: /config/config.yaml
       # These services are ungated in this file, so the toggles are fixed:
--- a/local-setup/docker-compose.registry.yml
+++ b/local-setup/docker-compose.registry.yml
@@ -14,9 +14,9 @@ services:
     volumes:
       - ./otel/otel-collector-config.yaml:/etc/otelcol-contrib/config.yaml:ro
     ports:
-      - "14317:4317"
-      - "14318:4318"
-      - "13133:13133"
+      - "127.0.0.1:14317:4317"
+      - "127.0.0.1:14318:4318"
+      - "127.0.0.1:13133:13133"
     networks:
       - egov-network
 
@@ -29,7 +29,7 @@ services:
       - ./otel/tempo-config.yaml:/etc/tempo/config.yaml:ro
       - tempo_data:/var/tempo
     ports:
-      - "13200:3200"
+      - "127.0.0.1:13200:3200"
     healthcheck:
       test: ["CMD", "wget", "--spider", "-q", "http://localhost:3200/ready"]
       interval: 10s
@@ -53,7 +53,7 @@ services:
       - ./otel/grafana/provisioning:/etc/grafana/provisioning:ro
       - grafana_data:/var/lib/grafana
     ports:
-      - "13000:3000"
+      - "127.0.0.1:13000:3000"
     healthcheck:
       test: ["CMD", "curl", "-sf", "http://localhost:3000/api/health"]
       interval: 10s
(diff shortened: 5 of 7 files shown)


singhalkarun:

```
```

singhalkarun:
@coderabbitai review

singhalkarun:
**Self-review addition** (automated review was rate-limited on this PR, so I went over the diff against the same issue classes it raised on #1605).

Found a gap: `docker-compose.registry.yml` carries the **same** problem this PR fixes, and the original commit didn't touch it. It isn't the Ansible path — `local-setup/README.md:324` is explicit that the playbook never references it — but it is a documented, runnable alternative stack, so leaving it alone would have kept a second one-command route to exactly what this PR closes.

Now fixed in the same PR. Scope note: that file has no `loki` / `prometheus` / `node-exporter` services, so six mappings apply there rather than eight.

Also checked and found clean:
- no docs, READMEs, or Jinja templates point operators at these ports on a non-loopback address (only `127.0.0.1` / `localhost` references exist)
- Ansible validate tasks already use `127.0.0.1`
- `docker-compose.yml` and `docker-compose.deploy.yaml` (the CI compose paths) define no observability services beyond Gatus, so CI is unaffected

singhalkarun:
**CI note — the `tilt-test` failure is a flake, not a regression from this PR.**

Worth showing the evidence, since "PR changes compose file → compose-based test fails" is the obvious suspicion.

**1. The file this PR edits is not loaded by that job.** The sidecar logs its own inputs:

```
digit-telemetry | [telemetry] files=.../local-setup/docker-compose.yml,.../local-setup/docker-compose.tilt.yml
```

`docker-compose.egov-digit.yaml` is not among them — and it is the only compose file that defines `prometheus` at all. The changed configuration is not present in the failing job.

**2. The failure is a timing race, and the step's own comment says so:**

```
# tilt ci exits as soon as resources are healthy, so the sidecar
# captures fewer events than the long-running docker compose job.
```

Observed counts: `Container start events: 0`, `Container healthy events: 2`. The sidecar was running and capturing — it simply attached after the containers had already started, so it saw the healthy transitions and missed the start ones.

**3. It fails on `master` too** — `tilt-ci.yaml` on `master` shows failure on 2026-08-05 with successes on either side, and it passes on #1605, #1607 and #1627 while failing here and on #1606. That distribution tracks runner timing, not file contents.

Not something this PR can fix. The underlying flake (`>= 1 start event` is not guaranteed when `tilt ci` exits early) deserves its own issue — the assertion should either wait for the sidecar to attach before starting containers, or drop the start-event requirement and rely on healthy events.

vinothrallapalli-eGov (commented):
Verified the two touched files: all 8 mappings in `docker-compose.egov-digit.yaml` and all 6 in `docker-compose.registry.yml` (otel-collector x3, tempo, grafana, prometheus, gatus, loki) correctly gained the `127.0.0.1:` prefix, matching the PR's table exactly — no sibling port on any of these services was left un-rebound within these two files. However, this is a **medium-severity partial fix at the repo level**: per `local-setup/README.md`'s own project-structure listing, `docker-compose.yml`, `docker-compose.deploy.yaml`, and `docker-compose.db-migrations.yml` are standalone sibling stacks (not overlays) that each define their own `gatus` service, and all three still publish `"18889:8080"` with no host IP — i.e. still on 0.0.0.0/[::]. `docker-compose.deploy.yaml` in particular is pushed to real target hosts via `performance/ansible/playbook-setup.yml` (`docker compose -f docker-compose.deploy.yaml up -d` over SSH with `become: true`), so this isn't dead code — it's a live deploy path the PR missed. Separately, checked the tracking issue's ask: no runbooks/scripts curl these ports via a box's public IP (all references are `localhost`/`127.0.0.1`, and `docs/2.12-beta/*.md` already describes them as "loopback ports"), so no doc updates are actually owed there. One unverified item: the PR description says these were previously reachable on both `0.0.0.0` and `[::]`; the fix only adds an IPv4 host address, which should implicitly drop the IPv6 listener, but the PR's own checklist ("time out when probed off-box") is unchecked and doesn't call out IPv6 specifically — worth confirming during the template-box run rather than assuming.

local-setup/docker-compose.egov-digit.yaml
```diff
@@ -1645,7 +1645,7 @@ services:
     volumes:
       - ./gatus/config.yaml:/config/config.yaml:ro
     ports:
-      - "18889:8080"
+      - "127.0.0.1:18889:8080"
```
vinothrallapalli-eGov, line 1648: This gatus service is duplicated (own `ports:` block, no anchor/extends) in three sibling compose files that are NOT touched by this PR and still expose `"18889:8080"` on all interfaces: `local-setup/docker-compose.yml:1132`, `local-setup/docker-compose.deploy.yaml:911`, and `local-setup/docker-compose.db-migrations.yml:1298`. `docker-compose.deploy.yaml` is actively deployed straight to hosts by `performance/ansible/playbook-setup.yml`, so this isn't a stale/legacy file — the gatus dashboard (health/status of internal services, `auth_enabled` not set) stays reachable from off-box there. Since these files are documented in `local-setup/README.md` as parallel full stacks rather than overlays of `docker-compose.egov-digit.yaml`, fixing this file alone doesn't propagate. Suggest the same `127.0.0.1:` prefix in those three files (or, better, factor gatus's port mapping into one place so this class of drift can't recur).

local-setup/docker-compose.egov-digit.yaml
```diff
@@ -14,9 +14,9 @@ services:
     volumes:
       - ./otel/otel-collector-config.yaml:/etc/otelcol-contrib/config.yaml:ro
     ports:
-      - "14317:4317"
-      - "14318:4318"
-      - "13133:13133"
+      - "127.0.0.1:14317:4317"
+      - "127.0.0.1:14318:4318"
+      - "127.0.0.1:13133:13133"
```
vinothrallapalli-eGov, line 19: Nit/verification: the PR body says these ports were previously bound to both `0.0.0.0` and `[::]` (IPv6). Specifying only an IPv4 host address here should drop the implicit IPv6 listener entirely (rather than rebind it to `::1`), but that's host/Docker-version dependent behavior worth confirming explicitly against the public IPv6 address during the template-box verification, not just assumed — the PR's own checklist item for off-box probing doesn't call out v4 vs v6 separately.

local-setup/docker-compose.egov-digit.yaml
singhalkarun: Valid, and worse than described — thanks. Fixed in 23700e6d.

Verified all three:

```
local-setup/docker-compose.yml:1132              - "18889:8080"
local-setup/docker-compose.deploy.yaml:911       - "18889:8080"
local-setup/docker-compose.db-migrations.yml:1298 - "18889:8080"
```

And your reasoning about `deploy.yaml` not being legacy holds — `performance/ansible/playbook-setup.yml:65` rsyncs it to `{{ digit_dir }}` and line 100 runs `docker compose -f docker-compose.deploy.yaml up -d`.

Gatus is the only observability service in those three files (`grep -nE '^  (grafana|prometheus|loki|tempo|otel-collector|gatus):'` matches `gatus:` and nothing else in each), so `18889` was the whole gap there.

**One you didn't catch:** the repo-root standalone `docker-compose.egov-digit.yaml` — the Docker Hub variant documented in `MAC-QUICKSTART.md` — still had **all eight** ports on `0.0.0.0` (14317/14318/13133/13200/13000/19090/18889/13100). That one is not the file the playbook ships (`playbook-deploy.yml:1090` uses `src: ../docker-compose.egov-digit.yaml`, which resolves to `local-setup/`), so it drifted independently. Fixed too.

On "factor it into one place": I couldn't. These are parallel full stacks in separate files and YAML anchors don't span files, so there's no shared node to point at. The substitute is a CI contract in `local-setup/tests/static/infra-contracts.test.ts` — it walks every `docker-compose*.y{a,}ml` in `local-setup/` plus the repo-root one, and fails on any published mapping whose host port is in the observability set without a `127.0.0.1:` prefix. Confirmed it actually bites by reverting one line:

```
● observability ports are loopback-bound in every compose file
  local-setup/docker-compose.deploy.yaml: "18889:8080" (gatus) is not loopback-bound
```

Deliberately left alone as out of scope for this PR: jupyter `18888:8888` (`--ServerApp.token=''`) and digit-mcp `13101:3000` (unauthenticated ADMIN). Both are real exposures but they aren't observability, and folding them in here would change what this PR is reviewable as.

`local-setup/tests` static suite green (37 tests), `check-gatus-coverage.py` green, all six touched compose files parse.

local-setup/docker-compose.egov-digit.yaml
singhalkarun: Fair, and I've taken the checklist half of it — the PR body now splits the off-box probe into IPv4 and IPv6 and adds an on-box `ss -lntp` assertion:

- [ ] … off-box over **IPv4** (`nc -4 -vz <public-v4> 13000 … 19090`)
- [ ] … off-box over **IPv6** (`nc -6 -vz <public-v6> …`)
- [ ] on-box `ss -lntp | grep -E ':(13000|13100|13133|13200|14317|14318|18889|19090)\b'` shows **only** `127.0.0.1:` rows, no `[::]:` / `*:`

On the mechanism, your read is right: it drops the v6 listener rather than rebinding it to `::1`. In a `-p` mapping, an *empty* HostIP is what means "all addresses" — dockerd expands that to `0.0.0.0` and, when the daemon has v6 on the bridge, also `::`, which is where the `[::]` row came from. Giving an explicit IPv4 host address yields exactly one port binding: one `docker-proxy` with `-host-ip 127.0.0.1`, one `nat/DOCKER` DNAT rule scoped to `-d 127.0.0.1`, and nothing written to `ip6tables`. There is no code path that maps a v4 host IP onto a v6 equivalent.

**But I have not verified that at runtime and I'm not going to claim I have.** I didn't deploy — other agents are on the devbox — and there's no Docker daemon on this machine to demonstrate it locally, so the above is reasoning about Docker's binding semantics, not an observation. That's exactly why it's a checklist item now rather than an assertion in the PR body. Treat the IPv6 row as genuinely unverified until the template-box run.

One operator-visible consequence worth flagging either way, which I've added to the body: the tunnel target has to be `127.0.0.1`, not `localhost`. On a dual-stack box `localhost` can resolve to `::1` first, and after this change nothing is listening there — so `ssh -L 13000:localhost:13000` can fail on a box where the pre-fix version worked. `ssh -L 13000:127.0.0.1:13000` is the safe form.

vinothrallapalli-eGov (approved):

PR #1607, opened 2026-08-05
fix(gateway): drop unauthenticated Kong observability routes

### LINKED ISSUE #1604: Remove unauthenticated Kong observability routes; gate nginx /grafana/ behind a flag

singhalkarun:
Fixes #1604

> [!IMPORTANT]
> **Depends on #1684 — merge that first, or merge the two together.**
>
> This PR only edits `kong/kong.yml`. On an *existing* deployment that file never reaches the running Kong container: the config-dir sync is a plain `synchronize`/rsync (`playbook-deploy.yml:1199`, `kong: ../kong`), rsync's default temp-file-plus-rename replaces the inode, and compose bind-mounts the file *individually* (`./kong/kong.yml:/kong/kong.yml:ro`), so the mount stays pinned to the old inode. Nothing in the playbook recreates or reloads Kong when only its config changed. The result is silent success: correct file on disk, green playbook, old routes still being served.
>
> #1684 fixes both halves (`--inplace` on the sync, plus a `kong reload` when `kong.yml` changed). Until it lands, this PR is only effective on a full stack recreate (`docker compose down` + `up`) or a fresh box.

## What

Kong proxied `/grafana` and `/tempo` straight to the observability stack with no auth plugin.

`/tempo` exposed the entire trace API and has no legitimate consumer — Grafana queries Tempo server-side over the container network, and the browser never touches it.

`/grafana` additionally created a **fall-through**: in static UI mode host nginx's `location /` catch-all forwards to Kong, so a tenant that removed or disabled the nginx `/grafana/` location was still served Grafana through Kong. Any nginx-level gating was cosmetic while that route existed.

## Changes

- **`kong/kong.yml`** — remove `grafana-service`/`grafana-route` (with its `request-transformer` plugin) and `tempo-service`/`tempo-route`. A comment records why, so the pattern isn't reintroduced. The `/otel/v1/*` ingest routes are untouched.
- **`nginx-site.conf.j2`** — wrap the `/grafana/` location in `{% if nginx_features.grafana | default(true) %}`.
- **`group_vars/digit.yml`** — add `grafana: true` to `nginx_features`.

## The `| default(true)` is load-bearing

`host_vars` files override `nginx_features` as a whole dict and Ansible's `hash_behaviour` is the default `replace`, so every tenant file written before this key existed (`pg.citya.yml`, `maputo.yml.example`, `localhost-full.yml.example`, …) has no `grafana` member. Without the default, rendering this template would fail on those tenants and take the deploy down. Same pattern as the existing `digit_ui_v2` and `root_to_digit_ui` flags.

## Default is on, deliberately

Grafana carries a real login as of #1602, and `/grafana/` is the team's dashboard entry point — defaulting it off would break the workflow for every tenant on upgrade. Tenants that want Grafana off the internet entirely set `nginx_features.grafana: false` and reach it over an SSH tunnel to `127.0.0.1:13000`.

## Testing

Not yet deployed. Checklist:

- [ ] `https://<domain>/tempo/api/search` → 404
- [ ] with the flag **false**, `https://<domain>/grafana/` → 404 (confirms no Kong fall-through)
- [ ] with the flag **true**, Grafana loads and shows its login page
- [ ] Kong still boots and its config parses after the route removal
- [ ] **with #1684 in place**, a plain redeploy (no stack recreate) is enough for the removed routes to stop answering — without it, verify only after `docker compose down && up`
- [ ] existing tenants (`pg.citya`) render nginx without error despite having no `grafana` key

## Notes

Pairs with #1602, which puts the login behind this path. Blocked on #1684 for delivery (see the note at the top) — that bug was found while verifying this PR on the devbox.

### DIFF
--- a/local-setup/ansible/templates/nginx-site.conf.j2
+++ b/local-setup/ansible/templates/nginx-site.conf.j2
@@ -124,7 +124,14 @@ server {
 {% endif %}
 
 {% endif %}
-  # Grafana dashboard
+{# `| default(true)` is load-bearing: host_vars override nginx_features as a
+   whole dict (hash_behaviour=replace), so every tenant file written before
+   this key existed has no `grafana` member. Without the default those deploys
+   would fail rendering this template. Same reason as digit_ui_v2 below. #}
+{% if nginx_features.grafana | default(true) %}
+  # Grafana dashboard. This is the ONLY public path to Grafana — the Kong
+  # /grafana route was removed in #1604, so turning this flag off makes the
+  # UI genuinely unreachable from outside rather than merely unlinked.
   location /grafana/ {
     proxy_pass http://{{ nginx_upstream_host | default('127.0.0.1') }}:{{ upstream_grafana_port }};
     proxy_set_header Host $host;
@@ -136,6 +143,7 @@ server {
     proxy_set_header Connection "upgrade";
   }
 
+{% endif %}
 {% if nginx_features.status %}
   # Gatus health page — public path /status/, SPA assets rewritten on the fly.
   # Gatus's index.html / JS reference assets and the API as absolute paths
--- a/local-setup/kong/kong.yml
+++ b/local-setup/kong/kong.yml
@@ -670,24 +670,6 @@ services:
     paths:
     - /jupyter
     strip_path: false
-- name: grafana-service
-  url: http://grafana:3000
-  tags:
-  - observability
-  host: grafana
-  port: 3000
-  routes:
-  - name: grafana-route
-    paths:
-    - /grafana
-    strip_path: false
-    preserve_host: true
-  plugins:
-  - name: request-transformer
-    config:
-      add:
-        headers:
-        - "X-Forwarded-Proto:https"
 - name: indexer-service
   url: http://egov-indexer:8080
   tags:
@@ -706,18 +688,16 @@ services:
     paths:
     - /inbox
     strip_path: false
-- name: tempo-service
-  url: http://tempo:3200
-  tags:
-  - observability
-  routes:
-  - name: tempo-route
-    paths:
-    - /tempo
-    strip_path: true
+# NOTE: the /grafana and /tempo routes were removed (#1604). Both proxied to
+# the observability stack with no auth, and the Grafana one also created a
+# fall-through: host nginx's `location /` catch-all forwards to Kong, so a
+# tenant that disabled the nginx /grafana/ location was still served Grafana
+# through here. Grafana is now reachable only via the nginx location (gated on
+# nginx_features.grafana) and Tempo only via Grafana's server-side datasource.
+#
 # Browser -> otel-collector OTLP/HTTP ingest for the dashboard render-lag
-# instrumentation (#1110). Same-origin (follows the /tempo and /grafana
-# observability precedent) so the SPA needs no CORS work and no new host port.
+# instrumentation (#1110). Same-origin so the SPA needs no CORS work and no
+# new host port.
 # Security posture (R8): EXACT paths, POST only, /v1/traces deliberately NOT
 # exposed; per-route 1MB size cap + 60/min-per-IP rate limit; both paths are
 # whitelisted in the pre-function's AUTH_OPTIONAL above. No ingest token in v1
--- a/local-setup/ansible/inventory/group_vars/digit.yml
+++ b/local-setup/ansible/inventory/group_vars/digit.yml
@@ -106,6 +106,11 @@ nginx_features:
   configurator: false       # /configurator + /configurator/assets
   configurator_caching: false  # /configurator/assets/ long-cache headers
   status: false             # /status/ Gatus board (sub_filter rewrites)
+  grafana: true             # /grafana/ dashboards — the only public path to
+                            # Grafana (#1604). Default on: it carries a real
+                            # login since #1602. Set false to take Grafana off
+                            # the internet entirely (reach it over an SSH
+                            # tunnel to 127.0.0.1:13000 instead).
   api_pgr: false            # /api/pgr/ proxied to MCP for the dashboard
   mcp: false                # /mcp HTTP transport — see CLAUDE.md security note
   # Note: `/` routing is derived from `digit_ui_mode`:


singhalkarun:

```
```

singhalkarun:
@coderabbitai review

singhalkarun:
**Self-review** (the automated reviewer was rate-limited on this PR, so I checked the diff against the same issue classes it raised on #1605).

Verified clean:
- **No dangling references** to `grafana-service` / `grafana-route` / `tempo-service` / `tempo-route` anywhere in `local-setup/` or `docs/` — the k3s tier defines no Kong config, so there is no second copy to keep in sync.
- **No docs** instruct anyone to reach Grafana or Tempo through the gateway (no `:18000/grafana`, no `/tempo` references outside the removed block).
- **`digit-mcp` is unaffected** — its tracing tools talk to `TEMPO_URL` / `GRAFANA_URL`, which default to host-side `localhost:13200` / `localhost:13000`, never through Kong.
- **`whitelist-parity` CI passes**, which is the check that would catch a mismatch between the Kong route set and `env.yaml`.
- Grafana's `/api/health` (used by `digit-mcp` and by the Ansible gate in #1605) stays unauthenticated in Grafana itself, so the auth change in #1605 does not break those probes.

Unlike #1605 and #1606, `docker-compose.registry.yml` needed nothing here — it has no Kong service and no nginx template, so the routes this PR removes never existed on that path.

singhalkarun:
**Second CI note — the `test` job failure is also pre-existing, and it is *not* the Kong change.**

Worth spelling out because `Localization module check` calls `http://localhost:18000`, i.e. through Kong, so "PR removes Kong routes → Kong-routed check fails" is the obvious suspicion. It isn't what happened:

1. **`master` fails the identical pair of steps** on its own latest run — `infra contract tests` and `Localization module check`.
2. **Kong starts and routes correctly on this PR.** In the same job, `Run health checks` and `Gatus health dashboard check` both pass, and both exercise Kong-routed services. A broken declarative config would have taken Kong down and failed those first — the failure would not have waited politely for the localization step.
3. The removed routes are `/grafana` and `/tempo`. `whitelist-parity` passes, confirming the remaining route set still matches `env.yaml`.

So: two red checks on this PR, both inherited from `master`. The `FROZEN_LATEST_DEBT` one is described in the comment above; this one needs its own fix, also unrelated to these PRs.

singhalkarun:
## Devbox verification — the change is correct, but it does not take effect on an existing deployment

Deployed this on the eGov devbox (`karun` tenant) on top of #1605/#1606 and verified the routes.

**The removal itself works.** After Kong picks up the new config, `grafana-service` and `tempo-service` are gone (34 → 32 services) and the routes stop proxying:

| Route via Kong `:18000` | Before | After |
|---|---|---|
| `/grafana/api/health` | 200 | **404** |
| `/tempo/api/echo` | 200 (body `echo`) | **404** |
| `/egov-mdms-service/health` | 200 | 200 (unaffected) |

The `otel-metrics-ingest` / `otel-logs-ingest` routes remain, which matches the PR's stated intent.

**But a plain redeploy does not apply it.** Straight after the ansible run, `/opt/digit/kong/kong.yml` had the routes removed while Kong was still serving them:

```
host      /opt/digit/kong/kong.yml   Aug  7 13:21   md5 2c8578b7…   grafana|tempo services: 0
container /kong/kong.yml             Jul 27 10:02   md5 4311e54f…   grafana|tempo services: 2
```

The container was serving an **11-day-old copy**. `kong reload` did *not* help — it reported "Kong reloaded" and kept serving the old routes, because it re-reads the same stale in-container file.

**Cause:** `kong.yml` is bind-mounted as a *single file* (`/opt/digit/kong/kong.yml → /kong/kong.yml`). Ansible's `template`/`copy` writes atomically — temp file plus rename — which replaces the inode. A single-file bind mount is bound to the original inode, so the container keeps seeing the old file forever.

Only `docker compose up -d --force-recreate kong` fixed it; afterwards the md5s matched and the routes 404'd.

**Why this matters beyond this PR:** it applies to *any* change to `kong.yml`, not just this one. A Kong config change can appear deployed — file on disk correct, playbook green, `kong reload` reporting success — while the gateway runs the old policy indefinitely. That's the same silent-success failure mode the RBAC work in #1255 depends on being able to trust.

**Suggested follow-up** (not done here, since it's outside this PR's scope): either have the playbook recreate Kong when `kong.yml` changes, or mount the *directory* rather than the file so atomic replacement propagates. Worth its own issue.

Everything else stayed healthy: 43 containers, 0 unhealthy, Prometheus targets `node=up otel-collector=up`, and #1605/#1606/#1625 all still holding.

local-setup/kong/kong.yml
```diff
@@ -706,18 +688,16 @@ services:
...
-- name: tempo-service
-  url: http://tempo:3200
-  tags:
-  - observability
-  routes:
-  - name: tempo-route
-    paths:
-    - /tempo
-    strip_path: true
+# NOTE: the /grafana and /tempo routes were removed (#1604). Both proxied to
```
vinothrallapalli-eGov, line 691: Route removal itself looks correct and complete (full deletion, not commented out) — confirmed no lingering `grafana`/`tempo` service/route/plugin entries elsewhere in the file. However, note that this change alone won't reach any already-running Kong container: `kong.yml` is delivered via the plain `synchronize` (rsync, default temp+rename) task in playbook-deploy.yml and bind-mounted as a single file into kong-gateway, and this diff adds no `kong reload` step. Per #1684's PR description, that combination leaves a running container reading a stale copy indefinitely even though the file on disk (and this diff) is correct. Recommend this PR explicitly states it depends on #1684 landing first (or merging together) before it's safe to rely on for any tenant that isn't doing a full stack recreate.

local-setup/ansible/templates/nginx-site.conf.j2
```diff
@@ -124,7 +124,14 @@ server {
 {% endif %}
 
 {% endif %}
-  # Grafana dashboard
+{# `| default(true)` is load-bearing: host_vars override nginx_features as a
+   whole dict (hash_behaviour=replace), so every tenant file written before
+   this key existed has no `grafana` member. Without the default those deploys
+   would fail rendering this template. Same reason as digit_ui_v2 below. #}
+{% if nginx_features.grafana | default(true) %}
```
vinothrallapalli-eGov, line 131: Gate looks correct and the `default(true)` reasoning is sound (matches the existing hash-replace behavior for `digit_ui_v2`/`root_to_digit_ui`). One thing worth double-checking in testing: this nginx flag is the only remaining public path to Grafana now that the Kong route is gone, so `nginx_features.grafana: false` needs to be verified end-to-end (404, not just unlinked) once #1684 is also in place — since nginx config itself is presumably also delivered via the same rsync path and template rendering, though nginx (unlike Kong) is likely recreated/reloaded routinely so may be less exposed to the inode-staleness issue than kong.yml is.

vinothrallapalli-eGov (commented):
Reviewed the diff directly (kong.yml, nginx-site.conf.j2, group_vars/digit.yml). The `grafana-service`/`grafana-route` and `tempo-service`/`tempo-route` blocks are fully deleted from kong.yml (not commented out), and the replacement comment correctly documents why. The nginx `/grafana/` location is now correctly gated behind `nginx_features.grafana | default(true)`, and no other file reintroduces an equivalent unauthenticated path to Grafana/Tempo — Tempo has no nginx location at all post-change, matching the PR's claim that it's only reached server-side from Grafana's datasource. The `default(true)` reasoning for host_vars hash-replace semantics checks out against the existing `digit_ui_v2`-style precedent in the same file.

On the tracked dependency: I confirmed this PR does **not** sidestep the rsync/bind-mount staleness bug that #1684 fixes. `kong.yml` is synced to the target host via the plain `synchronize` (rsync, default temp-file+rename) task in `playbook-deploy.yml`, then bind-mounted single-file (`./kong/kong.yml:/kong/kong.yml:ro`) into `kong-gateway`. This PR's diff adds no `kong reload` step and no `--inplace` sync flag — both are introduced only in #1684. So on an already-running deployment, this fix is inert until #1684 lands (or the Kong container is otherwise recreated): the health check only confirms Kong is up, not that it picked up the new declarative config. This exactly matches what #1684's own PR body reports finding while trying to verify #1607 on a devbox. The PR description should explicitly call out this hard dependency/merge-order requirement rather than leaving it implicit in the tracking issue.

Severity: correctness/process gap, not a security flaw in the change itself — the config change is right, but it needs #1684 to actually take effect on any live tenant.

local-setup/kong/kong.yml
singhalkarun: Correct on every point, and #1684 is exactly the fix — it was in fact found *while verifying this PR* on the devbox, where the routes were still answering after a green deploy because Kong was reading an 11-day-old `kong.yml`.

Confirmed the chain in this branch:

- `playbook-deploy.yml:1199` "Synchronize config directories" — plain `synchronize` over `digit_config_dirs`, which includes `kong: ../kong` (line 1209). Default rsync mode, so temp-file + rename → new inode.
- `docker-compose.egov-digit.yaml:1615` — `./kong/kong.yml:/kong/kong.yml:ro`, a *single-file* bind mount, which binds the inode rather than the path.
- No `kong reload`, `restart`, or forced recreate anywhere in the playbook: `grep -i kong playbook-deploy.yml | grep -iE 'reload|restart|recreate|force'` returns one unrelated hit ("wait for egov-user via Kong to be ready after restart", line 3539).

So on an existing deployment this diff is inert until the container is recreated. Done as you asked — the PR body now opens with an `[!IMPORTANT]` block stating the #1684 dependency and spelling out the inode mechanism, and the test checklist gained a row distinguishing "plain redeploy suffices (needs #1684)" from "only after `docker compose down && up`".

No code change here: the fix belongs in the delivery path, not in `kong.yml`. Thanks for confirming the removal itself is complete — `grep -i 'grafana\|tempo' local-setup/kong/kong.yml` now matches only the explanatory comment.

local-setup/ansible/templates/nginx-site.conf.j2
singhalkarun: Both halves check out, and your instinct that nginx is structurally less exposed than Kong is right — for a concrete reason, not just "it gets recreated more often".

**1. `grafana: false` is a real 404, and the 404 comes from Kong.**
With the location gated out, `/grafana/` falls into the catch-all at `nginx-site.conf.j2:504` (`location / { proxy_pass … upstream_kong_port }`), and since this PR deletes `grafana-route` there is no Kong route left to match — Kong answers `{"message":"no Route matched with those values"}`, 404. Nothing else in the tree can serve it: `upstream_grafana_port` (13000) appears in exactly one place, the gated block itself, and `grep -i grafana local-setup/kong/kong.yml` now matches only the comment. So it is genuinely unreachable, not merely unlinked — which is precisely the fall-through this PR was written to close.

**2. nginx is not exposed to the inode-staleness bug at all.**
Different delivery path from `kong.yml`. On Linux the vhost is written with ansible `template:` — a direct render to `/etc/nginx/{sites-available/<domain>,conf.d/<domain>.conf}` (lines 2861 and 2906) — with `notify: Reload nginx` on the render, the enable-symlink, and the default-site removal, followed by an explicit `meta: flush_handlers` so the reload happens *before* the validate tasks run. nginx reads its config by path at reload time, and it isn't a bind mount, so a changed inode is irrelevant. On the macOS container path (line 2952) it *is* a single-file bind mount (`{{ digit_dir }}/nginx/site.conf:/etc/nginx/conf.d/default.conf:ro`) — but the very next task is `docker rm -f digit-nginx` followed by `docker run`, unconditionally, every deploy. Recreation picks up the new inode by definition. Both paths are safe; only Kong had "config replaced, container never touched".

**3. Template rendering, all four `nginx_features` shapes.**
Rendered `nginx-site.conf.j2` through Jinja2 with the real `group_vars/digit.yml`:

| `nginx_features` shape | renders | `/grafana/` block |
|---|---|---|
| `grafana: true` | ✅ | present |
| `grafana: false` | ✅ | **absent** |
| key absent (post-#1604 hash-replace case) | ✅ | present |
| legacy 3-key dict, as an old `host_vars` would have | ✅ | present |

No exceptions on any shape — which is the `| default(true)` claim actually exercised rather than argued.

**What I did not verify, stated plainly:** the end-to-end 404. I didn't deploy — other agents are on the devbox — so #1607 + #1684 together against a live box with `grafana: false` is still an open checklist item, and I couldn't run `nginx -t` on the rendered output either (no nginx on this machine; the render check above is Jinja-level, not nginx-syntax-level). The PR checklist already carries the `grafana: false → 404` row and now also carries a row for the #1684 delivery dependency.

vinothrallapalli-eGov (approved):

Issue #1617, opened 2026-08-05
Root-level otel/ copy has drifted and its promtail config is fatally broken


singhalkarun:
Part of #1601 (monitoring stack umbrella). **P1 · effort: XS (part 1) + S (part 2)**

## Problem

The repo carries **two copies** of the observability config: the deployed one under `local-setup/`, and an older one at the repo root. The root copy has drifted, and one file in it is not merely stale — it is **fatally broken**.

Measured drift (root vs `local-setup/`):

| File | State |
|---|---|
| `otel/promtail-config.yaml` | **DIFFERS — root copy is fatal** |
| `otel/loki-config.yaml` | DIFFERS |
| `docker-compose.egov-digit.yaml` | DIFFERS |
| `otel/otel-collector-config.yaml`, `otel/prometheus.yml`, `otel/tempo-config.yaml`, `docker-compose.monitoring.yml` | identical |

The root `otel/promtail-config.yaml` still contains:

```yaml
limits_config:
  reject_old_samples: true
  reject_old_samples_max_age: 168h
```

We know exactly what that does, because the **fixed** copy documents it in place (`local-setup/otel/promtail-config.yaml:16-21`):

> `reject_old_samples` / `reject_old_samples_max_age` belong in Loki's loki-config.yaml, NOT here — promtail's `limits_config` schema doesn't accept them and rejects the whole config with a fatal: `field reject_old_samples not found in type limit.Config`

Someone hit this, diagnosed it, fixed the copy they were using, and left the broken one in the repo.

## Why it matters

Promtail does not warn or degrade — it refuses to start. Anyone deploying from the repo root (following an older doc, or reasonably assuming the root compose file is the primary one) gets a stack where **log collection is silently dead**. Everything else comes up healthy, Grafana works, and the logs dashboard is simply empty forever with nothing explaining why.

### Part 1 — defuse it now (separate, mergeable immediately)

Copy `local-setup/otel/promtail-config.yaml` over `otel/promtail-config.yaml`. One file, no behaviour change to anything currently deployed, no discussion required. This should not wait behind a repository-layout debate.

### Part 2 — decide whether the root copies should exist

Either delete the root `otel/`, `docker-compose.egov-digit.yaml` and `docker-compose.monitoring.yml`, or keep them and add a CI drift check. The repo already has this exact pattern for the Gatus tier files (`.github/workflows/gatus-coverage.yml`), so it is an established, accepted approach:

```yaml
[...]

PR #1624, opened 2026-08-06
fix(otel): sync the root promtail config, which was fatally invalid

singhalkarun:
Part 1 of #1617.

## What

The root-level `otel/promtail-config.yaml` still contained:

```yaml
limits_config:
  reject_old_samples: true
  reject_old_samples_max_age: 168h
```

Promtail's `limits_config` schema has no such fields. It does not warn or ignore them — it refuses the entire config:

```
field reject_old_samples not found in type limit.Config
```

and the process never starts.

The deployed copy under `local-setup/` was fixed some time ago and documents the reason in place. The root copy was left behind. Anyone deploying from the repo root — following an older doc, or reasonably assuming the root compose file is the primary one — gets a stack where **log collection is silently dead**: everything else comes up healthy, Grafana works, and the logs dashboard is simply empty forever with nothing explaining why.

## Change

One file, copied verbatim from `local-setup/otel/promtail-config.yaml`. The two are now byte-identical.

## Verification

```
$ diff otel/promtail-config.yaml local-setup/otel/promtail-config.yaml
(empty)

$ python3 -c "import yaml; d=yaml.safe_load(open('otel/promtail-config.yaml')); print(d.get('limits_config'))"
None
```

Top-level keys are now `server`, `positions`, `clients`, `scrape_configs` — no `limits_config` block at all. (Note the phrase `reject_old_samples` still appears in the file as *explanatory comment text*, which is intended: it documents why the keys must not be reintroduced.)

## Risk

None to anything deployed. The Ansible playbook uses the `local-setup/` copy, which is unchanged by this PR. This only repairs a file that is currently guaranteed to fail if used.

## Deliberately not included

Part 2 of #1617 — deciding whether the root copies should exist at all, and the remaining drift in `otel/loki-config.yaml` and `docker-compose.egov-digit.yaml` — is a separate change. It needs a discussion about repository layout, and defusing the landmine should not wait behind that.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated log collection configuration to remove unsupported sample-rejection settings.
  * Clarified that sample age limits are enforced by the log storage service.

### DIFF
--- a/otel/promtail-config.yaml
+++ b/otel/promtail-config.yaml
@@ -13,12 +13,12 @@ positions:
 clients:
   - url: http://loki:3100/loki/api/v1/push
 
-limits_config:
-  # Drop log entries older than 7 days (matches Loki's reject threshold) so
-  # the first sweep over historical /var/lib/docker/containers/*-json.log
-  # files doesn't hammer Loki with 400s.
-  reject_old_samples: true
-  reject_old_samples_max_age: 168h
+# Note: `reject_old_samples` / `reject_old_samples_max_age` belong in
+# Loki's loki-config.yaml, NOT here — promtail's limits_config schema
+# doesn't accept them and rejects the whole config with a fatal:
+#   "field reject_old_samples not found in type limit.Config"
+# Loki already enforces the 7-day rejection on its side, so there's
+# nothing to mirror in promtail.
 
 scrape_configs:
   - job_name: docker


singhalkarun:
Note: #1637 (part 2 of #1617) now **deletes** this file entirely, following the decision to remove the root-level duplicates rather than CI-guard them.

The two do not conflict mechanically — this one edits the file, that one removes it — but the outcome depends on merge order:

- **#1637 first** → this PR becomes unnecessary and can be closed unmerged.
- **this PR first** → #1637 deletes the file it just repaired. Also fine.

Either way the landmine is gone. Keeping them separate was deliberate: defusing a config that is guaranteed to fail should not have waited on a repository-layout decision.

vinothrallapalli-eGov (approved):

PR #1625, opened 2026-08-06
fix(prometheus): reload by SIGHUP, drop --web.enable-lifecycle

### LINKED ISSUE #1608: Prometheus accepts unauthenticated config reload and shutdown (--web.enable-lifecycle)
Part of #1601 (monitoring stack umbrella). **P0-security · effort: XS**

## Problem

Prometheus (the container storing our metrics — `digit-prometheus`, port 19090) is started with `--web.enable-lifecycle` (`local-setup/docker-compose.egov-digit.yaml:84`).

That option enables two **unauthenticated** HTTP endpoints:

- `POST /-/reload` — re-read the config file from disk
- `POST /-/quit` — **shut Prometheus down**

Neither requires a credential of any kind. Anyone who can reach the port can stop metrics collection with a single request. Today that port is published on `0.0.0.0` (#1603), so it is reachable from the internet; once #1603 lands the internet reach closes, but the capability remains available to anything running on the host — a compromised container, or any service that can be induced to make a request to `127.0.0.1:19090`.

## Why the flag is there

Exactly one deploy step uses it (`local-setup/ansible/playbook-deploy.yml:1850`), and the reason is legitimate. `docker compose up -d` recreates a container only when its service *spec* changes — editing the bind-mounted `otel/prometheus.yml` does not trigger a recreate. So on a redeploy of a running tenant, Prometheus keeps its old in-memory config and silently ignores a newly added scrape job. This is not hypothetical: it is what the node-exporter job in #1335 ran into. Posting to `/-/reload` after the deploy makes Prometheus re-read the file in place.

## Fix

Prometheus reloads its configuration on **SIGHUP**, which is equivalent to `POST /-/reload` and carries no HTTP surface at all.

1. Replace the reload task:

```yaml
- name: Reload Prometheus config (SIGHUP — lifecycle endpoint removed)
  ansible.builtin.command: docker kill -s HUP digit-prometheus
  register: prometheus_reload
  changed_when: prometheus_reload.rc == 0
  failed_when: false   # first deploy: Prometheus may not be up, and boots with fresh config anyway
```

(`docker kill` despite its name only delivers a signal; HUP tells Prometheus to reload, not to exit.)

2. Drop `--web.enable-lifecycle` from the Prometheus command in `docker-compose.egov-digit.yaml`, and update the comment block above the task, which currently explains the `/-/reload` mechanism.

The `failed_when: false` is retained deliberately: on a first deploy Prometheus may not be running yet, and in that case it has already booted with the current config, so a failed signal is not an error.

## Acceptance

- [ ] A redeploy that adds a scrape job shows the new target in Prometheus without a container recreate (the #1335 scenario)
- [ ] `POST http://127.0.0.1:19090/-/reload` returns 404 from **on the host** (proving the endpoint is gone, not merely unreachable)
- [ ] Existing metrics and dashboards are unaffected

## Notes

Unlike the Grafana password fallback in #1602, there is no tradeoff here — SIGHUP is a strictly equivalent replacement, so the endpoints can be removed outright rather than merely restricted. Complements #1603: that PR makes the port unreachable from outside, this one removes the capability itself.

singhalkarun:
Fixes #1608

## What

Prometheus ran with `--web.enable-lifecycle`, which enables two **unauthenticated** HTTP endpoints:

- `POST /-/reload` — re-read the config file
- `POST /-/quit` — **shut Prometheus down**

The second is a remote kill switch for the metrics store, available to anyone who can reach the port. Today that is the internet, since the port is published on `0.0.0.0` (fixed separately in #1606); after that lands, it is still available to anything running on the host.

## Why the flag was there

Exactly one deploy step used it, and for a real reason: `docker compose up -d` recreates a container only when its service *spec* changes, so editing the bind-mounted `otel/prometheus.yml` does not restart Prometheus. On a redeploy the live process keeps its old in-memory config and silently ignores a newly added scrape job — which is precisely what the node-exporter job in #1335 ran into.

## Change

Prometheus re-reads its configuration on **SIGHUP**, which is documented and equivalent to `POST /-/reload`, and carries no HTTP surface.

```diff
-    - name: Reload Prometheus config (pick up new scrape jobs without a recreate)
-      ansible.builtin.uri:
-        url: http://127.0.0.1:19090/-/reload
-        method: POST
-        status_code: [200]
+    - name: Reload Prometheus config (pick up new scrape jobs without a recreate)
+      ansible.builtin.command: docker kill -s HUP digit-prometheus
       register: prometheus_reload
-      changed_when: prometheus_reload.status == 200
+      changed_when: prometheus_reload.rc == 0
       failed_when: false
```

and `--web.enable-lifecycle` is removed from the Prometheus command, replaced by a comment recording why it must not come back.

`docker kill` only **delivers** a signal despite its name; HUP does not terminate the process.

`failed_when: false` is kept deliberately: on a first deploy Prometheus may not be up yet, and in that case it has already booted with the current config, so a failed signal is not an error.

## Also touches the root-level copy

The drifted root `docker-compose.egov-digit.yaml` carried the same flag. Leaving it would keep the exposed endpoint for anyone deploying from the repo root, so the same change is applied there. That copy's wider drift is tracked by issue #1617, and the file is **deleted outright by PR #1637** (`chore/1617-remove-root-duplicate-configs`, base `monitoring-fix`), which removes the root `docker-compose.egov-digit.yaml` and the root `otel/` tree. This PR does not attempt to reconcile the drift.

**Merge order:** #1637 and this PR both touch that file, so one direction is a modify/delete conflict — resolve it by **taking the delete**. If #1637 lands first, this hunk is moot and should simply be dropped; if this PR lands first, #1637's delete absorbs it (deleting a file whose comment differs is semantically identical).

## Verification

- [x] No live `--web.enable-lifecycle` remains in any compose file (only explanatory comments)
- [x] All three modified files parse as YAML
- [ ] Redeploy with a new scrape job → target appears in Prometheus without a container recreate (the #1335 scenario)
- [ ] `POST http://127.0.0.1:19090/-/reload` returns 404 **from on the host** — proving the endpoint is gone, not merely unreachable
- [ ] Existing dashboards and metrics unaffected

The last three need a real box; not yet deployed.

## Note

Unlike the Grafana password fallback in #1605, there is no tradeoff here — SIGHUP is a strictly equivalent replacement, so the capability is removed outright rather than restricted.

### DIFF
--- a/docker-compose.egov-digit.yaml
+++ b/docker-compose.egov-digit.yaml
@@ -81,7 +81,10 @@ services:
     command:
       - --config.file=/etc/prometheus/prometheus.yml
       - --storage.tsdb.retention.time=15d
-      - --web.enable-lifecycle
+      # NOTE: no --web.enable-lifecycle (#1608) — it exposes POST /-/reload
+      # AND POST /-/quit unauthenticated. The Ansible deploy reloads by
+      # SIGHUP instead. (This root-level copy is separately tracked for
+      # removal or CI drift-checking in #1617.)
     volumes:
       - ./otel/prometheus.yml:/etc/prometheus/prometheus.yml:ro
       - prometheus_data:/prometheus
--- a/docker-compose.egov-digit.yaml
+++ b/docker-compose.egov-digit.yaml
@@ -83,8 +83,10 @@ services:
       - --storage.tsdb.retention.time=15d
       # NOTE: no --web.enable-lifecycle (#1608) — it exposes POST /-/reload
       # AND POST /-/quit unauthenticated. The Ansible deploy reloads by
-      # SIGHUP instead. (This root-level copy is separately tracked for
-      # removal or CI drift-checking in #1617.)
+      # SIGHUP instead. (This whole root-level file is deleted by #1637,
+      # which closes drift issue #1617. If #1637 lands first this hunk is
+      # moot; if it lands second, resolve the modify/delete by taking the
+      # delete. The live copy is local-setup/docker-compose.egov-digit.yaml.)
     volumes:
       - ./otel/prometheus.yml:/etc/prometheus/prometheus.yml:ro
       - prometheus_data:/prometheus
--- a/local-setup/ansible/playbook-deploy.yml
+++ b/local-setup/ansible/playbook-deploy.yml
@@ -1844,16 +1844,22 @@
     # live Prometheus keeps its old in-memory config and never picks up a newly
     # added scrape job (e.g. the `node` job for node-exporter) until it happens
     # to restart. A fresh deploy is unaffected — Prometheus boots with the new
-    # config. Prometheus runs with --web.enable-lifecycle, so POST /-/reload
-    # rereads the file in place. Best-effort: on a first deploy Prometheus may
-    # not be up yet (and already loaded fresh config), so a failure here is fine.
+    # config.
+    #
+    # Reload is by SIGHUP, not POST /-/reload (#1608). The HTTP lifecycle
+    # endpoints are enabled by --web.enable-lifecycle, which also exposes
+    # POST /-/quit — an unauthenticated remote shutdown of the metrics store.
+    # SIGHUP is Prometheus's documented equivalent for re-reading config and
+    # carries no HTTP surface, so the flag is gone from the compose file.
+    # `docker kill` only DELIVERS the signal despite its name; HUP does not
+    # terminate the process.
+    #
+    # Best-effort: on a first deploy Prometheus may not be up yet (and has
+    # already loaded fresh config), so a failure here is fine.
     - name: Reload Prometheus config (pick up new scrape jobs without a recreate)
-      ansible.builtin.uri:
-        url: http://127.0.0.1:19090/-/reload
-        method: POST
-        status_code: [200]
+      ansible.builtin.command: docker kill -s HUP digit-prometheus
       register: prometheus_reload
-      changed_when: prometheus_reload.status == 200
+      changed_when: prometheus_reload.rc == 0
       failed_when: false
 
     # ==================== digit-ui mode convergence ====================
--- a/local-setup/docker-compose.egov-digit.yaml
+++ b/local-setup/docker-compose.egov-digit.yaml
@@ -81,7 +81,11 @@ services:
     command:
       - --config.file=/etc/prometheus/prometheus.yml
       - --storage.tsdb.retention.time=15d
-      - --web.enable-lifecycle
+      # NOTE: no --web.enable-lifecycle (#1608). That flag exposes
+      # POST /-/reload AND POST /-/quit with no authentication — a remote
+      # shutdown for the metrics store. The deploy reloads config by sending
+      # SIGHUP instead (playbook-deploy.yml, "Reload Prometheus config"),
+      # which is Prometheus's documented equivalent and has no HTTP surface.
     volumes:
       - ./otel/prometheus.yml:/etc/prometheus/prometheus.yml:ro
       - prometheus_data:/prometheus


singhalkarun:
**CI note — the `tilt-test` failure is a flake, not a regression from this PR.**

Worth showing the evidence, since "PR changes compose file → compose-based test fails" is the obvious suspicion.

**1. The file this PR edits is not loaded by that job.** The sidecar logs its own inputs:

```
digit-telemetry | [telemetry] files=.../local-setup/docker-compose.yml,.../local-setup/docker-compose.tilt.yml
```

`docker-compose.egov-digit.yaml` is not among them — and it is the only compose file that defines `prometheus` at all. The changed configuration is not present in the failing job.

**2. The failure is a timing race, and the step's own comment says so:**

```
# tilt ci exits as soon as resources are healthy, so the sidecar
# captures fewer events than the long-running docker compose job.
```

Observed counts: `Container start events: 0`, `Container healthy events: 2`. The sidecar was running and capturing — it simply attached after the containers had already started, so it saw the healthy transitions and missed the start ones.

**3. It fails on `master` too** — `tilt-ci.yaml` on `master` shows failure on 2026-08-05 with successes on either side, and it passes on #1605, #1607 and #1627 while failing here and on #1606. That distribution tracks runner timing, not file contents.

Not something this PR can fix. The underlying flake (`>= 1 start event` is not guaranteed when `tilt ci` exits early) deserves its own issue — the assertion should either wait for the sidecar to attach before starting containers, or drop the start-event requirement and rely on healthy events.

vinothrallapalli-eGov (commented):
Reviewed for correctness/security. The fix is sound: `--web.enable-lifecycle` is fully removed (not just commented) from both compose copies, so neither `/-/reload` nor `/-/quit` remains HTTP-reachable — confirmed no other script/compose file in the repo (Tilt/k8s paths, Grafana provisioning, kong-test.sh, CI) assumes the lifecycle endpoint is reachable. `docker kill -s HUP` targets the correct fixed container name (`container_name: digit-prometheus`, single-tenant-per-host model matches how `kong-gateway`/`egov-persister` are referenced elsewhere in the same playbook). The official `prom/prometheus` image uses an exec-form `ENTRYPOINT ["/bin/prometheus"]` with no shell wrapper, so the signal reaches PID 1 directly — no swallowed-signal risk there. One pre-existing gap, not introduced by this PR: the reload task runs after health-waits for Kong/persister/HRMS but Prometheus itself has no compose healthcheck (Grafana's `depends_on: prometheus: condition: service_started`, not `service_healthy`), so in principle the SIGHUP could race a slow TSDB/WAL replay on an unusually slow box; `failed_when: false` masks that either way, same as the old HTTP call did. Also flagging the real modify/delete conflict with #1637 for merge-order awareness. Overall: low severity findings, no blockers.

docker-compose.egov-digit.yaml
```diff
@@ -81,7 +81,10 @@ services:
     command:
       - --config.file=/etc/prometheus/prometheus.yml
       - --storage.tsdb.retention.time=15d
-      - --web.enable-lifecycle
+      # NOTE: no --web.enable-lifecycle (#1608) — it exposes POST /-/reload
```
vinothrallapalli-eGov, line 84: This root-level copy is still being edited here, and PR #1637 deletes this entire file (`docker-compose.egov-digit.yaml`, 1674 lines) as part of removing the root-level observability duplicate. That's a genuine modify/delete conflict — GitHub will flag it, and per the stated plan the resolution should "take the delete." The PR description cross-references #1617 (a drift-tracking issue) for this file's fate, but the actual deletion lives in #1637 — worth updating that reference so whoever resolves the conflict knows which PR wins. Not a defect in this PR's logic, just a merge-order note: if #1637 lands first, this hunk becomes moot; if this PR lands first, #1637's delete will need to absorb it (trivial, since deleting a file with these lines vs. the old flag doesn't change anything semantically).

local-setup/ansible/playbook-deploy.yml
```diff
@@ -1844,16 +1844,22 @@
...
+    # terminate the process.
+    #
+    # Best-effort: on a first deploy Prometheus may not be up yet (and has
+    # already loaded fresh config), so a failure here is fine.
     - name: Reload Prometheus config (pick up new scrape jobs without a recreate)
-      ansible.builtin.uri:
-        url: http://127.0.0.1:19090/-/reload
-        method: POST
-        status_code: [200]
+      ansible.builtin.command: docker kill -s HUP digit-prometheus
```
vinothrallapalli-eGov, line 1860: Verified this is safe as designed: the official `prom/prometheus` Docker image entrypoint is exec-form (`ENTRYPOINT ["/bin/prometheus"]`, no shell wrapper), so `docker kill -s HUP` reaches the Prometheus PID 1 directly and its SIGHUP handler is installed very early in `main()` — no swallowed-signal path here. One thing worth noting for the failure-mode question this PR raises: this task runs after health-waits for Kong/persister/HRMS, but Prometheus itself has no compose healthcheck and Grafana's `depends_on` on it uses `condition: service_started` (not `service_healthy`). So there's no explicit gate proving the Prometheus process has finished its startup/WAL-replay before the signal is sent — on a slow disk this could in principle race. In practice this is very unlikely (Prometheus installs its reload handler well before TSDB open completes, and by this point in the playbook several minutes have usually elapsed) and `failed_when: false` already accepts silent no-ops here by design, so this isn't a regression from the old HTTP-based reload (which had the identical gap). Low severity, just flagging since the review explicitly asked about this race.

docker-compose.egov-digit.yaml
singhalkarun: Confirmed, and both halves of the reference are now fixed.

Verified: PR #1637 (`chore/1617-remove-root-duplicate-configs` → `monitoring-fix`, still open) deletes `docker-compose.egov-digit.yaml` at the repo root along with the whole root `otel/` tree, and #1617 is the drift *issue*, not the deletion. So the note was pointing at the tracker rather than at the PR that actually decides the file's fate.

Also checked, so the record is complete:
- The root copy is the only other place the flag lived — `grep` for `enable-lifecycle` / `-/reload` / `-/quit` across the tree now returns explanatory comments only, no live flag (a1e9d7cc).
- Neither `docker-compose.monitoring.yml` overlay overrides `prometheus.command`, so nothing re-adds the flag at merge time.
- Both compose files still parse and both `prometheus.command` lists resolve to exactly `[--config.file=…, --storage.tsdb.retention.time=15d]`.

Changed:
- a1e9d7cc rewrites the inline note at that line to name #1637 as the deleting PR (keeping #1617 as the drift issue) and to state the resolution explicitly: take the delete.
- PR description updated with the same, plus a "Merge order" paragraph.

No functional change — agreed this is a merge-order note, not a defect. If #1637 lands first, drop this hunk entirely rather than reapplying it.

local-setup/ansible/playbook-deploy.yml
singhalkarun: Agreed on the verdict — no change made. I re-checked the specifics rather than taking them on trust:

- `prom/prometheus:v2.55.1` entrypoint is exec-form `["/bin/prometheus"]` with no shell wrapper, so `docker kill -s HUP` lands on PID 1. Confirmed.
- Prometheus has **no** `healthcheck` in `local-setup/docker-compose.egov-digit.yaml` (keys are `image, container_name, restart, command, volumes, ports, networks`), and Grafana's `depends_on: prometheus: condition: service_started` — exactly as described. So the gap is real.
- Permissions: the play runs `become: "{{ deploy_become | default(true) }}"`, and the neighbouring health-waits already shell out to bare `docker inspect`, so `docker kill` needs no extra privilege on either the Linux or the Darwin (`deploy_become: false`) path. No silent-EACCES variant of the no-op.
- `container_name: digit-prometheus` is explicit in both compose files, so the name is not compose-project-prefix dependent.
- CI: "Ansible static validation" passes with the `command` task; the one red check (`test` → "Localization module check") is unrelated to this change.

One refinement to the failure mode, which slightly changes its shape but not the severity: the default disposition of SIGHUP is **terminate**, not ignore. So a signal that genuinely beat `signal.Notify` would not be a silent no-op — it would kill the container. In practice that is still self-correcting: `restart: unless-stopped` brings it straight back, and the restarted process boots from the already-rsynced `otel/prometheus.yml`, i.e. the exact config the reload was trying to install. So even the losing side of the race converges on the intended state.

The window itself: the handler is registered while the run-group actors are constructed, before TSDB open, and this task fires after the Kong / persister / HRMS health-waits — minutes of wall clock after `up -d`. Not worth gating on.

If we ever want it closed properly, the clean fix is a `prometheus` healthcheck against `/-/healthy` (always-on, not gated by `--web.enable-lifecycle`) plus a `condition: service_healthy` and a wait-until-healthy before the signal — but that needs a real box to confirm the busybox-based image can actually run the probe, so I am not adding it blind here. Tracking it as a follow-up rather than folding it into #1608.

PR #1626, opened 2026-08-06
feat(dashboard): public unauthenticated dashboard entry + gateway allowlist

KDwevedi:
Puts a UI on the backend's existing PUBLIC floor, and makes that floor an explicit gateway contract instead of a side effect.

Split out of the original combined PR. The duplicate-copy removal is now **#1633**; the two share **no files**, so they can merge in either order.

> **Draft:** the code is complete but **not built locally** — `digit-ui-esbuild/node_modules` is empty in my environment, so `node esbuild.build.js` has not been run. The per-entry asset-matching logic was verified against a simulated metafile; the full build was not. Please confirm a real build emits `build/public-dashboard.html` + `build/public-dashboard.{js,css}` before this leaves draft.

## The gap this closes

Nothing has ever served an anonymous visitor a dashboard. Both shells gate on `hasDashboardSession()` and render `DashboardLogin`. The public capability that *does* exist is server-side — the 10 `rbac.visibleTo:["PUBLIC"]` KPI defs `AnalyticsService` serves to a role-less caller. This PR is the UI for it.

## `mode` replaces the `embedded` boolean

`embedded` conflated two independent things: *who owns the chrome* and *whether a session is required*.

| mode | chrome | session |
|---|---|---|
| `embedded` | host (DigitUI) | guaranteed by AppModules |
| `standalone` | own shell | own login gate |
| `public` | own shell | **none** — anonymous requests |

`embedded` is kept as a legacy alias, so `Module.js` is untouched. Public renders the standalone shell with `onSignOut` omitted — `Sidebar` already draws that control conditionally, so nothing downstream needed a new prop.

The session-expiry listener (#1466) is now bound **only** in standalone mode. Public has no session to expire, and dropping a public visitor onto a login form would be a regression, not a recovery.

## A second entry, not a fork

`src/public-dashboard.js` is a second esbuild entry over the **same** `products/dashboard` tree. It deliberately skips `initLibraries()` so no `window.Digit` runtime exists, which engages fallbacks the subtree already has:

- `useDashboardConfig` → `{config: null, loading: false}` (no MDMS gate)
- `i18n/localeRuntime` → standalone branch, fetching bundles from `/localization/messages/v1/_search`

The subtree imports no `@egovernments` modules and no react-router, so this stays a small independent bundle.

Both entries emit into the same `build/` docroot so hashed assets resolve against one `publicPath`. `/dashboard` redirects to `public-dashboard.html` — one redirect covering both `digit_ui_mode` branches without duplicating the `sub_filter` block.

**`generateHTML` had a latent bug** the second entry would have triggered: it injected *every* entry output into `index.html`. Correct with one entry; with two, each page would have loaded the other's bundle. Now matched per-entry via `metafile.entryPoint`.

## Gateway: making the public floor explicit

`/packs`, `/catalog/_search` and `/_query` are added to Kong's `AUTH_OPTIONAL` **and** env.yaml's mixed-mode whitelist.

Previously anonymous analytics worked *only* because Kong ships `ENFORCE_UNAUTH = false` (`kong.yml:87`) — a flag whose own comment says to flip it after an observation period — and would have **401'd outright on the K8s path**, where analytics appears in neither whitelist. Declaring the three paths turns the public floor into a contract that survives that flip.

**`/_schema` is deliberately not whitelisted.** The dashboard never calls it, and it enumerates every grain's columns including UUID dimensions, so it should start 401ing as soon as enforcement is on.

To be clear about where the boundary actually is: **not these lists.** The server-side PUBLIC floor is the control — a role-less caller resolves only `PUBLIC`-tagged defs and may invoke them by `kpiId` reference only; inline query grammar is refused (`AnalyticsService.java:133,172`). The whitelist entries only stop the gateway from rejecting the call before that logic runs.

## Reachability

| URL | Serves |
|---|---|
| `/digit-ui/employee/dashboard` | employee dashboard, MDMS role gate (unchanged) |
| `/dashboard` → `/digit-ui/public-dashboard.html` | public dashboard, anonymous |

## Known gap

**There is no `PUBLIC` `DashboardPack`.** An anonymous caller therefore hits the no-pack fallback (`AnalyticsController.java:96-98`) and receives all visible tiles with an empty `defaultLayout` — uncurated, default sizes, catalog order. Seed a `public-default` pack before this faces real users. Happy to add it here if you'd prefer it not ship separately.

## Verification

- ✅ `.github/scripts/check-gateway-whitelist-parity.py` → `OK: gateway auth-optional whitelists match (44 entries)`
- ✅ its `--self-test` → `drift is detected in both directions`
- ✅ `env.yaml` still parses as YAML
- ✅ per-entry asset matching: each page resolves only its own `.js`/`.css`
- ❌ full esbuild build (deps not installed here)
- ❌ runtime check against a live backend

Refs #29


KDwevedi:
Closing this. Review turned up a blocker in the gateway change, and the design is better settled on the ticket before any code lands.

The blocker: adding the analytics paths to `egov-mixed-mode-endpoints-whitelist` is not safe. The Spring gateway does not strip a client-supplied `RequestInfo.userInfo` on whitelisted paths — `AuthPreCheckFilterHelper`'s mixed-mode branch sets `doAuth=false` and returns the body unmodified, and `AuthCheckFilterHelper` (the only place userInfo is replaced with a token-vouched User) runs only when `doAuth` is true. Kong does strip it. Since the analytics PUBLIC floor is derived entirely from `userInfo.roles`, that whitelist entry would let an anonymous caller forge roles and bypass the floor. Kong-only is fine; K8s is not, and there is no Helm chart for `digit-ui-esbuild` anyway, so the entry bought nothing there.

Design notes and the full review findings will go on the public dashboard ticket.

Issue #1612, opened 2026-08-05
Observability containers run without memory limits and can starve the workload


singhalkarun:
Part of #1601 (monitoring stack umbrella). **P1 · effort: S to write, +1 week soak before defaulting**

## Measured starting point

In `local-setup/docker-compose.egov-digit.yaml` (the Ansible-deployed stack):

- **29 services carry `-Xmx`, totalling 12.8 GB of configured max heap.** Tenant boxes are commonly 16 GB.
- Only **3 of 60** services declare any memory constraint.
- **None of the observability containers do** — grafana, prometheus, loki, promtail, tempo, otel-collector, gatus, node-exporter all run uncapped.

For contrast, `docker-compose.yml` (the local/CI stack) caps **28 of 34** services, gatus included.

## This is not "they forgot"

The deployed stack omitting limits looks deliberate — `local-setup/README.md` describes a sibling file as the "deploy variant (no resource limits)". For **application** services that is a defensible choice: on a dedicated box, capping a JVM service means refusing citizen requests when it needs headroom.

The argument here is narrower, and rests on an asymmetry:

- A Java service consuming memory is **the workload doing its job**.
- Loki serving an expensive query, or Prometheus growing its TSDB, is **overhead**. It has no claim on memory the workload needs.

With 12.8 GB of heap configured against 16 GB of RAM, headroom is already thin. And these boxes have **no swap** (confirmed on Bomet), so memory pressure does not degrade gracefully — the kernel OOM-kills, and it targets the largest consumer, which will almost always be a JVM service rather than Loki. Net effect: **the monitoring misbehaves and a citizen-facing service is killed for it.** Exactly backwards.

## Proposed caps

Applied via `mem_limit` (compose-spec, no swarm required), ~2.2 GB ceiling in total:

| Container | mem_limit | Rationale |
|---|---|---|
| loki | 512m | largest real-world grower — chunks plus query working set |
| prometheus | 512m | 15d TSDB, scrape targets growing (#1601 adds Kong + Postgres) |
| tempo | 384m | 24h retention, small blocks |
| otel-collector | 320m | its internal `memory_limiter` is 200 MiB — this is the hard backstop above it |
| grafana | 256m | |
| promtail | 128m | |
| gatus | 64m | matches the k3s Deployment, which already sets 64Mi |
| node-exporter | 64m | |

[...]

PR #1627, opened 2026-08-06
feat(kong): expose gateway metrics to Prometheus + dashboard

### LINKED ISSUE #1614: Kong serves every request and produces no metrics (bundled prometheus plugin is off)
Part of #1601 (monitoring stack umbrella). **P1 · effort: XS config, S with dashboard polish**

## Problem

Kong is the front door — every citizen action (filing a complaint, login, dashboard load) transits it, and it is the only component with a view of all traffic. It currently contributes **zero metrics**.

We cannot answer, from any dashboard:

- how many requests are we serving?
- what fraction are failing, and on which route?
- which endpoints are slow, and are they getting slower?
- is the error rate climbing right now?

The only route to those answers today is grepping raw log lines in Loki. During an incident, "are we returning errors and where" is usually the *first* question.

Prometheus currently scrapes exactly two targets (`local-setup/otel/prometheus.yml`): `otel-collector:8889` and `node-exporter:9100`. Neither has any view of gateway traffic.

## Why this is cheap

Kong **bundles** a `prometheus` plugin — nothing to install, no new container, no image change. This is configuration only, enabling code already in the binary (`egovio/kong:3.6`).

It publishes on Kong's **status listener**, `KONG_STATUS_LISTEN: 0.0.0.0:8100` — a port Gatus already health-checks (`gatus/config.yaml:373-375`, `http://kong:8100/status`). The port is already reachable on `egov-network`; we would read a different path on it.

## Fix

1. **Enable the plugin** — global `plugins:` block in `local-setup/kong/kong.yml`:

```yaml
- name: prometheus
  config:
    status_code_metrics: true
    latency_metrics: true
    bandwidth_metrics: true
    upstream_health_metrics: true
```

2. **Scrape it** — add to `local-setup/otel/prometheus.yml`:

```yaml
  - job_name: kong
    static_configs:
      - targets: ['kong:8100']
```

(Prometheus reads `/metrics` by default, which is where the plugin serves.)

3. **Dashboard** — import the official Kong dashboard (grafana.com ID 7424) into `otel/grafana/provisioning/dashboards/`, same pattern as the vendored `node-exporter-full.json`. No dashboard authoring required.

## Scope — deliberately per-route, not per-consumer

Kong's plugin can emit per-consumer metrics. **Not proposed**: it multiplies series count for little incident-response value, and on a memory-tight box (#1612) metric cardinality is a real cost. Per-route status codes and latency are the right altitude.

## Acceptance

- [ ] `curl -s http://kong:8100/metrics | head` returns Prometheus-format output from inside the network
- [ ] `kong_http_requests_total` and latency histograms visible in Prometheus
- [ ] Dashboard renders per-route request rate, status-code breakdown and latency percentiles
- [ ] Kong's own request handling is unaffected (the plugin adds negligible overhead, but confirm the CI `test` job still passes end to end)
- [ ] Series count sanity-checked after a day — no cardinality blow-up from high-variance route names

## Relates to

- #1611 — gateway error rate is the best single proxy for "are users actually affected", and would make a strong alert rule once this data exists.
- #1612 — cardinality is why per-consumer metrics are out of scope.

singhalkarun:
Fixes #1614

## What

Kong is the front door — every citizen action transits it — and it produced no metrics. Prometheus scraped exactly two targets (`otel-collector:8889`, `node-exporter:9100`), neither with any view of gateway traffic. During an incident, "are we returning errors, and on which route?" could only be answered by grepping raw log lines in Loki.

## Change — config only

**1. Enable the bundled plugin** (`local-setup/kong/kong.yml`, global `plugins:`):

```yaml
- name: prometheus
  config:
    status_code_metrics: true
    latency_metrics: true
    bandwidth_metrics: true
    upstream_health_metrics: true
```

Nothing is installed — the plugin ships inside `egovio/kong:3.6`.

**2. Scrape it** (`local-setup/otel/prometheus.yml`):

```yaml
  - job_name: kong
    static_configs:
      - targets: ['kong:8100']
```

It serves on Kong's **status** listener (`KONG_STATUS_LISTEN 0.0.0.0:8100`) — the same port Gatus already health-checks at `/status` — not the proxy or admin port. Prometheus reaches it by DNS over `egov-network`, so **no host port is published**, consistent with #1606 closing the others.

**3. Dashboard** — the official "Kong (official)" board (grafana.com ID 7424), vendored alongside `node-exporter-full.json`.

## Two details worth reviewing

**The vendored dashboard needed fixing up.** As downloaded it carries `__inputs` / `__requires` import-wizard scaffolding and refers to its datasource as `${DS_PROMETHEUS}`. That placeholder is only substituted by Grafana's *interactive* import — under **file provisioning it is never resolved**, so every panel would render "datasource not found". Stripped the scaffolding and pinned every datasource to `{"type": "prometheus", "uid": "prometheus"}`, matching the convention `node-exporter-full.json` already uses and the uid declared in `provisioning/datasources/prometheus.yaml`. Also set a stable `uid: kong-gateway` and dropped the numeric `id`, so redeploys update in place rather than creating duplicates.

**`per_consumer` is deliberately off.** It multiplies series count for little incident-response value, and metric cardinality is a real cost on a memory-tight box (#1612). The four toggles enabled are precisely the ones the dashboard consumes — verified by extracting the metric names it queries:

```
kong_http_requests_total        <- status_code_metrics
kong_bandwidth_bytes            <- bandwidth_metrics
kong_{request,kong,upstream}_latency_ms_bucket  <- latency_metrics
kong_upstream_target_health     <- upstream_health_metrics
kong_memory_*, kong_nginx_connections_total     <- always exported
```

## Verification

- [x] `kong.yml` and `prometheus.yml` parse; plugin and scrape job present
- [x] Dashboard is valid JSON, 6 panels, no `${DS_` or `__inputs` remaining
- [x] Repo static tests: same 2 failures as clean `master`, none new (see below)
- [ ] `curl -s http://kong:8100/metrics` returns Prometheus-format output from inside the network
- [ ] `kong_http_requests_total` appears in Prometheus; dashboard populates
- [ ] Series count sanity-checked after a day — no cardinality blow-up from high-variance route names

The last three need a running box; not yet deployed.

## CI note

Two static tests fail — `no :latest image pins beyond the frozen debt list` and `seeds INTERNAL_USER on state_root after bootstrap` — **both reproduce on a clean `master` checkout** with no changes applied. This PR touches neither image pins nor the playbook. Same pre-existing failures noted on #1605–#1607.

### DIFF
--- a/local-setup/kong/kong.yml
+++ b/local-setup/kong/kong.yml
@@ -52,6 +52,22 @@ plugins:
   config:
     allowed_payload_size: 10
     size_unit: megabytes
+# Gateway metrics (#1614). Kong bundles this plugin — nothing to install.
+# It serves Prometheus-format metrics on the STATUS listener
+# (KONG_STATUS_LISTEN 0.0.0.0:8100, the port Gatus already health-checks)
+# at /metrics; Prometheus scrapes it by DNS over egov-network, so no host
+# port is published. Every citizen request crosses Kong, so this is the
+# only source of per-route status-code and latency data we have.
+#
+# per_consumer is deliberately NOT enabled: it multiplies series count for
+# little incident-response value, and metric cardinality is a real cost on
+# a memory-tight box (#1612).
+- name: prometheus
+  config:
+    status_code_metrics: true
+    latency_metrics: true
+    bandwidth_metrics: true
+    upstream_health_metrics: true
 - name: pre-function
   config:
     access:
--- a/local-setup/otel/grafana/provisioning/dashboards/kong-gateway.json
+++ b/local-setup/otel/grafana/provisioning/dashboards/kong-gateway.json
@@ -1,0 +1,3097 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": {
+          "type": "prometheus",
+          "uid": "prometheus"
+        },
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "description": "Dashboard that graphs metrics exported via Prometheus plugin in Kong (http://github.com/kong/kong)",
+  "editable": true,
+  "gnetId": 7424,
+  "graphTooltip": 0,
+  "iteration": 1662693484232,
+  "links": [
+    {
+      "asDropdown": false,
+      "icon": "info",
+      "includeVars": false,
+      "keepTime": false,
+      "tags": [],
+      "targetBlank": true,
+      "title": "Prometheus Plugin Config",
+      "tooltip": "",
+      "type": "link",
+      "url": "https://docs.konghq.com/hub/kong-inc/prometheus/"
+    }
+  ],
+  "panels": [
+    {
+      "collapsed": true,
+      "datasource": {
+        "type": "prometheus",
+        "uid": "prometheus"
+      },
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 0
+      },
+      "id": 38,
+      "panels": [
+        {
+          "aliasColors": {},
+          "bars": false,
+          "dashLength": 10,
+          "dashes": false,
+          "datasource": {
+            "type": "prometheus",
+            "uid": "prometheus"
+          },
+          "fieldConfig": {
+            "defaults": {
+              "links": [],
+              "custom": {
+                "fillOpacity": 10,
+                "showPoints": "never",
+                "spanNulls": true
+              },
+              "unit": "reqps"
+            },
+            "overrides": []
+          },
+          "fill": 1,
+          "fillGradient": 0,
+          "gridPos": {
+            "h": 7,
+            "w": 24,
+            "x": 0,
+            "y": 1
+          },
+          "hiddenSeries": false,
+          "id": 1,
+          "legend": {
+            "alignAsTable": false,
+            "avg": false,
+            "current": false,
+            "max": false,
+            "min": false,
+            "rightSide": false,
+            "show": true,
+            "total": false,
+            "values": false
+          },
+          "lines": true,
+          "linewidth": 1,
+          "links": [],
+          "nullPointMode": "null",
+          "options": {
+            "alertThreshold": true,
+            "legend": {
+              "calcs": [],
+              "displayMode": "hidden",
+              "placement": "bottom"
+            },
+            "tooltip": {
+              "mode": "multi",
+              "sort": "none"
+            }
+          },
+          "paceLength": 10,
+          "percentage": false,
+          "pluginVersion": "8.4.5",
+          "pointradius": 5,
+          "points": false,
+          "renderer": "flot",
+          "seriesOverrides": [],
+          "spaceLength": 10,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "expr": "sum(rate(kong_http_requests_total{instance=~\"$instance\"}[1m]))",
+              "format": "time_series",
+              "intervalFactor": 2,
+              "refId": "A",
+              "legendFormat": "Requests/second"
+            }
+          ],
+          "thresholds": [],
+          "timeFrom": null,
+          "timeRegions": [],
+          "timeShift": null,
+          "title": "Total requests per second (RPS)",
+          "tooltip": {
+            "shared": true,
+            "sort": 0,
+            "value_type": "individual"
+          },
+          "type": "timeseries",
+          "xaxis": {
+            "buckets": null,
+            "mode": "time",
+            "name": null,
+            "show": true,
+            "values": []
+          },
+          "yaxes": [
+            {
+              "format": "short",
+              "label": null,
+              "logBase": 1,
+              "max": null,
+              "min": null,
+              "show": true
+            },
+            {
+              "format": "s",
+              "label": null,
+              "logBase": 1,
+              "max": null,
+              "min": null,
+              "show": true
+            }
+          ],
+          "yaxis": {
+            "align": false,
+            "alignLevel": null
+          }
+        },
+        {
+          "aliasColors": {},
+          "bars": false,
+          "dashLength": 10,
+          "dashes": false,
+          "datasource": {
+            "type": "prometheus",
+            "uid": "prometheus"
+          },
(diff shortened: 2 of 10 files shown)


local-setup/kong/kong.yml
```diff
@@ -52,6 +52,22 @@ plugins:
...
+# It serves Prometheus-format metrics on the STATUS listener
+# (KONG_STATUS_LISTEN 0.0.0.0:8100, the port Gatus already health-checks)
+# at /metrics; Prometheus scrapes it by DNS over egov-network, so no host
+# port is published. Every citizen request crosses Kong, so this is the
+# only source of per-route status-code and latency data we have.
+#
+# per_consumer is deliberately NOT enabled: it multiplies series count for
+# little incident-response value, and metric cardinality is a real cost on
+# a memory-tight box (#1612).
+- name: prometheus
```
vinothrallapalli-eGov, line 65: Severity: high — scoping/exposure. This plugin serves `/metrics` on the STATUS listener (8100), and that listener is already published to the host as `18100:8100` (bind-all, not `127.0.0.1:18100:8100`) in `local-setup/docker-compose.yml:1072`, and duplicated in `docker-compose.db-migrations.yml:1271` and `docker-compose.registry.yml:1578`. None of those are touched here, but this PR is what turns that pre-existing open port into something worth exploiting: unauthenticated route names, per-route status-code counts and latency histograms for all live gateway traffic. Kong's status API does not execute ordinary plugins, so there's no way to bolt key-auth onto this port from `kong.yml` — the only real fix is network-level (bind the host mapping to `127.0.0.1` or drop it, across all three compose files).

Worth checking whether that host port is even needed: Gatus's own Kong check already hits `http://kong:8100/status` over `egov-network` (`local-setup/gatus/config.yaml:375`), not through the published port — so `18100:8100` looks like unused legacy exposure this change should close rather than build on top of, given the initiative this PR is part of is specifically about eliminating exactly this kind of gap.

local-setup/otel/grafana/provisioning/dashboards/kong-gateway.json
```diff
@@ -0,0 +1,3097 @@
...
+          "text": "Prometheus",
+          "value": "Prometheus"
+        },
+        "description": null,
+        "error": null,
+        "hide": 0,
+        "includeAll": false,
+        "label": "Datasource",
+        "multi": false,
+        "name": "DS_PROMETHEUS",
```
vinothrallapalli-eGov: Severity: low — nit. This `DS_PROMETHEUS` datasource template variable is leftover Grafana.com import-wizard scaffolding; nothing in the dashboard references `$DS_PROMETHEUS` anymore since every panel/target datasource is hardcoded to `{"type": "prometheus", "uid": "prometheus"}`. Harmless (Grafana just shows an unused dropdown), but the PR description says the scaffolding was fully stripped — this variable slipped through and could be deleted along with the rest.

vinothrallapalli-eGov (commented):
Checked the dashboard's panel `expr`s against the four Prometheus plugin toggles enabled in `kong.yml` — the metric names (`kong_http_requests_total`, `kong_bandwidth_bytes`, `kong_{kong,request,upstream}_latency_ms_bucket`, `kong_upstream_target_health`, plus the always-on `kong_nginx_connections_total`/`kong_memory_*`) are real Kong prometheus-plugin metrics, not stale/guessed ones, and the `$service`/`$route`/`$instance`/`$upstream` template variables resolve against them correctly. That part of the PR checks out. However, the PR's central security claim — "no host port is published, consistent with #1606 closing the others" — does not hold: Kong's status listener (8100) is already published as `18100:8100` (bound to 0.0.0.0) in `local-setup/docker-compose.yml`, `docker-compose.db-migrations.yml`, and `docker-compose.registry.yml`, none of which this PR touches. Gatus's own Kong healthcheck reaches it over the internal network (`http://kong:8100/status` per `local-setup/gatus/config.yaml:375`), so that host mapping isn't even required for the stated purpose — it's dead exposure this PR makes materially worse by turning it into a full `/metrics` endpoint. See inline comment for detail and a suggested fix. Also flagging a minor leftover from the dashboard import (unused `DS_PROMETHEUS` template var) and a heads-up about #1628/#1629 both editing the same anchor point in `prometheus.yml`.

local-setup/kong/kong.yml
singhalkarun: Confirmed, and it is broader than three files — `18100:8100` is published bind-all in **six**: `local-setup/docker-compose.yml:1072`, `db-migrations.yml:1271`, `registry.yml:1578`, `egov-digit.yaml:1621`, `deploy.yaml:871`, and the generated root `docker-compose.egov-digit.yaml:1497`. All six mount the same `kong/kong.yml`, so all six get the plugin.

Fixed in ace09996 — every one is now `127.0.0.1:18100:8100`, matching the openbao precedent already in the tree (`127.0.0.1:18200:8200`).

Nothing consumed the published port. Repo-wide grep for `18100` outside the port mappings returns exactly one hit: a Tiltfile comment (line 367), updated. The two real consumers both go over `egov-network` by DNS:
- Prometheus → `kong:8100` (the job this PR adds)
- Gatus → `http://kong:8100/status` (`gatus/config.yaml:375`)

So this is pure closure, no functional change. Verified with `yaml.safe_load` on all six files.

One thing deliberately **not** touched, flagging rather than silently expanding scope: the Admin API (`18001:8001`) and Kong Manager GUI (`18002:8002`) are published bind-all in the same blocks and are a strictly worse exposure than the status port — the Admin API is write-capable. Unlike 8100 they are plausibly used interactively from a browser, so loopback-binding them is a behaviour change that needs its own decision, and it is unrelated to this PR (no metrics involved). Worth a separate issue.

local-setup/otel/grafana/provisioning/dashboards/kong-gateway.json
singhalkarun: Confirmed and removed in ace09996.

`DS_PROMETHEUS` appeared exactly once in the file — its own definition in `templating.list`. Zero `$DS_PROMETHEUS` references remained; all 34 datasource references are hardcoded `{"type": "prometheus", "uid": "prometheus"}`. `__inputs`/`__requires` were already absent, so this variable was the last of the import scaffolding.

Diff is 20 lines removed, 0 added (the JSON round-trips byte-identically at 2-space indent otherwise), and the file still parses. 4 template variables left: `service`, `instance`, `route`, `upstream`.

vinothrallapalli-eGov (approved):

Issue #1623, opened 2026-08-06
Kafka/Redpanda has no metrics — consumer lag (silent data loss) is invisible


singhalkarun:
Part of #1601 (monitoring stack umbrella). **P1 · effort: S**

## Problem

The message broker is invisible. Prometheus scrapes two targets today (`otel-collector:8889`, `node-exporter:9100`); neither sees the broker. Gatus checks only that TCP port 9092 accepts a connection.

That check stays green in the failure that matters most: **consumer lag**.

`egov-persister` is what writes complaints to the database — it consumes from the broker rather than being called synchronously. If it falls behind or its consumer stalls:

- the API still returns **200**
- the service is still **up**
- Gatus is still **green**
- and the complaint **is not in the database**

There is no signal anywhere in the current stack that would reveal it. The citizen's complaint is simply missing, and the first report comes from a person, not a dashboard. Same shape for `egov-indexer` (search index silently stale) and `audit-service` (audit trail silently incomplete).

This is a **correctness** failure with no detection, which is why it is being separated from the Redis / Elasticsearch exporters that #1615 declines — those cost visibility, this costs data.

> Correction to #1615: that issue's "deliberately out of scope" section groups Kafka with Redis and Elasticsearch as "another always-on container for a question we are not currently asking". That reasoning was wrong for Kafka specifically, and this issue supersedes it.

## The broker is Redpanda — no exporter needed

`egovio/redpanda:v24.1.1`, not Apache Kafka. Redpanda **serves Prometheus metrics natively** from its admin API (`/public_metrics`, and the fuller `/metrics`), default port **9644**, including consumer-group lag. So this is the same shape as #1614: no new container, no `kafka_exporter`, just enable and scrape.

Note the current command does not mention the admin API:

```
redpanda start --smp 1 --memory 256M --overprovisioned --node-id 0
  --kafka-addr PLAINTEXT://0.0.0.0:9092 --advertise-kafka-addr PLAINTEXT://redpanda:9092
  --pandaproxy-addr PLAINTEXT://0.0.0.0:8082 --advertise-pandaproxy-addr PLAINTEXT://redpanda:8082
```

The admin API is on by default, but this should be confirmed on a running container before assuming — add explicit `--advertise-rpc-addr` / admin-addr flags if it is not listening where expected:

```bash
docker exec redpanda curl -s localhost:9644/public_metrics | head
```

[...]

Issue #1616, opened 2026-08-05
pgr-services analytics metrics are collected but no dashboard shows them


singhalkarun:
Part of #1601 (monitoring stack umbrella). **P2 · effort: S**

Unlike the rest of the umbrella, nothing here is broken or exposed. This is instrumentation that was **built, shipped, and verified working** — and that no dashboard displays, so in practice nobody looks at it.

## What already exists

`pgr-services` instruments its analytics query path (`backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsMetrics.java`), emitting two instruments with tags `kpi_id`, `grain`, `tenant`:

| Prometheus series | Meaning |
|---|---|
| `pgr_analytics_query_duration_ms_bucket` / `_sum` / `_count` | histogram — duration of one executed analytics SQL query |
| `pgr_analytics_query_rows_total` | counter — rows returned |

It is documented in `docs/observability/dashboard-metrics-server.md`. **No dashboard renders any of it.**

## Verified reaching Prometheus (not assumed)

- `pgr-services` runs with the OTEL javaagent attached and `OTEL_METRICS_EXPORTER: otlp` → `otel-collector:4317`. This matters: the code uses `GlobalOpenTelemetry`, which its own javadoc notes is "a no-op" without the agent.
- The collector's prometheus exporter serves `:8889`, already one of Prometheus's two scrape targets.
- The analytics commits (`70bd0d6a`, `277fed2b`, `b51b8bc6`) are all ancestors of `96dcf10`, the commit behind the pinned default image `egovio/pgr-services:2.12-beta-96dcf10` — so the code is in the image actually deployed, not just in the repo.
- Strongest evidence: `277fed2b` was written *from observing these series live on the bomet deploy* ("validated live on bomet: unit `ms` surfaced as `pgr_analytics_query_duration_ms_milliseconds_*`").

## Two things a dashboard author must know

**1. The names are exact — do not "fix" them.** `277fed2b` deliberately omits the OTLP unit field, because the collector's prometheus exporter appends a unit-derived suffix when unit is set. That is why the scraped names are clean (`pgr_analytics_query_duration_ms_*`, not `..._ms_milliseconds_*`). A unit test pins the name and empty unit for both instruments — if a future change re-adds `setUnit`, the series get renamed and the dashboard silently goes blank.

[...]

PR #1628, opened 2026-08-06
feat(redpanda): scrape broker metrics into Prometheus + dashboard

singhalkarun:
Partially addresses #1623 — **the consumer-lag panel is deliberately not included; see below.**

## What

The broker was invisible. Prometheus scraped only `otel-collector` and `node-exporter`; Gatus checked only that TCP 9092 accepts a connection. That check stays green in the failure that matters most: **consumer lag**, where `egov-persister` falls behind, the API keeps returning 200, Gatus stays green, and complaints are simply never written to the database.

## Change

**Scrape job** — the broker is **Redpanda**, not Apache Kafka, and it serves Prometheus metrics natively from its admin API. No `kafka_exporter`, no new container — the same shape as the bundled Kong plugin in #1627:

```yaml
  - job_name: redpanda
    metrics_path: /public_metrics
    static_configs:
      - targets: ['redpanda:9644']
```

`/public_metrics` rather than `/metrics`: the internal endpoint carries far higher cardinality, which is a real cost on a memory-tight box (#1612).

**Dashboard** — Redpanda's own "Default" board (grafana.com 18134, 19 panels), vendored like `node-exporter-full.json` with `__inputs`/`__requires` stripped and every datasource pinned to uid `prometheus`, plus a stable `uid: redpanda-broker`.

## A deliberate non-change

The compose command does **not** name the admin API, so this relies on Redpanda's default `0.0.0.0:9644` listener. I chose not to add an explicit `--admin-addr` flag:

- if the assumption is right, nothing needs adding;
- if it is wrong, the scrape target simply reports **down** and the broker is untouched;
- whereas a mistaken flag on `redpanda start` would fail the broker and take the whole stack with it.

Fail-safe in the direction that matters. The verification command is recorded in the config comment:

```
docker exec redpanda curl -s localhost:9644/public_metrics | head
```

## What this does NOT deliver

**No consumer-lag panel**, which is the headline reason #1623 exists. Being explicit rather than letting it look done:

- Neither official Redpanda dashboard (18134 or 18135) contains one — verified by extracting every metric each queries.
- Redpanda does not export lag directly. It must be derived as `redpanda_kafka_max_offset` − `redpanda_kafka_consumer_group_committed_offset`, joined on topic and partition.
- Both series are confirmed present (the official dashboard queries them), but it aggregates by `cluster`/`instance`/`shard`, so it never reveals the **group/topic/partition label names** — and those are what a lag query joins on.

Guessing label names would ship panels that render blank while looking complete. That is precisely the failure mode called out in #1616, so I have not done it here. #1623 stays open with the recipe recorded; the remaining step is one `curl` against a live box to read the labels, then one panel.

This PR still delivers real value on its own: throughput, partition counts, unavailable partitions, latency, memory, and consumer-group membership all become visible.

## Verification

- [x] `prometheus.yml` parses; three jobs with correct `metrics_path`
- [x] Dashboard is valid JSON, 19 panels, no `${DS_` or `__inputs`
- [ ] `docker exec redpanda curl -s localhost:9644/public_metrics | head` returns metrics (confirms the default listener)
- [ ] `redpanda_*` series appear in Prometheus and the dashboard populates
- [ ] Broker overhead unchanged — it runs on a tight `--smp 1 --memory 256M`

## Merge note

Touches `otel/prometheus.yml`, as does #1627 (Kong). Both append a job at the end of the file, so expect a trivial conflict depending on merge order — keep both jobs.

### DIFF
--- a/local-setup/otel/grafana/provisioning/dashboards/kafka-consumer-lag.json
+++ b/local-setup/otel/grafana/provisioning/dashboards/kafka-consumer-lag.json
@@ -47,7 +47,7 @@
     {
       "id": 1,
       "title": "Total lag by consumer group",
-      "description": "Messages produced but not yet committed by each group. This is the number that matters: when it climbs and stays up, records are being accepted and never processed.\n\nRedpanda does not export lag directly -- it is derived as max_offset - committed_offset, joined on topic and partition. The join uses group_right() because committed_offset carries redpanda_group and max_offset does not.",
+      "description": "Messages produced but not yet committed by each group. This is the number that matters: when it climbs and stays up, records are being accepted and never processed.\n\nRedpanda does not export lag directly -- it is derived as max_offset - committed_offset, joined on topic and partition. The join uses group_right() because committed_offset carries redpanda_group and max_offset does not.\n\nNegative values are possible and are diagnostic, not noise: they are left un-clamped on purpose. Both series come from the SAME Prometheus target and scrape (job \"redpanda\", redpanda:9644/public_metrics), so they are mutually consistent within a scrape and cannot skew against each other. A committed offset AHEAD of the partition high-water mark therefore means something real -- an offset reset, a topic truncation/recreate, or a series that has gone stale on one side. clamp_min(..., 0) would render exactly that condition as 0, i.e. as \"caught up\", which is the misreading it would be meant to prevent.\n\nmax_offset is collapsed with max by (namespace, topic, partition) before the join. On this single-broker, --smp 1 stack it is a no-op (one series per partition), but on a multi-broker cluster the same partition is exported by every replica and the on(topic, partition) join would fail with \"multiple matches for labels: many-to-one matching must be explicit\". max, not sum: the high-water mark of a partition is not the sum of its replicas' views of it.",
       "type": "timeseries",
       "datasource": {
         "type": "prometheus",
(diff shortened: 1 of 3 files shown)


vinothrallapalli-eGov (commented):
Checked the scrape config and dashboard for correctness. `redpanda:9644` is the right target: the `redpanda` service in docker-compose.yml sits on `egov-network` alongside `otel-collector`/`node-exporter`, and 9644 is Redpanda's default admin API port (not published to the host, but that's fine since Prometheus reaches it over the internal Docker network, matching the existing `node-exporter:9100` pattern). `/public_metrics` is the correct curated endpoint for this use case. Spot-checked ~45 panel `expr` queries in redpanda-broker.json against known Redpanda metric names (redpanda_cluster_brokers, redpanda_kafka_max_offset, redpanda_kafka_partitions, redpanda_kafka_request_latency_seconds_bucket, redpanda_cluster_unavailable_partitions, etc.) — all are real, and the `$node`/`$aggr_criteria` template variables resolve against standard `instance`/`shard` labels, consistent with this being the vendored official Redpanda dashboard. No `__inputs`/`${DS_...}` leftovers; datasource uid and dashboard `uid: redpanda-broker` are both pinned as claimed. The PR's stated scope (no consumer-lag panel) matches the diff — there is no lag query or lag-labeled panel anywhere in the JSON, and the PR body correctly explains why (label names for the join aren't derivable from the two official dashboards alone). Verified job_name collision risk across the three sibling PRs (#1627 `kong`, #1628 `redpanda`, #1629 `postgres`): all three append to the same anchor line in `otel/prometheus.yml` so they'll produce a trivial merge conflict against each other, but the `job_name` values themselves are distinct — no actual config collision. No correctness issues found.

vinothrallapalli-eGov (approved):

PR #1629, opened 2026-08-06
feat(postgres): export database metrics to Prometheus + dashboard

### LINKED ISSUE #1615: Postgres has no metrics — pool saturation and slow queries are invisible
Part of #1601 (monitoring stack umbrella). **P1 · effort: S**

## Problem

Postgres sits under every complaint, user lookup and workflow transition. What we currently know about it is **one bit**: whether a TCP connect succeeds (the Gatus check). That stays green right up to the point things are badly wrong.

Invisible today:

- **Connection-pool saturation.** PgBouncer exists precisely because connections are scarce. When the pool fills, requests queue and *every* service slows at once — which presents as a mysterious platform-wide slowdown rather than a database problem.
- **Slow queries, table/index bloat, autovacuum falling behind** — including the "fine at 10k complaints, degrading at 100k" case.
- **Database-internal disk and transaction pressure**, distinct from the host-level disk metrics node-exporter provides.

"The site is slow" is among the most common reports, and database saturation among the most common causes. Confirming or ruling it out currently means SSHing to the box and running queries by hand, during an incident.

## Fix

Add `postgres_exporter` — a small container that runs statistics queries against Postgres and publishes them for Prometheus — plus the standard community dashboard.

**Placement: `local-setup/docker-compose.monitoring.yml`.** That overlay exists for exactly this purpose and currently holds node-exporter; its header documents the pattern (always layered by the Ansible deploy, reached by DNS on `egov-network`, no host port published, scrape job in `otel/prometheus.yml`). This follows the #1335 precedent rather than inventing a new one — including using the official Docker Hub image rather than the registry mirror.

```yaml
  postgres-exporter:
    image: prometheuscommunity/postgres-exporter:v0.15.0
    container_name: digit-postgres-exporter
    restart: unless-stopped
    environment:
      DATA_SOURCE_NAME: "postgresql://${POSTGRES_USER:-postgres}:${POSTGRES_PASSWORD}@postgres-db:5432/postgres?sslmode=disable"
    networks:
      - egov-network
```

Scrape job in `otel/prometheus.yml`:

```yaml
  - job_name: postgres
    static_configs:
      - targets: ['postgres-exporter:9187']
```

Dashboard: grafana.com ID 9628, vendored into `otel/grafana/provisioning/dashboards/` like `node-exporter-full.json`.

## Critical detail — point it at `postgres-db`, not `postgres`

In this stack `postgres` is a **network alias for PgBouncer**, and `postgres-db` is the actual database (`docker-compose.egov-digit.yaml:95`). The Gatus config already warns about this trap in a comment:

> `postgres` is a network alias for PgBouncer, not for Postgres itself. […] Both are checked separately so a dead database cannot hide behind a healthy pooler.

Pointing the exporter at the pooler would report PgBouncer's view rather than the database's, producing misleading connection statistics — precisely the metric we are adding this for.

Credentials come from the existing `POSTGRES_PASSWORD` in `.env` (written from OpenBao), so no new secret is introduced.

## Deliberately out of scope

**No kafka / redis / elasticsearch exporters.** Each is another always-on container on a memory-constrained box (#1612), and none answers a question we are currently unable to answer *and* actually ask. Postgres and Kong (#1614) earn their place because they answer the two most frequent incident questions. Add the others when a concrete incident demands one.

## Acceptance

- [ ] `pg_stat_activity_count` and `pg_stat_database_*` visible in Prometheus
- [ ] Dashboard shows connections per database, transaction rate, cache hit ratio, longest-running query
- [ ] Exporter connects to `postgres-db` (verify the target is the database, not PgBouncer)
- [ ] No credential appears in `docker inspect` output beyond what `.env` already exposes
- [ ] Container memory footprint stays small enough not to matter against #1612's budget

## Relates to

- #1614 — the other half of the "answer the common incident question" pair.
- #1611 — pool saturation is a strong future alert-rule candidate once this data exists.

singhalkarun:
Fixes #1615

## What

Postgres is under every complaint, user lookup and workflow transition, and the only thing we knew about it was whether a TCP connect succeeded. That check stays green while the **connection pool is saturated** and every service queues behind it — presenting as a mysterious platform-wide slowdown rather than a database problem. Confirming or ruling it out today means SSHing to the box and running queries by hand, mid-incident.

## Change

**Exporter** — `postgres_exporter` added to `docker-compose.monitoring.yml`, following the node-exporter conventions from #1335: official Docker Hub image, **no published host port** (Prometheus reaches it by DNS over `egov-network`), always layered by the deploy.

**Scrape job** — `postgres-exporter:9187`.

**Dashboard** — grafana.com 9628, 35 panels: connections vs `max_connections`, transaction and tuple rates, cache hit ratio, locks by mode, and per-database breakdowns.

## Three details worth reviewing

**1. It points at `postgres-db`, not `postgres`.** In this stack `postgres` is a network alias for **PgBouncer**; the database itself is `postgres-db`. The gatus config already warns about this trap in a comment ("a dead database cannot hide behind a healthy pooler"). Aiming the exporter at the pooler would report PgBouncer's view of connections rather than the database's — the wrong answer for the exact metric this exists to provide.

Credentials reuse `postgres-db`'s own `egov`/`egov` and the `${POSTGRES_PASSWORD}` the deploy already writes into `.env` from OpenBao, so **no new secret is introduced**.

**2. The dashboard needed de-Kubernetising, not just vendoring.** As published it targets the Helm chart:

- 8 of its 40 queries filter on `release="$release"` — a label only the Helm deployment sets
- it defines `namespace` and `release` template variables sourced from labels a compose scrape does not have
- its `instance` variable was derived from `pg_up{release="$release"}`, so it too would have come back empty

Downloaded as-is, a large share of the panels would have rendered blank. Fixed by stripping the `release` filter, dropping the two Helm-only variables, and repointing `instance` at `label_values(pg_up, instance)`. Verified afterwards that no `$release` or `$namespace` reference survives, alongside the usual `__inputs` / `${DS_PROMETHEUS}` treatment used for the other vendored boards.

**3. The Gatus coverage CI needed an exemption.** Adding any compose service without a health check fails that job. Confirmed by running it locally:

```
$ python3 .github/scripts/check-gatus-coverage.py
FAIL: compose services with no Gatus check:
  - postgres-exporter
```

Exempted with a reason mirroring node-exporter's: the **database** it reads is already covered by the existing PostgreSQL check, so this container going down costs metrics, not service. After the change:

```
OK: 89 compose services, 23 k3s Services, 51 endpoints per tier, no drift, no dangling checks.
self-test OK: all detectors fire on their failure mode
```

## Verification

- [x] Compose overlay and `prometheus.yml` parse; exporter publishes no host port
- [x] Dashboard valid JSON, 35 panels, no `$release` / `$namespace` / `${DS_` remaining
- [x] `check-gatus-coverage.py` passes, including its self-test
- [ ] `pg_stat_activity_count` and `pg_stat_database_*` visible in Prometheus
- [ ] Dashboard populates; `instance` and `datname` variables offer real values
- [ ] Exporter connects to the database, not PgBouncer (confirm from its logs)

The last three need a running box; not yet deployed.

## Merge note

Touches `otel/prometheus.yml`, as do #1627 (Kong) and #1628 (Redpanda). All three append a job at the end — expect a trivial conflict depending on merge order, and keep all jobs.

### DIFF
--- a/.github/scripts/check-gatus-coverage.py
+++ b/.github/scripts/check-gatus-coverage.py
@@ -138,6 +138,7 @@ EXEMPT = {
     "tempo": "observability plumbing: trace store, not a serving dependency",
     "otel-collector": "observability plumbing: telemetry pipeline, not a serving dependency",
     "node-exporter": "observability plumbing: host-metrics exporter (#1335), not a serving dependency; scraped by prometheus, absent from k3s tier and docker-compose.yml",
+    "postgres-exporter": "observability plumbing: database-metrics exporter (#1615), not a serving dependency; scraped by prometheus, absent from k3s tier and docker-compose.yml. The DATABASE it reads is already covered by the PostgreSQL check; this container going down costs metrics, not service.",
     # Deploy-time only: nothing declares depends_on openbao, and ansible reads its
     # secrets during the deploy and injects them as env, so a runtime outage does
     # not break serving. Listens on 127.0.0.1 only, so Gatus could not reach it.
--- a/local-setup/docker-compose.monitoring.yml
+++ b/local-setup/docker-compose.monitoring.yml
@@ -69,6 +69,18 @@ services:
       # Credentials match postgres-db's own POSTGRES_USER/DB (egov/egov), with
       # the password coming from the same ${POSTGRES_PASSWORD} the deploy
       # writes into .env from OpenBao. No new secret is introduced.
+      #
+      # Least privilege: `egov` is the stack's superuser, and postgres_exporter
+      # only needs pg_monitor. Reusing it is not a new grant — every backend
+      # service already connects as `egov` through PgBouncer — but a dedicated
+      # role would be tighter. It is deliberately not done here: the only place
+      # this stack creates roles is postgres-db's /docker-entrypoint-initdb.d
+      # (db/full-dump.sql), which Postgres runs ONLY on an empty data
+      # directory. Adding the role there would take effect on fresh volumes
+      # only, so every existing deploy would come up with an exporter that
+      # cannot authenticate — a blank dashboard, which is the failure mode this
+      # work exists to remove. Doing it properly needs a migration that runs
+      # against live databases; tracked as follow-up, not smuggled in here.
       DATA_SOURCE_NAME: "postgresql://egov:${POSTGRES_PASSWORD:-egov123}@postgres-db:5432/egov?sslmode=disable"
     networks:
       - egov-network
--- a/local-setup/docker-compose.monitoring.yml
+++ b/local-setup/docker-compose.monitoring.yml
@@ -45,3 +45,30 @@ services:
       - /:/rootfs:ro,rslave
     networks:
       - egov-network
+
+  # Database metrics (#1615). Postgres is under every complaint, user lookup
+  # and workflow transition, and the only thing we knew about it was whether a
+  # TCP connect succeeded — which stays green while the connection pool is
+  # saturated and every service queues behind it.
+  #
+  # Same conventions as node-exporter above: official Docker Hub image, no
+  # published host port (Prometheus reaches it by DNS over egov-network), and
+  # always layered by the Ansible deploy.
+  postgres-exporter:
+    image: prometheuscommunity/postgres-exporter:v0.15.0
+    container_name: digit-postgres-exporter
+    restart: unless-stopped
+    environment:
+      # IMPORTANT: postgres-db, NOT postgres.
+      # `postgres` is a network alias for PgBouncer in this stack; the database
+      # itself is postgres-db (the gatus config carries the same warning: "a
+      # dead database cannot hide behind a healthy pooler"). Pointing this at
+      # the pooler would report PgBouncer's view of connections rather than the
+      # database's — misleading on the exact metric this exists for.
+      #
+      # Credentials match postgres-db's own POSTGRES_USER/DB (egov/egov), with
+      # the password coming from the same ${POSTGRES_PASSWORD} the deploy
+      # writes into .env from OpenBao. No new secret is introduced.
+      DATA_SOURCE_NAME: "postgresql://egov:${POSTGRES_PASSWORD:-egov123}@postgres-db:5432/egov?sslmode=disable"
+    networks:
+      - egov-network
--- a/local-setup/otel/prometheus.yml
+++ b/local-setup/otel/prometheus.yml
@@ -25,6 +25,11 @@ scrape_configs:
   # is PgBouncer — see the comment on the service definition. Connection-pool
   # saturation is the metric this exists for, and the pooler's own view of it
   # would be the wrong answer.
-  - job_name: postgres
+  #
+  # The job is named after the exporter, not `postgres`, for the same reason:
+  # `postgres` is the pooler's alias in this stack, so a job by that name would
+  # re-introduce exactly the ambiguity the comment above warns about for anyone
+  # filtering by job= in Prometheus or Grafana later.
+  - job_name: postgres-exporter
     static_configs:
       - targets: ['postgres-exporter:9187']
--- a/local-setup/otel/prometheus.yml
+++ b/local-setup/otel/prometheus.yml
@@ -16,3 +16,15 @@ scrape_configs:
   - job_name: node
     static_configs:
       - targets: ['node-exporter:9100']
+
+  # Database metrics (#1615) from postgres_exporter, added by the same
+  # docker-compose.monitoring.yml overlay as node-exporter. Feeds the
+  # "PostgreSQL Database" Grafana dashboard.
+  #
+  # It connects to postgres-db (the database), NOT the `postgres` alias, which
+  # is PgBouncer — see the comment on the service definition. Connection-pool
+  # saturation is the metric this exists for, and the pooler's own view of it
+  # would be the wrong answer.
+  - job_name: postgres
+    static_configs:
+      - targets: ['postgres-exporter:9187']
--- a/local-setup/otel/grafana/provisioning/dashboards/postgres.json
+++ b/local-setup/otel/grafana/provisioning/dashboards/postgres.json
@@ -1,0 +1,3181 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "description": "This dashboard works with postgres_exporter for prometheus",
+  "editable": true,
+  "gnetId": 9628,
+  "graphTooltip": 0,
+  "iteration": 1568751813814,
+  "links": [],
+  "panels": [
+    {
+      "collapsed": false,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 0
+      },
+      "id": 34,
+      "panels": [],
+      "title": "General Counters, CPU, Memory and File Descriptor Stats",
+      "type": "row"
+    },
+    {
+      "cacheTimeout": null,
+      "colorBackground": false,
+      "colorValue": true,
+      "colors": [
+        "#299c46",
+        "#7eb26d",
+        "#d44a3a"
+      ],
+      "datasource": {
+        "type": "prometheus",
+        "uid": "prometheus"
+      },
+      "format": "none",
+      "gauge": {
+        "maxValue": 100,
+        "minValue": 0,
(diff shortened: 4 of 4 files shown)


local-setup/otel/prometheus.yml
```diff
@@ -16,3 +16,15 @@ scrape_configs:
...
+
+  # Database metrics (#1615) from postgres_exporter, added by the same
+  # docker-compose.monitoring.yml overlay as node-exporter. Feeds the
+  # "PostgreSQL Database" Grafana dashboard.
+  #
+  # It connects to postgres-db (the database), NOT the `postgres` alias, which
+  # is PgBouncer — see the comment on the service definition. Connection-pool
+  # saturation is the metric this exists for, and the pooler's own view of it
+  # would be the wrong answer.
+  - job_name: postgres
```
vinothrallapalli-eGov: Minor naming nit: this job is named `postgres`, which is exactly the string this PR (and the gatus config) goes out of its way to warn means "PgBouncer, not the database" elsewhere in this stack. Someone filtering Prometheus/Grafana by `job="postgres"` later could reasonably assume they're looking at the pooler. Consider `job_name: postgres-exporter` (matching the target/service name) to avoid re-introducing the ambiguity this PR is otherwise careful to eliminate. Not a functional bug — the target itself (`postgres-exporter:9187`, which reads from `postgres-db`) is correct.

.github/scripts/check-gatus-coverage.py
```diff
@@ -138,6 +138,7 @@
     "tempo": "observability plumbing: trace store, not a serving dependency",
     "otel-collector": "observability plumbing: telemetry pipeline, not a serving dependency",
     "node-exporter": "observability plumbing: host-metrics exporter (#1335), not a serving dependency; scraped by prometheus, absent from k3s tier and docker-compose.yml",
+    "postgres-exporter": "observability plumbing: database-metrics exporter (#1615), not a serving dependency; scraped by prometheus, absent from k3s tier and docker-compose.yml. The DATABASE it reads is already covered by the PostgreSQL check; this container going down costs metrics, not service.",
```
vinothrallapalli-eGov, line 141: This exemption is added correctly and independently in this PR (good — it doesn't depend on #1636 to land). Flagging the interaction called out in the tracking issue for visibility: if #1636 rewrites this `EXEMPT` dict wholesale (rather than appending like this PR does), whichever of the two merges second could silently drop this `postgres-exporter` line depending on merge/rebase mechanics, since a dict rewrite doesn't produce a clean 3-way merge the way a single-line insertion does. Worth re-running `check-gatus-coverage.py` after both PRs are merged to confirm this entry survived.

local-setup/docker-compose.monitoring.yml
```diff
@@ -45,3 +45,30 @@ services:
...
+      # `postgres` is a network alias for PgBouncer in this stack; the database
+      # itself is postgres-db (the gatus config carries the same warning: "a
+      # dead database cannot hide behind a healthy pooler"). Pointing this at
+      # the pooler would report PgBouncer's view of connections rather than the
+      # database's — misleading on the exact metric this exists for.
+      #
+      # Credentials match postgres-db's own POSTGRES_USER/DB (egov/egov), with
+      # the password coming from the same ${POSTGRES_PASSWORD} the deploy
+      # writes into .env from OpenBao. No new secret is introduced.
+      DATA_SOURCE_NAME: "postgresql://egov:${POSTGRES_PASSWORD:-egov123}@postgres-db:5432/egov?sslmode=disable"
```
vinothrallapalli-eGov, line 84: Credentials check: this reuses the `egov` superuser (same one every backend service already connects with via PgBouncer), so it's not a new privilege escalation for this stack, and the password sourcing matches the existing `${POSTGRES_PASSWORD:-egov123}` fallback pattern used by postgres-db/pgbouncer. That said, postgres_exporter only needs `pg_monitor` (or a subset of SELECT grants on the stats views), so giving it the shared superuser is broader than the exporter itself requires. Not a regression given the existing convention, but worth a scoped monitoring role if this pattern gets revisited.

vinothrallapalli-eGov (commented):
Verified the two flagged risks: (1) the exporter's DATA_SOURCE_NAME correctly targets `postgres-db:5432` (the real database), not the `postgres` alias (PgBouncer) — matches the pattern the gatus config already documents. (2) This PR adds its own `postgres-exporter` EXEMPT entry to check-gatus-coverage.py as a one-line insertion (not waiting on #1636) — but see the inline note on merge-order risk with #1636 if that PR rewrites the same dict wholesale rather than appending. Dashboard is clean: no leftover `$release`/`$namespace`/`${DS_` tokens, 35 panels, `instance` variable correctly repointed at `label_values(pg_up, instance)`, and every `pg_*` metric referenced (pg_stat_activity_count, pg_locks_count, pg_stat_database_*, pg_stat_bgwriter_*, pg_settings_*) is a real default postgres_exporter collector metric — no custom queries.yaml needed. Two minor nits left inline: the scrape job is named `postgres` (same name this PR spends effort disambiguating from PgBouncer), and the exporter reuses the shared `egov` superuser rather than a scoped `pg_monitor` role (consistent with existing repo convention, so not a regression, just worth a future look).

local-setup/otel/prometheus.yml
singhalkarun: Agreed — renamed to `job_name: postgres-exporter` in 6b12c0a8, with the reasoning written into the comment block next to the existing "postgres is PgBouncer" warning so the two are read together.

Checked before renaming that nothing selects on the job label: `postgres.json` contains no `job=` matcher at all (all 30-odd panel queries are unqualified `pg_*` selectors), and there are no recording/alerting rules in `local-setup/otel/`. So the rename is inert today and only affects what a human sees in the target list.

Validated: `promtool check config` passes, and the three jobs are `otel-collector`, `node`, `postgres-exporter` with no duplicates.

.github/scripts/check-gatus-coverage.py
singhalkarun: Good flag, and I tested the exact scenario rather than reasoning about it. #1636 (`feat/1613-monitor-the-monitoring`) does rewrite the block wholesale — it deletes the six-line comment plus the `grafana`/`prometheus`/`loki`/`tempo`/`otel-collector` entries and rewrites the `node-exporter` line, i.e. the line this PR inserts immediately after.

The good news is that it fails loud, twice:

**1. Git refuses to auto-merge.** Test-merged both branches off their common base:

```
$ git merge-tree --write-tree --merge-base=origin/monitoring-fix \
    origin/feat/1615-postgres-exporter origin/feat/1613-monitor-the-monitoring
CONFLICT (content): Merge conflict in .github/scripts/check-gatus-coverage.py
```

Both sides touch overlapping lines in the same hunk, so there is no clean 3-way merge to silently take — whoever merges second must resolve by hand and will see the `postgres-exporter` line sitting inside the conflict markers.

**2. CI catches it even if the resolver drops the line.** Deleted the entry and re-ran the script:

```
FAIL: compose services with no Gatus check (add one to local-setup/gatus/config.yaml, or add to EXEMPT with a reason):
  - postgres-exporter
exit=1
```

That is the whole point of the checker, and it is wired into CI, so a dropped exemption cannot reach the default branch quietly. Re-running it after both merges is still the right habit, but it is a belt-and-braces check rather than the only line of defence. Current state on this branch: `OK: 89 compose services, 23 k3s Services, 51 endpoints per tier, no drift, no dangling checks.`

local-setup/docker-compose.monitoring.yml
singhalkarun: Right on both counts, and I have written the reasoning into the compose comment (6b12c0a8) rather than making the change — because the change is not safely makeable here, and it is worth recording why.

`pg_monitor` is the correct grant. The blocker is *where* a role could be created. The only place this stack creates database objects at provisioning time is `postgres-db`'s `/docker-entrypoint-initdb.d/01-full-dump.sql` (`docker-compose.yml:21`), and Postgres runs `docker-entrypoint-initdb.d` **only when the data directory is empty**. `postgres_data` is a named volume that survives redeploys, so a role added there would take effect on fresh volumes only. Every existing deploy would come up with an exporter that cannot authenticate — a dashboard that renders blank while looking installed, which is precisely the failure mode this whole initiative exists to eliminate (#1616).

Doing it properly means a migration that runs against live databases (`CREATE ROLE postgres_exporter ... GRANT pg_monitor`, plus a secret for it in OpenBao and the `.env` plumbing), which is a real change with its own review surface and does not belong smuggled into a dashboard PR.

So the position is the one you described: not a new grant — `egov` is what every backend service already connects as through PgBouncer, and `${POSTGRES_PASSWORD:-egov123}` is the same sourcing as `postgres-db` and `pgbouncer` — but broader than the exporter needs, and now flagged in-tree as follow-up rather than left implicit.

vinothrallapalli-eGov (approved):

Issue #1622, opened 2026-08-06
Nothing monitors TLS certificate expiry — Gatus has no external HTTPS check


singhalkarun:
Part of #1601 (monitoring stack umbrella). **P1 · effort: S**

## Problem

TLS certificates are Let's Encrypt, issued and renewed by Certbot's systemd timer (`local-setup/ansible/README.md:436-444`). Renewal is automatic — and unverified. **Nothing anywhere checks certificate validity.**

If renewal fails — expired ACME account, a firewall change blocking the HTTP-01 challenge, nginx not reloading after renewal, a DNS change, a disabled timer — the certificate expires silently. At that moment **every citizen and every employee gets a browser security warning**, and the platform is effectively down for everyone.

It is also the most predictable outage possible: the expiry date is known 90 days ahead, and Certbot renews at 30 days. Any check would give us weeks of warning.

## Why the existing monitoring cannot catch it

Gatus is entirely **inside** the container network. Every one of its ~51 endpoints is `http://` or `tcp://` against a service name on `egov-network` (verified: 41 `http`, 10 `tcp`, **zero** external HTTPS checks). Nothing traverses nginx, so nothing presents the public certificate.

Consequence: with an expired certificate, **Gatus stays entirely green** while the public site is unusable. The dashboard would actively mislead the person diagnosing it. The Ansible validation tasks have the same blind spot — they use `127.0.0.1` with a `Host:` header.

This is the sharpest example of a general gap: our monitoring watches the platform from the inside, and the one failure that is 100% user-visible is invisible to us.

## Fix — options

**Option A (recommended): a Gatus check against the public URL.**

Gatus supports certificate-expiry conditions natively, so this needs no new component:

```yaml
  - name: TLS certificate
    group: Public endpoint
    url: "https://{{ domain }}/"
    interval: 5m
    conditions:
      - "[STATUS] < 500"
      - "[CERTIFICATE_EXPIRATION] > 240h"   # ~10 days' warning
```

Two things make this more than a certificate check: it is the **only** check that exercises the real public path — DNS, the host firewall, nginx, TLS termination, and routing through to the app. Every existing check bypasses all of that.

Caveats to handle: it must be templated per tenant (`domain`), gated for deployments where `tls_enabled: false`, and Gatus must be able to resolve and reach the public hostname from inside the container (hairpin NAT is not always available — if that fails, fall back to Option B).

[...]

Issue #1618, opened 2026-08-05
Port the full monitoring stack to the k3s tier (currently Gatus-only)


singhalkarun:
Part of #1601 (monitoring stack umbrella). **P2 · effort: L — phased**

**Decision taken:** the k3s tier should run the full monitoring stack, not stay Gatus-only. This issue tracks the port.

## Current state

`local-setup/k8s/` contains `app-services/`, `config/`, `core-services/`, `infrastructure/`, `tools/`. The **only** observability manifest is `tools/gatus.yaml`. There is no Grafana, Prometheus, Loki, Promtail, Tempo, otel-collector or node-exporter.

The mismatch is confusing today because the Gatus config *is* deliberately kept byte-identical across tiers with CI enforcing it — signalling "these tiers should match" — while all other observability silently does not. `k8s/tools/gatus.yaml` half-acknowledges this, noting other stacks "have no manifests under local-setup/k8s/ yet" and that tier drift "is now merely *detected* … rather than prevented".

## What the port actually involves

Not a straight copy. Verified constraints:

### 1. Promtail cannot reuse its config
`local-setup/otel/promtail-config.yaml:25-26` discovers containers via
```yaml
    docker_sd_configs:
      - host: unix:///var/run/docker.sock
```
k3s runs **containerd**, not Docker — there is no docker socket. The k8s version needs a DaemonSet using `kubernetes_sd_configs` reading `/var/log/pods`, with relabeling to reproduce the existing labels (`compose_service` → its pod/container equivalents). The Loki-facing half of the config carries over; the discovery half is a rewrite.

### 2. No service currently emits telemetry in this tier
`grep OTEL_ local-setup/k8s/{core,app}-services/*.yaml` returns **nothing**. Deploying a collector alone yields empty dashboards: each service also needs the javaagent mount plus `OTEL_SERVICE_NAME` / `OTEL_EXPORTER_OTLP_ENDPOINT` / exporter vars, mirroring the compose tier's per-service settings — including the deliberate exceptions (`egov-localization` runs metrics-only to avoid ~200 MB of agent heap; the OTP/SMS services are off entirely).

### 3. Persistence is new ground here
There are **no PersistentVolumeClaims anywhere** in `local-setup/k8s/` today. Prometheus, Loki, Tempo and Grafana all need durable storage. k3s ships the local-path provisioner, so this is straightforward, but it establishes a pattern the tier has not needed until now.

[...]

Issue #1609, opened 2026-08-05
Gatus detects outages but notifies nobody — wire alerting to Slack


singhalkarun:
Part of #1601 (monitoring stack umbrella). **P0 · effort: M**

## Problem

Gatus checks ~45 endpoints every 30s — infrastructure, all core services, the gateway, the UI — and renders them green/red. It never notifies anyone. A service can be red for days and the only way to find out is for a human to open the page.

This is not hypothetical. `.github/workflows/gatus-coverage.yml:6-9` records the incident that motivated the coverage CI, verbatim:

> egov-url-shortening sat DOWN for days and the whole mcp stack ran unwatched, because nothing fails when a service simply isn't on the dashboard.

That incident produced a fix for **coverage** — every service is now on the dashboard, and CI keeps it that way. **Notification was never added.** The same outage today would still go unnoticed.

There is no alerting anywhere else in the stack either: no Alertmanager, no Prometheus `rule_files`, no Grafana `provisioning/alerting/`. Gatus's own config carries only a commented-out placeholder (`local-setup/gatus/config.yaml:18-21`):

```yaml
# Alerting (optional - configure as needed)
# alerting:
#   slack:
#     webhook-url: "https://hooks.slack.com/services/xxx/yyy/zzz"
```

## Constraint that shapes the fix

`local-setup/gatus/config.yaml` is kept **byte-identical** in endpoint catalogue with the k3s tier's ConfigMap (`local-setup/k8s/tools/gatus.yaml`), and `.github/workflows/gatus-coverage.yml` enforces it. The file header says so explicitly: *"Add an endpoint to one file, add it to the other."*

Enabling alerts the normal way means adding an `alerts:` block to each of the ~45 endpoints, in **both** files, kept in lockstep forever. Worse, an endpoint declaring `- type: slack` when no `alerting.slack` provider is configured risks failing Gatus config validation on tenants that have no webhook — i.e. the change would break every deployment that hasn't opted in.

## Proposed fix — inject at deploy time

1. **Repo config stays as it is today.** No `alerts:` blocks, no provider, no drift between tiers, coverage CI unaffected.

2. **New script** `local-setup/scripts/gatus-inject-alerts.py` (~30 lines): reads `gatus/config.yaml`, appends to every endpoint

   ```yaml
   alerts:
     - type: slack
   ```

   and prepends the provider:

   ```yaml
   alerting:
     slack:
[...]

Issue #1601, opened 2026-08-05
[Master] Monitoring stack: hardening, alerting and coverage


singhalkarun:
Tracking issue for the monitoring/observability improvement plan. Every work item has its own issue; PRs reference the individual issue (`Fixes #N`) and this issue tracks overall progress.

**18 issues across five groups, below.**

Note on completeness: the first 16 came from auditing what is *deployed*. A second pass asking "what would we want to know?" found gaps that audit could not surface — #1622 is the first of those. Known remaining candidates, not yet filed: telemetry coverage (41 of 60 services emit nothing), the client-side dashboard metrics having no dashboard (the browser-side twin of #1616), and business/SLA metrics.

**Scope:** items 1–4 below cover the **Ansible/compose tier** under `local-setup/`, which is the **small-scale** deployment path.

**`devops/deploy-as-code` is the production path for large-scale deployments** — it is NOT legacy, and an earlier revision of this issue wrongly described it that way. Its monitoring is tracked in #1619, whose goal is parity with what the Ansible tier provides. `local-setup/k8s/` (the k3s tier, #1618) is a third path again.

---

## Where the stack stands today

| Concern | Status |
|---|---|
| Metrics store (Prometheus, 15d) | present — only 2 scrape targets |
| Host metrics (node-exporter + dashboard) | done — #1335 |
| JVM metrics (OTEL javaagent → collector → Prometheus) | present on 15 of ~61 services |
| Logs (Promtail → Loki, 72h) | present |
| Traces (javaagent + Kong otel plugin → Tempo, 24h @ 100% sampling) | present |
| Docker log rotation (daemon.json, 100m slice / 1g cap) | already handled in Ansible — nothing to do |
| Uptime checks (Gatus, ~45 endpoints, CI-enforced coverage) | strong |
| **Alerting — anywhere** | **none** |
| Gateway + datastore metrics | none |
| Monitoring auth and exposure | open — #1602–#1604, #1608 |
| k3s-tier observability beyond Gatus | none — **decided: port it** (#1618) |

---

## 1 · Security lockdown

- [ ] #1602 — Grafana deploys with no authentication (anonymous Admin) · P0 · S — **PR #1605**
- [ ] #1603 — All monitoring ports bind 0.0.0.0, bypassing nginx/Kong · P0 · S — **PR #1606**
[...]

PR #1630, opened 2026-08-06
feat(gatus): persist uptime history to sqlite

### LINKED ISSUE #1610: Gatus loses all uptime history on every restart (storage: memory)
Part of #1601 (monitoring stack umbrella). **P1 · effort: XS**

## Problem

Gatus stores results in memory only — `local-setup/gatus/config.yaml:11-12` and the same block in the k3s ConfigMap (`local-setup/k8s/tools/gatus.yaml:31-32`):

```yaml
storage:
  type: memory
```

Every container restart — including every redeploy — discards all history. The dashboard can say what is happening now, but nothing about what happened before. Questions that go unanswerable:

- how long was that service actually down last night?
- is this endpoint flapping, or is this the first failure?
- did the fix we shipped change anything?

The compose service has no volume either (only the read-only config bind mount), so there is nowhere for history to live even if storage were enabled.

## Why it matters more after #1609

Once Slack alerts land, the first question on any alert is "how long, and has it done this before?" Today the dashboard cannot answer either. Persistence is what turns an alert into something diagnosable.

## Fix

**Compose** (`docker-compose.egov-digit.yaml`):

```yaml
# gatus service
    volumes:
      - ./gatus/config.yaml:/config/config.yaml:ro
      - gatus_data:/data          # new

# top-level volumes
  gatus_data:                     # new
```

**Config** (both tiers, kept in lockstep):

```yaml
storage:
  type: sqlite
  path: /data/gatus.db
```

**k3s** (`k8s/tools/gatus.yaml`): same storage block, plus an `emptyDir` mounted at `/data`. That survives container restarts within a pod, which is the meaningful win; a PVC is only worth it once that tier runs something real (tracked by the k3s scope decision in #1601).

## Coordination

This edits the config file that is byte-synced between tiers and enforced by `.github/workflows/gatus-coverage.yml`. Both files must change together. If #1609 is in flight, land them in a deliberate order to avoid a conflict in the same file — either PR first is fine, they touch different parts of it.

## Acceptance

- [ ] `docker compose restart gatus` preserves endpoint history in `/status/`
- [ ] A full redeploy preserves history
- [ ] `gatus-coverage` CI still passes (both tier files updated together)
- [ ] The sqlite file lands in the volume, not the container filesystem (`docker exec digit-gatus ls -la /data`)
- [ ] Uptime percentages and response-time graphs populate over time rather than resetting

## Notes

Not a security item — this is about being able to answer questions during an incident rather than only during it.

singhalkarun:
Fixes #1610

## What

Gatus stored results in memory only, so every container restart — including every redeploy — discarded all history. The dashboard could report what was happening at that moment and nothing about what came before:

- how long was that service actually down last night?
- is this endpoint flapping, or is this the first failure?
- did the fix we shipped change anything?

This matters most as soon as alerting lands (#1609). The first question on any alert is "how long, and has it done this before?" — and today the dashboard cannot answer either.

## Change

Both tiers move to sqlite, kept byte-identical as the tier-sync CI requires:

```yaml
storage:
  type: sqlite
  path: /data/gatus.db
```

and each gets a **writable volume**, because the existing config mount is read-only and the database needs somewhere of its own:

- **compose** — named volume `gatus_data:/data`, declared alongside the other `*_data` volumes
- **k3s** — an `emptyDir` mounted at `/data`

## Why emptyDir rather than a PVC on k3s

An `emptyDir` is tied to the pod: history survives container restarts but not a reschedule. That is the right trade while that tier runs nothing real, and it avoids introducing the first PersistentVolumeClaim in a tier that currently has none. #1618 revisits storage there properly.

## Verification

- [x] Both tier configs parse; storage blocks are identical between them
- [x] Compose: `gatus_data` mounted at `/data` and declared as a top-level volume
- [x] k3s: `gatus-data` volume and mount present on the Deployment
- [x] `check-gatus-coverage.py` passes, self-test included — no tier drift introduced
- [ ] `docker compose restart gatus` preserves endpoint history in `/status/`
- [ ] A full redeploy preserves history
- [ ] `docker exec digit-gatus ls -la /data` shows the db in the volume, not the container filesystem
- [ ] Uptime percentages and response-time graphs accumulate rather than resetting

The last four need a running box; not yet deployed.

## Merge note

Touches `local-setup/gatus/config.yaml`, which is byte-synced with `k8s/tools/gatus.yaml` and CI-enforced. #1609 (Slack alerting) and #1622 (TLS expiry check) also edit that pair — sequence them, and keep both tiers in step in whichever order they land.

### DIFF
--- a/local-setup/docker-compose.egov-digit.yaml
+++ b/local-setup/docker-compose.egov-digit.yaml
@@ -1644,6 +1644,9 @@ services:
         condition: service_healthy
     volumes:
       - ./gatus/config.yaml:/config/config.yaml:ro
+      # sqlite store (#1610) — uptime history survives redeploys. The config
+      # above is mounted read-only, so the database needs its own volume.
+      - gatus_data:/data
     ports:
       - "18889:8080"
     environment:
@@ -2531,3 +2534,4 @@ volumes:
   keycloak_postgres_data: null
   openbao_data: null
   novu_mongo_data: null
+  gatus_data: null
--- a/local-setup/gatus/config.yaml
+++ b/local-setup/gatus/config.yaml
@@ -12,6 +12,14 @@
 # every redeploy wiped it, so "how long was it down?" and "is this endpoint
 # flapping or is this the first failure?" were unanswerable — which is the
 # first question anyone asks when an alert fires.
+#
+# Retention is deliberately left unset: Gatus bounds it itself, per endpoint, at
+# maximum-number-of-results 100 and maximum-number-of-events 50 (upstream
+# storage/config.go, DefaultMaximumNumberOfResults / DefaultMaximumNumberOfEvents),
+# and the SQL store prunes past those thresholds. gatus.db therefore settles at a
+# steady-state size instead of growing forever. Set those keys here only if a
+# deployment wants a different window -- and check the pinned image supports them,
+# since Gatus rejects unknown config keys outright.
 storage:
   type: sqlite
   path: /data/gatus.db
--- a/local-setup/gatus/config.yaml
+++ b/local-setup/gatus/config.yaml
@@ -8,8 +8,13 @@
 # deployment actually runs. Add an endpoint to one file, add it to the other.
 # CI enforces this (.github/workflows/gatus-coverage.yml).
 
+# Uptime history survives restarts (#1610). With the previous `type: memory`
+# every redeploy wiped it, so "how long was it down?" and "is this endpoint
+# flapping or is this the first failure?" were unanswerable — which is the
+# first question anyone asks when an alert fires.
 storage:
-  type: memory
+  type: sqlite
+  path: /data/gatus.db
 
 ui:
   title: DIGIT Core Services - Health Dashboard
--- a/local-setup/k8s/tools/gatus.yaml
+++ b/local-setup/k8s/tools/gatus.yaml
@@ -501,6 +501,28 @@ data:
         conditions:
           - "[STATUS] == 200"
 ---
+# Backing store for the sqlite database (#1610). Without this the volume was an
+# emptyDir, which is destroyed with the pod -- so the uptime history this PR
+# exists to preserve was still lost on every rollout, drain or eviction.
+# Measured before the fix: after deleting the pod, the oldest row in the
+# database was timestamped one second after the new pod started.
+#
+# No storageClassName: k3s provisions this with local-path, the default. 1Gi is
+# generous -- Gatus stores one row per check per interval and prunes on read.
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: gatus-data
+  namespace: digit
+  labels:
+    app: gatus
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
+---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -510,6 +532,13 @@ metadata:
     app: gatus
 spec:
   replicas: 1
+  # Recreate, not the default RollingUpdate. The claim above is ReadWriteOnce
+  # and holds a sqlite file, so a rolling update would deadlock: the new pod
+  # cannot mount the volume until the old one releases it, and the old one is
+  # not terminated until the new one is ready. Recreate also keeps two Gatus
+  # processes off one sqlite file, which replicas: 1 assumes anyway.
+  strategy:
+    type: Recreate
   selector:
     matchLabels:
       app: gatus
@@ -581,7 +610,8 @@ spec:
           configMap:
             name: gatus-config
         - name: gatus-data
-          emptyDir: {}
+          persistentVolumeClaim:
+            claimName: gatus-data
 ---
 apiVersion: v1
 kind: Service
--- a/local-setup/k8s/tools/gatus.yaml
+++ b/local-setup/k8s/tools/gatus.yaml
@@ -32,6 +32,14 @@ data:
     # it is tied to the pod, so a restart keeps history while a reschedule does
     # not. That is the right trade while this tier runs nothing real — revisit
     # with a PVC when it does (#1618).
+    #
+    # Retention is deliberately left unset: Gatus bounds it itself, per endpoint, at
+    # maximum-number-of-results 100 and maximum-number-of-events 50 (upstream
+    # storage/config.go, DefaultMaximumNumberOfResults / DefaultMaximumNumberOfEvents),
+    # and the SQL store prunes past those thresholds. gatus.db therefore settles at a
+    # steady-state size instead of growing forever. Set those keys here only if a
+    # deployment wants a different window -- and check the pinned image supports them,
+    # since Gatus rejects unknown config keys outright.
     storage:
       type: sqlite
       path: /data/gatus.db
--- a/local-setup/k8s/tools/gatus.yaml
+++ b/local-setup/k8s/tools/gatus.yaml
@@ -28,8 +28,13 @@ metadata:
   namespace: digit
 data:
   config.yaml: |
+    # Uptime history survives container restarts (#1610). emptyDir, not a PVC:
+    # it is tied to the pod, so a restart keeps history while a reschedule does
+    # not. That is the right trade while this tier runs nothing real — revisit
+    # with a PVC when it does (#1618).
     storage:
-      type: memory
+      type: sqlite
+      path: /data/gatus.db
 
     ui:
       title: DIGIT Core Services - Health Dashboard
@@ -559,10 +564,16 @@ spec:
             - name: gatus-config
               mountPath: /config/config.yaml
               subPath: config.yaml
+            # sqlite store (#1610). The config mount is a read-only ConfigMap,
+            # so the database needs a writable volume of its own.
+            - name: gatus-data
+              mountPath: /data
       volumes:
         - name: gatus-config
           configMap:
             name: gatus-config
+        - name: gatus-data
+          emptyDir: {}
 ---
 apiVersion: v1
 kind: Service


local-setup/k8s/tools/gatus.yaml
```diff
@@ -559,10 +564,16 @@ spec:
...
+            # sqlite store (#1610). The config mount is a read-only ConfigMap,
+            # so the database needs a writable volume of its own.
+            - name: gatus-data
+              mountPath: /data
       volumes:
         - name: gatus-config
           configMap:
             name: gatus-config
+        - name: gatus-data
+          emptyDir: {}
```
vinothrallapalli-eGov, line 584: Informational, not blocking: `emptyDir` is scoped to the pod, not just the container, so this only survives an in-place container restart (crash/OOM-kill). A real redeploy on this tier — image bump, `kubectl rollout restart`, config change, node eviction/reschedule — creates a new pod and a fresh empty volume, wiping history exactly like `type: memory` did. The PR body and the comment above are upfront about this trade-off (PVC deferred to #1618), which is reasonable, but it means the original bug report ("every container restart — including every redeploy — discarded all history") is only fully fixed for the compose tier. Worth confirming the cited devbox verification (1650→22, 96→123) was measured against compose and not this tier, since a k3s redeploy would still show a reset.

local-setup/gatus/config.yaml
```diff
@@ -8,8 +8,13 @@
...
 # CI enforces this (.github/workflows/gatus-coverage.yml).
 
+# Uptime history survives restarts (#1610). With the previous `type: memory`
+# every redeploy wiped it, so "how long was it down?" and "is this endpoint
+# flapping or is this the first failure?" were unanswerable — which is the
+# first question anyone asks when an alert fires.
 storage:
-  type: memory
+  type: sqlite
+  path: /data/gatus.db
```
vinothrallapalli-eGov, line 25: Nit: no pruning/retention setting is configured here, so it's worth confirming this relies on Gatus's built-in default (`maximum-number-of-results: 200` per endpoint) rather than genuinely unbounded growth. That default does bound `gatus.db` size automatically, so this isn't a real risk, but making it explicit in the config (even just as a comment) would save the next person from having to check upstream defaults to be sure the file won't grow forever.

local-setup/docker-compose.egov-digit.yaml
```diff
@@ -1644,6 +1644,9 @@ services:
         condition: service_healthy
     volumes:
       - ./gatus/config.yaml:/config/config.yaml:ro
+      # sqlite store (#1610) — uptime history survives redeploys. The config
+      # above is mounted read-only, so the database needs its own volume.
+      - gatus_data:/data
```
vinothrallapalli-eGov, line 1649: Low risk, flagging for the deploy checklist: the `gatus` service doesn't set a `user:`, so whatever the `egovio/gatus:latest` image's default user is will own files written into this fresh named volume. Upstream `TwiN/gatus` has historically run as root by default (there's a long-open upstream issue requesting non-root support), so this is likely fine, but it's exactly the kind of thing that only shows up at runtime — matches the PR's own unchecked checklist item (`docker exec digit-gatus ls -la /data`). Worth a quick look once this lands on a box, just to confirm no permission-denied errors on writes to `/data/gatus.db`.

vinothrallapalli-eGov (commented):
Solid, minimal fix. The compose tier gets genuine persistence: `gatus_data` is a real top-level named volume, so history survives container restarts, recreates, and `down`/`up` — that fully matches the PR's own claim and the devbox restart-count numbers cited for verification, as long as those numbers were measured against this tier. The k3s tier only gets a partial fix (see inline note on the `emptyDir`). Checked for schema/merge collision with #1632 (TLS check) and #1636 (monitor-the-monitoring): both PRs diff cleanly against the same base and only append new endpoint blocks well below the `storage:` block this PR touches, so there's no line-level overlap or config-schema break — just the usual tier-sync coordination the author already flagged. No blocking issues found; two informational nits and one thing to confirm on deploy.

local-setup/k8s/tools/gatus.yaml
singhalkarun: Confirmed on both halves, and no change needed.

**The `emptyDir` reading is correct**, and it is the documented intent rather than an oversight. The ConfigMap comment this PR adds says it in the same terms — "it is tied to the pod, so a restart keeps history while a reschedule does not" — and the PR body's "Why emptyDir rather than a PVC on k3s" section makes the same trade explicit, with #1618 owning the PVC. So: compose gets real persistence, k3s gets restart-durability only, and the original bug report is fully fixed on one tier of the two.

**On the devbox numbers — they were compose, and could not have been anything else.** The integration pass on #1601 records every k8s-tier PR in that run as `⚠️ static only | no cluster to deploy on` (#1659, #1662, #1663, #1664, #1679). The box in that run is a 44-container Docker stack; there is no k3s cluster anywhere in it. So `restart 1650 → 22 before, 96 → 123 after` substantiates the compose tier and nothing else — exactly the caveat you asked to have confirmed. A k3s redeploy would still show a reset.

Resolving: the claim is accurate, the trade-off is already written down in both the code and the PR body, and the verification scope is now on the record here.

local-setup/gatus/config.yaml
singhalkarun: Valid nit, and worth writing down — with one correction: the built-in default is **100**, not 200.

Upstream `storage/config.go`:

```go
const (
	DefaultMaximumNumberOfResults = 100
	DefaultMaximumNumberOfEvents  = 50
)
```

`ValidateAndSetDefaults()` applies these whenever the values are zero or negative, and the SQL store prunes past those thresholds. So your substantive point holds exactly as stated: this is bounded by an upstream default, not unbounded growth, and `gatus.db` settles at a steady-state size.

Documented in both tiers in 8c200245, as a comment rather than as explicit config keys. That choice is deliberate: Gatus rejects unknown config keys outright, and `egovio/gatus:latest` is an unpinned mirror whose version I cannot confirm from here — so writing keys I cannot verify the image accepts would risk the same class of startup failure as #1622 for zero behavioural gain at the default values. The comment records the numbers, names the upstream symbol so the next reader can re-check it, and says what to verify before setting them for real.

local-setup/docker-compose.egov-digit.yaml
singhalkarun: Checked statically and against the recorded devbox run. No permission problem, and no change needed — but this is now verified rather than assumed.

**Static.** Upstream's Dockerfile has no `USER` directive; the final stage is `FROM scratch` with `ENTRYPOINT ["/gatus"]`, so the container runs as UID 0. A fresh Docker named volume is created root-owned, so root writing `/data/gatus.db` lines up. Your read of the upstream non-root issue is right, incidentally — it is still open, and root is still the default.

**Runtime — already exercised, which I think retires the checklist item.** The #1601 integration pass restarted gatus on the devbox with this exact named volume, and history *accumulated across the restart* (96 → 123) rather than resetting. That outcome is only reachable if sqlite opened `/data/gatus.db` read-write and committed to it, so the write path is proven, not merely likely. A permission-denied on `/data` would have shown up as the pre-fix behaviour (a reset), which is precisely what the numbers rule out.

Leaving `user:` unset. Given the image runs as root and the volume is root-owned, *adding* a `user:` is the change that would introduce a permission-denied here, not the absence of one.

One limitation worth stating: I could not re-run `docker exec digit-gatus ls -la /data` myself — no Docker daemon on this machine, and I deliberately did not touch the devbox (other agents are on it). The above is upstream source plus the recorded run, not a fresh observation.

singhalkarun:
## k3s tier: the sqlite database sits on an `emptyDir`, so history still does not survive

The compose half of this PR is verified on the devbox. The **k3s half is not equivalent** — I deployed `local-setup/k8s/tools/gatus.yaml` on a real cluster and the history is still lost.

The storage config is right:

```yaml
storage:
  type: sqlite
  path: /data/gatus.db
```

but the volume behind `/data` is not:

```yaml
- name: gatus-data
  emptyDir: {}
```

`emptyDir` is deleted with the pod. It survives a *container* restart inside the same pod, but not pod replacement — a rollout, a node drain, an eviction, an upgrade. Those are the cases you actually consult uptime history after.

Measured it rather than argued it. Deleted the pod and compared timestamps:

```
pod started at:        2026-08-10T12:30:56Z
oldest stored result:  2026-08-10T12:30:57.98Z   <- one second AFTER pod start
newest stored result:  2026-08-10T12:31:33.11Z
```

Every row in the database is post-restart. Nothing carried over — which is the same outcome as the in-memory store this PR replaces, for the failure mode the PR exists to address.

Compare with the deploy-as-code tier, where the same question came up on #1659 and the answer was different: there, a PVC-backed volume kept the marker file and the WAL across a pod delete.

**Fix:** back `/data` with a PersistentVolumeClaim instead of `emptyDir`, as #1659 does for Prometheus. On k3s the default `local-path` StorageClass is enough. Worth noting a single-replica constraint comes with it (`ReadWriteOnce` plus a sqlite file means the Deployment must not scale past 1 — it currently does not, but nothing enforces it).

singhalkarun:

`933d48f8` — `/data` is now a 1Gi ReadWriteOnce PVC instead of an `emptyDir`, and the Deployment uses `strategy: Recreate`.

Recreate is not cosmetic: a RollingUpdate against a RWO claim deadlocks, because the new pod cannot mount the volume until the old one releases it and the old one is not terminated until the new one is ready. It also keeps two Gatus processes off one sqlite file.

Re-verified with the same measurement that exposed the problem — delete the pod, compare timestamps:

```
before delete:  84 results, oldest 2026-08-10T16:31:05.736Z
new pod started at 16:32:38Z
after delete:  140 results, oldest 2026-08-10T16:31:05.736Z   <- unchanged
```

The oldest row is identical and predates the new pod, and the count grew rather than resetting. Before the fix the oldest row was one second *after* pod start.

vinothrallapalli-eGov (approved):

PR #1632, opened 2026-08-06
feat(gatus): check public TLS certificate expiry

### LINKED ISSUE #1622: Nothing monitors TLS certificate expiry — Gatus has no external HTTPS check

singhalkarun:
Fixes #1622

## What

Nothing watched TLS certificate validity. Certificates are Let's Encrypt, renewed by certbot's systemd timer — automatically, and unverified. If renewal fails (expired ACME account, blocked HTTP-01 challenge, nginx not reloaded after renewal, disabled timer), the certificate expires silently and **every citizen and employee gets a browser security warning**. It is also the most predictable outage possible: the expiry date is known 90 days ahead, and certbot renews at 30.

## The existing monitoring structurally cannot catch it

Every one of the ~51 Gatus checks is `http://` or `tcp://` against a service name on the container network — **41 http, 10 tcp, zero external**. The Ansible validate tasks have the same blind spot, using `127.0.0.1` with a `Host:` header. Nothing traverses nginx, so nothing ever presents the public certificate.

With an expired certificate the dashboard therefore stays **entirely green** while the platform is unusable — actively misleading whoever is diagnosing it, because from the inside everything genuinely is healthy.

## Change

One endpoint, added to both tier catalogues:

```yaml
  - name: TLS Certificate
    group: Public Endpoint
    enabled: ${GATUS_TLS}
    url: "${GATUS_PUBLIC_URL}"
    interval: 5m
    conditions:
      - "[CERTIFICATE_EXPIRATION] > 240h"
      - "[STATUS] < 500"
```

Gatus supports certificate-expiry conditions natively, so this needs **no new component**.

Beyond certificates, this is the only check that exercises the **real public path end to end** — DNS, host firewall, nginx, TLS termination, and routing through to the app. Everything else bypasses all of it. 240h ≈ 10 days of warning, against certbot's 30-day renewal, so it fires only once renewal has genuinely failed rather than on the normal cycle.

## Off unless a deployment opts in

Ansible derives both variables from existing inventory (`digit.env.j2`):

```jinja
GATUS_TLS={{ (gatus_tls_check | default(tls_enabled | default(true))) | lower }}
GATUS_PUBLIC_URL={{ 'https://' + domain + '/' if (tls_enabled | default(true)) else '' }}
```

`gatus_tls_check` is a host_vars escape hatch for boxes where the Gatus container cannot reach its own public hostname from inside — hairpin NAT is not universal, and a permanent false red is precisely how dashboards get ignored.

## Why all five compose files and the k8s Deployment changed

This is required, not tidying. **Gatus reads an unset `enabled` variable as `true`** — the config header already warns about this — so omitting the toggle anywhere would enable the check with an unsubstituted `${GATUS_PUBLIC_URL}`, producing a permanently-red square nobody can fix. Every file that mounts this config now declares both variables explicitly:

| File | GATUS_TLS |
|---|---|
| `docker-compose.egov-digit.yaml` (tenant path) | `${GATUS_TLS:-false}` — Ansible supplies it |
| `docker-compose.yml`, `deploy.yaml`, `registry.yml`, `db-migrations.yml` | `"false"` — no public hostname |
| `k8s/tools/gatus.yaml` Deployment | `"false"` — no ingress in that tier (#1618) |

## Verification

- [x] All seven touched files parse; both tier catalogues carry an identical endpoint
- [x] `check-gatus-coverage.py` passes with self-test: **52 endpoints per tier, no drift, no dangling checks**
- [x] Every compose file and the k8s Deployment declare both variables
- [ ] On a TLS tenant: the check appears and passes against the live certificate
- [ ] Gatus container can actually reach the public hostname from inside (the hairpin-NAT caveat)
- [ ] Forcing a near-expiry certificate turns it red, and alerts once #1609 lands

The last three need a real box; not yet deployed.

## Merge note

Touches `local-setup/gatus/config.yaml` / `k8s/tools/gatus.yaml`, as do #1630 (sqlite persistence) and #1609 (Slack alerting). Sequence them and keep both tiers in step in whichever order they land.

### DIFF
--- a/local-setup/ansible/templates/digit.env.j2
+++ b/local-setup/ansible/templates/digit.env.j2
@@ -120,7 +120,16 @@ NOVU_BRIDGE_CHANNEL={{ novu_bridge_channel | default('sms') }}
 # container cannot reach the public hostname from inside the host (hairpin NAT
 # is not universal), which would otherwise show as a false failure.
 GATUS_TLS={{ (gatus_tls_check | default(tls_enabled | default(true))) | lower }}
-GATUS_PUBLIC_URL={{ 'https://' + domain + '/' if (tls_enabled | default(true)) else '' }}
+# Always populated, even when the check is disabled. Gatus validates `url` on
+# EVERY endpoint regardless of `enabled:`, so an empty value here is not an inert
+# placeholder -- it makes gatus PANIC at startup ("you must specify an url for
+# each endpoint"), the container dies, and the whole health dashboard goes with
+# it (nginx /status/ 502s). Caught deploying this to a tenant with
+# tls_enabled: false.
+#
+# `enabled: ${GATUS_TLS}` is what actually gates the check. The URL only has to
+# be syntactically present; it is never dialled while the endpoint is disabled.
+GATUS_PUBLIC_URL={{ ('https://' if (tls_enabled | default(true)) else 'http://') + domain + '/' }}
 
 GATUS_PROFILE_OTP={{ enable_otp_services | default(false) | lower }}
 GATUS_PROFILE_SEARCH={{ enable_search_stack | default(false) | lower }}
--- a/local-setup/ansible/templates/digit.env.j2
+++ b/local-setup/ansible/templates/digit.env.j2
@@ -111,6 +111,17 @@ NOVU_BRIDGE_CHANNEL={{ novu_bridge_channel | default('sms') }}
 # Rendered unconditionally: the compose files read these as ${GATUS_PROFILE_*:-false},
 # and that fallback is what makes an unset toggle mean "off" rather than a red
 # check for a stack this tenant does not run.
+# Public TLS-certificate check (#1622). Every other Gatus check runs inside the
+# container network and never presents the public certificate, so an expired one
+# leaves the whole dashboard green while the site is unusable for everyone.
+#
+# Only meaningful when TLS actually terminates here, so it follows tls_enabled.
+# Override gatus_tls_check in host_vars to force it off — e.g. where the gatus
+# container cannot reach the public hostname from inside the host (hairpin NAT
+# is not universal), which would otherwise show as a false failure.
+GATUS_TLS={{ (gatus_tls_check | default(tls_enabled | default(true))) | lower }}
+GATUS_PUBLIC_URL={{ 'https://' + domain + '/' if (tls_enabled | default(true)) else '' }}
+
 GATUS_PROFILE_OTP={{ enable_otp_services | default(false) | lower }}
 GATUS_PROFILE_SEARCH={{ enable_search_stack | default(false) | lower }}
 GATUS_PROFILE_MCP={{ enable_mcp | default(false) | lower }}
--- a/local-setup/docker-compose.egov-digit.yaml
+++ b/local-setup/docker-compose.egov-digit.yaml
@@ -1659,6 +1659,14 @@ services:
       GATUS_AUDIT: "true"
       GATUS_PROXIES: "true"
       GATUS_SMS_PORT: "8005"
+      # Public TLS-certificate check (#1622). Ansible renders both of these into
+      # .env from `domain` + `tls_enabled`; the defaults keep the check OFF for a
+      # bare `docker compose up` outside Ansible, where there is no public
+      # hostname. Off is the safe default: an unset `enabled` var reads as TRUE to
+      # Gatus, which would leave a permanently-red check against a URL that never
+      # resolves — and a red square nobody can fix is how dashboards get ignored.
+      GATUS_TLS: ${GATUS_TLS:-false}
+      GATUS_PUBLIC_URL: ${GATUS_PUBLIC_URL:-}
     networks:
       - egov-network
 
--- a/local-setup/docker-compose.egov-digit.yaml
+++ b/local-setup/docker-compose.egov-digit.yaml
@@ -1666,7 +1666,14 @@ services:
       # Gatus, which would leave a permanently-red check against a URL that never
       # resolves — and a red square nobody can fix is how dashboards get ignored.
       GATUS_TLS: ${GATUS_TLS:-false}
-      GATUS_PUBLIC_URL: ${GATUS_PUBLIC_URL:-}
+      # The fallback is non-empty on purpose, and `:-` (not `-`) is load-bearing:
+      # it fires when the var is unset OR set-but-empty, so a stale .env carrying
+      # `GATUS_PUBLIC_URL=` cannot reintroduce the empty value. Gatus validates
+      # `url` on EVERY endpoint before it consults `enabled:`, so "" aborts
+      # startup with "you must specify an url for each endpoint" and takes the
+      # whole dashboard down (nginx /status/ then 502s). Only ever parsed here --
+      # with GATUS_TLS false the endpoint is never dialled.
+      GATUS_PUBLIC_URL: ${GATUS_PUBLIC_URL:-https://localhost/}
     networks:
       - egov-network
 
--- a/local-setup/gatus/config.yaml
+++ b/local-setup/gatus/config.yaml
@@ -472,3 +472,34 @@ endpoints:
     interval: 60s
     conditions:
       - "[STATUS] == 200"
+
+  # ==================== Public endpoint (TLS) ====================
+  # Every other check in this file runs INSIDE the container network and never
+  # traverses nginx, so none of them presents the public certificate. An expired
+  # cert therefore leaves this whole dashboard green while every citizen and
+  # employee gets a browser security warning — the dashboard would actively
+  # mislead whoever is diagnosing the outage (#1622).
+  #
+  # This is also the only check that exercises the REAL public path end to end:
+  # DNS, the host firewall, nginx, TLS termination and routing through to the
+  # app. Everything above bypasses all of it.
+  #
+  # Certificates are Let's Encrypt, renewed by certbot's systemd timer. Renewal
+  # is automatic and unverified: if it fails (expired ACME account, blocked
+  # HTTP-01 challenge, nginx not reloaded, disabled timer) the expiry is
+  # perfectly predictable 90 days ahead, and nothing currently notices.
+  #
+  # 240h ≈ 10 days of warning. Certbot renews at 30 days, so this only fires
+  # once renewal has genuinely failed, not on the normal renewal cycle.
+  #
+  # GATUS_TLS is false unless the deployment sets it: it needs a real public
+  # hostname AND the gatus container must be able to reach it from inside
+  # (hairpin NAT is not universal). Ansible enables it only when tls_enabled.
+  - name: TLS Certificate
+    group: Public Endpoint
+    enabled: ${GATUS_TLS}
+    url: "${GATUS_PUBLIC_URL}"
+    interval: 5m
+    conditions:
+      - "[CERTIFICATE_EXPIRATION] > 240h"
+      - "[STATUS] < 500"
--- a/local-setup/gatus/config.yaml
+++ b/local-setup/gatus/config.yaml
@@ -495,11 +495,27 @@ endpoints:
   # GATUS_TLS is false unless the deployment sets it: it needs a real public
   # hostname AND the gatus container must be able to reach it from inside
   # (hairpin NAT is not universal). Ansible enables it only when tls_enabled.
+  #
+  # GATUS_PUBLIC_URL is always rendered non-empty, even when GATUS_TLS is false.
+  # Gatus validates `url` on EVERY endpoint before it consults `enabled:`, so an
+  # empty value here does not disable the check -- it aborts startup entirely
+  # ("you must specify an url for each endpoint") and takes the dashboard with it.
+  #
+  # [CONNECTED] == true is what separates the two ways this can go red. On any
+  # TLS failure -- SAN mismatch, broken chain, wrong cert -- the handshake never
+  # completes, so [CERTIFICATE_EXPIRATION] resolves to 0s and [STATUS] to 0. With
+  # only the two conditions below, "0s > 240h" would fail and read exactly like a
+  # cert about to expire, sending on-call to chase a renewal that is not the
+  # problem. [CONNECTED] fails too in that case and not in the expiry case, so
+  # which condition is red says which fault it is. (It also closes a real gap:
+  # 0 < 500 is TRUE, so [STATUS] alone stays green on a dead connection.)
+  # Upstream pairs these two conditions in its own certificate examples.
   - name: TLS Certificate
     group: Public Endpoint
     enabled: ${GATUS_TLS}
     url: "${GATUS_PUBLIC_URL}"
     interval: 5m
     conditions:
+      - "[CONNECTED] == true"
       - "[CERTIFICATE_EXPIRATION] > 240h"
       - "[STATUS] < 500"
--- a/local-setup/docker-compose.db-migrations.yml
+++ b/local-setup/docker-compose.db-migrations.yml
@@ -1314,7 +1314,12 @@ services:
       # true, which would leave a permanently-red check for a URL that does
       # not resolve.
       GATUS_TLS: "false"
-      GATUS_PUBLIC_URL: ""
+      # Non-empty on purpose. Gatus validates `url` on EVERY endpoint before it
+      # ever consults `enabled:`, so "" is not an inert placeholder -- it aborts
+      # startup with "you must specify an url for each endpoint" and takes the
+      # entire dashboard down with it. This value is only ever parsed by the
+      # config loader; the check is disabled, so it is never dialled.
+      GATUS_PUBLIC_URL: "https://localhost/"
     deploy:
       resources:
         limits:
--- a/local-setup/docker-compose.db-migrations.yml
+++ b/local-setup/docker-compose.db-migrations.yml
@@ -1309,6 +1309,12 @@ services:
       GATUS_AUDIT: "false"
       GATUS_PROXIES: "false"
       GATUS_SMS_PORT: "8080"
+      # No public hostname on this path — the TLS check (#1622) is off.
+      # Declared explicitly because Gatus treats an UNSET `enabled` var as
+      # true, which would leave a permanently-red check for a URL that does
+      # not resolve.
+      GATUS_TLS: "false"
+      GATUS_PUBLIC_URL: ""
     deploy:
       resources:
         limits:
--- a/local-setup/docker-compose.deploy.yaml
+++ b/local-setup/docker-compose.deploy.yaml
@@ -927,7 +927,12 @@ services:
       # true, which would leave a permanently-red check for a URL that does
       # not resolve.
       GATUS_TLS: "false"
-      GATUS_PUBLIC_URL: ""
+      # Non-empty on purpose. Gatus validates `url` on EVERY endpoint before it
+      # ever consults `enabled:`, so "" is not an inert placeholder -- it aborts
+      # startup with "you must specify an url for each endpoint" and takes the
+      # entire dashboard down with it. This value is only ever parsed by the
+      # config loader; the check is disabled, so it is never dialled.
+      GATUS_PUBLIC_URL: "https://localhost/"
     networks:
       - egov-network
 
--- a/local-setup/docker-compose.deploy.yaml
+++ b/local-setup/docker-compose.deploy.yaml
@@ -922,6 +922,12 @@ services:
       GATUS_AUDIT: "false"
       GATUS_PROXIES: "false"
       GATUS_SMS_PORT: "8080"
+      # No public hostname on this path — the TLS check (#1622) is off.
+      # Declared explicitly because Gatus treats an UNSET `enabled` var as
+      # true, which would leave a permanently-red check for a URL that does
+      # not resolve.
+      GATUS_TLS: "false"
+      GATUS_PUBLIC_URL: ""
     networks:
       - egov-network
 
--- a/local-setup/docker-compose.registry.yml
+++ b/local-setup/docker-compose.registry.yml
@@ -1621,7 +1621,12 @@ services:
       # true, which would leave a permanently-red check for a URL that does
       # not resolve.
       GATUS_TLS: "false"
-      GATUS_PUBLIC_URL: ""
+      # Non-empty on purpose. Gatus validates `url` on EVERY endpoint before it
+      # ever consults `enabled:`, so "" is not an inert placeholder -- it aborts
+      # startup with "you must specify an url for each endpoint" and takes the
+      # entire dashboard down with it. This value is only ever parsed by the
+      # config loader; the check is disabled, so it is never dialled.
+      GATUS_PUBLIC_URL: "https://localhost/"
     networks:
       - egov-network
 
--- a/local-setup/docker-compose.registry.yml
+++ b/local-setup/docker-compose.registry.yml
@@ -1616,6 +1616,12 @@ services:
       GATUS_AUDIT: "false"
       GATUS_PROXIES: "true"
       GATUS_SMS_PORT: "8080"
+      # No public hostname on this path — the TLS check (#1622) is off.
+      # Declared explicitly because Gatus treats an UNSET `enabled` var as
+      # true, which would leave a permanently-red check for a URL that does
+      # not resolve.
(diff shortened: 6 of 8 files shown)


singhalkarun:
## Devbox test: this takes the health dashboard down on any tenant with `tls_enabled: false`

Deployed to the eGov devbox (`karun`, `tls_enabled: false`). Gatus **crash-looped** and the deploy failed at the `/status/` health gate with a 502.

```
panic: error parsing config: invalid endpoint public-endpoint_tls-certificate:
you must specify an url for each endpoint
```
→ `digit-gatus  Exited (2)`, `http://127.0.0.1/status/` → **502 Bad Gateway**

## Cause

`digit.env.j2` renders the URL empty when TLS is off:

```jinja
GATUS_PUBLIC_URL={{ 'https://' + domain + '/' if (tls_enabled | default(true)) else '' }}
```

and the endpoint is gated on `enabled:`:

```yaml
- name: TLS Certificate
  enabled: ${GATUS_TLS}        # false here
  url: "${GATUS_PUBLIC_URL}"   # "" here
```

**Gatus validates `url` on every endpoint regardless of `enabled:`.** So instead of skipping a disabled check, it refuses to start at all.

That makes the net effect worse than not having the check: on a non-TLS tenant this PR takes down **every other check** along with it — the dashboard that is supposed to tell you what is broken is itself the thing that is broken.

Worth noting because the pattern looks safe: the `GATUS_PROFILE_*` endpoints elsewhere in this file gate exactly the same way and are fine — their urls stay non-empty when disabled. This is the only endpoint whose url *also* empties, which is why the idiom fails here specifically.

## Fix

Pushed to `fix/1632-tls-url` (`c9542647`): always populate the URL, scheme following `tls_enabled`. The `enabled: ${GATUS_TLS}` gate is unchanged and is what actually stops the check; the URL only needs to be syntactically present, since it is never dialled while disabled.

## Verified after the fix

- gatus back **Up**, `/status/` no longer 502
- **TLS Certificate endpoint correctly absent** from the active set while `GATUS_TLS=false`
- the rest of the batch unaffected — #1630's sqlite persistence verified in the same run (history survived a restart: 96 → 123, where before the change it went 1650 → 22)

## Untested

The check firing **positively** — this tenant has TLS off, so the endpoint never runs. Someone should confirm on a `tls_enabled: true` tenant that it evaluates `[CERTIFICATE_EXPIRATION]` correctly and that the gatus container can actually reach the public hostname from inside (the hairpin-NAT caveat the config comment already flags).

vinothrallapalli-eGov (commented):
## P0 defect: CONFIRMED STILL PRESENT at the current head commit (f12b574b) — this is a release blocker

I independently verified the panic-on-non-TLS-tenant defect against the live diff, not just the PR description. It is real and unfixed on this branch as it stands.

**Root cause, traced through the diff:** `local-setup/ansible/templates/digit.env.j2` line 123 renders `GATUS_PUBLIC_URL` to the **empty string** whenever `tls_enabled` is false (or unset and defaulted false):
```
GATUS_PUBLIC_URL={{ 'https://' + domain + '/' if (tls_enabled | default(true)) else '' }}
```
`docker-compose.egov-digit.yaml` line 1669 (`GATUS_PUBLIC_URL: ${GATUS_PUBLIC_URL:-}`) and the four other compose files / the k8s Deployment hard-code it to `""` outright. That empty value then lands in `local-setup/gatus/config.yaml` line 501 (`url: "${GATUS_PUBLIC_URL}"`) on the new `TLS Certificate` endpoint.

Gatus validates the `url` field on **every** endpoint at config-parse time, unconditionally — the `enabled: ${GATUS_TLS}` gate on line 500 does not exempt it. Result on any non-TLS tenant:
```
panic: error parsing config: invalid endpoint public-endpoint_tls-certificate: you must specify an url for each endpoint
```
Gatus exits (code 2), the container crash-loops, and `/status/` 502s — taking down visibility into every *other* check along with it. This is strictly worse than not shipping the TLS check at all.

**Confirmed via commit history, not present on this branch:** a fix exists at `fix/1632-tls-url` (commit `c9542647`), branched directly off this PR's head (`f12b574b`) and titled `fix(gatus): never emit an empty url — it panics gatus at startup (#1622)`. Its commit message independently corroborates the exact panic text above and states it was devbox-verified (`/status/` recovered, TLS endpoint correctly absent from the active set while `GATUS_TLS` is false). The fix keeps `enabled: ${GATUS_TLS}` as the actual gate and instead makes the URL always syntactically non-empty:
```
GATUS_PUBLIC_URL={{ ('https://' if (tls_enabled | default(true)) else 'http://') + domain + '/' }}
```
i.e. it templates a scheme-appropriate but always-present URL rather than templating the endpoint out. **This fix has NOT been merged into PR #1632.** `git compare f12b574b...fix/1632-tls-url` shows the fix branch is 1 commit ahead / 0 behind this PR's head — it is sitting unintegrated. Do not merge #1632 as-is; it will 502 the dashboard on every non-TLS tenant.

## Secondary finding: CERTIFICATE_EXPIRATION-reads-as-0 on TLS error is still unhandled even after the fix

Separately from the panic: the conditions on this endpoint are only `[CERTIFICATE_EXPIRATION] > 240h` and `[STATUS] < 500`. Per Gatus's own documented behavior, when the TLS handshake itself errors (mis-issued cert, wrong SAN, broken chain, connection refused before the handshake completes), `[CERTIFICATE_EXPIRATION]` reads back as `0s` rather than surfacing a distinct error/failure state. `0s` is *less than* `240h`, so this endpoint will correctly go red for that case too — the failure mode isn't silently swallowed — but the dashboard will present a hard TLS misconfiguration identically to "cert expires in 3 days," i.e. exactly what the task brief flagged: an outright error gets classified as "imminent expiry" rather than a distinct error state. Neither this PR nor the `fix/1632-tls-url` branch adds a condition to disambiguate (e.g. asserting `[STATUS] == 200` more strictly, or a separate `[CONNECTED]`-style condition/alert message distinguishing handshake failure from real expiry). This is not blocking, but worth a follow-up so on-call doesn't misdiagnose a cert-mismatch outage as "renewal is 3 days away."

## Summary
- **Panic bug: PRESENT, not fixed, in the current diff at head f12b574b.** Flagging as P0 / release blocker per the review brief — do not merge until `fix/1632-tls-url` (or an equivalent always-non-empty-URL template) is folded in.
- Everything else about the PR (the `enabled:` gating pattern for the other 5 compose files/k8s, the coverage-CI enforcement of parity between the two catalogues, the 240h threshold rationale) is sound and consistent with the existing profile-gated-endpoint pattern in this file — the TLS endpoint was the only one whose `url` *also* went empty when disabled, which is exactly why it alone panics.
- Secondary: the 0s-on-TLS-error ambiguity is real per Gatus semantics and is not addressed by either this branch or the fix branch; recommend a follow-up condition/alert-message change, not a blocker for this PR.

local-setup/ansible/templates/digit.env.j2
```diff
@@ -111,6 +111,17 @@ NOVU_BRIDGE_CHANNEL={{ novu_bridge_channel | default('sms') }}
...
+# Public TLS-certificate check (#1622). Every other Gatus check runs inside the
+# container network and never presents the public certificate, so an expired one
+# leaves the whole dashboard green while the site is unusable for everyone.
+#
+# Only meaningful when TLS actually terminates here, so it follows tls_enabled.
+# Override gatus_tls_check in host_vars to force it off — e.g. where the gatus
+# container cannot reach the public hostname from inside the host (hairpin NAT
+# is not universal), which would otherwise show as a false failure.
+GATUS_TLS={{ (gatus_tls_check | default(tls_enabled | default(true))) | lower }}
+GATUS_PUBLIC_URL={{ 'https://' + domain + '/' if (tls_enabled | default(true)) else '' }}
```
vinothrallapalli-eGov: This is the root cause of the P0 panic. When `tls_enabled` is false (or defaulted false), this renders `GATUS_PUBLIC_URL` to the empty string. Gatus validates `url` on every endpoint at parse time regardless of `enabled:`, so an empty url here makes gatus panic at startup (`you must specify an url for each endpoint`), crash-loop, and take `/status/` down with it (502) on every non-TLS tenant.

Verified: this is exactly what the unmerged fix branch `fix/1632-tls-url` (commit c9542647) fixes, by always populating a syntactically-valid URL with a scheme that follows `tls_enabled` instead of ever emitting `''`:
```
GATUS_PUBLIC_URL={{ ('https://' if (tls_enabled | default(true)) else 'http://') + domain + '/' }}
```
That fix is not present on this branch — please pull it in before merge. This is a release blocker: it kills the whole dashboard, not just the TLS check, on any tenant with `tls_enabled: false`.

local-setup/gatus/config.yaml
```diff
@@ -472,3 +472,34 @@ endpoints:
...
+  # 240h ≈ 10 days of warning. Certbot renews at 30 days, so this only fires
+  # once renewal has genuinely failed, not on the normal renewal cycle.
+  #
+  # GATUS_TLS is false unless the deployment sets it: it needs a real public
+  # hostname AND the gatus container must be able to reach it from inside
+  # (hairpin NAT is not universal). Ansible enables it only when tls_enabled.
+  - name: TLS Certificate
+    group: Public Endpoint
+    enabled: ${GATUS_TLS}
+    url: "${GATUS_PUBLIC_URL}"
```
vinothrallapalli-eGov, line 516: This is where the empty `GATUS_PUBLIC_URL` (see digit.env.j2 line 123 and the compose files that default it to `""`) lands. Because Gatus validates `url` unconditionally at config-parse time — the `enabled: ${GATUS_TLS}` gate above does not exempt it from validation — an empty value here panics gatus at startup on any non-TLS tenant, rather than merely leaving this one check red. Comment above this block asserts 'GATUS_TLS is false unless the deployment sets it' but doesn't account for the fact that GATUS_PUBLIC_URL also goes empty in that same case, which is the actual trigger.

local-setup/gatus/config.yaml
```diff
@@ -472,3 +472,34 @@ endpoints:
...
+    interval: 5m
+    conditions:
+      - "[CERTIFICATE_EXPIRATION] > 240h"
```
vinothrallapalli-eGov, line 520: Secondary, non-blocking: per Gatus's documented behavior, `[CERTIFICATE_EXPIRATION]` reads back as `0s` when the TLS handshake itself errors (mis-issued cert, SAN mismatch, broken chain) rather than surfacing a distinct failure. `0s > 240h` is false, so this will correctly go red — but it will be visually indistinguishable from 'renewal is overdue,' when it's actually a hard misconfiguration. Consider a follow-up to disambiguate (e.g. a dedicated condition or alert-message branch) so on-call doesn't misdiagnose a cert-mismatch outage as an imminent-expiry warning.

local-setup/docker-compose.egov-digit.yaml
```diff
@@ -1659,6 +1659,14 @@ services:
...
       GATUS_PROXIES: "true"
       GATUS_SMS_PORT: "8005"
+      # Public TLS-certificate check (#1622). Ansible renders both of these into
+      # .env from `domain` + `tls_enabled`; the defaults keep the check OFF for a
+      # bare `docker compose up` outside Ansible, where there is no public
+      # hostname. Off is the safe default: an unset `enabled` var reads as TRUE to
+      # Gatus, which would leave a permanently-red check against a URL that never
+      # resolves — and a red square nobody can fix is how dashboards get ignored.
+      GATUS_TLS: ${GATUS_TLS:-false}
+      GATUS_PUBLIC_URL: ${GATUS_PUBLIC_URL:-}
```
vinothrallapalli-eGov: Same empty-URL trap as digit.env.j2: `${GATUS_PUBLIC_URL:-}` defaults to `""` whenever Ansible doesn't populate it (or renders it empty for non-TLS tenants), which panics gatus per the linked defect. Needs the same always-non-empty-URL fix as the template.

local-setup/ansible/templates/digit.env.j2
singhalkarun: Confirmed, and folded in — `fix/1632-tls-url` (c9542647) is now on this branch as a fast-forward, so the template renders a scheme-following URL and never `''`.

The root cause is confirmed at source, not just by symptom. `ValidateEndpointsConfig` iterates every endpoint with no `IsEnabled()` check before calling `ValidateAndSetDefaults`, which returns `ErrEndpointWithNoURL` — "you must specify an url for each endpoint". `enabled:` is never consulted on that path, exactly as you describe.

Rendering the fixed template both ways:

| `tls_enabled` | `GATUS_TLS` | `GATUS_PUBLIC_URL` |
|---|---|---|
| `true` | `true` | `https://<domain>/` |
| `false` | `false` | `http://<domain>/` |

Neither empty.

**One important caveat on "please pull it in before merge":** pulling it in was necessary but *not sufficient*. This template was one of six places feeding `GATUS_PUBLIC_URL`, and c9542647 fixed only this one — the other five still emitted an empty string, so the P0 would have survived the merge. Closed in 022df05e; details on the `docker-compose.egov-digit.yaml` thread.

local-setup/gatus/config.yaml
singhalkarun: Confirmed — and your framing is the more useful one. The `enabled: ${GATUS_TLS}` gate above genuinely does not exempt this endpoint from validation, which means the comment asserting "GATUS_TLS is false unless the deployment sets it" was answering a question nobody was asking. The toggle was never the thing standing between a non-TLS tenant and a crash-loop.

Fixed at both ends in 022df05e:

- **Every producer** of `GATUS_PUBLIC_URL` now emits a non-empty value — six paths in total, not just the template (see the `docker-compose.egov-digit.yaml` thread).
- **The comment block here** now states the actual invariant — `url` is validated before `enabled` is consulted, so it must always be populated — instead of implying the toggle is sufficient. That was the part most likely to cause someone to reintroduce this later.

Verified by expanding env vars into the config and parsing the result for every path: ansible `tls_enabled` true and false, bare `docker compose up`, a stale `.env` carrying an empty value, and the k3s Deployment env. No endpoint with an empty url in any of them. The same check run against `f12b574b` flags `TLS Certificate`, so it is a real test and not a tautology.

local-setup/gatus/config.yaml
singhalkarun: Confirmed, and addressed here rather than deferred — the fix turned out to be one line.

Upstream documents `[CERTIFICATE_EXPIRATION]` as resolving to `0` when there is no certificate to read, so on a handshake fault `0s > 240h` fails and is visually identical to an overdue renewal. Exactly the misdiagnosis you describe.

Added `[CONNECTED] == true` to the check, in both tiers. It fails on a handshake fault and passes on a genuine imminent expiry, so *which* condition is red now names the fault — no new endpoint, no alert-message branching. This is also upstream's own idiom rather than something invented here: both certificate-monitoring examples in the Gatus README pair `[CONNECTED] == true` with `[CERTIFICATE_EXPIRATION] > 48h`.

It closes a second gap the thread did not raise, which I think justifies doing it now rather than as a follow-up: **`[STATUS]` is `0` on a dead connection, and `0 < 500` is true.** So the existing `[STATUS] < 500` condition stayed *green* on a total connection failure, and the entire signal rested on the cert line alone. With `[CONNECTED]` present, a dead connection fails a condition that is unambiguously about connectivity.

**Not runtime-verified.** Confirming the display requires a `tls_enabled: true` tenant plus a deliberately mismatched host, and I did not deploy (other agents are on the devbox). The mismatched-host path was exercised in the #1601 integration run and behaved as expected, but against the two-condition version — the third condition's rendering is reasoned from upstream docs, not observed.

local-setup/docker-compose.egov-digit.yaml
singhalkarun: Confirmed — and this turned out to be the larger half of the defect. `fix/1632-tls-url` fixed **only** `digit.env.j2`, so pulling that branch in, as the other thread asked, would not by itself have closed the P0. Five further places still emitted an empty URL:

| file | before |
|---|---|
| `docker-compose.yml`, `.deploy.yaml`, `.registry.yml`, `.db-migrations.yml` | `GATUS_PUBLIC_URL: ""` |
| `docker-compose.egov-digit.yaml` | `${GATUS_PUBLIC_URL:-}` |
| `k8s/tools/gatus.yaml` (Deployment env) | `value: ""` |

Every one of them trips the identical mechanism — `ValidateEndpointsConfig` validates `url` on all endpoints before `enabled:` is ever read.

**The k3s row deserves separate billing:** it hardcodes `""`, with no env indirection at all, so that tier would have CrashLoopBackOff'd on first deploy unconditionally — not only on non-TLS tenants. Same P0, on a tier nobody had connected to this defect, and it would have been blamed on something else entirely.

All six now carry a syntactically valid placeholder that is parsed and never dialled. On `docker-compose.egov-digit.yaml` the `:-` form rather than `-` is load-bearing: `:-` substitutes when the variable is unset **or** set-but-empty, so a stale `.env` still carrying `GATUS_PUBLIC_URL=` from a previous deploy cannot reintroduce the panic. That is the realistic regression path once the template is fixed, and it is the one a plain `-` would have left open.

022df05e.

singhalkarun:
## Re-verification: fresh live run, all four claims reproduce

The earlier ✅ on this was overconfident — `c9542647` patched `digit.env.j2`, one of six producers of `GATUS_PUBLIC_URL`, and the devbox tenant happened to use that one path. `022df05e` closed the rest. This is the live run that was outstanding, executed against the **real shipped artifacts** with the real `twinproduction/gatus:latest` image. No cluster and no devbox involved.

### 1. The panic is real, and `enabled: false` does not protect you

Running `local-setup/gatus/config.yaml` **as shipped**, with the exact env block from `docker-compose.yml`, and only `GATUS_PUBLIC_URL` emptied:

```
panic: error parsing config: invalid endpoint public-endpoint_tls-certificate:
       you must specify an url for each endpoint
```

`GATUS_TLS=false` was set the whole time. Gatus validates `url` on every endpoint before it consults `enabled:`, exactly as the code comments claim. So yes — had this merged with any of the five unpatched producers in play, that tier would have died at startup.

### 2. The fix holds, and the disabled endpoint is never dialled

Same config, `GATUS_PUBLIC_URL=https://localhost/`:

```
[config.ValidateEndpointsConfig] Validated 52 endpoints
[controller.Handle] Listening on 0.0.0.0:8080
state: running restarts=0
```

52 endpoints validated, 26 registered for monitoring, and zero log lines mentioning `tls-certificate` — parsed only, never dialled, which is what the placeholder is for.

### 3. The k3s tier: same defect, same fix, both directions confirmed

Extracted the `gatus-config` ConfigMap and the Deployment's env verbatim from `k8s/tools/gatus.yaml` and ran them together:

| `GATUS_PUBLIC_URL` | Result |
|---|---|
| `https://localhost/` (as fixed) | `Validated 52 endpoints`, running, restarts=0 |
| `""` (as the manifest hardcoded before) | `panic: ... you must specify an url for each endpoint` |

So the CrashLoopBackOff on that tier was real and is now closed. Still **not** verified as an actual rollout — there is no cluster — but the cause is gone and the identical config+env starts clean.

### 4. `[STATUS] < 500` really does stay green on a dead connection

Two endpoints, same unreachable target, differing only in conditions:

```
status-only      success=True   errors=['... connect: connection refused']
      True  [STATUS] < 500
with-connected   success=False  errors=['... connect: connection refused']
      False  [CONNECTED] (false) == true
      False  [CERTIFICATE_EXPIRATION] (0s) > 240h
      True  [STATUS] < 500
```

`STATUS` resolves to `0` on a refused connection and `0 < 500` is true, so the check was green while nothing was reachable. `[CONNECTED] == true` is what makes it red.

It also confirms the diagnostic argument in the comment: on a dead connection `[CERTIFICATE_EXPIRATION]` reports `0s`, so a cert-only rule reads exactly like an imminent expiry and would send on-call chasing a renewal that is not the problem. With both conditions present, which one is red tells you which fault it is.

### 5. The check actually works when enabled

`GATUS_TLS=true`, `GATUS_PUBLIC_URL=https://example.com/`, real certificate:

```
public-endpoint_tls-certificate  success=True
    True [CONNECTED] == true
    True [CERTIFICATE_EXPIRATION] > 240h
    True [STATUS] < 500
```

### Producer inventory, current branch

| Producer | Value |
|---|---|
| `docker-compose.yml` | `https://localhost/` |
| `docker-compose.deploy.yaml` | `https://localhost/` |
| `docker-compose.registry.yml` | `https://localhost/` |
| `docker-compose.db-migrations.yml` | `https://localhost/` |
| `docker-compose.egov-digit.yaml` | `${GATUS_PUBLIC_URL:-https://localhost/}` |
| `ansible/templates/digit.env.j2` | Jinja, scheme + `domain` + `/` |
| `k8s/tools/gatus.yaml` (Deployment env) | `https://localhost/` |

All seven non-empty. The Ansible one cannot render empty: `domain` is a required tenant variable set in every `host_vars` example, and an undefined one fails the template task loudly rather than emitting an empty string. Worst case is `domain: ""` giving `https:///` — malformed, but non-empty, so a failing check rather than a dead container.

### What is still not covered

One path, and it is the one that needs a real deployment: **rendering `digit.env.j2` on a live tenant with `tls_enabled: false` and watching the container come up.** Everything upstream of that is now verified, and that producer was the only one already correct before `022df05e`. I have not touched the devbox.

vinothrallapalli-eGov (approved):

Issue #1365, opened 2026-07-21
local-setup: postgres-db missing shm_size, and no log rotation — both cause database failure under load


singhalkarun:
Two container resource defaults in the Docker Compose deployment cause database-level failures under sustained load. Found during PGR load testing on a single-host stack.

## Failures observed

| Failure signature | Count | Root cause | Confirmed | Why unresolved |
|---|---|---|---|---|
| `PGR_List` → HTTP 400, `DataAccessResourceFailureException … could not resize shared memory segment` | 1,412 | `/dev/shm` at Docker's 64 MB default, exhausted by parallel query workers | **Yes** | — |
| `Auth_Login` → HTTP 500, `Could not get JDBC Connection … the database system is in recovery mode` | 153,409 | Disk filled by unrotated container logs → Postgres PANIC → crash-loop | **Yes** | — |
| `Auth_Login` → HTTP 500, `Unknown error occurred in decryption process` | 18,387 | Unknown | No | Began 33 s into the run — 6 min before the disk filled and 30 min before the Postgres PANIC — so it is independent of the row above despite occurring in the same run. No ERROR-level log is emitted for it: the exception is handled and returned to the caller silently. Aggregated logs for the start of the window had already passed retention, and it did not recur after a stack rebuild. |
| `PGR_Assign` / `PGR_Resolve` → HTTP 400, `INVALID ACTION — Action <X> not found in config for the businessService` | 116,568 / 3,111 | Unknown | No | Not uniform — 6,576 assigns succeeded in the same run, so the action was valid at least intermittently. Three hypotheses were tested and falsified: async workflow lag (at 1 VU, a 3 s delay before retry did not help), role or workflow-config drift (neither the role-assignment nor workflow-action tables showed modification in the window), and load correlation (the two failing runs were the earliest, not the highest-rate). The database was later restored from a backup predating the run, so process-instance and role state were unrecoverable. |

The last two are **not** claimed as upstream defects — listed for completeness, and may be specific to the deployment tested.

## Postgres `/dev/shm` exhaustion

1,412 of 3,831 concurrent complaint-list requests (36.9%) failed with:

```
ERROR: could not resize shared memory segment "/PostgreSQL.NNNNNNNNNN"
       to 8388608 bytes: No space left on device
```

[...]

PR #1634, opened 2026-08-06
feat(grafana): dashboard for the pgr-services analytics metrics

### LINKED ISSUE #1616: pgr-services analytics metrics are collected but no dashboard shows them

singhalkarun:
Fixes #1616

## What

`pgr-services` has been emitting analytics query metrics all along — duration histogram and rows-scanned counter, tagged by `kpi_id`, `grain` and `tenant`. No dashboard rendered any of it, so in practice nobody looked. The instrumentation was built, shipped, and verified live; only the last step was missing.

Unlike the rest of this umbrella, nothing here was broken — this is unrealised value.

## The dashboard

Provisioned as `pgr-analytics.json` (uid `pgr-analytics`, folder `DIGIT`), 8 panels:

- duration percentiles (p50/p95/p99)
- **p95 by KPI** — which KPI is expensive
- query rate and rows-scanned rate by KPI
- duration by **planner grain** (`facts`/`events`/`daily`) — whether the planner is choosing the cheap pre-aggregated path or falling back to raw facts
- slowest-KPIs table, and range totals

Three template variables (`tenant`, `kpi_id`, `grain`), all multi-select with an "All" default. `grain` was added in review — the panel that groups by it had no way to isolate one — and is applied to all eight panels, not just that one.

**Every query uses only the four series the metric documentation names**, and the two headline panels are that document's own example queries rather than invented ones. Verified programmatically rather than by eye:

```
metrics in QUERIES: ['pgr_analytics_query_duration_ms_bucket',
                     'pgr_analytics_query_duration_ms_count',
                     'pgr_analytics_query_duration_ms_sum',
                     'pgr_analytics_query_rows_total']
all documented: True
```

That mattered because a dashboard built on a guessed series name renders blank while looking finished — the failure mode this issue exists to avoid.

## Two traps recorded in the dashboard itself

**`tenant` is the state-root tenant** (`ke`), never the city (`ke.bomet`). That is deliberate upstream: the raw request `tenantId` is caller-controlled, so tagging it would let an anonymous caller mint a Prometheus series per request. A panel built expecting city-level granularity would look broken while behaving exactly as designed — so the dashboard description says so.

**The board is blank today, and this PR does not change that.** The OTEL instruments register lazily on first use, so before any analytics query executes the four series do not exist at all — which also means the three template variables come back empty, since they are `label_values()` over those same series. Exercising the path needs KPIs seeded on the tenant, and no test tenant has them. Expected, not a fault, but it does mean the parts below marked unverified are genuinely unverified rather than merely untidy.

## Docs merge

`dashboard-metrics-server.md` carried a merge note asking to fold it into `dashboard-metrics.md` "once both PRs have landed". They have, so this does it.

Worth reviewing one detail: headings are demoted **only outside code fences**. A naive pass also rewrites the `#` comment lines inside the PromQL and shell examples into headings — I hit exactly that on the first attempt and redid it. Verified afterwards that fences are balanced and the code comments survive as comments.

All five references to the removed file are repointed: two dashboard panel descriptions, the `AnalyticsMetricsTest` javadoc, and two in the 2.12-beta migration guide. `grep` for the old filename now returns nothing.

## Verification

Verified statically:

- [x] Dashboard is valid JSON; every datasource pinned to uid `prometheus`
- [x] Queries reference only documented series (re-checked after adding `$grain`)
- [x] All eight panels carry all three template filters
- [x] Docs merged; fences balanced; no dangling references to the deleted file
- [x] `check-gatus-coverage.py` passes

Verified in Docker:

- [x] Dashboard provisions and appears in the `DIGIT` folder

**Not verified, and not verifiable on any current tenant:**

- [ ] Panels populate — needs KPIs seeded so the analytics path executes; none exist
- [ ] `tenant` / `kpi_id` / `grain` variables offer real values — same blocker, since
      they read `label_values()` off the series the panels need

The instrumentation itself is not in question: it is in the pinned image (`7c8dc988`
and `277fed2b` both predate `master-8116812`). What is unproven is that these panels
render data, and nothing in this PR should be read as claiming otherwise.


singhalkarun:
## Devbox test: the dashboard provisions, but its data path is unverified

Deployed on the eGov devbox (`karun`). The dashboard loads in Grafana (9 dashboards total, *DIGIT — PGR Analytics Queries* among them) and Grafana stays healthy.

**However, it renders blank here, and I could not prove it ever populates.** No `pgr_analytics_*` series exist — 0 of 926 metric names in Prometheus:

```
pgr_analytics_query_duration_ms_bucket   0 series
pgr_analytics_query_duration_ms_count    0 series
pgr_analytics_query_rows_total           0 series
```

### This is not a defect in the PR — I chased it down

- **The instrumentation is in the deployed image.** `AnalyticsMetrics.java` arrives via `7c8dc988` (16 Jul) and `277fed2b` (17 Jul); the deploy pins `egovio/pgr-services:master-8116812` (3 Aug), and both are ancestors of it.
- **The metrics are lazily registered.** They are OTEL `LongCounter` / histogram instruments recorded inside `recordQuery(...)`, so they do not appear on `/actuator/prometheus` until a query actually completes. Confirmed: 0 `pgr_analytics` lines there too.
- **No analytics query can succeed on this tenant.** The catalog is empty and a query is rejected:
  ```
  POST /pgr-services/v2/analytics/catalog/_search  ->  200  {"tiles":[],"total":0}
  POST /pgr-services/v2/analytics/_query           ->  400  {"error":"kpi_forbidden",
                                                             "message":"KPI not found or not authorized"}
  ```
  No KPIs are seeded, so `recordQuery` is never reached and the instruments never register.

So the blank dashboard here is environmental, not a fault in the panels.

### What still needs verifying, and by whom

The queries in this dashboard have **never been run against real series** — only against a schema I reasoned about. That is precisely the failure mode #1616 exists to prevent, so I do not want "dashboard provisions cleanly" recorded as "dashboard works".

Someone should confirm on a tenant with KPIs seeded and analytics actually in use that:
- [ ] the four metric names match what pgr-services emits (`_bucket` / `_sum` / `_count` / `_rows_total`) — note `277fed2b` exists specifically because the OTLP unit suffix once mangled these names
- [ ] the histogram panels resolve (`_bucket` requires `le` labels to survive the OTLP → Prometheus conversion)
- [ ] the `kpiId` / `grain` / `tenant` attributes arrive as the labels the panels group by

### Incidental finding

`POST` to a **non-existent** analytics path returns **400**, not 404 — the catch-all `@ControllerAdvice` from #1365. I briefly read that as evidence the endpoint existed. Worth knowing for anyone probing this service by status code.

### Also verified in the same run

Deploy clean (`ok=125, changed=20, failed=0`), 44 containers, 0 unhealthy, Grafana healthy, and the other batch-D PRs passed: #1673's 6 alert rules all provisioned with `health=ok` against real Prometheus/Loki, and #1672's Gatus alerting failed open as designed with no webhook set.

local-setup/otel/grafana/provisioning/dashboards/pgr-analytics.json
```diff
@@ -0,0 +1,353 @@
...
+      ],
+      "fieldConfig": {
+        "defaults": {
+          "custom": {},
+          "unit": "ms"
+        },
+        "overrides": []
+      },
+      "options": {},
+      "description": "facts / events / daily. Shows whether the planner is choosing the cheap pre-aggregated grain or falling back to raw facts."
```
vinothrallapalli-eGov, line 263: Nit (non-blocking): "Duration by planner grain" groups by `grain` but there's no `$grain` template variable to isolate a single grain (only `tenant`/`kpi_id` exist). Not a bug — the legend already splits by grain — just flagging in case a future iteration wants a filter here too.

local-setup/otel/grafana/provisioning/dashboards/pgr-analytics.json
```diff
@@ -0,0 +1,353 @@
...
+    "list": [
+      {
+        "name": "tenant",
+        "label": "Tenant (state root)",
+        "type": "query",
+        "datasource": {
+          "type": "prometheus",
+          "uid": "prometheus"
+        },
+        "query": "label_values(pgr_analytics_query_rows_total, tenant)",
```
vinothrallapalli-eGov, line 27: Nit (non-blocking): both template variables (`tenant` here and `kpi_id` below) source their `label_values(...)` from `pgr_analytics_query_rows_total` only, not the duration histogram. Verified in `AnalyticsMetrics.recordQuery` that both instruments are always recorded together with identical attribute sets in the same call, so this is safe today — just noting the implicit coupling in case that invariant ever changes.

vinothrallapalli-eGov (commented):
Static review only (no seeded tenant available here, so panel rendering itself remains functionally unverified). Cross-checked every PromQL series name in `pgr-analytics.json` against `backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsMetrics.java`: `pgr_analytics_query_duration_ms_bucket/_sum/_count` and `pgr_analytics_query_rows_total` match the actual OTLP instrument names (`pgr.analytics.query.duration.ms`, `pgr.analytics.query.rows`) and the Prometheus-exporter naming convention exactly, and the `kpi_id`/`grain`/`tenant` labels used in every query match the `AttributeKey`s set in `recordQuery(...)`. Datasource `uid: "prometheus"` matches `local-setup/otel/grafana/provisioning/datasources/prometheus.yaml` and is used consistently with the other provisioned dashboards (`jvm-services.json`, `node-exporter-full.json`), and the dashboard `uid`/title don't collide with any existing file in that directory. On the tracking issue's tenant caveat: the dashboard gets this right — the `tenant` template variable is explicitly labeled "Tenant (state root)", the dashboard-level `description` states the state-root-vs-city distinction verbatim, and no panel expression or variable treats `tenant` as city-level. Also verified all 5 repointed references to the deleted `dashboard-metrics-server.md` (test javadoc, 2 migration-guide lines, 2 dashboard panel descriptions) resolve to real content in the merged `dashboard-metrics.md`, and the merged doc's code fences are balanced (12 total, 6 pairs). No incorrect metric names, mislabeled datasource, or tenant-semantics bugs found; only a couple of low-severity/non-blocking observations left inline.

local-setup/otel/grafana/provisioning/dashboards/pgr-analytics.json
singhalkarun: Good catch — implemented in 1d0091fd rather than deferred.

The asymmetry is what decided it: `tenant` and `kpi_id` both had selectors, `grain`
did not, and this is the one panel whose whole point is the grain dimension. "The
legend splits it" is fine at three series and stops being fine the moment you want to
answer *is the daily rollup path slow, specifically* — which is the question the panel
exists for.

Added `grain` as a third template variable in the same shape as the other two (query
/ multi / `includeAll` / `allValue: ".*"` / `refresh: 1` / `sort: 1`), and applied
`grain=~"$grain"` to **all eight** panels rather than only this one. Filtering on a
dimension in one panel and silently ignoring it in the other seven is a worse trap
than not having the filter — the header row should mean the same thing everywhere.

Re-validated after the change:

```
JSON OK; vars: ['tenant', 'kpi_id', 'grain']
metrics used: ['pgr_analytics_query_duration_ms_bucket',
               'pgr_analytics_query_duration_ms_count',
               'pgr_analytics_query_duration_ms_sum',
               'pgr_analytics_query_rows_total']
all documented: True
grain filter on all panels: True
```

Worth flagging in the same breath, because it bears on how much any of this has been
demonstrated: this board is **known-blank**. The OTEL instruments register lazily on
first use, so until an analytics query actually executes the four series do not
exist — which also means all three variables come back empty, since they are
`label_values()` over those same series. Exercising the path needs KPIs seeded on the
tenant and no test tenant has them. So the new variable is verified as valid JSON and
correct PromQL, not as a working dropdown. The dashboard description and
`dashboard-metrics.md` now say that in those terms instead of the softer "empty until
the analytics query path is exercised", and I corrected the PR description too — it
was claiming more than had been shown.

local-setup/otel/grafana/provisioning/dashboards/pgr-analytics.json
singhalkarun: Confirmed, and the invariant holds for the reason you found — no change made, but
worth stating why not, since I added a third variable on the same source in
1d0091fd (`grain`, from your other thread).

`AnalyticsMetrics.recordQuery` builds **one** `Attributes` object and hands the same
reference to both instruments:

```java
Attributes attrs = Attributes.of(
        ATTR_KPI_ID, orDash(kpiId),
        ATTR_GRAIN,  orDash(grain),
        ATTR_TENANT, orDash(tenant));
queryDuration.record(tookMs, attrs);
queryRows.add(rowCount, attrs);
```

That is stronger than "recorded together with identical attribute sets": there is no
second construction site to drift, and the two calls are adjacent inside one `try`.
Divergence would take editing that method, which is also the only place either
instrument is touched.

I kept all three variables on `pgr_analytics_query_rows_total` rather than splitting
them across the two series. Sourcing from the histogram would need
`..._duration_ms_count` — the same three labels, one more chance for a typo, and no
independence gained, since both series come out of the block above. Uniformity is the
cheaper property to check.

One thing that does weaken the coupling in practice: the counter is the safer of the
two to source from. `_bucket` carries an extra `le` dimension and `label_values()` on
a histogram series is the more common way to end up with a surprising result, so if
this ever does need to change the direction is toward the counter, not away from it.

vinothrallapalli-eGov (approved):

Issue #1611, opened 2026-08-05
No alert rules for capacity or error trends — only after-the-fact up/down


singhalkarun:
Part of #1601 (monitoring stack umbrella). **P1 · effort: M**

## Problem

#1609 makes Gatus tell us when a service stops responding. That is a binary, after-the-fact signal. Nothing warns us about the gradual failures that actually take these boxes down:

- **Disk filling** — logs, DB growth, images. Fine until it is abruptly not.
- **Memory exhaustion** — sharper here than usual: the tenant boxes have **no swap**, so the kernel goes straight to OOM-killing processes instead of degrading gracefully.
- **JVM heap creep** — a service climbing toward its limit until it dies and restarts.
- **Error floods** — a service can answer health checks perfectly while failing every real request. Gatus shows it green throughout.

All of this data is already being collected. Prometheus has host CPU/memory/disk (node-exporter, #1335) and JVM metrics via the OTEL pipeline; Loki has every container's logs. The *DIGIT JVM Services* dashboard already charts heap usage against limits. **We gather the evidence and nobody watches it.**

There is no alerting engine of any kind: no Alertmanager, no Prometheus `rule_files`, no `provisioning/alerting/` directory under `local-setup/otel/grafana/provisioning/` (only `dashboards/` and `datasources/` exist).

## Proposed approach — Grafana unified alerting, not Alertmanager

Recommend **Grafana's built-in alerting** over adding Alertmanager:

- no new always-on container on a memory-tight box (see #1601's resource-limit item)
- Grafana already has **both** datasources provisioned — Prometheus *and* Loki. The error-flood rule needs Loki; Alertmanager cannot see it at all.
- rules ship as provisioned files, so they are reviewable in git and deploy automatically, consistent with how dashboards and datasources are already handled

New directory `local-setup/otel/grafana/provisioning/alerting/` with `contactpoints.yaml`, `policies.yaml`, `rules.yaml`.

## Starter rules

Metric names below are taken from the **existing** `jvm-services.json` queries, not invented — but every expression must still be checked against a live box before merge (see caveat).

| Rule | Source | Expression sketch | For |
|---|---|---|---|
| ScrapeTargetDown | Prom | `up == 0` | 5m |
| HostDiskLow | Prom | `node_filesystem_avail_bytes{mountpoint="/"} / node_filesystem_size_bytes{mountpoint="/"} < 0.15` | 10m |
[...]

PR #1635, opened 2026-08-06
feat(compose): cap observability container memory

### LINKED ISSUE #1612: Observability containers run without memory limits and can starve the workload

singhalkarun:
Fixes #1612

> ⚠️ **Do not roll this out everywhere on merge.** It needs a one-week soak on a single tenant first — see "Rollout" below.

## What

No observability container had a memory ceiling. Measured on `docker-compose.egov-digit.yaml`: **29 services declare 12.8 GB of max heap**, on boxes that are commonly 16 GB and have **no swap**.

Without swap the kernel does not degrade gracefully — it OOM-kills, and it targets the **largest** consumer, which is a JVM service rather than the Loki query that caused the pressure. So today the monitoring can misbehave and a **citizen-facing service dies for it**. That is backwards.

## This does not change the deployed stack's philosophy

The deployed stack running application services uncapped is deliberate — only 3 of its 60 services carry limits, while `docker-compose.yml` caps 28 of 34. This PR leaves that alone.

The argument is narrower: **application memory is workload, Loki and Prometheus memory is overhead.** A JVM service consuming memory is the platform doing its job; an expensive Loki query has no claim on memory the workload needs.

## Ceilings

| Container | mem_limit | Why |
|---|---|---|
| loki | 512m | largest real-world grower — chunks plus query working set |
| prometheus | 512m | 15d TSDB; scrape targets grow as exporters land (#1614/#1615/#1623) |
| tempo | 384m | 24h retention, small blocks |
| otel-collector | 320m | sits above its own 200 MiB internal `memory_limiter` |
| grafana | 256m | query proxy + dashboard rendering |
| promtail | 128m | tails files, small buffers |
| gatus | 64m | **not invented** — the k3s Deployment already sets `64Mi` |
| node-exporter | 64m | thin reader of /proc and /sys |

Total ≈ **2.2 GB**.

## Rollout — the important part

A ceiling set too tight converts "slow dashboard" into "dead dashboard": the container is OOM-killed and restart-loops, and the first thing you lose is the monitoring that would have told you.

1. Deploy to **one** tenant.
2. Leave it a week.
3. Check the *Node Exporter Full* dashboard for OOM-kill events before making it the default.

**Loki is the most likely to need raising** — its usage is query-driven, not steady-state, so a wide time-range query is what would trip it.

## Verification

- [x] Diff is **purely additive** — no line removed
- [x] `mem_limit` is the only key added, to exactly the seven intended services (verified by diffing the parsed YAML against `origin/master`, not by eye)
- [x] Service count unchanged: 60 before, 60 after
- [x] `check-gatus-coverage.py` passes
- [ ] `docker inspect` shows the limits applied
- [ ] No monitoring container OOM-killed over 7 days on the pilot box
- [ ] Grafana still renders at normal query ranges (Node Exporter Full is the heaviest board)
- [ ] Loki queries across the full 72h retention still complete

The last four are the soak; not yet deployed.

## Note on process

The first attempt at this inserted each limit by searching forward from `container_name` to the next `restart:` line. The gatus block has no `restart:` line, so the search ran on and placed **gatus's 64m limit inside the Loki service** — which would have capped Loki at 64m and OOM-looped it. Caught by verifying the parsed result rather than the diff, and redone by anchoring the insertion immediately after `container_name` so it cannot cross a service boundary. Mentioning it because it is exactly the kind of change that reads fine in review.

### DIFF
--- a/local-setup/docker-compose.egov-digit.yaml
+++ b/local-setup/docker-compose.egov-digit.yaml
@@ -9,6 +9,11 @@ services:
   otel-collector:
     image: egovio/opentelemetry-collector-contrib:0.114.0
     container_name: digit-otel-collector
+    # Memory ceiling (#1612). Observability is overhead, not workload:
+    # uncapped it competes with ~12.8 GB of JVM heap on 16 GB boxes that
+    # have no swap, and the OOM killer takes the biggest process — a JVM
+    # service — not the monitoring that misbehaved.
+    mem_limit: 320m   # internal memory_limiter is 200 MiB; this is the hard backstop above it
     restart: unless-stopped
     command: ["--config=/etc/otelcol-contrib/config.yaml"]
     volumes:
@@ -23,6 +28,11 @@ services:
   tempo:
     image: egovio/tempo:2.6.1
     container_name: digit-tempo
+    # Memory ceiling (#1612). Observability is overhead, not workload:
+    # uncapped it competes with ~12.8 GB of JVM heap on 16 GB boxes that
+    # have no swap, and the OOM killer takes the biggest process — a JVM
+    # service — not the monitoring that misbehaved.
+    mem_limit: 384m   # 24h retention, small blocks
     restart: unless-stopped
     command: ["-config.file=/etc/tempo/config.yaml"]
     volumes:
@@ -41,6 +51,11 @@ services:
   grafana:
     image: egovio/grafana:11.4.0
     container_name: digit-grafana
+    # Memory ceiling (#1612). Observability is overhead, not workload:
+    # uncapped it competes with ~12.8 GB of JVM heap on 16 GB boxes that
+    # have no swap, and the OOM killer takes the biggest process — a JVM
+    # service — not the monitoring that misbehaved.
+    mem_limit: 256m   # query proxy + dashboard rendering
     restart: unless-stopped
     environment:
       GF_AUTH_ANONYMOUS_ENABLED: "true"
@@ -77,6 +92,11 @@ services:
   prometheus:
     image: prom/prometheus:v2.55.1
     container_name: digit-prometheus
+    # Memory ceiling (#1612). Observability is overhead, not workload:
+    # uncapped it competes with ~12.8 GB of JVM heap on 16 GB boxes that
+    # have no swap, and the OOM killer takes the biggest process — a JVM
+    # service — not the monitoring that misbehaved.
+    mem_limit: 512m   # 15d TSDB; scrape targets grow as exporters are added
     restart: unless-stopped
     command:
       - --config.file=/etc/prometheus/prometheus.yml
@@ -1637,6 +1657,11 @@ services:
   gatus:
     image: egovio/gatus:latest
     container_name: digit-gatus
+    # Memory ceiling (#1612). Observability is overhead, not workload:
+    # uncapped it competes with ~12.8 GB of JVM heap on 16 GB boxes that
+    # have no swap, and the OOM killer takes the biggest process — a JVM
+    # service — not the monitoring that misbehaved.
+    mem_limit: 64m   # matches the 64Mi already set on the k3s Deployment
     depends_on:
       pgbouncer:
         condition: service_healthy
@@ -1666,6 +1691,11 @@ services:
   loki:
     image: grafana/loki:3.4.2
     container_name: digit-loki
+    # Memory ceiling (#1612). Observability is overhead, not workload:
+    # uncapped it competes with ~12.8 GB of JVM heap on 16 GB boxes that
+    # have no swap, and the OOM killer takes the biggest process — a JVM
+    # service — not the monitoring that misbehaved.
+    mem_limit: 512m   # largest real-world grower: chunks plus query working set
     restart: unless-stopped
     command: ["-config.file=/etc/loki/config.yaml"]
     volumes:
@@ -1689,6 +1719,11 @@ services:
   promtail:
     image: grafana/promtail:3.4.2
     container_name: digit-promtail
+    # Memory ceiling (#1612). Observability is overhead, not workload:
+    # uncapped it competes with ~12.8 GB of JVM heap on 16 GB boxes that
+    # have no swap, and the OOM killer takes the biggest process — a JVM
+    # service — not the monitoring that misbehaved.
+    mem_limit: 128m   # log shipper: tails files, small buffers
     restart: unless-stopped
     command: ["-config.file=/etc/promtail/config.yaml"]
     volumes:
--- a/local-setup/docker-compose.egov-digit.yaml
+++ b/local-setup/docker-compose.egov-digit.yaml
@@ -9,11 +9,6 @@ services:
   otel-collector:
     image: egovio/opentelemetry-collector-contrib:0.114.0
     container_name: digit-otel-collector
-    # Memory ceiling (#1612). Observability is overhead, not workload:
-    # uncapped it competes with ~12.8 GB of JVM heap on 16 GB boxes that
-    # have no swap, and the OOM killer takes the biggest process — a JVM
-    # service — not the monitoring that misbehaved.
-    mem_limit: 320m   # internal memory_limiter is 200 MiB; this is the hard backstop above it
     restart: unless-stopped
     command: ["--config=/etc/otelcol-contrib/config.yaml"]
     volumes:
@@ -22,17 +17,28 @@ services:
       - "14317:4317"
       - "14318:4318"
       - "13133:13133"
+    # Memory ceiling (#1612). Observability is overhead, not workload:
+    # uncapped it competes with ~12.8 GB of JVM heap on 16 GB boxes that
+    # have no swap, and the OOM killer takes the biggest process — a JVM
+    # service — not the monitoring that misbehaved.
+    #
+    # Written as deploy.resources.limits.memory rather than the legacy
+    # top-level mem_limit key, because that is already the convention here:
+    # egov-otp, user-otp and egov-notification-sms use it in this file, and
+    # 28 of docker-compose.yml's services do. Compose V2 applies it outside
+    # Swarm exactly as it applies mem_limit — egov-otp's cap was raised
+    # 256M→512M precisely because it was taking effect — so this is one
+    # convention rather than a second one for the same job.
+    deploy:
+      resources:
+        limits:
+          memory: 320M   # internal memory_limiter is 200 MiB; this is the hard backstop above it
     networks:
       - egov-network
 
   tempo:
     image: egovio/tempo:2.6.1
     container_name: digit-tempo
-    # Memory ceiling (#1612). Observability is overhead, not workload:
-    # uncapped it competes with ~12.8 GB of JVM heap on 16 GB boxes that
-    # have no swap, and the OOM killer takes the biggest process — a JVM
-    # service — not the monitoring that misbehaved.
-    mem_limit: 384m   # 24h retention, small blocks
     restart: unless-stopped
     command: ["-config.file=/etc/tempo/config.yaml"]
     volumes:
@@ -45,17 +51,18 @@ services:
       interval: 10s
       timeout: 5s
       retries: 5
+    # Memory ceiling (#1612) — see the note on otel-collector at the top of
+    # docker-compose.egov-digit.yaml.
+    deploy:
+      resources:
+        limits:
+          memory: 384M   # 24h retention, small blocks
     networks:
       - egov-network
 
   grafana:
     image: egovio/grafana:11.4.0
     container_name: digit-grafana
-    # Memory ceiling (#1612). Observability is overhead, not workload:
-    # uncapped it competes with ~12.8 GB of JVM heap on 16 GB boxes that
-    # have no swap, and the OOM killer takes the biggest process — a JVM
-    # service — not the monitoring that misbehaved.
-    mem_limit: 256m   # query proxy + dashboard rendering
     restart: unless-stopped
     environment:
       GF_AUTH_ANONYMOUS_ENABLED: "true"
@@ -83,6 +90,12 @@ services:
         condition: service_healthy
       prometheus:
         condition: service_started
+    # Memory ceiling (#1612) — see the note on otel-collector at the top of
+    # docker-compose.egov-digit.yaml.
+    deploy:
+      resources:
+        limits:
+          memory: 256M   # query proxy + dashboard rendering
     networks:
       - egov-network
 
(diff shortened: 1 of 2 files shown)


local-setup/docker-compose.egov-digit.yaml
```diff
@@ -1666,6 +1691,11 @@ services:
   loki:
     image: grafana/loki:3.4.2
     container_name: digit-loki
+    # Memory ceiling (#1612). Observability is overhead, not workload:
+    # uncapped it competes with ~12.8 GB of JVM heap on 16 GB boxes that
+    # have no swap, and the OOM killer takes the biggest process — a JVM
+    # service — not the monitoring that misbehaved.
+    mem_limit: 512m   # largest real-world grower: chunks plus query working set
```
vinothrallapalli-eGov: The PR body itself flags Loki as "the most likely to need raising" because usage is query-driven. But `otel/loki-config.yaml`'s `limits_config` only sets `max_query_length: 721h` — there's no `max_query_series`, `max_entries_limit_per_query`, or querier concurrency cap to bound a single expensive query's memory footprint before it hits this 512m hard ceiling. Without that, an oversized query still gets the container hard OOM-killed (and restart-looped) rather than cleanly rejected — worth adding as a fast follow, since Loki can lose not-yet-flushed in-memory chunks on an OOM-kill, which is a small but real cost beyond "dashboard goes dark for a bit."

local-setup/docker-compose.egov-digit.yaml
```diff
@@ -77,6 +92,11 @@ services:
   prometheus:
     image: prom/prometheus:v2.55.1
     container_name: digit-prometheus
+    # Memory ceiling (#1612). Observability is overhead, not workload:
+    # uncapped it competes with ~12.8 GB of JVM heap on 16 GB boxes that
+    # have no swap, and the OOM killer takes the biggest process — a JVM
+    # service — not the monitoring that misbehaved.
+    mem_limit: 512m   # 15d TSDB; scrape targets grow as exporters are added
```
vinothrallapalli-eGov: 512m for Prometheus is plausible for today's 15d-retention/current-cardinality snapshot, but the PR's own "Why" column already flags scrape targets growing (#1614/#1615/#1623). No `--query.max-samples` / `--query.max-concurrency` override here either, though Prometheus's built-in defaults for those are more conservative than Loki's, so this is lower risk than the Loki case. Given the tracking issue (#1601) explicitly calls this out as a single measured snapshot rather than validated over time, it'd be worth wiring a Gatus/alerting check for OOM-kills on this container specifically during the one-week soak mentioned in the PR description, since Prometheus is the one most likely to grow past its cap as more exporters land.

local-setup/docker-compose.egov-digit.yaml
```diff
@@ -9,6 +9,11 @@ services:
   otel-collector:
     image: egovio/opentelemetry-collector-contrib:0.114.0
     container_name: digit-otel-collector
+    # Memory ceiling (#1612). Observability is overhead, not workload:
+    # uncapped it competes with ~12.8 GB of JVM heap on 16 GB boxes that
+    # have no swap, and the OOM killer takes the biggest process — a JVM
+    # service — not the monitoring that misbehaved.
+    mem_limit: 320m   # internal memory_limiter is 200 MiB; this is the hard backstop above it
```
vinothrallapalli-eGov: Confirmed: `otel/otel-collector-config.yaml` line 15 does set `memory_limiter.limit_mib: 200`, so the "backstop above the internal limiter" comment is accurate — this is the best-designed of the seven caps, since the collector can shed load internally before ever touching the hard 320m ceiling. Minor style note: this repo's `docker-compose.yml` already establishes `deploy.resources.limits.memory` as the convention for capping container memory (28 of 34 services). This PR introduces the legacy top-level `mem_limit:` key instead for `docker-compose.egov-digit.yaml`/`docker-compose.monitoring.yml`. Both work fine with the Compose V2 CLI outside Swarm, so this isn't a functional issue — just worth deciding if the codebase wants one convention or is fine with per-file variance.

local-setup/docker-compose.monitoring.yml
```diff
@@ -22,6 +22,10 @@ services:
     # prom/prometheus) — not the registry.preview mirror.
     image: prom/node-exporter:v1.8.2
     container_name: digit-node-exporter
+    # Memory ceiling (#1612) — see the note on the observability services in
+    # docker-compose.egov-digit.yaml. node-exporter is a thin reader of /proc
+    # and /sys; 64m is generous for it.
+    mem_limit: 64m
```
vinothrallapalli-eGov: Heads up for sequencing: PR #1666 ("make the observability stack modular") adds a `profiles:` key at this exact same anchor point (immediately after `container_name:`) on all 7 services this PR touches in `docker-compose.egov-digit.yaml` plus this `node-exporter` block here. I verified with a three-way `git merge-file` against the common `monitoring-fix` base that this produces real conflicts on all 7 blocks — and that's not hypothetical: the actual `integration/all-monitoring` merge commit (d935fdfd5) lists both these files under `# Conflicts:` in its message. It resolved cleanly there (both `mem_limit` and `profiles` kept side by side), so this is informational rather than a blocker — just flagging that whichever of #1635/#1666 lands second in `monitoring-fix` will need a manual (if trivial) conflict resolution pass rather than a clean auto-merge.

vinothrallapalli-eGov (commented):
Verified the core claims: otel-collector's internal `memory_limiter` really is 200 MiB (otel/otel-collector-config.yaml:15), so its 320m cap is a genuine backstop above a soft limiter rather than a bare hard cap. The other six containers (tempo, grafana, prometheus, gatus, loki, promtail, node-exporter) have no internal soft limit or `mem_reservation` — a hard `mem_limit` only — but that matches this repo's existing convention (docker-compose.yml also caps memory with no reservations anywhere), so it's not a new gap. The values themselves look reasonable for the stated workload, with one real risk: Loki's `limits_config` (otel/loki-config.yaml) only sets `max_query_length`, nothing bounding per-query memory (`max_query_series`, `max_entries_limit_per_query`, concurrency) — so a wide-range query, which the PR itself flags as the likely trigger, can still balloon past 512m and get hard OOM-killed with no softer failure mode in between, risking loss of not-yet-flushed chunks rather than a clean query rejection. Separately, I confirmed by three-way `git merge-file` (and cross-checked against the real `integration/all-monitoring` merge commit d935fdfd5, whose message literally lists `local-setup/docker-compose.egov-digit.yaml` and `local-setup/docker-compose.monitoring.yml` under `# Conflicts:`) that this PR and #1666 both insert immediately after `container_name:` on the same 7 services, producing real merge conflicts in both files. They resolved cleanly there (both lines kept), so this is low-severity/informational, but whichever of the two merges second into `monitoring-fix` should expect manual conflict resolution on every shared service block.

local-setup/docker-compose.egov-digit.yaml
singhalkarun: Confirmed the gap. `otel/loki-config.yaml`'s `limits_config` is exactly two keys:

```yaml
limits_config:
  retention_period: 72h
  max_query_length: 721h
```

No `max_query_series`, no `max_entries_limit_per_query`, no querier concurrency cap. So there is nothing bounding an expensive query below the 512M cgroup ceiling, and your framing of the failure shape is the right one: a *rejected* query is a 400 to one person, an OOM-kill is a restart that also drops whatever chunks had not been flushed yet.

Not taking it in this PR, deliberately. This one only adds ceilings, and those limit values are the kind of number that should come from the same measured basis the caps came from rather than from taste — a `max_query_series` picked blind risks rejecting legitimate dashboard queries, which is a worse day-one outcome than the OOM it is meant to prevent. Recorded on #1601 with the specific keys so it does not evaporate with this thread.

local-setup/docker-compose.egov-digit.yaml
singhalkarun: Agreed on the risk ranking, and the "single measured snapshot" caveat is exactly right — Prometheus came in under 40% of 512M, but that is one reading against today's **two** scrape jobs (`otel-collector` and `node`), and #1614/#1615/#1623 each add more.

One thing worth knowing before someone tries to build the OOM-kill check, because it is not as close to hand as it looks: there is no cAdvisor in this stack and `otel/prometheus.yml` scrapes only those two jobs, so there is **no per-container memory series to alert on at all** right now. The nearest thing that already exists is node-exporter's `node_vmstat_oom_kill` — host-wide rather than per-container, but it counts cgroup OOM kills as well as global ones on modern kernels, so `increase(node_vmstat_oom_kill[10m]) > 0` would at least turn "one of these caps was hit" from invisible into visible. (Worth confirming on a real box that a container hitting its cgroup limit does move that counter there — I have not run it.)

Also: that belongs in #1611's rules file rather than as a Gatus endpoint. Gatus probes HTTP, and a Prometheus that OOM-restarts is usually answering again by the time the next probe lands — this specifically needs a counter, not a health check, which is the same reason #1611 exists alongside #1609.

Logged on #1601 rather than expanded here.

local-setup/docker-compose.egov-digit.yaml
singhalkarun: Taken — all eight switched to `deploy.resources.limits.memory` in b9fd37d8.

It is a bit stronger than "per-file variance", as it turns out: the variance was **inside one file**. `docker-compose.egov-digit.yaml` already caps three services this way — `egov-otp` (L1323), `user-otp` (L1367), `egov-notification-sms` (L1402) — so this PR was adding a second spelling next to an existing one rather than establishing a first for that file. Those three are also the evidence the key takes effect outside Swarm: `egov-otp`'s cap was raised 256M→512M precisely *because* 256M was OOM-killing it.

Verified after the change with `docker compose config` — exit 0, no warnings, all eight resolving to the intended byte values:

```
gatus 64 MiB · grafana 256 · loki 512 · node-exporter 64
otel-collector 320 · prometheus 512 · promtail 128 · tempo 384
```

Numbers, units and the per-service sizing rationale are unchanged. Two side effects, both welcome: the blocks now sit before `networks:` where the file's existing `deploy:` blocks sit, which is what resolves your sequencing comment below; and the four-line rationale that was pasted verbatim seven times is now stated once on `otel-collector` and referenced from the rest.

local-setup/docker-compose.monitoring.yml
singhalkarun: Your merge analysis was right, and the convention fix above happens to dissolve the problem: moving the caps out of the `container_name:` anchor into `deploy:` blocks before `networks:` vacates the exact line #1666 inserts `profiles:` at.

Re-ran your test from `monitoring-fix`, merging #1666 first and then this branch:

| | result |
|---|---|
| before (7ffe28e9) | `CONFLICT` in `docker-compose.egov-digit.yaml` **and** `docker-compose.monitoring.yml` |
| after (b9fd37d8) | Auto-merging both — **clean** |

And the merged tree keeps both changes side by side, which was the thing actually worth protecting:

```
otel-collector  mem=320M  profiles=['obs-metrics']
tempo           mem=384M  profiles=['obs-traces']
grafana         mem=256M  profiles=['obs-metrics']
prometheus      mem=512M  profiles=['obs-metrics']
gatus           mem=64M   profiles=None
loki            mem=512M  profiles=['obs-logs']
promtail        mem=128M  profiles=['obs-logs']
node-exporter   mem=64M   profiles=['obs-metrics']
```

So whichever of the two lands second no longer needs the manual pass. (#1687 subsequently un-profiles `otel-collector` for an unrelated reason — that changes the profiles column above, not the merge result.)

vinothrallapalli-eGov (approved):

PR #1636, opened 2026-08-06
feat(gatus): watch the monitoring stack itself

### LINKED ISSUE #1613: Nothing monitors the monitoring — observability failures are silent
Part of #1601 (monitoring stack umbrella). **P1 · effort: S** (rises in importance once #1609 and #1611 land)

## Problem

If the observability stack stops, nothing tells us. Three gaps combine:

### 1. Gatus deliberately skips it

`.github/scripts/check-gatus-coverage.py` — the script that fails CI when a compose service has no health check — exempts the whole stack:

```python
"grafana": "observability plumbing: dashboards UI, not a serving dependency",
"prometheus": "observability plumbing: metrics store, not a serving dependency",
"loki": "observability plumbing: log store, not a serving dependency",
"tempo": "observability plumbing: trace store, not a serving dependency",
"otel-collector": "observability plumbing: telemetry pipeline, not a serving dependency",
```

The reasoning is sound as far as it goes — a dead Grafana does not stop a citizen filing a complaint. But the comment above that block records that the question was **parked, not settled**:

> Monitoring the monitoring is a separate decision -- if it is ever wanted, they need a GATUS_OBSERVABILITY toggle, since none of them exist in the k3s tier or in docker-compose.yml

### 2. The one health gate that exists is advisory

`local-setup/ansible/playbook-deploy.yml` waits for Loki with `ignore_errors: yes`, prints `Loki: UNHEALTHY (non-critical)`, and the deploy proceeds to success. There is **no gate at all** for Grafana, Prometheus, Tempo, otel-collector or Promtail.

### 3. The docs claim otherwise

`local-setup/ansible/README.md:656` lists the deploy's health gates as:

> **Health gates** — wait for kong / persister / hrms / ui / mcp / loki / grafana

Grafana has no gate, and Loki's does not gate anything.

## Why this gets worse after the alerting work

Once #1609 (Slack alerts) and #1611 (capacity rules) land, **silence starts to mean "all clear"**. But silence looks identical if Loki has crashed and Grafana has stopped evaluating rules. That is strictly worse than today: right now nobody relies on the alerting, so nobody is misled by its absence.

## Fix

1. **New `GATUS_OBSERVABILITY` endpoint group**, in both tier files, each entry gated `enabled: "${GATUS_OBSERVABILITY:-false}"`:

   | Component | Check |
   |---|---|
   | Grafana | `http://grafana:3000/api/health` |
   | Prometheus | `http://prometheus:9090/-/healthy` |
   | Loki | `http://loki:3100/ready` |
   | Tempo | `http://tempo:3200/ready` |
   | otel-collector | `http://otel-collector:13133` |

   `digit.env.j2` renders `GATUS_OBSERVABILITY=true` (the Ansible tier always deploys these); the compose default stays `false` so `docker-compose.yml` — which has none of them — does not go red; the k3s Deployment hard-sets `"false"` for the same reason. Exactly the existing `GATUS_PROFILE_*` pattern.

2. **Remove those five from `EXEMPT`** in `check-gatus-coverage.py`, so coverage CI keeps them honest from here on. `promtail` and `node-exporter` stay exempt — they expose no HTTP port to check.

3. **Fix the Ansible gates**: give Grafana, Prometheus, Tempo and otel-collector the same wait Loki has, and correct `ansible/README.md:656` to describe reality.

## Design decision to confirm — should a dead Grafana fail the deploy?

**Proposed: no.** These are not serving dependencies, and failing a tenant deploy over a dashboard is disproportionate. Instead:

- the deploy **warns loudly** and names the component (Loki keeps its current non-fatal behaviour, the others match it)
- **Gatus alerts** on it via #1609, which is the path that actually reaches a human

This keeps deploys robust while closing the "silently dead" gap. Note this differs from the reset-password gate in #1605, which *is* fatal — there the failure silently breaks access to a thing this work just secured.

## Acceptance

- [ ] `docker stop digit-loki` on a test box → Gatus shows it red within 60s → Slack alert (once #1609 lands)
- [ ] The same for grafana / prometheus / tempo / otel-collector
- [ ] `docker-compose.yml` (local) and the k3s tier do **not** show red observability checks
- [ ] `gatus-coverage` CI passes with the five removed from `EXEMPT`
- [ ] Deploy output names any unhealthy observability component clearly
- [ ] `ansible/README.md` health-gate list matches what the playbook actually does

## Relates to

- #1609 — the alerting path these checks would ride on. Low value until that lands.
- #1610 — same tier-synced `gatus/config.yaml`; sequence the three.

singhalkarun:
Fixes #1613

## What

If the observability stack stopped, nothing said so. Three gaps combined:

1. **Gatus exempted all five components.** `check-gatus-coverage.py` listed grafana/prometheus/loki/tempo/otel-collector as "observability plumbing… not a serving dependency".
2. **The only Ansible gate was advisory.** Loki's wait was `ignore_errors: yes` and printed one easily-missed line; grafana, prometheus, tempo and the collector had no gate at all.
3. **The docs claimed otherwise.** `ansible/README.md:653` listed gates for "loki / grafana" — grafana had none, and Loki's gated nothing.

This gets worse the moment alerting lands (#1609). From then on **silence means "all clear"** — and silence looks identical if Loki has crashed and Grafana has stopped evaluating rules. That is strictly worse than today, when nobody relies on the alerting yet.

## Change

**Observability endpoint group** in both tier catalogues, behind `GATUS_OBSERVABILITY`:

| Component | Check |
|---|---|
| Grafana | `/api/health` |
| Prometheus | `/-/healthy` |
| Loki | `/ready` |
| Tempo | `/ready` |
| OTEL Collector | `:13133` |

Ansible renders the toggle **true** (that tier always deploys the stack). The four non-tenant compose paths and the k3s Deployment hard-set **false** — none of them runs these, and Gatus reads an **unset** enabled-var as `true`, which would make them red forever.

**Exemptions removed** from `check-gatus-coverage.py`, closing the TODO that comment had parked:

> Monitoring the monitoring is a separate decision -- if it is ever wanted, they need a GATUS_OBSERVABILITY toggle

Two stay exempt, with sharper reasons: **promtail** has no listener to probe, and **node-exporter** is scraped directly by Prometheus — a dead exporter surfaces as a stale `node` target, not a missing check.

**Ansible gates** now cover all five and name each one in the output.

## The design decision worth reviewing: still non-fatal

The gate deliberately stays `ignore_errors`. Failing a tenant deploy because a dashboard is down is disproportionate — these are not serving dependencies. What was wrong before was **silence**, not the absence of a hard failure.

So: the deploy warns loudly and names the component, and the path that actually reaches a human is Gatus → Slack (#1609). This differs from the fatal reset in #1605, where a silent failure would have broken access to something that PR had just secured.

## Verification

- [x] Both tier catalogues carry an identical Observability group
- [x] `GATUS_OBSERVABILITY` declared in all five compose files and the k8s Deployment
- [x] `check-gatus-coverage.py` passes with self-test: **56 endpoints per tier** (was 51), no drift, no dangling — the five removed exemptions are covered by real checks rather than becoming "unmonitored" failures
- [x] Playbook parses; the status loop uses the plain `item.item.name` form (an earlier `zip`-filter version would have templated fine in lint and failed at runtime)
- [ ] `docker stop digit-loki` → Gatus red within 60s → Slack alert (once #1609 lands)
- [ ] Same for grafana / prometheus / tempo / otel-collector
- [ ] `docker-compose.yml` and the k3s tier show no red observability checks
- [ ] Deploy output names any unhealthy component

The last four need a running box; not yet deployed.

## Merge note

Touches the tier-synced `gatus/config.yaml` pair, as do #1630 (sqlite) and #1632 (TLS check). Third in that queue — sequence them and keep both tiers in step.


.github/scripts/check-gatus-coverage.py
```diff
@@ -125,19 +125,14 @@
...
-    "otel-collector": "observability plumbing: telemetry pipeline, not a serving dependency",
-    "node-exporter": "observability plumbing: host-metrics exporter (#1335), not a serving dependency; scraped by prometheus, absent from k3s tier and docker-compose.yml",
+    # Observability plumbing that genuinely cannot be probed. The five components
+    # that CAN be (grafana/prometheus/loki/tempo/otel-collector) are no longer
+    # exempt: #1613 added them behind a GATUS_OBSERVABILITY toggle, which is the
+    # decision this comment used to defer. They are defined in
+    # docker-compose.egov-digit.yaml (plus grafana/tempo/otel-collector in
+    # registry.yml) and absent from docker-compose.yml and the k3s tier, so the
+    # toggle is what keeps them from being red where they do not exist.
+    "node-exporter": "observability plumbing: host-metrics exporter (#1335); no check because prometheus scrapes it directly -- a dead exporter surfaces as a stale `node` target, not a missing endpoint. Absent from k3s tier and docker-compose.yml",
```
vinothrallapalli-eGov, line 135: Confirmed: this diff does not add a `postgres-exporter` exemption. Per the #1601 tracking issue's own merge-order notes, #1629 (postgres-exporter) is expected to merge before this one, and the integration pass explicitly forecast/observed `#1629 ↔ #1636 — postgres-exporter fails the coverage guard` with the recommended resolution being "add a postgres-exporter exemption mirroring node-exporter's" *in this PR*. Right now that entry isn't here, so whichever of the two lands second will fail `check-gatus-coverage.py` (or a merge will silently drop one side's exemption). Suggest adding a `"postgres-exporter": "..."` entry alongside `node-exporter` here so the guard survives either merge order.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -4133,18 +4133,42 @@
       when: enable_mcp | default(false)
 
     # ==================== Wait for Loki ====================
```
vinothrallapalli-eGov: Minor: the old `# ==================== Wait for Loki ====================` banner is left in place directly above the new `# ==================== Observability stack health (#1613) ====================` banner — looks like a leftover from the rename that should be deleted rather than stacked.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -4133,18 +4133,42 @@
...
+    # all, its failure printed a single easily-missed line, and ansible/README.md
+    # claimed gates for "loki / grafana" that did not exist.
+    #
+    # The path that actually reaches a human is Gatus -> Slack (#1609/#1613);
+    # this loop exists so the deploy operator sees it immediately, and names
+    # every component rather than only the one that happened to be checked.
+    - name: Wait for observability stack health (up to 2 min each)
+      shell: "curl -sf {{ item.url }} -o /dev/null"
+      register: observability_health
       retries: 12
```
vinothrallapalli-eGov: Minor/worth noting rather than blocking: `retries: 12` / `delay: 10` combined with `loop` over 5 items means each component retries independently (confirmed by the per-item `item.rc` read in the debug loop below, which is correct) — but that also means a fully-down observability stack now stretches this gate to up to 10 minutes sequentially (5 × 2 min), versus 2 minutes for the single Loki check it replaces. Given it's explicitly non-fatal, that's a deploy-time cost more than a correctness bug, but if the stack is commonly slow to come up (e.g. Grafana provisioning) this could meaningfully lengthen every deploy's tail. A shorter per-item timeout or `async`/parallel dispatch would keep the same signal for less wall-clock time.

vinothrallapalli-eGov (commented):
Reviewed the diff directly (not the integration branch). The five new Gatus checks are genuinely new coverage (nothing else in config.yaml probes Grafana/Prometheus/Loki/Tempo/OTEL), use the existing 60s/[STATUS]==200 idiom consistently across both tiers, and the toggle wiring (`GATUS_OBSERVABILITY` unconditional-true on the one compose path that runs the stack, hard `false` everywhere else, including the k3s Deployment) is internally consistent and matches host ports. Spot-checked both gatus config files for the reported merge-conflict-resolution hazard (`=======` inside a `# ==== banner ====` comment getting treated as a real conflict marker) — found no mangled or joined comment lines in this PR's actual diff; both files parse/read cleanly. One confirmed gap and two minor nits below.

.github/scripts/check-gatus-coverage.py
singhalkarun: Confirmed and added — 659ac7d9.

I also verified the merge mechanism rather than assuming it, and the result is better than just "entry added". Merging `origin/feat/1615-postgres-exporter` into this branch **does** conflict in `check-gatus-coverage.py`, exactly as the integration pass forecast: this PR deletes the five observability lines above `node-exporter` and rewrites `node-exporter` itself, while #1629 inserts immediately after it.

The useful part is what happens next. Because both sides now carry a byte-identical `postgres-exporter` line, git resolves it **outside** the conflict hunk — it lands as common context below the markers. So the exemption survives whichever way someone resolves the conflict, including a careless "take ours". That is precisely the failure mode you named ("a merge will silently drop one side's exemption"), and it is now closed structurally instead of relying on the merger noticing.

Proven end to end on a throwaway merge:

- #1629 merged, conflict resolved toward this branch → `OK: 89 compose services, 23 k3s Services, 56 endpoints per tier, no drift, no dangling checks`
- same tree, `postgres-exporter` line deleted → `FAIL: compose services with no Gatus check: postgres-exporter`

So the entry is load-bearing, not decorative. (#1629 defines the service in `docker-compose.monitoring.yml`, which is already inside the guard's perimeter.)

Landing it ahead of #1629 is safe: `EXEMPT` is only ever consulted via `name in EXEMPT` for services actually found in a compose file, so a key for a service that does not exist yet is inert. Nothing to follow up once #1629 arrives. Guard and its `--self-test` both pass on this branch as it stands.

local-setup/ansible/playbook-deploy.yml
singhalkarun: Confirmed — leftover from the rename. The diff shows the old `# ==================== Wait for Loki ====================` line surviving as unchanged context directly above the new banner, so both were being rendered.

Removed in 659ac7d9.

local-setup/ansible/playbook-deploy.yml
singhalkarun: Confirmed, including the per-item reading of the retry semantics: `until` retries are sequential *per loop item*, so 12 × 10s across five components is up to 10 minutes, against 2 minutes for the single Loki check it replaces. The old budget cannot simply be reused five times over.

Cut to `retries: 6` / `delay: 5` — 30s per component, 2.5 min worst case for all five — in 659ac7d9.

**Why 30s is sufficient here is placement, not optimism.** This gate is a readiness re-check, not a startup wait. `docker compose up -d` runs at roughly line 1744 of the play; this task sits at roughly line 4146, and the tasks in between include the keycloak bootstrap, which alone allows 5 minutes just to bind. The stack has had many minutes to settle before this ever runs. A component still silent at this point is down rather than slow — and since the gate is explicitly non-fatal, the cost of being wrong is one inaccurate line, weighed against 7.5 minutes on the tail of every deploy where the stack is down.

**On async/parallel, which was your other suggestion:** I went with the shorter budget instead, deliberately. Async dispatch would preserve the full per-component window and is the better answer in principle, but it introduces a new construct into the deploy path, and I have not deployed this — the devbox has other agents on it. Between a change that can only be *slightly* wrong (a component that needed 45s gets reported UNHEALTHY, non-fatally) and one that could fail in a new way in the deploy path unverified, I took the first. If someone is deploying this anyway, async is a reasonable follow-up.

Flagging honestly: the timing behaviour is unexercised outside a real run. Statically I have only confirmed the playbook parses.

**One adjacent thing this thread surfaced, not part of your comment:** this task carries no `observability_level` gate, so a metrics-only tenant will probe Loki and Tempo and duly report them UNHEALTHY — the same class of false alarm #1666 hit. #1687 adds that gating. Worth checking the two compose cleanly, since both touch this region.

singhalkarun:
## k3s tier: all five observability checks fail on DNS — exactly as #1682 predicted

#1682 flagged this in its "deliberately not done" section:

> the endpoints will need namespaced DNS — Gatus runs in `digit`, this stack in `monitoring`, so `http://grafana:3000` does not resolve; it needs `grafana.monitoring.svc.cluster.local`

Confirmed on a real cluster. Gatus is `namespace: digit` (line 28), and the checks use bare service names:

```yaml
url: "http://grafana:3000/api/health"
url: "http://prometheus:9090/-/healthy"
url: "http://loki:3100/ready"
url: "http://tempo:3200/ready"
url: "http://otel-collector:13133"
```

With `GATUS_OBSERVABILITY=true` and the monitoring stack running in `monitoring`, every one of them fails:

```
Grafana         success=False  lookup grafana on 10.96.0.10:53: ...
Loki            success=False  lookup loki ...
OTEL Collector  success=False  lookup otel-collector ...
Prometheus      success=False  lookup prometheus ...
Tempo           success=False  lookup tempo ...
```

These are **DNS failures, not connection failures** — the name never resolves, because a bare name resolves within the querying pod's own namespace (`grafana.digit.svc.cluster.local`).

The counterfactual, from a pod in `digit`, against a Grafana that is genuinely up and healthy in `monitoring`:

```
http://grafana:3000/api/health             -> DNS-FAIL
http://grafana.monitoring:3000/api/health  -> 200
```

So the fix is precisely the namespace qualifier.

**Severity is latent, not live:** `GATUS_OBSERVABILITY` defaults to `false`, so nothing is red today. But the feature this PR adds does not work the moment anyone turns it on — and it would turn on as five permanently-red checks, which is the worst kind of introduction for a monitoring feature.

Two things worth deciding alongside the fix: the namespace should probably not be hardcoded either (`grafana.monitoring` breaks if the stack moves), and the compose tier's service names differ, so the two catalogues can no longer be literally identical here — worth a comment saying why.

singhalkarun:

`447f89fe` — the five Observability URLs are now namespace-qualified via `GATUS_OBSERVABILITY_NS` (default `monitoring`, set non-empty on the Deployment for the reason #1622 documents).

Re-verified with `GATUS_OBSERVABILITY=true` against a real monitoring stack:

- **Grafana flipped `DNS-FAIL` → `success=True`**
- Prometheus went green too, once a service carrying the k3s-tier name existed in `monitoring`

Loki, Tempo and the OTEL collector still fail in my cluster, but for an unrelated reason: it runs the *helm* tier, whose services are named `kube-prometheus-stack-prometheus`, `otel-collector-opentelemetry-collector`, and it has no Loki or Tempo at all. The two that share a name between the tiers both resolve.

vinothrallapalli-eGov (approved):

PR #1637, opened 2026-08-06
chore: delete the root-level duplicate observability configs

### LINKED ISSUE #1617: Root-level otel/ copy has drifted and its promtail config is fatally broken

singhalkarun:
Closes #1617 (part 2 — part 1 is #1624)

## What

The repo carried **two copies** of the observability config: the deployed one under `local-setup/`, and an older one at the repo root. Nothing referenced the root copies, and they had drifted far enough to be actively dangerous — deploying from the repo root produced a stack with **silently dead log collection** (the fatal promtail config, defused separately in #1624).

Deletes 15 tracked files: `otel/`, `docker-compose.egov-digit.yaml`, `docker-compose.monitoring.yml`.

## Verified unreferenced, not assumed

| Suspected consumer | Result |
|---|---|
| Ansible config sync (`:1206`) | `otel: ../otel` from `playbook_dir` → **`local-setup/otel`** |
| Ansible javaagent download (`:1165`) | `{{ playbook_dir }}/../otel/download-agent.sh` → **`local-setup/otel`** |
| `check-gatus-coverage.py` | enumerates `LS / "..."` — **`local-setup/` only** |
| CI workflow `paths:` filters | **none** name the root copies |
| `performance/ansible/playbook-setup.yml` | looked like a consumer — it syncs `otel` and `gatus`. It is **not**: `src: ../{{ item }}/` resolves to `performance/otel` and `performance/gatus`, **neither of which exists** |
| Remaining doc mentions | bare filenames used from inside `local-setup/` or `/opt/digit`, not root-relative paths |

## Verified nothing unique was lost

Every deleted file has a `local-setup/` counterpart. Ten were byte-identical; three differed (`promtail-config.yaml`, `loki-config.yaml`, `docker-compose.egov-digit.yaml`), and in each case the root side was the **stale** one.

One finding worth recording: the root compose was an older **generation**, not merely a drifted copy — **46 services against 60**, including five legacy seed containers absent from `local-setup/docker-compose.egov-digit.yaml`:

```
db-seed, localization-seed, mdms-security-seed, mdms-tenant-seed, mdms-workflow-seed
```

I checked each rather than assuming they were obsolete. All five still exist in `local-setup/docker-compose.db-migrations.yml` and `docker-compose.registry.yml`, so no service definition is lost by this deletion.

## Also

Corrects the `.gitignore` comment, which pointed at the now-removed `otel/download-agent.sh` instead of `local-setup/otel/download-agent.sh`.

## Verification

- [x] `check-gatus-coverage.py` passes with self-test — 88 compose services, no drift, no dangling
- [x] `check-flyway-dump-alignment.py` passes with self-test
- [x] No `otel/` or `docker-compose.*` remains at the repo root
- [x] Every deleted file confirmed to have a `local-setup/` counterpart

## Note on #1624

#1624 repairs the fatal root promtail config; this PR removes that file entirely. They do not conflict (one edits, one deletes) but the resolution differs by merge order — if this lands first, #1624 becomes unnecessary and can be closed; if #1624 lands first, this deletes the file it just fixed. Either outcome is correct. #1624 was deliberately kept separate so the landmine could be defused immediately without waiting for this decision.


vinothrallapalli-eGov (commented):
Verified independently: grepped the whole repo (code search, not just the diff) for every deleted path (`docker-compose.egov-digit.yaml`, `docker-compose.monitoring.yml`, `otel/...`) and found no surviving root-relative consumer. The Ansible sync tasks at `local-setup/ansible/playbook-deploy.yml` resolve `../otel` and `../docker-compose.egov-digit.yaml` from `playbook_dir` (`local-setup/ansible/`) to the `local-setup/` copies, matching the PR body's claim. The only two non-`local-setup` mentions of `docker-compose.egov-digit.yaml` in the tree (`MAC-QUICKSTART.md`, `digit-mcp/src/services/snapshot.ts`) both point at the *deployed* file under `/opt/digit` or `~/digit`, which originates from `local-setup/`, not the repo root — so nothing breaks. I also fetched the post-merge `local-setup/otel/promtail-config.yaml` and confirmed it has no `limits_config` (the fatal key is genuinely absent), and `local-setup/otel/loki-config.yaml` still carries the working `limits_config` this PR's root copy lacked — the 'root was the stale one' claim checks out. The diff itself is a clean deletion plus a correct `.gitignore` comment fix; nothing to flag there.

Two non-blocking process notes for whoever merges this, not defects in the diff: (1) #1624 (patches the file this PR deletes) is still open — worth closing as superseded once this lands, or rebasing it away. (2) #1625 edits this same root compose file and is open/mergeable now (mergeStateStatus CLEAN here), so a real modify/delete conflict will surface depending on merge order; per the tracking issue the resolution is 'take the delete', which is straightforward since #1625's root-file edit is functionally superseded by this deletion.

vinothrallapalli-eGov (approved):

vinothrallapalli-eGov (approved):

vinothrallapalli-eGov (approved):

Issue #1648, opened 2026-08-06
The entire ingress-nginx block in env.yaml is inert: no security headers, no TLS restrictions, no gateway metrics


singhalkarun:
Part of #1619 (Kubernetes monitoring parity). **P0-security · effort: M** *(re-scoped — see Revision)*

## The whole `ingress-nginx:` block in env.yaml is inert

Not just the metrics keys. **All seven keys** the block sets are silently ignored, and the consequences go well beyond monitoring.

`ingress-nginx` is a **vendored upstream chart** (v4.10.0). It does not use the `common` library's chart-name hoisting and reads values directly — 39 of its templates reference `.Values.controller.*`. The release passes the whole env tree (`backboneservices-helmfile.yaml:126-129`), so `env.yaml`'s block lands at `.Values["ingress-nginx"].*` — one level too deep. Verified: **no template anywhere in the chart indexes `.Values` by the chart name**, so there is no hoisting mechanism.

### What is consequently NOT applied

**Security headers — the chart default is `controller.addHeaders: {}` (empty):**

```yaml
    addHeaders:
      X-Content-Type-Options: "nosniff"
      X-Frame-Options: "SAMEORIGIN"
      X-XSS-Protection: "1; mode=block"
```

None of these reach any ingress. A reviewer reading `env.yaml` would reasonably conclude they are set.

**TLS restrictions — chart default is `controller.config: {}` (empty):**

```yaml
  ssl-protocols: "TLSv1.2 TLSv1.3"
  ssl-ciphers: "EECDH+CHACHA20:EECDH+AES"
  ssl-ecdh-curve: "X25519:prime256v1:secp521r1:secp384r1"
```

Not applied either. The controller falls back to its own defaults, which are reasonable — this is a hardening gap, not an open door — but it is not what the file says.

**Gateway metrics** — the original subject of this issue:

| Setting | env.yaml intends | Chart default | Effective |
|---|---|---|---|
| `controller.metrics.enabled` | `false` | **`true`** | **enabled** |
| `controller.metrics.serviceMonitor.enabled` | `true` | **`false`** | **disabled** |

So the metrics port and Service **do** exist; nothing scrapes them. The scrape job that would (`values/prometheus.yaml:3776`) targets `nginx-ingress-metrics.egov:10254`, while the Service is `<release>-ingress-nginx-controller-metrics` — a name that has never resolved.

### Two things it is fortunate are NOT applied

```yaml
[...]

Issue #1645, opened 2026-08-06
Prometheus and Alertmanager run without persistent storage; the env.yaml block that would fix it is mis-pathed


singhalkarun:
Part of #1619 (Kubernetes monitoring parity). **P1 · effort: XS to fix, but see the verification note**

## Problem

Prometheus and Alertmanager have **no persistent storage**, so every metric and all notification state is lost whenever the pod restarts — an upgrade, a node drain, a reschedule.

Someone intended otherwise. `charts/environments/env.yaml:308-321` is unambiguous:

```yaml
prometheus:
  retention: 7d
  storageSpec:
    volumeClaimTemplate:
      spec:
        storageClassName: gp3
        accessModes: ["ReadWriteOnce"]
        resources:
          requests:
            storage: 15Gi
```

**It has no effect.** kube-prometheus-stack reads these at `prometheus.prometheusSpec.retention` and `prometheus.prometheusSpec.storageSpec`. The block above sets them one level shallower, at `prometheus.retention` / `prometheus.storageSpec`.

Verified the strongest way available: the chart (`kube-prometheus-stack` 56.6.2) was pulled and its templates grepped. Only `.Values.prometheus.prometheusSpec.*` is referenced —
- `templates/prometheus/prometheus.yaml:128` → `retention: {{ .Values.prometheus.prometheusSpec.retention | quote }}`
- `templates/prometheus/prometheus.yaml:260-262` → `{{- if .Values.prometheus.prometheusSpec.storageSpec }}`

The shallower paths appear nowhere in the chart.

So the effective values are the **untouched upstream defaults** in the vendored file:
- `charts/monitoring/values/prometheus.yaml:3602` → `retention: 10d`
- `charts/monitoring/values/prometheus.yaml:3707` → `storageSpec: {}` → no volume → emptyDir

Alertmanager is in the same position — `values/prometheus.yaml:877` `retention: 120h`, `:882` `storage: {}` — so silences and notification state are lost on restart too.

There is no fallback: `thanos: {}`, `remoteWrite: []`, `thanosRuler.enabled: false`, and no backup of any kind. Losing the pod loses the history outright.

## Why nothing catches it

**Helm does not warn about values nobody reads.** A key at the wrong path is silently ignored — it is not an error, and there is no output to notice. The mis-pathed block sits in the file looking exactly like working configuration, and every review reads it as "we have 15Gi of persistent metrics".

[...]

Issue #1640, opened 2026-08-06
Alertmanager SMTP credential committed outside SOPS scope; consolidate onto the encrypted config


singhalkarun:
Part of #1601 (monitoring stack umbrella). **P0-security · effort: S**

> The credential itself is understood to have been rotated already, so this is not an incident response. It is the structural fix that stops the same thing recurring — and the value is deliberately not reproduced here.

## Problem

`devops/deploy-as-code/charts/monitoring/values/prometheus.yaml:454` carries an SMTP `auth_password` in cleartext, inside the Alertmanager `email_configs` block, for `unified.alerts@egovernments.org` on `smtp.gmail.com:587`. It is a Google app-password-format string, committed in a **public** repository.

It was never protected, because `charts/.sops.yaml` scopes encryption to exactly one file:

```yaml
creation_rules:
  - path_regex: environments/env\-secrets\.yaml$
```

`charts/monitoring/values/` is outside that pattern, so nothing in the tooling would ever have flagged it. The same applies to `charts/monitoring/values/apm-server.yaml:16`, which holds an Elasticsearch password (in a release that is not installed, but the exposure is identical).

## What makes this more than a stray secret

The repo **already has a correctly-placeholdered Alertmanager config in the right, SOPS-covered file** — `charts/environments/env-secrets.yaml:115-192`, using `<slack_api_url>`, `<smtp_host>`, `<smtp_username>`, `<smtp_password>` and so on.

It is inert. It sits under the top-level `secrets:` key, so the helmfile's `{{ .Values | toYaml }}` injection lands it at `.Values.secrets.alertmanager.config` — and kube-prometheus-stack reads `alertmanager.config`. Grep confirms **no template anywhere under `charts/` references `secrets.alertmanager`**.

So there are two Alertmanager configs: the safe one that nothing reads, and the one in `values/prometheus.yaml` that actually applies and contains the plaintext credential. Consolidating onto the first is both the security fix and a correctness fix.

## Proposed fix

**1. Delete the hardcoded config from `values/prometheus.yaml`.** Remove the `alertmanager.config` block (lines ~399-475) so that file stops being a place secrets can live.

[...]

PR #1652, opened 2026-08-06
ci: validate deploy-as-code values keys and duplicate YAML keys

### LINKED ISSUE #1651: Nothing in CI validates devops/deploy-as-code — the production deploy path
Part of #1619 (Kubernetes monitoring parity) — the root cause behind #1645 and #1648.

## Problem

A pull request touching `devops/deploy-as-code/` — the **production, large-scale deploy path** — triggers **zero** CI checks.

The repo has 19 workflows and they are good ones: they stand the whole local stack up, enforce that every service has a health check, verify the DB dump matches the migrations, catch frontend lockfile drift. Several are custom Python guards written after specific incidents.

Every one of them is pointed somewhere else — `local-setup/`, the frontend, the backend. The single exception is `gateway-whitelist-parity.yml`, which reads exactly two list keys out of `env.yaml`. Nothing examines charts, templates, or any other values.

## What that cost

Two long-lived defects in the production path, both found only by manual review:

- **#1645** — Prometheus and Alertmanager run on `emptyDir`, losing every metric on pod restart. The `env.yaml` block specifying 7d retention and a 15Gi gp3 volume sits one level too shallow (`prometheus.*` rather than `prometheus.prometheusSpec.*`), so the chart never reads it.
- **#1648** — ingress-nginx exposes metrics that nothing scrapes, because its override block is nested under a chart-name key the vendored chart does not read. The chart defaults win, and they are the inverse of the intent.

Both read as correct configuration. Neither produces any output, warning or exit code.

## Why the obvious tools would not have caught either

**`helm template` does not catch mis-pathed values.** Helm merges values into a plain map and silently ignores keys nothing references. It fails only on template syntax errors, explicit `required`/`fail` calls, or a `values.schema.json` violation — and these charts ship no schema. This is precisely why both defects survived.

Also worth recording, since it sounds like an explanation and is not: `disableValidation: true` on the kube-prometheus-stack release is **unrelated**. That flag controls helmfile's manifest-vs-cluster/CRD validation (it is used the same way on cert-manager). Removing it would surface nothing here.

## What this adds

**1. `check-helm-values-paths.py`** — compares the key paths `env.yaml` sets against the paths the target chart's `values.yaml` defines, failing on anything the chart cannot read.

Scoped deliberately to **external/vendored charts**, whose `values.yaml` is a complete reference. DIGIT's own charts are excluded: they read values via `index .Values "key"` without declaring them, so comparing against `values.yaml` there produces false positives — verified against `hrms-dev-mode`, `memory_limits` and `images`, all genuinely consumed but undeclared. Extending coverage to them needs the reference set to include template references, which is a later refinement. A guard that cries wolf gets switched off, which would cost more than it saves.

Ships with a `KNOWN_UNREAD` baseline of the six existing mis-paths, each naming its issue — the same shape as `FROZEN_LATEST_DEBT` in `infra-contracts.test.ts`. So it lands **green** and fails on anything new, rather than being merged red and ignored. The list may only shrink: an entry that stops reproducing fails as a **stale baseline**, so it cannot outlive the bug it was holding.

**2. `yamllint` with `key-duplicates`** over the values files this repo authors. A duplicated key is silently last-wins in PyYAML *and* in Helm, so it is invisible to check 1, to review, and to the deploy — only a linter reading the raw document sees it. Vendored upstream values are excluded; linting third-party defaults is noise.

## Verified before proposing

- self-test passes
- green on `master`, listing the six known mis-paths as outstanding
- injecting a new mis-path (`prometheus.retentionSize`) **fails** it
- simulating a fix for #1645 **fails** it as a stale baseline
- the yamllint step passes today (no duplicates currently)

## Deliberately not included

- **Rendering the charts and asserting on output** — the definitive check for #1645's class, but it needs network access and dummy values for the SOPS-encrypted secrets. Worth doing later; not required to close the gap this issue describes.
- **Making secret scanning blocking.** CodeRabbit already runs gitleaks, but advisory-only and diff-only, so it would never see a plaintext value that has been sitting in a file for a year. Relevant to #1640.
- **Dependabot/Renovate.** Neither exists in this repo at all, so the pinned chart versions (kube-prometheus-stack 56.6.2, grafana 7.3.0, loki-stack 2.10.1) age silently. Separate concern.

## Acceptance

- [ ] A PR touching `devops/deploy-as-code/**` runs checks
- [ ] A new mis-pathed values key fails the build
- [ ] Fixing #1645 or #1648 requires shrinking the baseline, and forgetting to fails the build
- [ ] Duplicate YAML keys fail the build

singhalkarun:
Fixes #1651

## What

A PR touching `devops/deploy-as-code/` — the production, large-scale deploy path — currently triggers **zero** checks. This adds two, both cheap and offline.

## Why these two specifically

They are the checks that would have caught the defects actually found there:

- **#1645** — Prometheus/Alertmanager on `emptyDir`; the `env.yaml` block specifying a 15Gi volume sits one level too shallow to be read.
- **#1648** — ingress-nginx metrics exposed but never scraped; its override block is nested under a key the vendored chart does not read.

Both look correct in review, and **`helm template` would not have caught either** — Helm silently drops values nothing reads, failing only on template errors, `required`/`fail`, or a schema violation, and these charts ship no schema. Purpose-built guard required.

## 1. `check-helm-values-paths.py`

Compares the key paths `env.yaml` sets against the paths the target chart's `values.yaml` defines; fails on anything the chart cannot read.

**Scoped to external/vendored charts on purpose.** Their `values.yaml` is a complete reference. DIGIT's own charts read values via `index .Values "key"` without declaring them, so comparing against `values.yaml` there yields false positives — I verified this against `hrms-dev-mode`, `memory_limits` and `images`, all genuinely consumed but undeclared. Covering them needs the reference set to include template references; that is a later refinement, not a blocker. A guard that cries wolf gets disabled, which costs more than it saves.

**Lands green, via a baseline.** `KNOWN_UNREAD` lists the six existing mis-paths, each naming its issue — same shape as `FROZEN_LATEST_DEBT` in `infra-contracts.test.ts`. The guard therefore fails on anything *new* instead of being merged red and ignored.

The list may only shrink. An entry that stops reproducing fails the check as a **stale baseline** — otherwise the baseline outlives the bug and quietly re-opens the hole it was holding.

## 2. `yamllint` with `key-duplicates`

A duplicated YAML key is silently last-wins in PyYAML *and* in Helm, so it is invisible to check 1, to review, and to the deploy. Only a linter reading the raw document sees it.

Scoped to the values files this repo authors. Vendored upstream chart values are excluded — thousands of lines of third-party defaults, and linting them reports other people's style, which is noise.

## Verified end to end before committing

```
$ check-helm-values-paths.py --self-test
self-test OK: detects mis-paths and orphaned blocks, silent on correct configuration

$ check-helm-values-paths.py                     # on master
OK: 2 external chart(s) checked, no new unread key paths.
6 known mis-path(s) still outstanding: ...        exit 0

# inject a new mis-path (prometheus.retentionSize)
FAIL: env.yaml sets values no chart reads ...     exit 1

# simulate #1645 being fixed, without shrinking the baseline
FAIL: KNOWN_UNREAD lists paths that no longer reproduce ...  exit 1

$ yamllint -d '{rules: {key-duplicates: enable}}' <authored values>
(no findings)                                     exit 0
```

Worth noting the first attempt at the injection test used `prometheus.scrapeInterval`, which the guard correctly did **not** flag — that key genuinely exists at that path. The test was wrong, not the guard; re-run with a truly unread key it fails as intended.

## Deliberately not included

- **Rendering charts and asserting on output** — definitive for #1645's class, but needs network and dummy values for the SOPS-encrypted secrets. Worth a follow-up.
- **Blocking secret scanning** — CodeRabbit runs gitleaks today, but advisory-only and diff-only, so it cannot see a plaintext value that has sat in a file for a year (see #1640).
- **Dependabot/Renovate** — neither exists in this repo, so pinned chart versions age silently. Separate concern.

## Note

The guard is green on `master` today, so this PR does not depend on #1645 or #1648 landing first. When they do land, whoever fixes them must delete the matching `KNOWN_UNREAD` entries — and the check tells them so if they forget.

## Summary by CodeRabbit

* **New Features**
  * Added automated validation for deployment configuration values and Helm chart paths.
  * Added duplicate-key detection for deployment YAML files.
  * Validation runs automatically for relevant pull requests and updates to primary branches.
  * Added clear diagnostics for missing, unreadable, or unused configuration references.

* **Tests**
  * Added self-tests covering valid paths, misconfigured paths, wrapped values, and missing configuration blocks.

### DIFF
--- a/.github/scripts/check-helm-values-paths.py
+++ b/.github/scripts/check-helm-values-paths.py
@@ -80,7 +80,7 @@ KNOWN_UNREAD = {
     "prometheus.externalLabels": "#1645 -- reads at prometheus.prometheusSpec.externalLabels",
     "prometheus.additionalScrapeConfigs": "#1645 -- reads at prometheus.prometheusSpec.additionalScrapeConfigs",
     "prometheus.alertmanager": "#1645 -- chart reads alertmanager.enabled at top level",
-    "ingress-nginx": "#1648 -- vendored chart reads .Values.controller.* at root, not under a chart-name block",
+    "ingress-nginx": "#1648 -- vendored chart reads .Values.controller.* at root; the helmfile now re-paths this block deliberately (PR #1679), so the env.yaml keys are read via templating rather than by the chart",
 }
 
 
--- a/.github/scripts/check-helm-values-paths.py
+++ b/.github/scripts/check-helm-values-paths.py
@@ -105,6 +105,29 @@ def load(path):
         raise SystemExit(2)
 
 
+def passthrough_paths(node, prefix=""):
+    """Paths whose reference value is an empty map or null.
+
+    Upstream charts use these for free-form blocks handed straight to
+    `toYaml` -- `storageSpec: {}`, `resources: {}`, `nodeSelector: {}`. Any
+    structure underneath is valid and IS read, so descending into them and
+    demanding each child appear in the reference reports pure noise.
+
+    Found by fixing #1645: re-pathing storageSpec correctly then tripped this
+    guard on `storageSpec.volumeClaimTemplate`, which is exactly the shape the
+    chart expects. A guard that fails a correct fix is worse than no guard.
+    """
+    out = set()
+    if isinstance(node, dict):
+        for k, v in node.items():
+            p = f"{prefix}.{k}" if prefix else str(k)
+            if v is None or (isinstance(v, dict) and not v):
+                out.add(p)
+            else:
+                out |= passthrough_paths(v, p)
+    return out
+
+
 def orphans_for(env_tree, block, reference):
     """Key paths the env block sets that the reference chart never defines.
 
@@ -116,7 +139,14 @@ def orphans_for(env_tree, block, reference):
         return []
     ours = key_paths({block: env_tree[block]})
     theirs = key_paths(reference)
-    unread = {p for p in ours if p not in theirs and p not in ALLOW}
+    free = passthrough_paths(reference)
+    unread = {
+        p for p in ours
+        if p not in theirs
+        and p not in ALLOW
+        # anything below a free-form block is the chart's to interpret
+        and not any(p.startswith(f + ".") for f in free)
+    }
     return sorted(p for p in unread if not any(p.startswith(q + ".") for q in unread))
 
 
@@ -193,6 +223,22 @@ def self_test():
 
         # 4. absent block is not a failure
         assert orphans_for({}, "foo", reference) == []
+
+        # 5. free-form passthrough blocks: the chart declares `storageSpec: {}`
+        #    and hands whatever is underneath to toYaml, so children are valid
+        #    and must NOT be reported. Regression test for the false positive
+        #    that fixing #1645 exposed.
+        ref2 = tmp / "ref2.yaml"
+        ref2.write_text(yaml.safe_dump({"foo": {"barSpec": {"storageSpec": {}}}}))
+        reference2 = load(ref2)
+        deep = {"foo": {"barSpec": {"storageSpec": {
+            "volumeClaimTemplate": {"spec": {"resources": {"requests": {"storage": "15Gi"}}}}}}}}
+        got = orphans_for(deep, "foo", reference2)
+        assert got == [], f"false positive under a free-form block: {got}"
+
+        # ...but a genuine mis-path OUTSIDE the free-form block still fires
+        got = orphans_for({"foo": {"storageSpec": {"x": 1}}}, "foo", reference2)
+        assert got == ["foo.storageSpec"], f"mis-path masked by passthrough logic: {got}"
 
     print("self-test OK: detects mis-paths and orphaned blocks, "
           "silent on correct configuration")
--- a/.github/scripts/check-helm-values-paths.py
+++ b/.github/scripts/check-helm-values-paths.py
@@ -1,0 +1,203 @@
+#!/usr/bin/env python3
+"""CI guard: a values key that no chart reads is silently ignored, not an error.
+
+This is the failure mode nothing else in the toolchain catches. Helm merges
+values into a plain map; a key nothing references is dropped without a warning,
+an exit code, or a line of output. `helm template` does NOT catch it either --
+it fails only on template syntax errors, explicit required/fail calls, or a
+values.schema.json violation, and the charts here ship no schema. So a setting
+at the wrong path looks exactly like working configuration, in the file and in
+review, forever.
+
+Two real defects in this repo, both long-lived, both invisible to every other
+check, are what this guard exists for:
+
+  1. charts/environments/env.yaml sets prometheus.retention and
+     prometheus.storageSpec (a 15Gi gp3 volume). kube-prometheus-stack reads
+     those at prometheus.prometheusSpec.*. Effective result: Prometheus and
+     Alertmanager run on emptyDir and lose every metric on pod restart, while
+     the file states 15Gi of persistent storage.
+
+  2. The same file sets ingress-nginx.controller.metrics.*. ingress-nginx is a
+     vendored upstream chart that reads .Values.controller.* directly, so the
+     whole block is inert -- and the chart's own defaults, which are the
+     inverse of the intent, win instead.
+
+SCOPE, and why it is narrow on purpose
+--------------------------------------
+Only EXTERNAL / vendored charts are checked -- ones whose values.yaml is the
+complete upstream reference for what the chart reads.
+
+DIGIT's own charts are deliberately excluded. They read values inside templates
+via `index .Values "some-key"` without declaring them in values.yaml, so
+values.yaml is not a complete reference for them and comparing against it
+produces false positives (verified: hrms-dev-mode, memory_limits and images are
+all genuinely consumed but undeclared). Covering them needs the reference set to
+include template references too -- a later refinement, not a blocker. A guard
+that cries wolf gets disabled, which would cost more than it saves.
+
+Run with --self-test to verify the detection logic itself catches the failure
+mode and stays quiet on correct configuration.
+"""
+import pathlib
+import sys
+import tempfile
+
+import yaml
+
+ROOT = pathlib.Path(__file__).resolve().parents[2]
+DAC = ROOT / "devops" / "deploy-as-code"
+ENV = DAC / "charts" / "environments" / "env.yaml"
+
+# env.yaml top-level block -> the chart values file that defines what that
+# chart actually reads. Only external/vendored charts; see SCOPE above.
+#
+# Both entries receive the whole env.yaml tree via the helmfile's
+# `{{ .Values | toYaml }}`, so a key must sit at the path the chart reads.
+CASES = [
+    ("prometheus", DAC / "charts/monitoring/values/prometheus.yaml"),
+    ("ingress-nginx", DAC / "charts/backbone-services/ingress-nginx/values.yaml"),
+]
+
+# Keys consumed by helmfile/templating rather than by the chart itself.
+# Every entry needs a reason -- "it was noisy" is not one.
+ALLOW = {
+    # helmfile's own release scoping, not a chart value
+    "prometheus.namespace": "helmfile release field, not read by the chart",
+}
+
+# Mis-paths that already exist, each with the issue that fixes it. Same shape as
+# FROZEN_LATEST_DEBT in local-setup/tests/static/infra-contracts.test.ts: the
+# guard goes in green today and fails on anything NEW, instead of being merged
+# red and ignored.
+#
+# This list may only shrink. A path here that no longer reproduces is a STALE
+# BASELINE and fails the check too -- otherwise the baseline outlives the bug
+# and quietly re-opens the hole it was meant to hold.
+KNOWN_UNREAD = {
+    "prometheus.retention": "#1645 -- reads at prometheus.prometheusSpec.retention",
+    "prometheus.storageSpec": "#1645 -- reads at prometheus.prometheusSpec.storageSpec",
+    "prometheus.externalLabels": "#1645 -- reads at prometheus.prometheusSpec.externalLabels",
+    "prometheus.additionalScrapeConfigs": "#1645 -- reads at prometheus.prometheusSpec.additionalScrapeConfigs",
+    "prometheus.alertmanager": "#1645 -- chart reads alertmanager.enabled at top level",
+    "ingress-nginx": "#1648 -- vendored chart reads .Values.controller.* at root, not under a chart-name block",
+}
+
+
+def key_paths(node, prefix=""):
+    """Every dotted key path in a mapping, parents included."""
+    out = set()
+    if isinstance(node, dict):
+        for k, v in node.items():
+            p = f"{prefix}.{k}" if prefix else str(k)
+            out.add(p)
+            out |= key_paths(v, p)
+    return out
+
+
+def load(path):
+    """Parse YAML, failing loudly. For a coverage guard every uncertainty
+    resolves toward a false green, so an unreadable input is a hard error."""
+    try:
+        return yaml.safe_load(path.read_text()) or {}
+    except (OSError, yaml.YAMLError) as e:
+        print(f"FAIL: cannot read {path}: {e}")
+        raise SystemExit(2)
+
+
+def orphans_for(env_tree, block, reference):
+    """Key paths the env block sets that the reference chart never defines.
+
+    Only the SHALLOWEST path of each branch is reported: if `prometheus.storageSpec`
+    is unread then every path beneath it is too, and listing them all buries the
+    one line a human needs to act on.
+    """
+    if block not in env_tree:
+        return []
+    ours = key_paths({block: env_tree[block]})
+    theirs = key_paths(reference)
+    unread = {p for p in ours if p not in theirs and p not in ALLOW}
+    return sorted(p for p in unread if not any(p.startswith(q + ".") for q in unread))
+
+
+def run():
+    env_tree = load(ENV)
+    seen = set()
+    new = []
+    for block, ref_path in CASES:
+        if not ref_path.exists():
+            print(f"FAIL: reference values file missing: {ref_path}")
+            return 2
+        for p in orphans_for(env_tree, block, load(ref_path)):
+            seen.add(p)
+            if p not in KNOWN_UNREAD:
+                new.append((p, block, ref_path))
+
+    stale = sorted(set(KNOWN_UNREAD) - seen)
+    rc = 0
+
+    if new:
+        rc = 1
+        print("FAIL: env.yaml sets values no chart reads. These are silently "
+              "ignored at deploy time -- the file says one thing, the cluster "
+              "does another.\n")
+        for p, block, ref_path in new:
+            print(f"  {p}")
+            print(f"      block `{block}:`, chart reference "
+                  f"{ref_path.relative_to(ROOT)}")
+        print("\nMove each key to the path the chart actually reads, then confirm "
+              "by RENDERING the chart and checking the output -- reading the "
+              "values file is what let the existing ones through.\n")
+
+    if stale:
+        rc = 1
+        print("FAIL: KNOWN_UNREAD lists paths that no longer reproduce. If they "
+              "were fixed, delete them -- a baseline that outlives its bug "
+              "quietly re-opens the hole it was holding.\n")
+        for p in stale:
+            print(f"  {p}    ({KNOWN_UNREAD[p]})")
+        print()
+
+    if rc == 0:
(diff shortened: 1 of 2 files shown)


singhalkarun:
Pushed a fix found by using the guard on real work.

Implementing #1645 — moving `storageSpec` to `prometheus.prometheusSpec.storageSpec`, which is the correct path — made the guard **fail on `storageSpec.volumeClaimTemplate`**. That is a false positive: upstream charts declare free-form blocks as an empty map (`storageSpec: {}`, `resources: {}`, `nodeSelector: {}`) and pass whatever is underneath straight to `toYaml`, so any structure there is valid and read.

A guard that fails a correct fix is worse than no guard — the next person works around it or disables it. Now a reference path whose value is `{}` or null terminates the walk, with self-test cases pinning both directions: children of a passthrough block are silent, a mis-path outside one still fires.

Worth noting the sequence, since it is the argument for the guard: it caught a real bug in #1645, then a real bug in itself, both within an hour of being used.

.github/scripts/check-helm-values-paths.py
```diff
@@ -0,0 +1,203 @@
...
+# BASELINE and fails the check too -- otherwise the baseline outlives the bug
+# and quietly re-opens the hole it was meant to hold.
+KNOWN_UNREAD = {
+    "prometheus.retention": "#1645 -- reads at prometheus.prometheusSpec.retention",
+    "prometheus.storageSpec": "#1645 -- reads at prometheus.prometheusSpec.storageSpec",
+    "prometheus.externalLabels": "#1645 -- reads at prometheus.prometheusSpec.externalLabels",
+    "prometheus.additionalScrapeConfigs": "#1645 -- reads at prometheus.prometheusSpec.additionalScrapeConfigs",
+    "prometheus.alertmanager": "#1645 -- chart reads alertmanager.enabled at top level",
+    "ingress-nginx": "#1648 -- vendored chart reads .Values.controller.* at root, not under a chart-name block",
+}
```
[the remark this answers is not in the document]
singhalkarun: Confirmed and fixed in c47bd062.

The hole was exactly as described: `stale = set(KNOWN_UNREAD) - seen` only ever
looked at the current `env.yaml`, so listing a newly-introduced mis-path in
`KNOWN_UNREAD` put it in `seen` and silenced both checks. The "may only shrink"
comment was a request, not a rule.

`run()` now also compares the baseline against the revision at
`VALUES_BASELINE_REF` — the keys are parsed out of that revision of this file
with `ast.literal_eval`, never executed — and fails on additions. The workflow
sets that ref to `github.event.pull_request.base.sha`, with its own
`git fetch --depth=1 origin "$BASE_SHA"` step (the default depth-1 checkout does
not contain the base), deliberately placed so a broken fetch turns that step red
instead of silently downgrading the guard to "skipped".

Outside CI the ref is unset and the check is skipped, loudly, on stdout — a
working tree can sit anywhere relative to any branch, and a guard that fails a
legitimate local run gets disabled. To run it locally:
`VALUES_BASELINE_REF=$(git merge-base HEAD origin/develop)`.

Verified by simulating the exact scenario in the finding — add
`prometheus.bogusRetention: 30d` to `env.yaml` *and* the matching
`KNOWN_UNREAD` entry:

```
=== A) no baseline ref (the old behaviour) ===
rc=0                          <- growth would have merged green

=== B) same tree, VALUES_BASELINE_REF=HEAD ===
FAIL: KNOWN_UNREAD has GROWN. ...
  prometheus.bogusRetention    (#9999 -- sneaked in)
baseline-growth check: KNOWN_UNREAD vs HEAD (6 entries there, 7 here).
rc=1
```

And a legitimate removal (drop `prometheus.externalLabels` from both files) stays
green — the list may still shrink:

```
OK: 2 external chart(s) checked, no new unread key paths. 5 known mis-path(s) ...
baseline-growth check: KNOWN_UNREAD vs HEAD (6 entries there, 5 here).
rc=0
```

Degradation paths all return "skipped", never a crash: ref absent from history
(the case on this very PR, where the file does not exist at `monitoring-fix`
yet), unknown ref, and a copy of the tree outside any git checkout — all
verified, all `rc=0` with the SKIPPED line printed.

`--self-test` covers the new logic (parse, addition fires, removal silent,
syntax-error / absent / non-literal / bad-ref all degrade to unknown) and both
CI commands pass:

```
self-test OK: detects mis-paths and orphaned blocks, silent on correct configuration, and rejects baseline growth
OK: 2 external chart(s) checked, no new unread key paths. 6 known mis-path(s) still outstanding: ...
```

.github/workflows/deploy-as-code-validation.yml
```diff
@@ -0,0 +1,82 @@
...
+      - 'devops/deploy-as-code/**'
+      - '.github/scripts/check-helm-values-paths.py'
+      - '.github/workflows/deploy-as-code-validation.yml'
+
+jobs:
+  values-paths:
+    name: values keys must be read by their chart
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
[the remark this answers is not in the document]
singhalkarun: Confirmed and fixed in c47bd062. Checked first — the workflow had no
`permissions:` block at any level, so nothing was duplicated.

- added top-level `permissions: contents: read`
- `with: persist-credentials: false` on both checkouts (values-paths and
  duplicate-keys)

Verified with zizmor 1.x locally, before vs after on the same file:

```
=== previous revision ===
   4 error[unpinned-uses]
   2 warning[artipacked]: credential persistence through GitHub Actions artifacts
   3 warning[excessive-permissions]: overly broad permissions

=== after ===
   4 error[unpinned-uses]
```

Both audits you flagged are clear. `unpinned-uses` is left alone: every workflow
in this repo references actions by tag (`actions/checkout@v4` ×20), so
hash-pinning is a repo-wide convention change, not something to do in one file
in this PR.

One note on the interaction with the other thread: the new baseline-growth check
needs the PR base commit, which the depth-1 checkout does not have, so there is
now a `git fetch --no-tags --depth=1 origin "$BASE_SHA"` step. With
`persist-credentials: false` that fetch is anonymous — fine for this public repo,
and the base sha comes in through `env:` rather than being interpolated into the
run script.

.github/scripts/check-helm-values-paths.py
```diff
@@ -0,0 +1,362 @@
...
+# mis-path and lists it here in the same breath makes that path `seen`, so
+# neither the new-path check nor the stale check fires.
+KNOWN_UNREAD = {
+    "prometheus.retention": "#1645 -- reads at prometheus.prometheusSpec.retention",
+    "prometheus.storageSpec": "#1645 -- reads at prometheus.prometheusSpec.storageSpec",
+    "prometheus.externalLabels": "#1645 -- reads at prometheus.prometheusSpec.externalLabels",
+    "prometheus.additionalScrapeConfigs": "#1645 -- reads at prometheus.prometheusSpec.additionalScrapeConfigs",
+    "prometheus.alertmanager": "#1645 -- chart reads alertmanager.enabled at top level",
+    "ingress-nginx": "#1648 -- vendored chart reads .Values.controller.* at root; the helmfile now re-paths this block deliberately (PR #1679), so the env.yaml keys are read via templating rather than by the chart",
+}
```
vinothrallapalli-eGov, line 114: Verified the #1659 interaction described in that PR's description: `prometheus.retention`, `prometheus.storageSpec`, `prometheus.externalLabels`, `prometheus.additionalScrapeConfigs`, and `prometheus.alertmanager` are exactly the 5 entries #1659 fixes (it moves them under `prometheus.prometheusSpec.*` / top-level `alertmanager`). #1659 is still open, not merged, so as of this diff these 5 entries are still live and correct here. But once #1659 lands, `stale = sorted(set(KNOWN_UNREAD) - seen)` will correctly flip these to a hard FAIL ("KNOWN_UNREAD lists paths that no longer reproduce") — which is the intended enforcement, not a bug — but it does mean whichever of the two PRs merges second needs a follow-up commit deleting these 5 lines (leaving only `ingress-nginx`, tracked separately against #1648). Confirming this isn't already handled anywhere else in this diff — flagging so it doesn't get lost as a required manual step tied to merge order between #1652 and #1659.

.github/scripts/check-helm-values-paths.py
```diff
@@ -0,0 +1,362 @@
...
+        for k, v in node.items():
+            p = f"{prefix}.{k}" if prefix else str(k)
+            if v is None or (isinstance(v, dict) and not v):
+                out.add(p)
+            else:
+                out |= passthrough_paths(v, p)
+    return out
+
+
+def orphans_for(env_tree, block, reference):
```
vinothrallapalli-eGov, line 161: Minor/future-scope note: `key_paths`/`orphans_for` only recurse into `dict` nodes (`isinstance(node, dict)`), not `list`. For the two current `CASES` this isn't a live gap — the one list-valued key in scope, `additionalScrapeConfigs`, is caught at the top-level path (and is baselined wholesale via #1645/#1659, not descended into). But if coverage is later extended to a chart/block where a mis-keyed field lives inside a list item's dict (e.g. an `extraVolumes`/container-array style key), it would silently escape detection the same way DIGIT's own charts were excluded for a different reason. Worth a one-line note in the SCOPE docstring for the next person extending `CASES`, not a blocker here.

.github/workflows/deploy-as-code-validation.yml
```diff
@@ -0,0 +1,105 @@
...
+      - run: pip install pyyaml
+
+      # Verify the detector itself before trusting its verdict — same idiom as
+      # check-gatus-coverage.py and check-flyway-dump-alignment.py.
+      - name: self-test the guard
+        run: python3 .github/scripts/check-helm-values-paths.py --self-test
+
+      - name: env.yaml keys must match paths the chart reads
+        env:
+          VALUES_BASELINE_REF: ${{ github.event.pull_request.base.sha }}
```
vinothrallapalli-eGov, line 79: On `push` events (develop/master), `github.event.pull_request.base.sha` is undefined, so `VALUES_BASELINE_REF` resolves empty and `baseline_growth()` SKIPs loudly rather than comparing. That looks intentional and benign (push only fires post-merge, after the PR-time check already gated growth), just confirming it's a deliberate skip and not an oversight — the loud skip message makes it visible either way.

vinothrallapalli-eGov (commented):
Reviewed the guard logic (`orphans_for`/`key_paths`/`passthrough_paths`) against the self-tests: it correctly avoids the naive-diff false-positive trap — it only flags paths env.yaml actually *sets* that the reference values.yaml never declares, so legitimately-optional undeclared keys never trigger, and free-form passthrough blocks (`storageSpec: {}` etc.) are correctly excluded (test 5/6 in `self_test()` reproduces the exact #1645 false positive and confirms it's fixed). Duplicate-YAML-key handling is delegated to `yamllint -d key-duplicates` in a separate job reading the raw document, which is the right call — `check-helm-values-paths.py` itself uses `yaml.safe_load`, which would silently last-wins a duplicate before the values-path logic ever saw it, so catching this requires exactly the raw-document linter this PR adds. The `KNOWN_UNREAD` growth enforcement via `known_unread_at`/`baseline_growth` (parsing the prior revision's AST rather than executing it) is sound and already covered by self-test case 6. Two items worth a look, both minor/informational — see inline.

.github/scripts/check-helm-values-paths.py
singhalkarun: Confirmed exactly, and recorded in 04347c9b. Not handled anywhere else in this diff —
you were right to flag it.

Reproduced rather than reasoned about: dropped #1659's `env.yaml` onto this tree and
ran the guard.

```
$ git show origin/fix/1645-prometheus-storage:devops/deploy-as-code/charts/environments/env.yaml \
    > devops/deploy-as-code/charts/environments/env.yaml
$ python3 .github/scripts/check-helm-values-paths.py
FAIL: KNOWN_UNREAD lists paths that no longer reproduce. ...
  prometheus.additionalScrapeConfigs
  prometheus.alertmanager
  prometheus.externalLabels
  prometheus.retention
  prometheus.storageSpec
rc=1
```

Precisely those five, nothing more — `ingress-nginx` survives, tracked against #1648,
as you said.

Nothing to change in the baseline now: both PRs target `monitoring-fix`, neither has
merged, and the five still reproduce here, so deleting them today would flip them to
*new unread paths* and fail this PR instead. What I could fix is the way this was
only going to exist as a review thread. Two places:

**1. A merge-order note above `KNOWN_UNREAD`** — the five entries are marked as one
unit that leaves together with #1659, with the verification above cited.

**2. The stale-baseline FAIL message**, which is the one a human will actually be
looking at:

```
This is the EXPECTED outcome when the fixing pull request lands: delete the listed
entries from KNOWN_UNREAD in that same change. The five prometheus.* entries all go
together with #1659.
```

Worth separating two things here: the step was never at risk of being *skipped* —
CI goes red and names the five paths with their issue numbers. The risk was that
whoever hits that red has no way to tell an expected handoff from a regression, and
"the guard is being awkward, add them back" is a plausible wrong move. That is what
the message now forecloses.

.github/scripts/check-helm-values-paths.py
singhalkarun: Confirmed, and noted in the SCOPE docstring in 04347c9b — agreed it is not a blocker,
and agreed it is exactly the kind of thing that turns into coverage-in-name-only for
whoever extends `CASES`.

Your read of the current state is right: `additionalScrapeConfigs` is the one
list-valued key in scope (`type(env['prometheus']['additionalScrapeConfigs'])` is
`list`), it is caught at its own top-level path, and it is baselined wholesale rather
than descended into — so there is no live gap.

Added:

> One more limit, for whoever extends CASES: key_paths/orphans_for recurse into
> `dict` nodes only, never into `list` items. A mis-keyed field inside a list element
> — an `extraVolumes:` / container-array style block — is therefore NOT detected; only
> the list-valued key itself is, at its own path. […] a chart whose overrides live
> inside lists needs list recursion added first, or it will be covered in name only.

Left the behaviour alone deliberately. Adding list recursion now would be untested
against any real case in `CASES`, and the interaction with `passthrough_paths` is not
obvious — an upstream chart declaring `extraVolumes: []` is the list analogue of the
free-form `{}` blocks, and getting that wrong reintroduces exactly the false positive
that fixing #1645 first exposed. Better done with a chart that needs it.

.github/workflows/deploy-as-code-validation.yml
singhalkarun: Deliberate, confirmed — and documented in the workflow in 04347c9b, since a comment
is the right place for it rather than a resolved review thread.

Your mechanism is right: on `push` there is no `pull_request` context,
`${{ github.event.pull_request.base.sha }}` renders empty, and `baseline_growth()`
returns the loud SKIP line. The `git fetch` step above is separately guarded with
`if: github.event_name == 'pull_request'`.

Two reasons rather than the one you named, both now in the file:

1. Push only fires post-merge, by which time the pull-request run has already gated
   growth — the check is redundant there, not missing.
2. There is no honest baseline for a merged commit anyway. The parent of a merge
   commit on `develop` is not "the base of the change"; comparing against it would
   either pass trivially or fail for reasons unrelated to the change.

Added: the other two checks (new unread paths, stale baseline) *do* run on push, so
the push run is not decorative — it is the one that catches a mis-path arriving on
`develop` through some path other than a PR touching these files.

singhalkarun:
Heads-up from resolving review on #1663/#1679 — two things here need adjusting, neither urgent while this is open.

**1. The `ingress-nginx` KNOWN_UNREAD reason is factually wrong.** The reason text says "vendored chart reads `.Values.controller.*` at root". It does, but this vendored copy also carries a **local modification** to `templates/_helpers.tpl` (lines 6-9, absent from upstream 4.10.0):

```gotemplate
```

`mergeOverwrite` mutates the root context in place, so `.Values["ingress-nginx"]` is hoisted over `.Values` at template time — after values files and after `--set`. The `ingress-nginx.controller.*` subtree is therefore genuinely read; only the siblings of `controller:` (`ssl-ciphers`, `ssl-protocols`, `ssl-ecdh-curve`, `default-backend-service`, `cert-issuer`, `setSslTls`) and the misnamed `replicas` are not.

This matters for more than wording: the guard's model is "compare env.yaml key paths against the chart's `values.yaml` key paths", and a chart that rewrites its own root context at render time is invisible to that model. Worth at least a comment in `CASES` noting the check is structurally blind to this pattern.

**2. The five `prometheus.*` entries.** #1659 fixes all five; run against that branch the guard is red with "KNOWN_UNREAD lists paths that no longer reproduce", naming exactly those five. Whichever of #1652 and #1659 merges second must delete them in the same commit. The baseline is behaving as designed — flagging so it doesn't land as a surprise red build.

vinothrallapalli-eGov (approved):

Issue #1649, opened 2026-08-06
No external or functional uptime checks in the production tier (blackbox exporter is installed: false)


singhalkarun:
Part of #1619 (Kubernetes monitoring parity). **P1 · effort: M**

## The gap, stated carefully

The Ansible tier runs Gatus: **51 endpoints** across 10 groups (Infrastructure 5, Core Services 15, OTP 3, Search 3, MCP 2, Keycloak 3, Notifications 9, API Gateway 5, Application 3, API Tests 3), with a CI guard that fails the build when a deployed service has no check.

The production tier has **no Gatus and no equivalent**. But "port Gatus" is the wrong conclusion, because Kubernetes already provides part of what Gatus provides, and the production tier already has one thing the Ansible tier only just gained. Worth separating the three layers:

### Layer 1 — per-pod health: **already covered here, natively**

25 of the 41 application charts define liveness/readiness probes (`charts/common/templates/_deployment.yaml:127-131`, driven by each chart's `healthChecks` block). Kubernetes restarts unhealthy pods and removes them from service — something Gatus cannot do, it only observes.

kube-prometheus-stack's `defaultRules.rules.kubernetesApps` is enabled (the only rule group that is), covering CrashLoopBackOff, pod-not-ready and deployment-replica mismatches. Custom rules add `KubernetesPodNotHealthy`, `KubernetesNodeNotReady`, node memory/disk/network pressure and PVC alerts (`charts/monitoring/values/prometheus.yaml:180-288`).

So the "is this service up" question is genuinely answered here — arguably better than by Gatus, since k8s acts on it.

### Layer 2 — external / synthetic checks: **missing**

Nothing verifies the platform responds *from outside*, through DNS, the ingress and TLS. The chart for it exists and is switched off — `blackbox` is `installed: false` (`charts/monitoring/monitoring-helmfile.yaml:77-78`) — while two scrape jobs and a Grafana dashboard already reference it, and one of those jobs probes a hardcoded foreign domain (`https://unified-uat.digit.org/citizen/`, `values/prometheus.yaml:3782`).

This is the layer that catches "pods are all healthy and citizens still cannot reach the site".

### Layer 3 — functional API checks: **missing**

[...]

PR #1659, opened 2026-08-07
fix(prometheus): put the storage/retention block where the chart reads it

### LINKED ISSUE #1645: Prometheus and Alertmanager run without persistent storage; the env.yaml block that would fix it is mis-pathed

singhalkarun:
Fixes #1645

## What

Prometheus and Alertmanager have been running on **emptyDir**, losing every metric on pod restart, while `env.yaml` stated 7d retention on a 15Gi gp3 volume. The block sat one level too shallow: kube-prometheus-stack reads `prometheus.prometheusSpec.*`, not `prometheus.*`.

Verified by pulling chart 56.6.2 and grepping its templates — only `.Values.prometheus.prometheusSpec.{retention,storageSpec,externalLabels}` are referenced. The shallower paths appear nowhere.

Moves `retention`, `storageSpec` and `externalLabels` under `prometheusSpec`, and `alertmanager.enabled` to the top level.

## The part worth reviewing: `additionalScrapeConfigs` is deliberately NOT moved

Re-pathing that list would have **added broken targets rather than fixing anything**:

- `redis-exporter` → `prometheus-redis-exporter.backbone:9121`, a Service that does not exist (the deployed redis chart ships no exporter)
- `blackbox` → probes `https://unified-uat.digit.org/citizen/`, a **foreign domain** carried in by the upstream import, via an exporter that is `installed: false`
- `nginx-ingress-metrics` → a Service name that does not resolve

So the naive fix would have started a permanently-failing probe against someone else's environment. The effective list stays the one in `values/prometheus.yaml`; the block is removed from `env.yaml` with a comment recording why, and the two targets that should exist are tracked in #1648 (gateway) and #1649 (blackbox).

## Risk

`storageClassName: gp3` — the same class is already used by pgadmin in this file, so it is present on the target clusters. Worth confirming on the specific cluster before rollout: if `gp3` is absent, Prometheus moves from "loses data on restart" to "does not schedule". That is a louder failure, but it is a failure.

This is also the first time Prometheus will claim a PVC on these clusters. Expect a one-off pod recreation on the next apply.

## Merge-order dependency with #1652

This interacts with the values-path guard in #1652, and the interaction is by design.

That guard ships a `KNOWN_UNREAD` baseline listing these five mis-paths. Once this PR lands they no longer reproduce, so the guard fails with **"KNOWN_UNREAD lists paths that no longer reproduce"** and names the exact lines to delete. Whichever of the two merges second, someone must remove those five entries from `.github/scripts/check-helm-values-paths.py`:

```
prometheus.retention
prometheus.storageSpec
prometheus.externalLabels
prometheus.additionalScrapeConfigs
prometheus.alertmanager
```

Leaving `ingress-nginx`, which #1648 will clear.

This is the baseline behaving as intended — it refuses to outlive the bug it was holding — but it does mean a red build until the entries go.

## Alertmanager gets a volume too

Added after review. The original change moved only the Prometheus block, but `alertmanager.alertmanagerSpec.storage` is `{}` in the vendored values (`values/prometheus.yaml:882`) and #1645 asks for both — acceptance criterion 2 is "The rendered `Alertmanager` resource shows a storage block". Without it a restart drops every active silence and the notification log, and re-pages for everything currently suppressed.

2Gi, not the 50Gi in the chart's commented-out example: Alertmanager persists silences and nflog, not time series.

## Verified / not verified

- [x] `env.yaml` parses; keys land at `prometheus.prometheusSpec.*` and top-level `alertmanager`
- [x] The corrected values-path guard reports no unread paths for this block
- [x] **Rendered**, not just grepped. `helm template` of kube-prometheus-stack 56.6.2 in the helmfile's merge order (`values/prometheus.yaml` then `env.yaml`) gives the `Prometheus` resource `retention: 7d`, `storage.volumeClaimTemplate` (gp3, 15Gi) and `externalLabels.cluster: urban`, and the `Alertmanager` resource `storage.volumeClaimTemplate` (gp3, 2Gi)
- [ ] **Not verified on a cluster** — this tier has none
- [ ] A pod restart preserves metric history
- [ ] `gp3` resolves on the target cluster

## Related

While implementing this I hit a false positive in the guard from #1652 — it flagged `storageSpec.volumeClaimTemplate`, because upstream declares `storageSpec: {}` as a free-form passthrough. Fixed on that PR; a guard that fails a correct fix is worse than no guard.

### DIFF
--- a/devops/deploy-as-code/charts/environments/env.yaml
+++ b/devops/deploy-as-code/charts/environments/env.yaml
@@ -305,44 +305,42 @@ grafana:
 # <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
 
 ### Prometheus & Alertmanager >>>>>>>>>>>>>>>>>>>>
+# kube-prometheus-stack reads these under prometheus.prometheusSpec.*, NOT
+# directly under prometheus.* (#1645). Set one level too shallow they are
+# silently ignored -- helm drops values nothing reads, with no warning and no
+# exit code -- so this block previously read as "15Gi of persistent metrics"
+# while Prometheus actually ran on an emptyDir and lost everything on restart.
 prometheus:
-  retention: 7d
-  storageSpec:
-    volumeClaimTemplate:
-      spec:
-        storageClassName: gp3
-        accessModes: ["ReadWriteOnce"]
-        resources:
-          requests:
-            storage: 15Gi
-  alertmanager:
-    enabled: true
-  externalLabels:
-    cluster: urban
-  additionalScrapeConfigs:
-    - job_name: 'nginx-ingress-metrics'
-      static_configs:
-        - targets: [ 'release-name-ingress-nginx-controller-metrics.egov:10254' ]
-    - job_name: 'redis-exporter'
-      static_configs:
-        - targets: [ 'prometheus-redis-exporter.backbone:9121' ]
-    - job_name: 'blackbox'
-      metrics_path: /probe
-      params:
-        module: [ http_2xx ]
-      static_configs:
-        - targets:
-            - https://unified-uat.digit.org/citizen/
-      relabel_configs:
-        - source_labels: [ __address__ ]
-          target_label: __param_target
-        - source_labels: [ __param_target ]
-          target_label: instance
-        - target_label: __address__
-          replacement: prometheus-blackbox-exporter:9115
-    - job_name: 'blackbox_exporter'
-      static_configs:
-        - targets: [ 'prometheus-blackbox-exporter:9115' ]
+  prometheusSpec:
+    retention: 7d
+    storageSpec:
+      volumeClaimTemplate:
+        spec:
+          storageClassName: gp3
+          accessModes: ["ReadWriteOnce"]
+          resources:
+            requests:
+              storage: 15Gi
+    externalLabels:
+      cluster: urban
+
+  # NOTE: additionalScrapeConfigs deliberately NOT moved here (#1645).
+  # The list that used to sit at this path named four jobs, and making them
+  # effective would have ADDED broken targets rather than fixed anything:
+  #   - redis-exporter -> prometheus-redis-exporter.backbone:9121, a Service
+  #     that does not exist (the deployed redis chart ships no exporter)
+  #   - blackbox -> probes https://unified-uat.digit.org/citizen/, a FOREIGN
+  #     domain carried in by the upstream import, against a blackbox exporter
+  #     that is installed: false
+  #   - nginx-ingress-metrics -> a Service name that does not resolve either
+  # The effective list remains the one in charts/monitoring/values/prometheus.yaml.
+  # Correcting the gateway target is #1648; enabling blackbox is #1649.
+
+# Top-level, not under prometheus: (#1645). Already true in
+# values/prometheus.yaml, so this is belt-and-braces rather than a behaviour
+# change -- but at the previous path it asserted nothing at all.
+alertmanager:
+  enabled: true
 # <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
 
 ### loki-stack >>>>>>>>>>>>>>>>>>>>
--- a/devops/deploy-as-code/charts/environments/env.yaml
+++ b/devops/deploy-as-code/charts/environments/env.yaml
@@ -341,6 +341,20 @@ prometheus:
 # change -- but at the previous path it asserted nothing at all.
 alertmanager:
   enabled: true
+  # Alertmanager was in exactly the same position as Prometheus: the vendored
+  # values leave alertmanager.alertmanagerSpec.storage at {} (values/prometheus.yaml:882),
+  # so the silence list and the notification log lived on an emptyDir and were
+  # lost on every restart -- a restart could re-page for everything currently
+  # silenced. 2Gi is ample: Alertmanager stores silences and nflog, not series.
+  alertmanagerSpec:
+    storage:
+      volumeClaimTemplate:
+        spec:
+          storageClassName: gp3
+          accessModes: ["ReadWriteOnce"]
+          resources:
+            requests:
+              storage: 2Gi
 # <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
 
 ### loki-stack >>>>>>>>>>>>>>>>>>>>


devops/deploy-as-code/charts/environments/env.yaml
```diff
@@ -305,44 +305,42 @@ grafana:
...
+    storageSpec:
+      volumeClaimTemplate:
+        spec:
+          storageClassName: gp3
+          accessModes: ["ReadWriteOnce"]
+          resources:
+            requests:
+              storage: 15Gi
+    externalLabels:
+      cluster: urban
```
vinothrallapalli-eGov, line 325: Confirmed against the vendored chart defaults in `values/prometheus.yaml` (kube-prometheus-stack 56.6.2): `retention` (base default `10d`, line ~3602), `storageSpec` (base default `{}`, line ~3707) and `externalLabels` (base default `cluster: unified-qa`, line ~3430) are all read from `prometheus.prometheusSpec.*`. This block now lands exactly there, so the override will actually take effect via helmfile's deep-merge of the two values files (base `values/prometheus.yaml` then this `env.yaml`). Good fix — this was silently a no-op at the old path.

devops/deploy-as-code/charts/environments/env.yaml
```diff
@@ -305,44 +305,42 @@ grafana:
...
+  #     that is installed: false
+  #   - nginx-ingress-metrics -> a Service name that does not resolve either
+  # The effective list remains the one in charts/monitoring/values/prometheus.yaml.
+  # Correcting the gateway target is #1648; enabling blackbox is #1649.
+
+# Top-level, not under prometheus: (#1645). Already true in
+# values/prometheus.yaml, so this is belt-and-braces rather than a behaviour
+# change -- but at the previous path it asserted nothing at all.
+alertmanager:
+  enabled: true
```
vinothrallapalli-eGov, line 343: Correct — `alertmanager.enabled` is top-level in the chart (confirmed at line 362 of the vendored `values/prometheus.yaml`, already `true` there too, so this is belt-and-braces as the comment says, not a behavior change). One thing this move does *not* do: give Alertmanager its own persistent storage. The chart's `alertmanager.alertmanagerSpec.storage` also defaults to `{}` (ephemeral) in the vendored values (~line 882), and neither the old nor new block in this file sets it. So Alertmanager will still lose silences/notification log on pod restart even after this fix lands — only Prometheus gets the PVC. Worth confirming that's intentionally out of scope for this PR (vs. an assumption baked into the issue title) rather than something expected to be covered here.

vinothrallapalli-eGov (commented):
Verified the core claim independently: this repo vendors the full kube-prometheus-stack 56.6.2 default `values.yaml` at `devops/deploy-as-code/charts/monitoring/values/prometheus.yaml`, and in it `retention` (line 3602), `storageSpec` (3707), `externalLabels` (3430) and `additionalScrapeConfigs` (3773) all live under `prometheus.prometheusSpec.*`, while `alertmanager.enabled` (362) is genuinely top-level. The new `env.yaml` block matches that nesting exactly (checked with `yaml.safe_load`, no duplicate keys, parses clean), and `gp3` is indeed already used by pgadmin elsewhere in the same file. So the retention/storage fix for Prometheus is correctly targeted, not just plausible-looking, and Alertmanager's `enabled` flag (the only Alertmanager setting that ever existed in this file) is not half-fixed. One gap worth flagging: neither the old nor the new block gives Alertmanager its own persistent storage (`alertmanager.alertmanagerSpec.storage`, which also defaults to `{}`/ephemeral in the vendored chart values at line 882) — so despite the PR/issue framing of 'Prometheus and Alertmanager' losing data on restart, this change only fixes Prometheus; Alertmanager silences/notification state still won't survive a restart. Not a regression (that gap predates this PR), but worth a follow-up issue if durable Alertmanager state is actually wanted. The #1652 KNOWN_UNREAD staleness is well-documented in the PR body and correctly left out of this diff since the guard script doesn't exist on the base branch yet.

devops/deploy-as-code/charts/environments/env.yaml
singhalkarun: Confirmed, and re-verified independently rather than taken on trust. Rendering `kube-prometheus-stack` 56.6.2 with the helmfile's own merge order (`values/prometheus.yaml` then `env.yaml`) produces:

```
==== Prometheus
      retention: "7d"
      storage:
        volumeClaimTemplate:
                storage: 15Gi
      externalLabels:
        cluster: urban
```

Your line references all check out against the vendored file: `retention: 10d` at 3602, `storageSpec: {}` at 3707, `externalLabels.cluster: unified-qa` at 3430, `alertmanager.enabled: true` at 362.

One thing worth flagging for merge ordering, since it isn't visible from this diff. #1652's `KNOWN_UNREAD` baseline in `.github/scripts/check-helm-values-paths.py` holds exactly the five `prometheus.*` paths this PR fixes, and that guard fails on a baseline entry that no longer reproduces, not only on new ones. I ran it against this branch with the script copied in:

```
FAIL: KNOWN_UNREAD lists paths that no longer reproduce.
  prometheus.additionalScrapeConfigs
  prometheus.alertmanager
  prometheus.externalLabels
  prometheus.retention
  prometheus.storageSpec
```

Nothing to do here today — #1652 is still open and the script is not on this branch, so neither PR is red on its own. But whichever of the two lands second has to delete those five entries in the same commit.

devops/deploy-as-code/charts/environments/env.yaml
singhalkarun: You're right on the facts and right to push on the scope question — it wasn't intentional, it was an omission. Fixed in 6e73858.

`alertmanager.alertmanagerSpec.storage: {}` is at line 882 exactly as you say, and #1645 is explicit about it: the body says "Alertmanager is in the same position — `values/prometheus.yaml:877` `retention: 120h`, `:882` `storage: {}`", and acceptance criterion 2 is "The rendered `Alertmanager` resource shows a storage block". So this was the issue title being accurate and the PR only covering half of it.

The branch now adds a PVC alongside the Prometheus one:

```yaml
alertmanager:
  enabled: true
  alertmanagerSpec:
    storage:
      volumeClaimTemplate:
        spec:
          storageClassName: gp3
          accessModes: ["ReadWriteOnce"]
          resources:
            requests:
              storage: 2Gi
```

Re-rendered with the same merge order:

```
==== Alertmanager
  retention: "120h"
  storage:
    volumeClaimTemplate:
        storageClassName: gp3
            storage: 2Gi
```

2Gi rather than the 50Gi in the chart's commented-out example: Alertmanager persists the silence list and the notification log, not time series. The failure mode this closes is the one you named — a restart dropping every active silence and re-paging for everything currently suppressed.

KDwevedi:
Changes look fine. One nit: this should be rebased on #1652 if it is merged first; otherwise, rebase #1652 onto this PR. The values-path CI will fail because its `KNOWN_UNREAD` baseline still lists the five `prometheus.*` paths that this PR fixes, and the guard deliberately rejects stale baseline entries.

KDwevedi (approved):

Issue #1647, opened 2026-08-06
Kafka has no metrics in the production tier: consumer lag is invisible


singhalkarun:
Part of #1619 (Kubernetes monitoring parity). **P1 · effort: S**

Mirrors #1623, which added broker metrics to the Ansible tier.

## Problem

Kafka is installed and serving (`charts/backbone-services/kafka-kraft/values.yaml`, release at `backboneservices-helmfile.yaml:109-113`, `installed: true`), and it produces **no metrics**. Both exporters the chart ships are switched off:

- `values.yaml:1629` — `metrics.kafka.enabled: false` (the standalone Kafka exporter)
- `values.yaml:1880` — `metrics.jmx.enabled: false` (the JMX exporter)
- `values.yaml:2010` — `metrics.serviceMonitor.enabled: false`

`charts/environments/env.yaml` does not override any of them — its only Kafka references are the broker address (`:14`) and the kafka-ui cluster block (`:276-282`). Note the helmfile passes the whole env tree into the chart, so a top-level `metrics:` key there *would* have worked; none exists.

## Why this matters more than a missing dashboard

The failure it hides is **consumer lag**, and in this platform lag is silent data loss rather than slowness.

`egov-persister` is the sole write path from Kafka to Postgres — its topic-to-SQL map covers PGR complaints, workflow transitions, HRMS employees, MDMS, boundaries and audit (`env.yaml:143`). It runs `replicas: 1` (`env.yaml:140`). If it stalls or falls behind:

- the API still returns **200** with a complaint ID
- the pod is **Running** and healthy
- and the record **is not in the database**

Nothing in the current setup would show this. `egov-indexer` (also `replicas: 1`) has the same shape for the search index — the PGR inbox silently goes stale.

## Which exporter is required

**The Kafka exporter, not the JMX one.** Consumer-group lag is not a broker JMX MBean — it comes from the consumer-group offsets API, which is what `kafka-exporter` reads, emitting `kafka_consumergroup_lag`. Enabling JMX alone would give broker health (under-replicated partitions, request rates, JVM) but **not lag**.

Both are worth having; only one answers the question that motivates this issue.

## Two amplifying findings from the same review

Not strictly parity items, but they change how urgent lag visibility is, and they are cheap to fix alongside:

[...]

Issue #1646, opened 2026-08-06
No application metrics in the production tier: adopt the OTEL agent as the Ansible tier does


singhalkarun:
Part of #1619 (Kubernetes monitoring parity). **P1 · effort: M** *(revised down from L — see "Revision" at the end)*

**Decision: mirror the Ansible tier — OpenTelemetry Java agent, not micrometer/ServiceMonitors.**

**This is additive to the upstream stack, not a replacement.** kube-prometheus-stack, Loki, Promtail and Grafana are adopted as-is; the only new component is one OTEL collector.

## Problem

The production tier produces **no application metrics**. Not "they are not collected" — they are not produced. Two independent gaps:

### 1. Nothing collects

- **0 of 41 application charts** define a `ServiceMonitor` or `PodMonitor`.
- `charts/common/templates/_servicemonitor.yaml` defines `common.servicemonitor` and **nothing includes it** — repo-wide search returns one hit, the definition. It would also need three fixes to work: it emits no `release: kube-prometheus-stack` label (which `serviceMonitorSelectorNilUsesHelmValues: true` requires), it references `.Values.namespace` which is never set for app charts, and it scrapes the service root with no `path:`.
- The annotation route is equally dead: `_service.yaml:9-12` emits `prometheus.io/*` only if a chart sets `metrics:`; none does; and no `kubernetes_sd_configs` job exists that would honour them.

### 2. Nothing to collect from — this is what decides the approach

Even a perfect ServiceMonitor would scrape a 404. There is **no micrometer registry and no actuator Prometheus endpoint in any of the 41 charts**. The only management setting is `charts/common/values.yaml:136-137` → `MANAGEMENT_ENDPOINTS_WEB_BASE_PATH: "/"`, with no `MANAGEMENT_ENDPOINTS_WEB_EXPOSURE_INCLUDE`, so Spring Boot's default applies: health and info only.

Upstream does not solve this either — its services have the same gap. So something must first *produce* metrics.

## Why the OTEL agent — and why it is smaller than it looks

The Ansible tier does not use micrometer. It attaches the OpenTelemetry Java agent, which emits JVM metrics **and** traces down one pipeline:

```yaml
JAVA_TOOL_OPTIONS: -javaagent:/otel/opentelemetry-javaagent.jar
OTEL_TRACES_EXPORTER: otlp
OTEL_METRICS_EXPORTER: otlp
[...]

Issue #1619, opened 2026-08-05
Bring Kubernetes monitoring to parity with the Ansible tier (production/deploy-as-code)


singhalkarun:
**Goal: bring Kubernetes monitoring to parity with the Ansible tier.**

`devops/deploy-as-code` is the **production path for large-scale deployments**; `local-setup/ansible` (Docker Compose) is for small-scale. This issue is the context and rationale for the parity work; the individual gaps are separate issues, listed below.

> **Correction (kept deliberately).** An earlier revision framed `deploy-as-code` as a "legacy" tier and offered "retire it" as an option. That was wrong, and the reasoning that produced it is worth recording so it is not repeated: the monitoring helmfile is commented out, the values are stale, and nothing in CI touches the directory — which reads as abandonment until you learn it is the production path.

---

## Headline conclusion: adopt the upstream stack, do not rebuild it

Tracing the provenance settled both the "why is it off?" question and the "what should we build?" question.

**`deploy-as-code` was imported wholesale from upstream `egovernments/DIGIT-DevOps`** in commit `2f5649c9` (2025-07-09) — a prune of 2,323 files, +25,199 / −102,444 lines. The monitoring line **arrived already commented**, immediately beside:

```
#  - path: ./charts/sanitation/sanitation-helmfile.yaml
#  - path: ./charts/monitoring/monitoring-helmfile.yaml
```

`sanitation` is an unrelated DIGIT business module. So the file's convention is "commented = a module this deployment does not use" — **it was never a local decision to disable monitoring.** Since the import, exactly one commit has touched `charts/monitoring/`: the `Devops` → `devops` rename. Upstream still ships the `monitoring` chart directory, so this is maintained code, not abandoned code.

That import also deleted the only README under `deploy-as-code`, which is why none of this was discoverable in-tree.

It also explains the stale values that recur across these issues — `unified-qa`, `unified-uat.digit.org`, `#unified-qa-alerts`, a `pgr-demo-db` RDS host. They are another environment's settings, carried in by the import and never adapted. Nobody left them behind; nobody ever went through them.

**Therefore: the work is adoption and configuration, not construction.** Deleting the stack would diverge from upstream and mean rebuilding by hand roughly 80% of the parity target that is already sitting in the tree.

---

## What that means in practice

| | |
|---|---|
[...]

PR #1662, opened 2026-08-07
feat(kafka): enable broker and consumer-lag metrics

singhalkarun:
Fixes #1647

## What

Kafka is installed and serving and produces **no metrics**. The chart ships both a kafka exporter and a JMX exporter; both are disabled. So the message bus every DIGIT service depends on has no visibility at all.

## Why consumer lag is the point

`egov-persister` is the sole Kafka→Postgres write path and runs a **single replica**. When it stalls:

- the API still returns **200** with a complaint ID
- the pod stays **Running** and healthy
- the record is **not in the database**

Nothing in this tier would show that today. `egov-indexer` has the same shape for the search index.

**Only the kafka exporter yields `kafka_consumergroup_lag`.** Consumer lag is not a broker JMX MBean — it comes from the consumer-group offsets API. Enabling JMX alone would give broker health (under-replicated partitions, request rates, JVM) and leave lag invisible, which would look like a fix and not be one. Both are enabled.

## Two implementation details worth reviewing

**1. Scoped to this release, not `env.yaml`.** The chart reads `.Values.metrics.*`. A top-level `metrics:` key in `env.yaml` is injected into **every** release, and `charts/common/templates/_service.yaml` emits `prometheus.io/*` scrape annotations for any chart that defines `metrics` — so it would silently touch all 41 application services. Same class of leak as the `controller:` key in #1648. Using the release's existing `set:` block avoids it.

**2. The ServiceMonitor carries `release=kube-prometheus-stack`.** Prometheus runs with `serviceMonitorSelectorNilUsesHelmValues: true` and an empty selector, which renders as `matchLabels {release: kube-prometheus-stack}`. Without that label the ServiceMonitor is created and **never scraped** — the trap documented in #1646. Note this chart spells the key `labels`; ingress-nginx spells it `additionalLabels` (see #1648).

## Cost, and both new containers are bounded

Two additional containers: a kafka-exporter Deployment and a jmx-exporter sidecar on the broker pod. Both default to `resources: {limits: {}, requests: {}}` in this chart, so review rightly pushed back on adding two more uncapped containers next to #1612's finding about the broker itself. They now carry:

| Container | requests | limits |
|---|---|---|
| kafka-exporter (Go) | 50m / 64Mi | 200m / 128Mi |
| jmx-exporter (JVM) | 50m / 128Mi | 250m / 384Mi |

384Mi is deliberately loose. The chart hardcodes `-XX:MaxRAMPercentage=100` on that container (`kafka-kraft/templates/controller-eligible/statefulset.yaml:326`) and does not expose it as a value, so the cgroup limit *is* the JVM's maximum heap, with nothing left for metaspace, thread stacks or GC structures. Worth knowing before anyone tightens it.

The broker container is untouched — that is #1612's call, and capping a broker is a different risk conversation from capping its exporters.

## Verified / not verified

- [x] Both exporter templates gate on exactly the keys set here (`metrics.kafka.enabled` + `metrics.serviceMonitor.enabled`)
- [x] `metrics.serviceMonitor.labels` is a real key in this chart's values schema
- [x] Confirmed a top-level `metrics:` key would leak into `common/_service.yaml` — the reason for the `set:` approach
- [x] Rendered: two Services (`kafka-metrics`, `kafka-jmx-metrics`), a `kafka-exporter` Deployment and **two** ServiceMonitors — the jmx sidecar gets its own from `templates/metrics/jmx-servicemonitor.yaml` — both labelled `release: kube-prometheus-stack`, and both exporters carrying requests/limits
- [x] `helmfile v0.140.0 build` on the backbone helmfile: all thirteen `set:` entries resolve as intended
- [ ] `kafka_consumergroup_lag` present in Prometheus, with `egov-persister` and `egov-indexer` groups visible
- [ ] Inducing lag (pause the persister, publish, resume) shows it rise and drain

## Not included

- **A lag alert.** A dashboard nobody watches will not catch a stalled persister. That belongs with the alerting decision in #1619.
- **A dashboard.** The Ansible tier's board (#1628) is for Redpanda and will not match Kafka's metric names.
- **The partition/replication defaults.** Topics auto-create with 1 partition and RF=1, so the persister cannot be scaled out to drain lag even once it is visible. Tracked in the issue.

### DIFF
--- a/devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
+++ b/devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
@@ -117,6 +117,30 @@ releases:
   set:
     - name: kraft.clusterId
       value: {{ .Values.secrets.kafka.clusterID }}
+    # Broker metrics (#1647). Scoped to THIS release on purpose: the chart
+    # reads .Values.metrics.*, and a top-level `metrics:` key in env.yaml would
+    # be injected into every release — including charts/common/_service.yaml,
+    # which emits prometheus.io/* scrape annotations for any chart that sets
+    # `metrics`, i.e. all 41 application services.
+    #
+    # kafka exporter is the one that yields consumer-group LAG
+    # (kafka_consumergroup_lag). Lag is the failure that matters here:
+    # egov-persister is the sole Kafka->Postgres write path and runs a single
+    # replica, so when it stalls the API still returns 200 while complaints are
+    # never persisted. JMX gives broker health (under-replicated partitions,
+    # request rates) but NOT lag — both, or the exercise misses its point.
+    - name: metrics.kafka.enabled
+      value: true
+    - name: metrics.jmx.enabled
+      value: true
+    # Prometheus only selects ServiceMonitors labelled with its release name:
+    # serviceMonitorSelectorNilUsesHelmValues is true with an empty selector,
+    # which renders as matchLabels {release: kube-prometheus-stack}. Without
+    # this label the object is created and silently never scraped.
+    - name: metrics.serviceMonitor.enabled
+      value: true
+    - name: metrics.serviceMonitor.labels.release
+      value: kube-prometheus-stack
 
 - name: ingress-nginx
   installed: true
--- a/devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
+++ b/devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
@@ -114,6 +114,46 @@ releases:
   values:
     - ./kafka-kraft/values.yaml
     - {{ .Values | toYaml | nindent 8 }}
+    # Exporter images (#1647). BOTH images the chart pins were retired from
+    # Docker Hub -- bitnami/jmx-exporter:0.19.0-debian-11-r95 and
+    # bitnami/kafka-exporter:1.7.0-debian-11-r132 are 404 today -- so enabling
+    # metrics without this turns a working release into ImagePullBackOff.
+    # Verified on a cluster; every plausible source was checked first:
+    #
+    #   jmx   -> bitnamilegacy/*, where Bitnami moved the legacy tags. Same
+    #            pinned tag, still published (updated 2025-07-02).
+    #   kafka -> danielqsj/*, the upstream project Bitnami repackaged, at the
+    #            SAME version. NOT egovio/kafka-exporter: that mirror carries
+    #            exactly one tag, `latest`, last pushed 2020 -- unpinnable and
+    #            six years stale. bitnamilegacy/kafka-exporter does not exist.
+    #
+    # The durable fix is mirroring both into egovio/ with real version tags, as
+    # was done for the broker (egovio/bitnami-kafka:3.6.0-debian-11-r0); that
+    # needs registry push rights this change does not have.
+    - metrics:
+        jmx:
+          image:
+            repository: bitnamilegacy/jmx-exporter
+        kafka:
+          image:
+            repository: danielqsj/kafka-exporter
+            tag: v1.7.0
+          # The chart runs the exporter through a shell: `command: [bash]` with
+          # every flag in one `-ce` string. Bitnami's image had bash; the
+          # upstream one is a scratch image whose entrypoint IS the binary, so
+          # under the chart's default it exits 128 with no logs. Invoking it
+          # directly is what makes a shell-less image usable.
+          #
+          # This repeats the chart's own broker list, so it must track
+          # controller.replicaCount (3 in kafka-kraft/values.yaml). Guarded by
+          # the fact that a wrong list shows up immediately as a down target.
+          command:
+            - /bin/kafka_exporter
+          args:
+{{- range $i := until 3 }}
+            - --kafka.server=kafka-kraft-controller-{{ $i }}.kafka-kraft-controller-headless.backbone.svc.cluster.local:9092
+{{- end }}
+            - --web.listen-address=:9308
   set:
     - name: kraft.clusterId
       value: {{ .Values.secrets.kafka.clusterID }}
--- a/devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
+++ b/devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
@@ -141,6 +141,35 @@ releases:
       value: true
     - name: metrics.serviceMonitor.labels.release
       value: kube-prometheus-stack
+    # Both exporters default to `resources: {limits: {}, requests: {}}`, so
+    # without these they would join the broker as unbounded containers -- the
+    # pattern #1612 flags -- and the jmx one is a sidecar on the broker pod
+    # itself, where an unbounded exporter competes with the broker for the
+    # node. Sized for what they are: kafka-exporter is a small Go binary that
+    # polls consumer offsets; jmx-exporter is a JVM.
+    - name: metrics.kafka.resources.requests.cpu
+      value: 50m
+    - name: metrics.kafka.resources.requests.memory
+      value: 64Mi
+    - name: metrics.kafka.resources.limits.cpu
+      value: 200m
+    - name: metrics.kafka.resources.limits.memory
+      value: 128Mi
+    # 384Mi and not tighter because the chart hardcodes
+    # `-XX:MaxRAMPercentage=100` on this container
+    # (kafka-kraft/templates/controller-eligible/statefulset.yaml:326, not
+    # exposed as a value), so the memory limit IS the JVM's max heap with no
+    # allowance for metaspace, thread stacks or GC structures. The live set of
+    # jmx_prometheus_httpserver is tens of MB; the headroom is there so a heap
+    # that expands under GC pressure cannot reach the cgroup ceiling.
+    - name: metrics.jmx.resources.requests.cpu
+      value: 50m
+    - name: metrics.jmx.resources.requests.memory
+      value: 128Mi
+    - name: metrics.jmx.resources.limits.cpu
+      value: 250m
+    - name: metrics.jmx.resources.limits.memory
+      value: 384Mi
 
 - name: ingress-nginx
   installed: true


devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
```diff
@@ -117,6 +117,30 @@ releases:
...
+    # which emits prometheus.io/* scrape annotations for any chart that sets
+    # `metrics`, i.e. all 41 application services.
+    #
+    # kafka exporter is the one that yields consumer-group LAG
+    # (kafka_consumergroup_lag). Lag is the failure that matters here:
+    # egov-persister is the sole Kafka->Postgres write path and runs a single
+    # replica, so when it stalls the API still returns 200 while complaints are
+    # never persisted. JMX gives broker health (under-replicated partitions,
+    # request rates) but NOT lag — both, or the exercise misses its point.
+    - name: metrics.kafka.enabled
```
vinothrallapalli-eGov, line 132: Confirmed via local `helm template` against the vendored `kafka-kraft` (Bitnami kafka 26.2.0) chart: `metrics.kafka.enabled=true` + `metrics.serviceMonitor.enabled=true` correctly renders the kafka-exporter Deployment, its Service, and a ServiceMonitor (`<release>-kafka-metrics`) scraping `/metrics` on the `http-metrics` port — these are real value paths the chart reads, all default `false` upstream. Good.

devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
```diff
@@ -117,6 +117,30 @@ releases:
...
+      value: true
+    - name: metrics.jmx.enabled
```
vinothrallapalli-eGov, line 134: Minor/non-blocking: `metrics.kafka.resources` and `metrics.jmx.resources` both default to `{limits: {}, requests: {}}` in this chart, so the new kafka-exporter Deployment and jmx-exporter sidecar land with no requests/limits — the same unbounded-container pattern #1612 already flagged for the broker itself. Worth setting alongside that fix rather than adding two more uncapped containers to broker nodes.

vinothrallapalli-eGov (commented):
Verified beyond static reading: pulled the vendored `kafka-kraft` chart (Bitnami `kafka` 26.2.0, real KRaft-mode Apache Kafka, not Redpanda) and ran `helm template` locally with these exact `set:` overrides. Both ServiceMonitors (`*-kafka-metrics` and `*-kafka-jmx-metrics`) render with `release: kube-prometheus-stack`, the kafka-exporter Deployment renders, and the jmx-exporter sidecar renders inside the broker StatefulSet — so every value path this PR writes is one the chart actually reads (`metrics.kafka.enabled`, `metrics.jmx.enabled`, `metrics.serviceMonitor.enabled/labels`), all default `false`/`{}` upstream, exactly as claimed. The `metrics:` top-level-key leak into `charts/common/templates/_service.yaml` is real (that template does `{{- if .Values.metrics }}` and expects a different shape, `.path`/`.port`), so scoping via `set:` on this release rather than `env.yaml` is the right call. `serviceMonitorNamespaceSelector: {}` in the monitoring chart confirms the `backbone` namespace isn't excluded. The consumer-lag reasoning is also correct and appropriately distinguished from #1678's Redpanda approach (native `redpanda_kafka_*` metrics) — real Kafka has no lag MBean, so kafka-exporter is genuinely required here, not just mirrored for consistency. One minor, non-blocking item: both new exporters inherit `resources: {limits: {}, requests: {}}` from the chart defaults, same unbounded pattern #1612 flagged for the broker itself — worth a follow-up now that two more containers are landing on broker nodes.

devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
singhalkarun: Confirmed the same way, against the vendored `kafka-kraft` (Bitnami kafka 26.2.0) in-tree. `metrics.kafka.enabled`, `metrics.jmx.enabled` and `metrics.serviceMonitor.enabled` are all `false` in `values.yaml`, and enabling them renders:

```
Service         kafka-jmx-metrics
Service         kafka-metrics
Deployment      kafka-exporter
ServiceMonitor  kafka-jmx-metrics    (port http-metrics, path /,        release: kube-prometheus-stack)
ServiceMonitor  kafka-metrics        (port http-metrics, path /metrics, release: kube-prometheus-stack)
```

Two ServiceMonitors rather than one, both carrying the release label from `metrics.serviceMonitor.labels.release` — the jmx sidecar gets its own from `templates/metrics/jmx-servicemonitor.yaml`. Both selectors match the `app.kubernetes.io/component` labels on their respective Services (`metrics` and `cluster-metrics`).

devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
singhalkarun: Valid, and taken — 638eee4. Reproduced first: with the PR's values as they were, the render is

```
Deployment  kafka-exporter  container=kafka-exporter  resources={'limits': {}, 'requests': {}}
StatefulSet kafka-controller container=kafka          resources={'limits': {}, 'requests': {}}
StatefulSet kafka-controller container=jmx-exporter   resources={'limits': {}, 'requests': {}}
```

so exactly the two new containers you named, one of them a sidecar on the broker pod itself. Now pinned:

- kafka-exporter: 50m/64Mi requests, 200m/128Mi limits
- jmx-exporter: 50m/128Mi requests, 250m/384Mi limits

One thing that came out of sizing the jmx one, worth recording because it is a trap for anyone tightening these later. The chart hardcodes `-XX:MaxRAMPercentage=100` on that container (`kafka-kraft/templates/controller-eligible/statefulset.yaml:326`) and does not expose it as a value, so the cgroup memory limit doubles as the JVM's maximum heap with nothing set aside for metaspace, thread stacks or GC structures. 384Mi is deliberately loose for that reason; the live set of `jmx_prometheus_httpserver` is tens of MB, so the headroom exists purely so an expanding heap cannot reach the ceiling.

The broker container is left alone — that's #1612's call, and capping a broker is a different risk conversation from capping its exporters.

singhalkarun:
## Cluster verification: works, but **it cannot deploy as written** — the pinned exporter images are gone

Ran this on a real cluster (kind on the devbox, kube-prometheus-stack 56.6.2 from the actual monitoring helmfile). Good news first, then a blocker that `helm template` cannot see.

### Confirmed working

- Both ServiceMonitors created, both labelled `release: kube-prometheus-stack`, and **all four targets scraping**:
  ```
  job=kafka-kraft-jmx-metrics  health=up  (x3, one per controller)
  job=kafka-kraft-metrics      health=up
  ```
- The resource pins land on exactly the two new containers; the broker keeps `{}` (still #1612's call):
  ```
  kafka-exporter  {limits:{cpu:200m,memory:128Mi} requests:{cpu:50m,memory:64Mi}}
  jmx-exporter    {limits:{cpu:250m,memory:384Mi} requests:{cpu:50m,memory:128Mi}}
  kafka           {}
  ```
- **Consumer lag genuinely works.** Created a topic, produced 5, consumed 2, and Prometheus agrees with the CLI exactly:
  ```
  $ kafka-consumer-groups.sh --describe --group verify-group
    verify-group  verify-lag  0   CURRENT-OFFSET 2   LOG-END-OFFSET 5   LAG 3

  kafka_consumergroup_lag{consumergroup="verify-group",topic="verify-lag"} = 3
  ```
  148 `kafka_*` series names in total.

### BLOCKER: both exporter images 404 on Docker Hub

```
Failed to pull image "docker.io/bitnami/jmx-exporter:0.19.0-debian-11-r95":
  docker.io/bitnami/jmx-exporter:0.19.0-debian-11-r95: not found
```

Bitnami retired the legacy tags. Checked every plausible source:

| Image | Status |
|---|---|
| `bitnami/jmx-exporter:0.19.0-debian-11-r95` | **MISSING** |
| `bitnami/kafka-exporter:1.7.0-debian-11-r132` | **MISSING** |
| `bitnamilegacy/jmx-exporter:0.19.0-debian-11-r95` | EXISTS |
| `bitnamilegacy/kafka-exporter:*` | **repo does not exist** |
| `public.ecr.aws/bitnami/kafka-exporter:*` | MISSING |
| `egovio/bitnami-kafka:3.6.0-debian-11-r0` (broker) | EXISTS |

The broker survives only because someone already re-pointed it at the `egovio/` mirror. The exporters were never mirrored — they were never enabled. **So this PR turns a working release into a `ImagePullBackOff` on first deploy.** That is the whole reason to run these on a cluster.

### A validated fix

`bitnamilegacy` covers jmx. For kafka-exporter there is no Bitnami image anywhere, and the drop-in candidates fail for a non-obvious reason: the chart invokes the exporter **through a shell** —

```yaml
command: [bash]
args: ["-ce", "kafka_exporter \\\n --kafka.server=..."]
```

— and both `danielqsj/kafka-exporter` and `egovio/kafka-exporter` are shell-less (`Entrypoint=[/bin/kafka_exporter]`), so they `StartError` with exit 128 and no logs. The chart does expose `metrics.kafka.command` / `metrics.kafka.args`, and overriding them works:

```
metrics.jmx.image.repository = bitnamilegacy/jmx-exporter
metrics.kafka.image.repository = egovio/kafka-exporter
metrics.kafka.image.tag = latest
metrics.kafka.command[0] = /bin/kafka_exporter
metrics.kafka.args[*] = --kafka.server=<each controller>:9092, --web.listen-address=:9308
```

→ `kafka-kraft-exporter 1/1 Running`, target `up`, lag series present. That is the configuration all the evidence above came from.

**Two caveats before anyone commits that**, both worth a decision rather than my judgement:

1. `egovio/kafka-exporter:latest` is a **floating tag** — precisely the skew hazard #1023 documents. A versioned `egovio/` push (mirroring `bitnamilegacy/jmx-exporter` and a pinned kafka-exporter) is the cleaner fix and matches what was already done for the broker.
2. The `args` override **duplicates the chart's broker-list logic** in the helmfile, so it silently goes stale if `controller.replicaCount` changes.

### One operational side effect, unrelated to images

With `metrics.jmx.enabled=true` the broker container gets `JMX_PORT` set, which makes **every Kafka CLI tool fail inside the pod** — `kafka-topics.sh` and friends try to bind the already-occupied RMI port and die with a stack trace. Workaround is `JMX_PORT= kafka-topics.sh ...`. Worth a line in the runbook; it will confuse whoever first debugs a broker after this lands.

### Metric-name note for whoever writes the dashboard

The JMX names are not what you would guess. Under-replicated partitions is:

```
kafka_server_replicamanager_total_underreplicatedpartitions_value
```

not `kafka_server_replicamanager_underreplicatedpartitions`. Writing that panel from memory yields a blank one.

vinothrallapalli-eGov (approved):

PR #1663, opened 2026-08-07
feat(gateway): scrape ingress-nginx metrics

singhalkarun:
Part 1 of #1648.

> **Revised after review.** The first version of this PR pinned the values with a release-scoped `set:` and did not work: through helmfile it rendered **zero** metrics objects. Cause and fix below; the review threads have the full trail ([1](https://github.com/egovernments/Citizen-Complaint-Resolution-System/pull/1663#discussion_r3748087599), [2](https://github.com/egovernments/Citizen-Complaint-Resolution-System/pull/1663#discussion_r3748089346)).

## What

Every citizen request crosses the ingress-nginx controller, and it contributes nothing to Prometheus — so "are we returning errors, and on which route?" can only be answered by reading logs.

## The metrics are exposed by the image; nothing collects them

| Setting | env.yaml | Chart default | Effective before this PR |
|---|---|---|---|
| `controller.metrics.enabled` | `false` | `true` | **`false`** — env.yaml wins, see below |
| `controller.metrics.serviceMonitor.enabled` | `true` | `false` | irrelevant while metrics are off |
| `controller.metrics.prometheusRule.enabled` | `true` | `false` | gated off by `metrics.enabled` |

The static scrape job in `values/prometheus.yaml` targets `nginx-ingress-metrics.egov:10254`, while the Service is `<release>-ingress-nginx-controller-metrics` — a name that has never resolved. So there was no collection path at all.

## Why `set:` did not work

The vendored chart carries a **local modification** to `templates/_helpers.tpl` (lines 6-9), absent from upstream ingress-nginx 4.10.0:

```gotemplate
```

`mergeOverwrite` mutates its destination in place, and the destination is the root context. The first template that renders a label therefore merges `.Values["ingress-nginx"]` over `.Values` for the entire render — at **template** time, which is after every values file and after `--set`.

Reproduced minimally with `helm template`:

| Inputs | metrics Service rendered |
|---|---|
| `ingress-nginx.controller.metrics.enabled: false` + `--set controller.metrics.enabled=true` | **no** |
| `--set controller.metrics.enabled=true`, no `ingress-nginx:` key | yes |
| same as row 1, chart renamed to `zzz-nginx` | yes |

So `env.yaml`'s `false` beat the release's `set:`. Row 3 pins the behaviour to `.Chart.Name` and rules out anything Helm does natively.

## What this PR now does

Writes the pins into the `ingress-nginx:` subtree — the path that actually wins — and corrects `env.yaml` so the operator-facing file agrees with the deployed state instead of contradicting it:

```yaml
- ingress-nginx:
    controller:
      metrics:
        enabled: true
        serviceMonitor:
          enabled: true
          additionalLabels:
            release: kube-prometheus-stack
        prometheusRule:
          enabled: false
```

`release: kube-prometheus-stack` is required by Prometheus's default selector (`serviceMonitorSelectorNilUsesHelmValues: true` with an empty selector). Note this chart spells the key `additionalLabels`; kafka-kraft (#1662) spells it `labels`.

This cannot move to a root `controller:` key either: the subtree would overwrite it, and **kafka-kraft reads `.Values.controller` too** (13 templates, its own controller block) off the same shared passthrough.

## `prometheusRule` must be pinned false

`env.yaml` asks for `true`, and — per the hoist above — that value was live all along. It was harmless only because the template is gated on `metrics.enabled`, which was false. **Turning metrics on un-gates it**, and the chart's `prometheusRule.rules` is `[]`, so the template emits the key with no body:

```yaml
kind: PrometheusRule
spec: null
```

The CRD (`crd-prometheusrules.yaml`, kube-prometheus-stack 56.6.2) declares `required: ['spec']` with `spec.type: object`, so the API server rejects that and the whole release apply fails. Left alone, this PR would have broken the deploy rather than merely rendering an empty rule.

## Deliberately not in this PR

The security headers and TLS restrictions in the same `env.yaml` block — that is #1679, stacked on this branch. It also corrects two values naming a foreign environment (`Access-Control-Allow-Origin`, `setSslTls`), which need per-key review rather than a wholesale apply.

## Verified / not verified

- [x] `helmfile v0.140.0 -e env -l name=ingress-nginx template` renders `controller-service-metrics` and a ServiceMonitor labelled `release: kube-prometheus-stack`, and no PrometheusRule
- [x] The same command on the previous commit renders **none** of those — the regression this revision fixes
- [x] `_helpers.tpl` hoist diffed against upstream 4.10.0 and isolated with a chart rename
- [x] `.github/scripts/check-helm-values-paths.py` (#1652) against this branch: no new unread key paths
- [ ] **Not verified on a cluster** — this tier has none
- [ ] `nginx_ingress_controller_requests` present in Prometheus
- [ ] Per-route status codes and latency visible
- [ ] The stale `nginx-ingress-metrics.egov` static job removed or corrected once the ServiceMonitor works (#1649 replaces that list)

### DIFF
--- a/devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
+++ b/devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
@@ -127,31 +127,50 @@ releases:
     - ./ingress-nginx/values.yaml
     - {{ .Values | toYaml | nindent 8 }}
     - domain: {{ .Values.global.domain }}
-  # Gateway metrics (#1648). Every citizen request crosses this controller and
-  # it contributes nothing to Prometheus, so "are we returning errors, and on
-  # which route?" can only be answered by reading logs.
-  #
-  # Set HERE, not in env.yaml. This is a vendored upstream chart that reads
-  # .Values.controller.* directly (39 templates) with no chart-name hoisting,
-  # so env.yaml's `ingress-nginx:` block lands one level too deep and is
-  # ignored in full — which is the wider defect #1648 tracks. And it must not
-  # move to a top-level `controller:` key either: kafka-kraft reads
-  # .Values.controller too (13 templates, its own controller block), so the
-  # override would leak into the broker.
-  set:
-    # Already effectively true via the chart default, but the env.yaml block
-    # tries to set it FALSE. Pinning it here means the deployed state stops
-    # depending on which of two ignored/honoured paths wins.
-    - name: controller.metrics.enabled
-      value: true
-    - name: controller.metrics.serviceMonitor.enabled
-      value: true
-    # Prometheus selects only ServiceMonitors carrying its release name
-    # (serviceMonitorSelectorNilUsesHelmValues true + empty selector renders as
-    # matchLabels {release: kube-prometheus-stack}). Without this the object
-    # exists and is never scraped.
-    - name: controller.metrics.serviceMonitor.additionalLabels.release
-      value: kube-prometheus-stack
+    # Gateway metrics (#1648). Every citizen request crosses this controller and
+    # it contributes nothing to Prometheus, so "are we returning errors, and on
+    # which route?" can only be answered by reading logs.
+    #
+    # Written into the `ingress-nginx:` subtree, and NOT via `set:`, because of
+    # a local modification to this vendored chart. templates/_helpers.tpl:6-9
+    # (absent from upstream 4.10.0) does, inside "ingress-nginx.name":
+    #
+    #   $envOverrides := index .Values .Chart.Name
+    #   mustMergeOverwrite . (dict "Values" (mustMergeOverwrite $baseValues $envOverrides))
+    #
+    # mergeOverwrite mutates its destination, and the destination here is the
+    # root context -- so the first template to render a label permanently
+    # merges .Values["ingress-nginx"] OVER .Values for the whole render. That
+    # happens at TEMPLATE time, after every values file and after --set, so
+    # anything named in env.yaml's block wins over both. Verified: with
+    # env.yaml present, `--set controller.metrics.enabled=true` renders NO
+    # metrics Service, because env.yaml says `enabled: false` one level down.
+    #
+    # Which is also why this cannot move to a root `controller:` key: the
+    # subtree would still overwrite it, and kafka-kraft reads .Values.controller
+    # for its own broker block off the same shared passthrough.
+    - ingress-nginx:
+        controller:
+          metrics:
+            enabled: true
+            serviceMonitor:
+              enabled: true
+              # Prometheus selects only ServiceMonitors carrying its release
+              # name (serviceMonitorSelectorNilUsesHelmValues true + empty
+              # selector renders as matchLabels {release: kube-prometheus-stack}).
+              # Without this the object exists and is never scraped.
+              additionalLabels:
+                release: kube-prometheus-stack
+            # MUST stay false while `prometheusRule.rules` is [] in
+            # ingress-nginx/values.yaml. The template emits `spec:` with no
+            # body when the list is empty, so the object renders as
+            # `spec: null` -- and the PrometheusRule CRD marks spec required
+            # and typed object, so the API server rejects it and the whole
+            # release apply fails. env.yaml asks for true; it was harmless only
+            # because metrics.enabled gated it off. Turning metrics on is what
+            # makes it dangerous.
+            prometheusRule:
+              enabled: false
 
 - name: novu
   installed: true
--- a/devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
+++ b/devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
@@ -127,6 +127,31 @@ releases:
     - ./ingress-nginx/values.yaml
     - {{ .Values | toYaml | nindent 8 }}
     - domain: {{ .Values.global.domain }}
+  # Gateway metrics (#1648). Every citizen request crosses this controller and
+  # it contributes nothing to Prometheus, so "are we returning errors, and on
+  # which route?" can only be answered by reading logs.
+  #
+  # Set HERE, not in env.yaml. This is a vendored upstream chart that reads
+  # .Values.controller.* directly (39 templates) with no chart-name hoisting,
+  # so env.yaml's `ingress-nginx:` block lands one level too deep and is
+  # ignored in full — which is the wider defect #1648 tracks. And it must not
+  # move to a top-level `controller:` key either: kafka-kraft reads
+  # .Values.controller too (13 templates, its own controller block), so the
+  # override would leak into the broker.
+  set:
+    # Already effectively true via the chart default, but the env.yaml block
+    # tries to set it FALSE. Pinning it here means the deployed state stops
+    # depending on which of two ignored/honoured paths wins.
+    - name: controller.metrics.enabled
+      value: true
+    - name: controller.metrics.serviceMonitor.enabled
+      value: true
+    # Prometheus selects only ServiceMonitors carrying its release name
+    # (serviceMonitorSelectorNilUsesHelmValues true + empty selector renders as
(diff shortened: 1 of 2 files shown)


devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
```diff
@@ -127,6 +127,31 @@ releases:
...
+    - name: controller.metrics.enabled
+      value: true
+    - name: controller.metrics.serviceMonitor.enabled
+      value: true
+    # Prometheus selects only ServiceMonitors carrying its release name
+    # (serviceMonitorSelectorNilUsesHelmValues true + empty selector renders as
+    # matchLabels {release: kube-prometheus-stack}). Without this the object
+    # exists and is never scraped.
+    - name: controller.metrics.serviceMonitor.additionalLabels.release
+      value: kube-prometheus-stack
```
vinothrallapalli-eGov: Confirmed correct: rendering this chart locally with `controller.metrics.serviceMonitor.additionalLabels.release=kube-prometheus-stack` produces a ServiceMonitor labeled `release: kube-prometheus-stack`, which is exactly what `kube-prometheus-stack`'s `serviceMonitorSelectorNilUsesHelmValues: true` + empty `serviceMonitorSelector` requires for auto-discovery (verified against `charts/monitoring/values/prometheus.yaml` and the release name `kube-prometheus-stack` in `monitoring-helmfile.yaml`). Not a silent no-op like the pattern this initiative has been finding elsewhere.

vinothrallapalli-eGov (commented):
Verified statically (no cluster access, but confirmed with a local `helm template` render of the vendored `ingress-nginx` chart using these exact `--set` values): `controller-servicemonitor.yaml` renders a valid ServiceMonitor carrying `release: kube-prometheus-stack`, which matches this cluster's actual discovery mechanism — `charts/monitoring/values/prometheus.yaml` sets `serviceMonitorSelectorNilUsesHelmValues: true` with an empty `serviceMonitorSelector`, and the Prometheus Operator release is named `kube-prometheus-stack` in `monitoring-helmfile.yaml`. So this does not fall into the 'looks configured, isn't' class this initiative keeps finding — the target will actually be picked up. On the second focus question: the metrics Service stays `ClusterIP` (confirmed via the same render of `controller-service-metrics.yaml`), so `/metrics` is not newly exposed on the internet-facing LB — this is not a repeat of #1602-#1608. One pre-existing, out-of-scope item worth a follow-up: `charts/environments/env.yaml`'s `prometheus.additionalScrapeConfigs` still has a static `nginx-ingress-metrics` job pointed at `release-name-ingress-nginx-controller-metrics.egov:10254` (an unsubstituted 'release-name' placeholder, never resolvable) — this PR correctly leaves it alone but it will now sit as a permanently-down duplicate target once the ServiceMonitor is live; worth removing in a follow-up. Also checked overlap with #1679: it's stacked directly on this PR's branch (`base=feat/1648-gateway-metrics`) and only adds a new `values:` list entry for `controller.replicaCount`/`addHeaders`/`config` — a disjoint subtree from this PR's `set:` overrides (`controller.metrics.*`) — so the two compose via deep-merge with no key collision and no textual conflict when merged in sequence.

devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
```diff
@@ -127,6 +127,31 @@ releases:
...
+  # which route?" can only be answered by reading logs.
+  #
+  # Set HERE, not in env.yaml. This is a vendored upstream chart that reads
+  # .Values.controller.* directly (39 templates) with no chart-name hoisting,
+  # so env.yaml's `ingress-nginx:` block lands one level too deep and is
+  # ignored in full — which is the wider defect #1648 tracks. And it must not
+  # move to a top-level `controller:` key either: kafka-kraft reads
+  # .Values.controller too (13 templates, its own controller block), so the
+  # override would leak into the broker.
+  set:
```
vinothrallapalli-eGov: Minor, non-blocking: `controller.metrics.prometheusRule.enabled` (which env.yaml's ignored block also intends as `true`) isn't pinned here alongside `metrics.enabled`/`serviceMonitor.enabled`. Not actually consequential though — the chart's `prometheusRule.rules` list is empty (`[]`) in this vendored chart's values, so even enabling it would render a PrometheusRule with no rule groups. Flagging only for completeness since this PR is explicit about pinning every path env.yaml fails to reach.

devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
singhalkarun: Your render is right, and it's also the render that hid the bug — thank you for stating the method, because repeating it *with `env.yaml` in the picture* is what exposed it. Fixed in cea5667.

Standalone, exactly as you had it, the ServiceMonitor appears with `release: kube-prometheus-stack`. Through the real helmfile it did not appear at all:

```
$ helmfile -f backboneservices-helmfile.yaml -e env -l name=ingress-nginx template   # at 6f1c999
   0 matches for controller-service-metrics / servicemonitor
```

Zero metrics objects, despite all three `set:` entries. Cause is a **local modification to the vendored chart** — `templates/_helpers.tpl:6-9`, which I diffed against upstream ingress-nginx 4.10.0 and which upstream does not have. Inside `ingress-nginx.name`:

```gotemplate
```

`mergeOverwrite` mutates its destination, and the destination is the root context, so the first template that renders a label permanently merges `.Values["ingress-nginx"]` over `.Values` for the whole render. That happens at **template** time — after every values file and after `--set`. Minimal reproduction:

```
values file with only `ingress-nginx.controller.metrics.enabled: false`, plus --set controller.metrics.enabled=true  -> 0 metrics Services
same --set, no `ingress-nginx:` key at all                                                                          -> 1 metrics Service
chart renamed to zzz-nginx, same inputs                                                                             -> 1 metrics Service
```

So env.yaml's `controller.metrics.enabled: false` was beating the `set:`. The pins now live in the `ingress-nginx:` subtree, which is the path that actually wins, and env.yaml is corrected to agree rather than contradict. Through helmfile the release now renders `controller-service-metrics` plus a ServiceMonitor labelled `release: kube-prometheus-stack`.

This also means your point on #1679 is right for a sharper reason than Helm's coalescing — replied there.

devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
singhalkarun: Right that it's unpinned and right that `prometheusRule.rules` is `[]` — but the consequence is worse than "renders an empty PrometheusRule", and chasing this comment is what turned up the bigger defect on the sibling thread. Now handled explicitly in cea5667.

Two corrections to "not actually consequential":

**1. env.yaml's `prometheusRule.enabled: true` was never ignored.** The vendored chart's locally-patched `_helpers.tpl:6-9` hoists the whole `ingress-nginx.controller.*` subtree onto the root context at template time (details on the other thread). It was inert only because `controller-prometheusrules.yaml` is gated on `and .Values.controller.metrics.enabled .Values.controller.metrics.prometheusRule.enabled` and `metrics.enabled` was false. **Turning metrics on is precisely what un-gates it.**

**2. With `rules: []` the object does not render empty — it renders invalid.** The template is:

```gotemplate
spec:
{{- if .Values.controller.metrics.prometheusRule.rules }}
  groups:
  ...
{{- end }}
```

so the key is emitted with no body:

```yaml
kind: PrometheusRule
metadata:
  name: ingress-nginx-controller
spec: null
```

and the CRD in kube-prometheus-stack 56.6.2 (`crd-prometheusrules.yaml`) has `required: ['spec']` with `spec.type: object`. The API server rejects `spec: null`, so this would have failed the whole release apply — not produced a harmless empty rule.

Pinned `false` with that reasoning recorded in both the helmfile and env.yaml, so nobody "completes" the pinning later. Post-fix render: metrics Service + ServiceMonitor, no PrometheusRule.

vinothrallapalli-eGov (approved):

Issue #1650, opened 2026-08-06
Grafana: GitHub OAuth client id is set under a non-existent env var, no admin password, and all dashboards are fetched from GitHub master at pod start


singhalkarun:
Part of #1619 (Kubernetes monitoring parity). **P1 · effort: S**

Two related defects in the same component. Mirrors #1602 (Grafana auth) and the vendoring pattern the Ansible tier uses for dashboards.

---

## Part 1 — Grafana login does not work as configured, and there is no fallback password

`charts/monitoring/monitoring-helmfile.yaml:68-70`:

```yaml
  - envRenderSecret:
      GF_AUTH_GITHUB_YAMLAPPLICATIONCONFIG_AUTH_OAUTH2_CLIENT_GITHUB_CLIENTID: {{ .Values.secrets.grafana.clientID }}
      GF_AUTH_GITHUB_CLIENT_SECRET: {{ .Values.secrets.grafana.clientSecret }}
```

Grafana's override convention is `GF_<SECTION>_<KEY>`. For `auth.github` / `client_id` that is **`GF_AUTH_GITHUB_CLIENT_ID`**. The name used here matches no Grafana setting and is ignored — note the *next line* uses the correct form for the secret, which makes it self-evidently a typo. The token `YAMLAPPLICATIONCONFIG_AUTH_OAUTH2_CLIENT_GITHUB_CLIENTID` is copy-paste debris from kafka-ui's `yamlApplicationConfig.auth.oauth2.client.github.clientId` block.

`GF_AUTH_GITHUB_CLIENT_ID` is set **nowhere else** — not in `values/grafana.yaml` (its `auth.github` block at `:836-847` has no `client_id`), not in `env.yaml`, not in `env-secrets.yaml` (which carries only `clientID`/`clientSecret` under `secrets.grafana`).

So GitHub OAuth cannot complete: with an empty client id, the provider either fails at github.com or Grafana declines to enable it.

**And there is no admin password to fall back to.** `values/grafana.yaml:415-422`:
```yaml
adminUser: admin
# adminPassword: strongpassword

admin:
  existingSecret: ""
```
Neither is set anywhere. The chart falls back to its own default behaviour for the `admin` account, on an instance published at `/monitoring` with **no ingress-level auth** (the annotations block at `:256-259` is entirely commented out).

The Ansible tier fixed the equivalent in #1602 — password from OpenBao, login form enabled, anonymous access off.

---

[...]

PR #1665, opened 2026-08-07
fix(grafana): set the OAuth client id under a real env var, and an admin password

singhalkarun:
Part 1 of #1650 — the auth half. Dashboards are not touched here.

## What

Grafana's GitHub login could not complete. The client id was passed as:

```
GF_AUTH_GITHUB_YAMLAPPLICATIONCONFIG_AUTH_OAUTH2_CLIENT_GITHUB_CLIENTID
```

That matches no Grafana setting and is silently ignored. Grafana's convention is `GF_<SECTION>_<KEY>`, so `auth.github` / `client_id` is **`GF_AUTH_GITHUB_CLIENT_ID`**.

The line immediately below it already used the correct form for the secret, which makes this copy-paste debris from kafka-ui's `yamlApplicationConfig.auth.oauth2.client.github.clientId` — the two values files sit in the same directory. So the provider had a secret and no client id.

## Why this is more than a broken login button

**There was no fallback either.** `adminPassword` is commented out in `values/grafana.yaml` and `admin.existingSecret` is empty, so the chart used its own default for the `admin` account — on an instance published at `/monitoring` with **no ingress-level auth** (the annotations block is entirely commented out).

So a broken OAuth provider left a local login form as the only door, with an unmanaged password behind it.

This PR adds `GF_SECURITY_ADMIN_PASSWORD` from `secrets.grafana.adminPassword`, and declares that key in `env-secrets.yaml` — the SOPS-covered file — so it is discoverable rather than implied.

## One judgement call worth reviewing

My first version defaulted the admin password to the OAuth `clientSecret` to avoid an unset value. I removed that. An OAuth client secret and a login password have different rotation cadences and different blast radius; reusing one for the other trades a visible gap for an invisible one. The key is now declared explicitly with a placeholder, so an operator sets it deliberately.

## Verified / not verified

- [x] Both files parse; `secrets.grafana` now carries `clientID`, `clientSecret`, `adminPassword`
- [x] `GF_AUTH_GITHUB_CLIENT_ID` is the documented Grafana variable for this setting
- [ ] GitHub OAuth completes end to end
- [ ] The admin account uses the configured password on a fresh install *and* on an existing one — note the Ansible-tier equivalent (#1605) needed an explicit reset step, because that variable only seeds a **new** Grafana database. The same caveat probably applies here and is worth checking before assuming an upgrade takes effect.

## Not in this PR

The other half of #1650: all eight dashboards are fetched from GitHub `master` at pod start, three of them from an account outside this organisation. That is a separate change — vendoring them in-repo — and #1650 stays open for it.

### DIFF
--- a/devops/deploy-as-code/charts/environments/env-secrets.yaml
+++ b/devops/deploy-as-code/charts/environments/env-secrets.yaml
@@ -112,13 +112,34 @@ secrets:
     grafana:
         clientID: cdabbaabcdefghi
         clientSecret: 8292723f0e1d596d234abc123def456
-        # Local admin login. Grafana is published at /monitoring with no
-        # ingress-level auth, so this is a real credential, not a fallback --
-        # set it per environment. Left unset, the chart uses its own default
-        # for the `admin` account (#1650). Deliberately NOT reusing
-        # clientSecret: an OAuth client secret and a login password are
-        # different secrets with different rotation and blast radius.
-        adminPassword: change-me-strong
+        # Local admin login (#1650). Grafana is published at /monitoring with
+        # no ingress-level auth, so this is a real credential, not a fallback.
+        #
+        # It is deliberately NOT written here. This file is committed and
+        # unencrypted -- there is no SOPS anywhere under devops/deploy-as-code,
+        # and .gitignore covers only the ansible tier's per-tenant secrets -- so
+        # a password put here enters git history and stays there after rotation.
+        #
+        # Instead, create a Kubernetes Secret in the monitoring namespace with
+        # keys `admin-user` and `admin-password` (sealed-secret,
+        # external-secret, or OpenBao/vault-synced -- whichever this
+        # environment already uses) and name it below. Grafana's chart reads
+        # both keys via secretKeyRef and then creates no Secret of its own:
+        #
+        #   kubectl -n monitoring create secret generic grafana-admin \
+        #     --from-literal=admin-user=admin \
+        #     --from-literal=admin-password="$(openssl rand -base64 24)"
+        #
+        adminExistingSecret: grafana-admin
+        # Escape hatch for throwaway clusters only: blank adminExistingSecret
+        # above and put a literal here. Anything written here is committed in
+        # cleartext -- treat it as public and rotate it.
+        #
+        # Leaving BOTH empty is a render error rather than a default, because
+        # the chart's own fallback is `randAlphaNum 40` evaluated at render
+        # time: every helmfile run would mint a different admin password and
+        # the operator would never hold a working one.
+        adminPassword: ""
     alertmanager:
         config:
             global:
--- a/devops/deploy-as-code/charts/environments/env-secrets.yaml
+++ b/devops/deploy-as-code/charts/environments/env-secrets.yaml
@@ -112,6 +112,13 @@ secrets:
     grafana:
         clientID: cdabbaabcdefghi
         clientSecret: 8292723f0e1d596d234abc123def456
+        # Local admin login. Grafana is published at /monitoring with no
+        # ingress-level auth, so this is a real credential, not a fallback --
+        # set it per environment. Left unset, the chart uses its own default
+        # for the `admin` account (#1650). Deliberately NOT reusing
+        # clientSecret: an OAuth client secret and a login password are
+        # different secrets with different rotation and blast radius.
+        adminPassword: change-me-strong
     alertmanager:
         config:
             global:
--- a/devops/deploy-as-code/charts/monitoring/monitoring-helmfile.yaml
+++ b/devops/deploy-as-code/charts/monitoring/monitoring-helmfile.yaml
@@ -113,15 +113,27 @@ releases:
   - envRenderSecret:
       GF_AUTH_GITHUB_CLIENT_ID: {{ .Values.secrets.grafana.clientID }}
       GF_AUTH_GITHUB_CLIENT_SECRET: {{ .Values.secrets.grafana.clientSecret }}
-{{- if not $grafanaAdminSecret }}
-      GF_SECURITY_ADMIN_PASSWORD: {{ $grafanaAdminPassword | quote }}
-{{- end }}
   set:
 {{- if $grafanaAdminSecret }}
     # Chart wires GF_SECURITY_ADMIN_USER/PASSWORD from this Secret's
     # admin-user / admin-password keys and creates no Secret of its own.
     - name: admin.existingSecret
       value: {{ $grafanaAdminSecret | quote }}
+{{- else }}
+    # `adminPassword`, NOT GF_SECURITY_ADMIN_PASSWORD via envRenderSecret.
+    #
+    # envRenderSecret becomes an `envFrom: secretRef` on the container, while
+    # the chart declares GF_SECURITY_ADMIN_PASSWORD explicitly in `env` from its
+    # own generated Secret. Kubernetes gives explicit `env` precedence over
+    # `envFrom`, so the value set here would be present in the pod's environment
+    # and simply lose. Verified on a cluster: the configured password returned
+    # 401 while the chart's generated one returned 200, and the two hashes
+    # differed -- the operator would set a password and be locked out by it.
+    #
+    # `adminPassword` is the value the chart itself writes into that Secret, so
+    # setting it here is what the explicit env then reads.
+    - name: adminPassword
+      value: {{ $grafanaAdminPassword | quote }}
 {{- end }}
     - name: ingress.hosts[0]
       value: {{ .Values.global.domain | quote }}
--- a/devops/deploy-as-code/charts/monitoring/monitoring-helmfile.yaml
+++ b/devops/deploy-as-code/charts/monitoring/monitoring-helmfile.yaml
@@ -5,6 +5,36 @@ environments:
       - ../environments/env-secrets.yaml
       - ../environments/env.yaml
 ---
+{{- /* Grafana admin credentials (#1650).
+
+       Two mutually exclusive paths, resolved once here so the release block
+       below stays readable:
+
+         adminExistingSecret  -- preferred. Names a Kubernetes Secret created
+                                 out of band (sealed-secret / external-secret /
+                                 OpenBao-synced) with keys admin-user and
+                                 admin-password. Nothing credential-shaped
+                                 lands in git, and rotation is a Secret update
+                                 rather than a commit.
+         adminPassword        -- escape hatch. Committed in cleartext to
+                                 env-secrets.yaml, which is neither encrypted
+                                 nor gitignored.
+
+       Read with `index` rather than `.Values.secrets.grafana.x` on purpose:
+       helmfile renders with missingkey=error, so the dotted form is a hard
+       parse failure against an env-secrets.yaml written before these keys
+       existed, while `index` yields "" for an absent key.
+
+       Neither set is a render error, not a silent default: with both empty the
+       chart falls back to `randAlphaNum 40` evaluated per render, so every
+       helmfile run would rewrite the admin Secret with a different password
+       and no operator would ever hold a working one. */}}
+{{- $gsec := .Values.secrets.grafana }}
+{{- $grafanaAdminSecret := index $gsec "adminExistingSecret" | default "" }}
+{{- $grafanaAdminPassword := index $gsec "adminPassword" | default "" }}
+{{- if and (not $grafanaAdminSecret) (not $grafanaAdminPassword) }}
+{{- fail "grafana admin login is unset: set secrets.grafana.adminExistingSecret (preferred) or secrets.grafana.adminPassword in charts/environments/env-secrets.yaml" }}
+{{- end }}
 repositories:
 - name: grafana
   url: https://grafana.github.io/helm-charts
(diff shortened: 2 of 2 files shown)


devops/deploy-as-code/charts/environments/env-secrets.yaml
```diff
@@ -112,6 +112,13 @@ secrets:
     grafana:
         clientID: cdabbaabcdefghi
         clientSecret: 8292723f0e1d596d234abc123def456
+        # Local admin login. Grafana is published at /monitoring with no
+        # ingress-level auth, so this is a real credential, not a fallback --
+        # set it per environment. Left unset, the chart uses its own default
+        # for the `admin` account (#1650). Deliberately NOT reusing
+        # clientSecret: an OAuth client secret and a login password are
+        # different secrets with different rotation and blast radius.
+        adminPassword: change-me-strong
```
vinothrallapalli-eGov: This is a literal, weak, guessable admin password committed in cleartext to a file that isn't SOPS-encrypted or gitignored anywhere in this repo (grep for `sops` under `devops/deploy-as-code` finds nothing, and `.gitignore` only excludes the ansible-tier per-tenant secrets, not this file). Every other secret in this file (db password, mail password, OAuth clientSecret, even an RSA private key) is already committed the same way, so this isn't a new pattern — but it is a chance to not extend it. Grafana's chart already supports `admin.existingSecret` / `admin.passwordKey` (see `values/grafana.yaml` ~L419-424) for pulling the admin password from a real Kubernetes Secret (sealed-secret / external-secret / vault-synced). That would avoid a real credential sitting in git history, and would match the stronger pattern #1605 used at the compose/ansible tier (OpenBao-generated password + Viewer-only escape hatch) instead of falling back to a hardcoded placeholder that operators may forget to rotate.

devops/deploy-as-code/charts/monitoring/monitoring-helmfile.yaml
```diff
@@ -65,9 +65,23 @@ releases:
...
+  # a secret and no client id: the provider could never complete a login.
+  #
+  # GF_SECURITY_ADMIN_PASSWORD is set from the same secrets file. Without it the
+  # chart falls back to its own default for the `admin` account, on an instance
+  # published at /monitoring with no ingress-level auth — so a broken OAuth
+  # provider left a local login form as the only door, with an unmanaged
+  # password behind it.
   - envRenderSecret:
-      GF_AUTH_GITHUB_YAMLAPPLICATIONCONFIG_AUTH_OAUTH2_CLIENT_GITHUB_CLIENTID: {{ .Values.secrets.grafana.clientID }}
+      GF_AUTH_GITHUB_CLIENT_ID: {{ .Values.secrets.grafana.clientID }}
```
vinothrallapalli-eGov, line 114: Correct fix. `GF_AUTH_GITHUB_CLIENT_ID` matches Grafana's `GF_<SECTION>_<KEY>` env override convention for `auth.github.client_id`, and is now consistent with the already-correct `GF_AUTH_GITHUB_CLIENT_SECRET` on the next line — good root-causing of the kafka-ui copy-paste origin.

vinothrallapalli-eGov (commented):
Static review only, per #1601 sweep methodology. The env var fix is correct: `GF_AUTH_GITHUB_CLIENT_ID` matches Grafana's documented `GF_<SECTION>_<KEY>` override convention (auth.github / client_id), and it's now consistent with the sibling `GF_AUTH_GITHUB_CLIENT_SECRET` that was already correct in this same block — good catch on the kafka-ui copy-paste debris. `GF_SECURITY_ADMIN_PASSWORD` is also the real, documented Grafana variable. One substantive concern (medium severity, inline below): the new `adminPassword` is a literal `change-me-strong` string committed to `env-secrets.yaml`. That file is not SOPS-encrypted or gitignored anywhere in this repo (no `sops` tooling found under `devops/deploy-as-code`, unlike the ansible-tier secrets which are OpenBao-generated and gitignored except for `_example.yml`), so this is effectively a hardcoded weak default in version control — a regression against the bar #1605 set (login-gated + OpenBao-generated password + Viewer-only fallback). The Grafana subchart already exposes `admin.existingSecret`/`passwordKey` (grafana.yaml ~L419-424) for referencing a real K8s Secret; that would have been the idiomatic fix instead of piping a plaintext value through `envRenderSecret`. Also worth tracking before merge: the PR's own checklist flags that `GF_SECURITY_ADMIN_PASSWORD` only seeds a *new* Grafana DB, so existing clusters may need the same manual reset step #1605 required for its ansible-tier equivalent.

devops/deploy-as-code/charts/environments/env-secrets.yaml
singhalkarun: Valid, and fixed in 0d4cb047.

Confirmed every part of the premise: `grep -rin sops devops/deploy-as-code` returns nothing, `.gitignore` has only `secrets/` (a directory that does not contain this file) plus the ansible tier's per-tenant entries, and `git ls-files` shows `charts/environments/env-secrets.yaml` tracked. So `change-me-strong` was a live credential in git history.

One correction to my own justification while I was in there, because it changes what the fix has to do. I wrote that without `GF_SECURITY_ADMIN_PASSWORD` "the chart falls back to its own default". It does not fall back to a *fixed* default — chart 7.3.0's `templates/secret.yaml` uses `randAlphaNum 40`, evaluated at render time:

```
$ helm template g grafana/grafana --version 7.3.0 | grep admin-password:
  admin-password: "YmVNVElucVVVVlBvMWw5TE83ZENjRDNwQTFnTTladVc2VnA5QXNlcA=="   # -> beMTInqUUVPo1l9LO7dCcD3pA1gM9ZuW6Vp9Asep
```

A different value on every run. So the pre-#1650 state was not "weak known password", it was "the admin password changes under the operator on every `helmfile apply`" — which is why a hardcoded placeholder felt like the fix and why it is not.

Took the mechanism you pointed at. `env-secrets.yaml` now names a Secret instead of holding a password:

```yaml
adminExistingSecret: grafana-admin
adminPassword: ""
```

Verified the chart honours it — with `admin.existingSecret` set it switches both env vars to `secretKeyRef` and creates no Secret of its own:

```
$ helm template g grafana/grafana --version 7.3.0 --set admin.existingSecret=grafana-admin
  - name: GF_SECURITY_ADMIN_USER
    valueFrom: {secretKeyRef: {name: grafana-admin, key: admin-user}}
  - name: GF_SECURITY_ADMIN_PASSWORD
    valueFrom: {secretKeyRef: {name: grafana-admin, key: admin-password}}
$ ... | grep -c 'admin-password:'
0
```

The Secret is produced however the environment already produces secrets — sealed-secret, external-secret, OpenBao-synced — which is the #1605 pattern you referenced, with the generation step outside this repo where it belongs.

Three details worth flagging:

- The inline password survives as an explicit escape hatch for throwaway clusters, documented as committed-in-cleartext, but it now defaults to `""` so nothing credential-shaped ships in git.
- **Both empty is a render error, not a default** — otherwise the `randAlphaNum` behaviour above quietly returns. The message names both keys and the file.
- Both are read with `index $gsec "..."` rather than `.Values.secrets.grafana.x`, because helmfile renders with `missingkey=error` and the dotted form is a hard parse failure against an `env-secrets.yaml` written before these keys existed. With `index`, that operator gets the guidance message instead of `map has no entry for key`.

Exercised all four paths with helmfile v0.140.0 `build`:

| env-secrets.yaml | result |
|---|---|
| shipped default | `admin.existingSecret: grafana-admin`, no password env var |
| `adminExistingSecret: ""` + a literal | `GF_SECURITY_ADMIN_PASSWORD: s3kr1t`, no `set` entry |
| both empty | `fail` with the guidance message |
| neither key present (legacy file) | same `fail`, not a parse error |

`check-helm-values-paths.py` is clean.

Not addressed here: the rest of this file. The RSA private key on line ~80 is the one I would not leave for a later pass — rotating it is a separate change with a blast radius outside this repo, so I have not folded it in.

devops/deploy-as-code/charts/monitoring/monitoring-helmfile.yaml
singhalkarun: Thanks — and one amendment now that I have checked the chart rather than assumed it.

The blast radius of the wrong env var name was larger than "OAuth could not complete". With no working `GF_SECURITY_ADMIN_PASSWORD` either, chart 7.3.0 falls back to `randAlphaNum 40` in `templates/secret.yaml`, evaluated per render — so the local admin password was regenerating on every `helmfile apply`. Broken SSO *and* a login nobody held the password to, on an instance published at `/monitoring` with no ingress-level auth.

The admin half of this has since moved to `admin.existingSecret` in 0d4cb047, per your other comment. The `GF_AUTH_GITHUB_CLIENT_ID` rename is unchanged.

singhalkarun:
## Cluster-verified: the OAuth half works, the admin-password half is inert

Ran this on a real cluster. Splitting the two changes because they land differently.

### OAuth client id — confirmed fixed

`GF_AUTH_GITHUB_CLIENT_ID` is present in the running Grafana container's environment, under the real Grafana name. The diagnosis in the PR body holds.

### Admin password — does not take effect

Two sources define `GF_SECURITY_ADMIN_PASSWORD` with **different values**, and the chart's own wins:

```
value the running process sees = 4f3721a9c246
  chart secret/grafana         = 4f3721a9c246   <- in force
  this PR's grafana-env        = 97bfcf35e759   <- inert
```

Authentication confirms it: the chart secret's password returns `200`, this PR's returns `401`.

The mechanism is Kubernetes precedence, not Grafana:

```
container env:     GF_SECURITY_ADMIN_PASSWORD <- secretKeyRef: grafana
container envFrom: secretRef: grafana-env
```

The chart declares the variable explicitly in `env` (from `secret/grafana`), and this PR supplies it via `envRenderSecret`, which becomes `envFrom`. **Explicit `env` beats `envFrom`**, so the value from `env-secrets.yaml` never reaches Grafana.

Net effect: the operator sets a password that silently does not apply, and the account is reachable only with the chart-generated one in `secret/grafana`. Not a security hole — I verified the effective password is **not** the chart's documented `strongpassword` default — but it defeats the stated intent, and an operator following this PR will be locked out of an instance they believe they configured.

Fix would be to set it where the chart reads it — `adminPassword` in `values/grafana.yaml`, or `admin.existingSecret` pointing at a secret you control — rather than through `envRenderSecret`.

singhalkarun:
## Correction: the admin-password finding above is wrong — it tested a stale checkout

Retracting the "admin password is inert" half of my previous comment. The OAuth
half stands; this half does not apply to this branch.

That comment describes a `grafana-env` Secret supplying
`GF_SECURITY_ADMIN_PASSWORD` via `envFrom`, losing to the chart's explicit
`env`. That was the **original** shape of this PR. It is not what the branch
contains now.

At head `0d4cb047`:

- `grafana-env` does not exist anywhere on the branch — the only match for the
  string is an unrelated commented-out `secretProviderClass` example in
  `values/grafana.yaml:569`.
- The password is passed through the chart's own mechanism instead
  (`charts/monitoring/monitoring-helmfile.yaml:120-124`):

```yaml
{{- if $grafanaAdminSecret }}
    - name: admin.existingSecret
      value: {{ $grafanaAdminSecret | quote }}
```

  with `adminExistingSecret: grafana-admin` set in `env-secrets.yaml`.

`admin.existingSecret` makes the chart point its own `env` entry at the named
Secret, so there are no longer two competing sources — which is exactly the
precedence problem the previous comment described. There is nothing for the
chart's value to win against.

The reported `4f3721a9c246` vs `97bfcf35e759` split is real, but it is evidence
from a pre-`0d4cb047` tree, not from this branch.

Apologies for the noise. The mistake was mine: verification ran against a
checkout that predated the fix, and I posted the result without confirming the
branch state it was taken from.

singhalkarun:

`b1b8a108` — the inline fallback now sets `adminPassword` (the value the chart writes into its own Secret) instead of `GF_SECURITY_ADMIN_PASSWORD` via `envRenderSecret`, which becomes `envFrom` and loses to the chart's explicit `env`.

Re-verified with a clean install so the password is set at DB init:

- the configured password now authenticates — it returned **401** before
- `secret/grafana`'s `admin-password` **is** that value, so there is one source rather than two competing ones

The `admin.existingSecret` path was already correct and is unchanged.

singhalkarun:
## Withdrawing my retraction — the original finding was right

Earlier I posted a "correction" claiming the *admin password is inert* finding was
measured against a stale checkout. **That correction was wrong, and I'm withdrawing
it.** The original finding was accurate. Apologies to whoever read the retraction
and concluded there was nothing here.

### What I got wrong

I searched the branch for a `grafana-env` Secret, found no such string, and
concluded no `envFrom` path existed — so there could be no `env`-beats-`envFrom`
conflict.

But that Secret is **generated by helmfile at render time** from the
`envRenderSecret:` block. It is never a literal in the repo. I checked for a
string instead of reading the mechanism that produces one.

### The defect, precisely

At `0d4cb047`, `monitoring-helmfile.yaml` had:

```yaml
  - envRenderSecret:
      ...
{{- if not $grafanaAdminSecret }}
      GF_SECURITY_ADMIN_PASSWORD: {{ $grafanaAdminPassword | quote }}
{{- end }}
```

So on the **escape-hatch path** — `adminExistingSecret` blank, `adminPassword`
set, which the file documents for throwaway clusters — the password reaches the
container via `envFrom`. With no `admin.existingSecret`, the chart generates its
own Secret and declares `GF_SECURITY_ADMIN_PASSWORD` explicitly in `env`.
Kubernetes gives explicit `env` precedence over `envFrom`, so the operator's
configured password is present in the pod and loses.

The consequence is worse than "inert": the operator sets a password and is
**locked out by it**, because the chart's `randAlphaNum 40` is what's actually in
force. That is the same lockout this PR exists to fix.

My retraction only examined the **default** path (`adminExistingSecret:
grafana-admin`), where the chart reads the named Secret directly and no second
source exists. That path was always fine. I generalised from it to the branch I
had not read.

### State

`b1b8a108` fixes it by setting the chart's own `adminPassword` value on the
escape-hatch path, so the chart writes the operator's password into the Secret
its explicit `env` reads. One source, no precedence race.

Tracker rows corrected to match.

vinothrallapalli-eGov (approved):

Issue #1657, opened 2026-08-07
Ansible tier deploys the whole observability stack or nothing — make it modular


singhalkarun:
Part of #1601 (monitoring umbrella). **P1 · effort: M**

## Problem

The Ansible tier deploys the **entire** observability stack on every tenant, with no supported way to run a subset.

Verified: all seven observability services in `docker-compose.egov-digit.yaml` — grafana, prometheus, loki, promtail, tempo, otel-collector, gatus — have **no `profiles:` key**, so they always start. `docker-compose.monitoring.yml` (node-exporter) is layered unconditionally by the playbook (`playbook-deploy.yml:1446`). The only `enable_*` flag in `group_vars/digit.yml` is `enable_search_stack`.

That matters because operators do not all have the same capacity. A 16 GB tenant box already carries **12.8 GB of configured JVM heap** across 29 services (measured in #1612), and these boxes have **no swap** — so memory pressure means OOM kills, not slowdown. An operator who wants uptime checks and metrics but cannot afford log and trace storage currently has no supported option other than editing the compose file by hand, which the next deploy overwrites.

## The mechanism already exists

This does not need inventing. Five optional stacks are already gated exactly this way:

- `docker-compose.egov-digit.yaml` uses `profiles:` for `keycloak`, `mcp`, `notifications`, `otp`, `search`
- each is driven by an `enable_*` flag in `group_vars`, assembled into `COMPOSE_PROFILES` by the playbook's "Compute compose profiles for this tenant"
- the matching Gatus checks are toggled in lockstep via `GATUS_PROFILE_*`, a pattern `digit.env.j2:101-118` documents at length: *"the same enable_\* flag that starts an optional stack has to switch on that stack's Gatus checks, or a stack we deployed goes unmonitored and looks identical to one that was never deployed."*

Observability was simply never wired into it.

## Proposed levels

Sizing uses the per-container ceilings from #1612, so these are real numbers an operator can plan against:

| Level | Adds | Cumulative footprint | What you give up by stopping here |
|---|---|---|---|
| **uptime** | gatus | **~64 MB** | no metrics, logs or traces — you know *that* something is down, not why |
| **metrics** | prometheus, node-exporter, grafana | **~900 MB** | no logs or traces; JVM/host dashboards work, investigation is limited |
| **logs** | loki, promtail | **~1.5 GB** | no traces; can read errors but not follow a request across services |
[...]

PR #1666, opened 2026-08-07
feat(ansible): make the observability stack modular

singhalkarun:
Fixes #1657

## What

Every tenant gets the whole observability stack or nothing. All seven services had no `profiles:` key so they always start, and the monitoring overlay is layered unconditionally.

That matters on these boxes: 29 services declare **~12.8 GB** of configured JVM heap on machines commonly 16 GB, with **no swap** — so memory pressure means OOM kills, not slowdown. An operator who wants uptime and metrics but cannot afford log and trace storage had no supported option short of editing compose by hand, which the next deploy overwrites.

## Levels

Cumulative, using the profile mechanism already proven for the five optional stacks (`search`, `mcp`, `notifications`, `keycloak`, `otp`):

| `observability_level` | Adds | Footprint |
|---|---|---|
| `metrics` | gatus, prometheus, node-exporter, grafana, **collector** | ~1.2 GB |
| `logs` | + loki, promtail | ~1.8 GB |
| `traces` *(default)* | + tempo | ~2.2 GB |

**Default is `traces` — the full stack, exactly what runs today.** This only ever lets an operator take less; existing tenants see no change.

Footprints are the `mem_limit` ceilings from #1612, not measured usage. #1657's acceptance asks for real numbers from a box before publishing these as guidance.

## Why the collector is in `metrics`, not `traces`

JVM metrics flow OTLP → otel-collector → `:8889`, which is the endpoint Prometheus scrapes. Dropping the collector costs **per-service JVM metrics**, not just traces. Tempo is the only traces-exclusive component. Getting this backwards would have produced a "metrics" level with no application metrics.

## Two things that had to change for the lower levels to work rather than fail

**1. Grafana's `depends_on` would have failed the deploy.** It required `tempo` and `loki` as `service_healthy`. Compose refuses to start a service whose dependency is excluded by the active profiles, so `metrics` or `logs` would have failed outright rather than degraded. Reduced to `prometheus`.

This is a fix in its own right: the old condition made **the dashboards unreachable whenever the log store was unhealthy** — which is a reason to open Grafana, not to withhold it. An unreachable datasource surfaces as a panel error and recovers on its own.

**2. The agents would have exported into the void.** `OTEL_TRACES_EXPORTER` / `OTEL_METRICS_EXPORTER` now come from `.env` (defaulting to `otlp`), so a level without Tempo stops services shipping spans to an absent backend and the collector logging export failures on a loop. A bare `docker compose up` outside Ansible is unchanged.

## Deliberately NOT included: a gatus-only `uptime` level

It would be the most useful level for a constrained operator (~64 MB), and I have left it out.

**16 application services declare `depends_on: otel-collector`.** Whether Compose auto-enables a dependency's profile or errors is version-dependent, and I have no way to test it here. Getting it wrong means the level does not shrink the deploy — it **fails** it, on a tenant box.

The right fix is to remove that dependency: a telemetry sink should never gate application startup, and today a collector that fails to start blocks 16 core services. That deserves its own change and its own verification.

## Verified / not verified

- [x] All five touched files parse
- [x] Level → profile mapping simulated for every level
- [x] No cross-tier `depends_on` violations remain among observability services
- [x] Audited every non-observability service for dependencies on profiled ones — that is how the 16-service collector dependency surfaced
- [x] `check-gatus-coverage.py` passes; static suite 35/35
- [ ] **Not deployed.** Needs a real run at each level, confirming the expected containers start and nothing else changes
- [ ] Measured footprints at each level, to replace the estimates

## Follow-ups this creates

- **Gatus checks should track the level.** At `metrics`, the Loki and Tempo checks added by #1636 would go red for components deliberately not deployed — the exact failure the profile pattern exists to prevent. #1636 is not in this branch yet; wiring `GATUS_OBSERVABILITY` to the level belongs with whichever lands second.
- The `uptime` level, once the collector dependency is removed.

### DIFF
--- a/local-setup/ansible/inventory/group_vars/digit.yml
+++ b/local-setup/ansible/inventory/group_vars/digit.yml
@@ -48,6 +48,39 @@ digit_ui_mode: static
 # Off by default; tenants that use the inbox opt in via host_vars.
 enable_search_stack: false
 
+# ──────────────────────────────────────────────────────────────────────
+# How much observability to deploy (#1657)
+# ──────────────────────────────────────────────────────────────────────
+# Cumulative levels — each includes the ones above it. Default `traces`
+# is the whole stack, i.e. exactly what every tenant runs today, so this
+# knob only ever lets an operator take LESS.
+#
+#   metrics  gatus, prometheus, node-exporter, grafana, collector  ~1.2 GB
+#   logs     + loki, promtail                                       ~1.8 GB
+#   traces   + tempo                                                ~2.2 GB
+#
+# A gatus-only `uptime` level (~64 MB) is deliberately NOT offered yet.
+# 16 application services declare `depends_on: otel-collector`, so a level
+# that omits the collector risks failing the whole deploy rather than
+# shrinking it — whether compose auto-enables a dependency's profile or
+# errors is version-dependent, and that is not a thing to guess at on a
+# tenant box. Removing that dependency is the right fix (a telemetry sink
+# should never gate application startup) and is tracked separately.
+#
+# (Footprints are the mem_limit ceilings from #1612, not measured usage.)
+#
+# Why the collector sits in `metrics` and not `traces`: JVM metrics flow
+# OTLP → otel-collector → /metrics on :8889, which is what Prometheus
+# scrapes. Drop the collector and you lose per-service JVM metrics, not
+# just traces. Tempo is the only traces-exclusive component.
+#
+# This matters because the boxes are small: 29 services declare ~12.8 GB
+# of max heap on machines that are commonly 16 GB and have NO swap, so
+# memory pressure means OOM kills rather than slowdown. An operator who
+# wants uptime and metrics but cannot afford log and trace storage should
+# be able to say so.
+observability_level: traces
+
 # SPA configuration — stable cross-tenant defaults.
 context_path: digit-ui
 gmaps_api_key: ""
--- a/local-setup/ansible/playbook-deploy.yml
+++ b/local-setup/ansible/playbook-deploy.yml
@@ -1413,7 +1413,32 @@
     # COMPOSE_PROFILES drives which optional service groups are included.
     # Today: `search` covers elasticsearch + egov-indexer + inbox-v2.
     # Default off; tenants opt in via `enable_search_stack: true`.
+    # Observability levels are CUMULATIVE (#1657): each includes the ones
+    # above it. Default `traces` yields all three profiles, i.e. exactly the
+    # stack every tenant runs today — this knob only ever removes.
+    - name: "Validate observability_level"
+      ansible.builtin.fail:
+        msg: >-
+          observability_level must be one of metrics|logs|traces, got
+          '{{ observability_level }}'. Levels are cumulative; `traces` is the
+          full stack and the default. A gatus-only level is not offered yet —
+          16 services depend_on otel-collector, so omitting it risks failing
+          the deploy rather than shrinking it.
+      when: (observability_level | default('traces')) not in ['metrics', 'logs', 'traces']
+
     - name: "Compute compose profiles for this tenant"
+      vars:
+        _obs: "{{ observability_level | default('traces') }}"
+        # collector lives in `metrics`, not `traces`: JVM metrics reach
+        # Prometheus via otel-collector:8889, so dropping it costs metrics too.
+        _obs_profiles: >-
+          {{
+            ([]
+              + (['obs-metrics'] if _obs in ['metrics', 'logs', 'traces'] else [])
+              + (['obs-logs']    if _obs in ['logs', 'traces']            else [])
+              + (['obs-traces']  if _obs == 'traces'                      else [])
+            )
+          }}
       ansible.builtin.set_fact:
         compose_profiles: >-
           {{
@@ -1423,6 +1448,7 @@
               + (['notifications'] if enable_novu          | default(false) else [])
               + (['keycloak']      if enable_keycloak      | default(false) else [])
               + (['otp']           if enable_otp_services  | default(false) else [])
+              + _obs_profiles
             ) | join(',')
           }}
 
--- a/local-setup/docker-compose.egov-digit.yaml
+++ b/local-setup/docker-compose.egov-digit.yaml
@@ -9,6 +9,8 @@ services:
   otel-collector:
     image: egovio/opentelemetry-collector-contrib:0.114.0
     container_name: digit-otel-collector
+    # JVM metrics flow through it, not just traces (#1657)
+    profiles: [obs-metrics]
     restart: unless-stopped
     command: ["--config=/etc/otelcol-contrib/config.yaml"]
     volumes:
@@ -23,6 +25,8 @@ services:
   tempo:
     image: egovio/tempo:2.6.1
     container_name: digit-tempo
+    # trace store — the only traces-only component (#1657)
+    profiles: [obs-traces]
     restart: unless-stopped
     command: ["-config.file=/etc/tempo/config.yaml"]
     volumes:
@@ -41,6 +45,8 @@ services:
   grafana:
     image: egovio/grafana:11.4.0
     container_name: digit-grafana
+    # the only UI for any of it (#1657)
+    profiles: [obs-metrics]
     restart: unless-stopped
     environment:
       GF_AUTH_ANONYMOUS_ENABLED: "true"
@@ -61,11 +67,17 @@ services:
       interval: 10s
       timeout: 5s
       retries: 5
+    # Only prometheus, which shares grafana's profile (#1657). tempo and loki
+    # were hard dependencies here, which breaks two ways:
+    #   1. compose refuses to start a service whose depends_on target is
+    #      excluded by the active profiles — so `metrics` or `logs` level would
+    #      fail the deploy outright, not degrade;
+    #   2. even at full stack, `condition: service_healthy` made the dashboards
+    #      unreachable whenever the LOG store was unhealthy — a trace/log
+    #      backend being down is a reason to open Grafana, not to withhold it.
+    # Grafana needs neither at boot: an unreachable datasource surfaces as a
+    # panel error and recovers on its own once the backend appears.
     depends_on:
-      tempo:
-        condition: service_healthy
-      loki:
-        condition: service_healthy
       prometheus:
         condition: service_started
     networks:
@@ -77,6 +89,8 @@ services:
   prometheus:
     image: prom/prometheus:v2.55.1
     container_name: digit-prometheus
+    # metrics store (#1657)
+    profiles: [obs-metrics]
     restart: unless-stopped
     command:
       - --config.file=/etc/prometheus/prometheus.yml
@@ -306,11 +320,11 @@ services:
       SECURITY_BASIC_ENABLED: 'false'
       MANAGEMENT_SECURITY_ENABLED: 'false'
       MANAGEMENT_HEALTH_DB_ENABLED: 'false'
-      OTEL_TRACES_EXPORTER: otlp
+      OTEL_TRACES_EXPORTER: ${OTEL_TRACES_EXPORTER:-otlp}
       OTEL_EXPORTER_OTLP_ENDPOINT: http://otel-collector:4317
       OTEL_EXPORTER_OTLP_PROTOCOL: grpc
       OTEL_SERVICE_NAME: egov-indexer
-      OTEL_METRICS_EXPORTER: otlp
+      OTEL_METRICS_EXPORTER: ${OTEL_METRICS_EXPORTER:-otlp}
       OTEL_LOGS_EXPORTER: none
       # heap-tuned via JAVA_OPTS by patch-jvm-heaps.py for egov-indexer
       JAVA_OPTS: "-Xms192m -Xmx512m"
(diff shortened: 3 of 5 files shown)


singhalkarun:
## Devbox test: the level is honoured for a *fresh* deploy, but lowering it does not tear down the disabled tier

Tested on the eGov devbox (`karun`) by setting `observability_level: metrics` on a box that had been running the full stack, then redeploying.

**What worked.** The playbook computed the right profiles — `Active compose profiles for karun: mcp,obs-metrics` — and the exporter env followed the level exactly:

```
OTEL_TRACES_EXPORTER=none      # traces tier off
OTEL_METRICS_EXPORTER=otlp     # metrics tier on
```

**What did not.** The traces and logs containers kept running:

| Container | Expected | Actual |
|---|---|---|
| `digit-tempo` (`obs-traces`) | gone | **Up 8 minutes** |
| `digit-loki` (`obs-logs`) | gone | **Up 8 minutes** |
| `digit-promtail` (`obs-logs`) | gone | **Up 7 minutes** |

**Cause:** `docker compose up -d` only declines to *start* services outside the active profiles — it does not *stop* ones already running. Confirmed directly:

```
$ COMPOSE_PROFILES=mcp,obs-metrics docker compose <files> up -d
  tempo before: Up 9 minutes (healthy)
  tempo after:  Up 9 minutes (healthy)
```

`--remove-orphans` would not help either: these are not orphans. They are declared in the compose file and simply not selected, which compose treats as "leave alone".

## Why this matters for this PR specifically

The stated purpose is letting capacity-limited sites run a subset. On a **fresh** box that works. But on any box that has already run a higher level — which is every existing deployment — lowering the level **frees no memory at all**. The containers stay up and keep their allocation; only the telemetry being *sent to* them stops.

So the operator sees a config that says metrics-only, a deploy that reports success, and a box still running the full stack. That is the same silent-mismatch shape as #1683: config correct, runtime unchanged, nothing surfacing the gap.

Combined with #1635's limits, the memory that a downgrade *should* release here is tempo 384m + loki 512m + promtail 128m ≈ **1 GB** still held.

## Suggested fix

After computing the active profiles, explicitly stop and remove the containers belonging to now-inactive `obs-*` profiles — e.g. `docker compose --profile obs-traces --profile obs-logs rm -sf tempo loki promtail` when those tiers are off, or a targeted `stop`/`rm` derived from the same `_obs_profiles` fact that already exists. The computation is right; only the teardown is missing.

## Also verified in the same run

- #1635's limits applied to all 8 observability containers, none uncapped, and headroom is comfortable: tempo 57%, otel-collector 38%, promtail 38%, grafana 28%, prometheus 16%, loki 16%, gatus 18%, node-exporter 14%.
- The remaining metrics stack stayed healthy throughout — all 5 Prometheus targets `up`, Grafana 200, 44 containers, 0 unhealthy.

singhalkarun:
## Second finding: every ad-hoc `docker compose` command breaks

Separate from the teardown gap above, and in practice the more disruptive of the two.

Putting `otel-collector` behind `obs-metrics` leaves **12 unprofiled core services** declaring `depends_on: otel-collector`:

```
egov-user            egov-workflow-v2    egov-persister     egov-filestore
egov-localization    egov-idgen          egov-enc-service   egov-accesscontrol
boundary-service     mdms-backend        egov-hrms          pgr-services
```

Compose rejects the whole project when a selected service depends on one excluded by profile:

```
service "egov-filestore" depends on undefined service "otel-collector": invalid compose project
```

### What that costs

| Command | without `COMPOSE_PROFILES` | with it |
|---|---|---|
| `docker compose config -q` | ❌ invalid project | ✅ |
| `docker compose ps` | ❌ | ✅ |
| `docker compose logs --tail=1 kong` | ❌ | ✅ |
| `docker compose up -d --no-deps kong` | ❌ | ✅ |
| `docker compose down --help` | ✅ | ✅ |

**Deploys are fine** — the playbook computes and passes the profiles. What breaks is everything a human types: checking `ps`, tailing `logs`, recreating one service. That is the debugging path, and it fails exactly when you need it.

Six documents under `docs/` hand operators raw `docker compose -f …` commands that now error, including `docs/notification-onboarding/RUNBOOK.md` and `docs/migration/operator-runbook.md`.

Found by accident: I tried to restart Gatus on the devbox to verify #1632 and hit it immediately — roughly how an operator would meet it.

### The underlying tension

**If 12 core services depend on `otel-collector`, it is not optional**, and profiling it contradicts the dependency graph.

The cleanest fix is probably to leave `otel-collector` unprofiled and profile the *consumers* of telemetry instead — Prometheus, Grafana, Tempo, Loki — rather than the collector everything feeds into. That also matches the tiering intent: a site running `observability_level: metrics` still wants the collector; what it does not want is Tempo and Loki.

Dropping the 12 `depends_on` edges would also work — the OTEL agent buffers and retries, so the ordering is soft — but it changes startup behaviour more broadly and is the larger change of the two.

Either way this wants resolving before merge: a modularity feature that makes the stack harder to operate by hand is a poor trade, and the failure mode gives no hint that a profile variable is the cause.

vinothrallapalli-eGov (commented):
Verified all three devbox-reported defects independently against the current diff (head 9ad09bb). **None of the three are fixed here — companion PR #1687 is still a required merge partner, not redundant.**

1. **Stale services keep running on downgrade — STILL PRESENT.** `compose_profiles`/`_obs_profiles` (playbook-deploy.yml) only feeds `COMPOSE_PROFILES` into `docker compose up -d` invocations. `up` only declines to *start* services outside the active profile set; it never stops ones already running from a prior, higher-level deploy. I grepped the whole playbook for `tempo`/`promtail`/any profile-aware stop or prune task — the only `down --remove-orphans` in the file is the pre-existing `force_clean: true` opt-in teardown, unrelated to level changes, and `--remove-orphans` wouldn't catch tempo/loki/promtail anyway since they're still declared services, not orphans. No teardown logic was added.

2. **otel-collector profile breaks unprofiled dependents — STILL PRESENT.** `otel-collector` now carries `profiles: [obs-metrics]`. I counted the `depends_on: otel-collector: condition: service_started` sites: 16 total, of which 3 are already profile-gated elsewhere (egov-indexer/inbox under `search`, novu-bridge under `notifications`) but **13 core services remain unprofiled** (egov-enc-service, mdms-backend, egov-idgen, egov-user, egov-workflow-v2, egov-localization, boundary-service, egov-accesscontrol, egov-persister, egov-filestore, egov-hrms, pgr-services, digit-config-service). Within Ansible this is masked because `obs-metrics` is included at every valid level (metrics/logs/traces), but `COMPOSE_PROFILES` is only ever exported inline on the specific `docker compose` shell lines in the playbook — it is never persisted into `/opt/digit/.env` (only `GATUS_PROFILE_*` and the `OTEL_*_EXPORTER` vars are). Any bare, ad-hoc `docker compose up <service>` run later on the box (debugging, a manual restart) now runs with no profiles active, otel-collector won't start, and all 13 core services' hard `service_started` dependency breaks — a regression this PR introduces (those services previously depended on an always-on, unprofiled otel-collector).

3. **Unconditional 2-minute Loki wait — STILL PRESENT.** `playbook-deploy.yml`'s "Wait for Loki health (up to 2 min)" task (`retries: 12`, `delay: 10`, curl against `127.0.0.1:13100/ready`) has no `when:` gate on `observability_level`, unlike the MCP health task right above it which is gated by `enable_mcp`. A `metrics`-level tenant (where Loki never starts) will still burn the full 2-minute retry budget and then print "Loki: UNHEALTHY (non-critical)". This task predates this PR's diff (it isn't a changed line), so it wasn't touched here — but the PR that adds the `observability_level` knob is exactly where it should have been gated, and it wasn't.

Other notes: the PR body is candid about the collector/depends_on risk and explicitly declines to add a `uptime` level for this reason — good judgement — but that same reasoning applies with equal force to the ad-hoc-compose-command case above, which isn't mentioned. No existing review comments found to duplicate.

local-setup/docker-compose.egov-digit.yaml
```diff
@@ -9,6 +9,8 @@ services:
   otel-collector:
     image: egovio/opentelemetry-collector-contrib:0.114.0
     container_name: digit-otel-collector
+    # JVM metrics flow through it, not just traces (#1657)
+    profiles: [obs-metrics]
```
vinothrallapalli-eGov, line 13: Defect #2 (still present): once otel-collector is profile-gated, the 13 unprofiled core services below (pgr-services, egov-user, egov-workflow-v2, egov-persister, egov-hrms, mdms-backend, egov-idgen, egov-accesscontrol, egov-filestore, egov-enc-service, egov-localization, boundary-service, digit-config-service) still declare `depends_on: otel-collector: condition: service_started` unchanged. That's masked inside Ansible because `obs-metrics` is part of every valid `observability_level`, but `COMPOSE_PROFILES` is only exported inline on the specific `docker compose` shell lines in playbook-deploy.yml — it's never written into `/opt/digit/.env`. Any bare `docker compose up <service>` run later on the box (outside the ansible-orchestrated command) gets no active profiles, otel-collector doesn't start, and all 13 dependents fail to start. This is a regression versus today (otel-collector was previously always-on and unprofiled).

local-setup/ansible/playbook-deploy.yml
```diff
@@ -1413,7 +1413,32 @@
...
+          16 services depend_on otel-collector, so omitting it risks failing
+          the deploy rather than shrinking it.
+      when: (observability_level | default('traces')) not in ['metrics', 'logs', 'traces']
+
     - name: "Compute compose profiles for this tenant"
+      vars:
+        _obs: "{{ observability_level | default('traces') }}"
+        # collector lives in `metrics`, not `traces`: JVM metrics reach
+        # Prometheus via otel-collector:8889, so dropping it costs metrics too.
+        _obs_profiles: >-
```
vinothrallapalli-eGov, line 1434: Defect #1 (still present): this only computes which profiles to pass to the next `docker compose up -d`. `up` never stops services outside the active profile set that are already running from a prior deploy at a higher level. Lowering `observability_level` from `traces` to `metrics` on an existing tenant will leave tempo/loki/promtail running (~1GB held) — I found no teardown/stop task anywhere in this playbook keyed off `_obs`/`observability_level`, and the only `down --remove-orphans` in the file is the unrelated `force_clean` opt-in (which also wouldn't help here since these aren't orphans, just profile-excluded).

local-setup/ansible/inventory/group_vars/digit.yml
```diff
@@ -48,6 +48,39 @@ digit_ui_mode: static
...
+# OTLP → otel-collector → /metrics on :8889, which is what Prometheus
+# scrapes. Drop the collector and you lose per-service JVM metrics, not
+# just traces. Tempo is the only traces-exclusive component.
+#
+# This matters because the boxes are small: 29 services declare ~12.8 GB
+# of max heap on machines that are commonly 16 GB and have NO swap, so
+# memory pressure means OOM kills rather than slowdown. An operator who
+# wants uptime and metrics but cannot afford log and trace storage should
+# be able to say so.
+observability_level: traces
```
vinothrallapalli-eGov, line 82: Defect #3 (still present, not addressed by this change): the pre-existing "Wait for Loki health (up to 2 min)" task in playbook-deploy.yml (`retries: 12, delay: 10`, curl to `127.0.0.1:13100/ready`) has no `when:` gate on this new `observability_level` var — unlike the MCP wait task right above it, which is gated on `enable_mcp`. A `metrics`-level tenant, where Loki is intentionally never started, will still block for the full 2 minutes before logging 'Loki: UNHEALTHY (non-critical)'. Since this PR is exactly where the level-aware gating should be introduced, this looks like a real gap rather than out-of-scope.

local-setup/docker-compose.egov-digit.yaml
singhalkarun: Confirmed, and fixed in **#1687** (`fix/1666-profile-defects`, stacked on this branch) rather than duplicated here — commit 7e032928 un-profiles `otel-collector` again.

Your diagnosis holds all the way down to the mechanism. The one correction is the count: **12** unprofiled services declare `depends_on: otel-collector`, not 13 — `digit-config-service` exists but is not among them. Verified by walking the parsed compose file:

```
otel-collector : 12 unprofiled dependents
  boundary-service, egov-accesscontrol, egov-enc-service, egov-filestore,
  egov-hrms, egov-idgen, egov-localization, egov-persister, egov-user,
  egov-workflow-v2, mdms-backend, pgr-services
tempo / grafana / prometheus / loki / promtail : 0 each
```

And compose rejects the **whole project**, not just the dependent, which is why the blast radius was everything rather than one service:

```
service "egov-filestore" depends on undefined service "otel-collector": invalid compose project
```

`config -q`, `ps`, `logs`, `up -d --no-deps kong` all failed without `COMPOSE_PROFILES` — including six runbooks under `docs/`. Deploys were fine, as you say, because the playbook passes the profiles inline.

Un-profiling was chosen over propagating profiles to the 12 dependents, and that is the conceptually right line rather than just the smaller diff: the tiers gate what **consumes** telemetry, not the pipeline everything emits into. A site on `observability_level: metrics` still needs the collector — what it does not need is Tempo and Loki. It stays cheap because `OTEL_TRACES_EXPORTER`/`OTEL_METRICS_EXPORTER` already follow the level, so at a lower tier it simply receives less.

local-setup/ansible/playbook-deploy.yml
singhalkarun: Confirmed, and fixed in **#1687** (commit 94c419c5) rather than duplicated here.

You were right on both halves, including the part that is easy to get wrong: `--remove-orphans` does not help, because these services are *declared and merely unselected*, which compose treats as "leave alone" rather than orphaned. They have to be named explicitly. On the devbox the downgrade released **nothing** — tempo was `Up 9 minutes (healthy)` before and after the `metrics`-level deploy, ~1 GB still held while the config said metrics-only and the run reported success.

The fix is a teardown task that runs after the stack is up, maps each tier to its services, and only acts when those containers actually exist — so a steady-state redeploy reports `ok` rather than needlessly stopping and starting things.

Worth flagging one trap it walked into, since it is the same class you are pointing at here: the service list has to derive from `compose_profiles` (a registered fact) and **not** from `_obs_profiles`, which is a task-scoped `vars:` entry on the `set_fact` further up and simply does not exist by the time this task runs. Testing the Jinja in isolation would have looked fine.

Result at `observability_level: metrics`: tempo/loki/promtail **RUNNING → GONE** (44 → 41 containers), `changed` when there was work, `ok` on rerun.

local-setup/ansible/inventory/group_vars/digit.yml
singhalkarun: Confirmed, and fixed in **#1687** rather than duplicated here — and agreed this PR is the right place for it, not out of scope: it introduces the var the gate needs, and the MCP wait immediately above already establishes the pattern.

Gated with `when: "'obs-logs' in (compose_profiles | default('')).split(',')"`. Note it had to go on **both** tasks, not just the wait: with the wait skipped, `loki_health` is never registered, so the "Show Loki health status" debug below it would error on `.rc` — an ungated status line turns a fixed hang into a failed run.

Devbox at `observability_level: metrics`: `skipping`, 0 retries (was 12), `ignored=0` (was 1). Your read that `ignore_errors` was letting it degrade quietly rather than announce itself is exactly what made it survive this long.

vinothrallapalli-eGov (approved):

Issue #1658, opened 2026-08-07
No documentation on how to enable monitoring, or what each part costs, on either tier


singhalkarun:
Part of #1601 (monitoring umbrella). **P1 · effort: S**

## Problem

There is no documentation, on either tier, telling an operator **what monitoring exists, what it costs, or how to turn parts of it on and off**. Operators have very different capacity — a 16 GB single-tenant box and a large-scale Kubernetes cluster are not the same decision — and today neither is served.

The gap is different on each tier:

**Ansible** — no documentation *and* no choice. All seven observability services always start (see the modularity issue filed alongside this one). Worse, what documentation exists is **wrong**: `local-setup/README.md`'s "What the Ansible playbook deploys" table has a single row labelled *Tracing* listing the collector, Tempo and Grafana — and omits **Loki, Promtail, Prometheus and node-exporter entirely**, though all four deploy on every box. `ansible/README.md` separately claims health gates for "loki / grafana" that do not exist (being corrected in #1613).

**Kubernetes** — the choice exists but is invisible. Every release in `charts/monitoring/monitoring-helmfile.yaml` has its own `installed:` flag, so it is already modular per component. But the whole helmfile is commented out of `digit-helmfile.yaml:18`, nothing explains that, and **the original import deleted the only README under `deploy-as-code`** — so an operator has no way to discover any of it short of reading the charts.

## What to write

A single page covering both tiers, structured around the operator's actual question — *"how much monitoring can I afford, and what do I lose?"*

**1. What each component gives you, and what you give up without it**

| Component | Answers | Without it |
|---|---|---|
| Gatus (Ansible) / blackbox (k8s) | is it up? | you learn from users |
| Prometheus + node-exporter | is the box healthy? is a service leaking memory? | no capacity signal, no trend |
| Grafana | somewhere to look | data with no view |
| Loki + Promtail | what did it say when it broke? | errors are unreachable |
| Collector + Tempo | where did this request spend its time? | cross-service latency is guesswork |

**2. Cost, in numbers.** Use the ceilings from #1612 — uptime only ~64 MB, plus metrics ~900 MB, plus logs ~1.5 GB, plus traces ~2.2 GB — with the context that a 16 GB tenant box already carries 12.8 GB of configured JVM heap and **has no swap**, so overcommitting means OOM kills rather than slowdown.

**3. How to enable, per tier.**
[...]

Issue #1620, opened 2026-08-06
Docs misdescribe the deployed monitoring stack and claim health gates that do not exist


singhalkarun:
Part of #1601 (monitoring stack umbrella). **P2 · effort: XS — but land it LAST**

Small, but it is part of *why* several other items in this umbrella went unnoticed.

## 1. The deploy table omits half the observability stack

`local-setup/README.md:338-345`, under "What the Ansible playbook deploys":

```
| **Tracing** | OpenTelemetry Collector, Tempo, Grafana |
```

That row is the only observability entry, and it is both **incomplete** and **mislabelled**:

- **Missing entirely: Loki, Promtail, Prometheus, node-exporter** — all four are deployed on every Ansible box.
- "Tracing" is the wrong heading for a row containing Grafana (a UI) and, once corrected, a metrics store and a log pipeline. It should read **Observability**.

Proposed:

```
| **Observability** | OpenTelemetry Collector, Tempo (traces), Prometheus + node-exporter (metrics), Loki + Promtail (logs), Grafana (dashboards) |
```

*(Note: the compose `-f` list immediately above the table, at :329-333, is correct — it does include `docker-compose.monitoring.yml`. The defect is the table only.)*

## 2. The health-gate list describes gates that do not exist

`local-setup/ansible/README.md:656`:

> **Health gates** — wait for kong / persister / hrms / ui / mcp / loki / grafana

**Grafana has no health gate at all**, and Loki's is `ignore_errors: yes` — it prints `UNHEALTHY (non-critical)` and the deploy proceeds. This is the specific line that let #1613 go unnoticed: anyone reading the docs would reasonably conclude both were covered.

Must be corrected in step with #1613, which adds the gates this line currently imagines.

## 3. An operationally useful fact is undocumented

Kong traces **100% of requests** (`KONG_TRACING_SAMPLING_RATE: "1.0"`) and Tempo retains them for **24h** (`block_retention: 24h`). Fine at current volumes — but sampling rate is the first knob to reach for if a box gets tight, and today nobody would know it exists without reading the compose file. Worth a line in the ops docs alongside the retention figures for Loki (72h) and Prometheus (15d).

## Sequencing — land this last

Several of these lines describe things other issues in this umbrella will change:

- #1613 adds the health gates that item 2 currently misdescribes
- #1614 / #1615 add scrape targets the table should then list
- #1617 may remove the root-level copies referenced in the directory map

[...]

PR #1667, opened 2026-08-07
docs(observability): how to enable monitoring, and what each part costs

singhalkarun:
Fixes #1658
Fixes #1620

Written as **one change** because the enabling guide subsumes most of the README corrections. Splitting them risks two inconsistent descriptions of the same thing, which is how the current docs got wrong in the first place.

## The docs were not merely absent — they were wrong

**`local-setup/README.md`** — "What the Ansible playbook deploys" had a single observability row labelled *Tracing*, listing the collector, Tempo and Grafana. It omitted **Loki, Promtail, Prometheus and node-exporter** entirely, though all four deploy on every box. Gatus was filed under *Tools* alongside Jupyter, as if it were a dev utility.

Renamed to **Observability**, completed, and Gatus moved where it belongs.

**`local-setup/ansible/README.md`** claimed health gates for `loki / grafana`. Grafana had none, and Loki's gated nothing (`ignore_errors: yes`). Now describes the fatal / non-fatal split as it actually behaves.

## The new guide

`docs/observability/enabling-monitoring.md`, structured around the question an operator actually has — **"how much can I afford, and what do I lose by skipping a piece?"** — rather than around a component list.

- what each component answers, **and what you give up without it**
- footprints per level, stated as `mem_limit` **ceilings rather than measured usage**, because presenting estimates as measurements is how sizing guidance goes wrong
- how to enable and verify, per tier, including the commands to check what actually started
- what is deliberately *not* monitored, and why — a gap you chose is different from one you missed

Two facts I put up front because both are easy to get backwards:

- **The collector is not only for traces.** JVM metrics travel through it to the endpoint Prometheus scrapes, so dropping it costs metrics too. Tempo is the only traces-exclusive component.
- **The two tiers differ in kind**, not just scale: Ansible is on-by-default and sized by a knob; Kubernetes is off-by-default behind one commented line.

## The missing README

`devops/deploy-as-code/README.md` did not exist — **the import that brought that tree in deleted the original**, which is why nothing about the production tier was discoverable.

It records what the tier is, that monitoring is commented out of the entrypoint and **why** (inherited from upstream, never a local decision — beside an unrelated `sanitation` module), and why some values name other environments.

It also carries the two traps that produced real defects there, so the next person meets them in prose rather than in production:

- **values are injected wholesale into every release**, so a top-level key intended for one chart can land in another that reads the same name — `controller:` is read by both `ingress-nginx` and `kafka-kraft`
- **a mis-pathed key is silent**: helm ignores values nothing reads, and `helm template` does not catch it either

## Verified

- [x] Every relative link in the new and edited files resolves to a real path
- [x] No `Tracing` row remains
- [x] `check-gatus-coverage.py` still passes

## Scope note

Describes the end state of the `monitoring-fix` branch, which is stated at the top of the guide. `observability_level` (#1666) and the Gatus observability checks (#1636) arrive with that work; if either is dropped, the guide needs a corresponding edit.


vinothrallapalli-eGov (commented):
Good, useful doc that genuinely closes the gap from #1658 — the tier comparison, the OTEL-collector-carries-JVM-metrics point, and the 12.8GB-heap-on-16GB-box rationale (matching the tracking issue's own figures) all read accurately. I spot-checked several concrete claims against the current repo state and found a few things that don't match today's code, plus one internal-consistency gap in the footprint numbers versus the issue. None of these are severe (the doc's own banner says it partly describes an unmerged `monitoring-fix` branch), but three of the four flagged claims live in files/sections that carry no such forward-looking caveat, so a reader would reasonably expect them to be true today.

devops/deploy-as-code/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Two things to know before changing anything here
+
+**Values are injected wholesale into every release.** Each release passes `{{ .Values | toYaml }}`, so the *entire* merged `env.yaml` + `env-secrets.yaml` tree reaches every chart. A top-level key intended for one chart can land in another that happens to read the same name — `controller:` is read by both `ingress-nginx` and `kafka-kraft`, and `metrics:` by the `common` library chart used by all 41 services. **Prefer a release-scoped `set:` block over a new top-level key.**
+
+**A mis-pathed values key is silent.** Helm ignores values nothing reads: no warning, no exit code, and `helm template` will not catch it either — it fails only on template errors, `required`/`fail`, or a `values.schema.json` violation, and these charts ship no schema. So a key at the wrong depth looks exactly like working configuration.
+
+Two long-lived defects came from precisely this, which is why `.github/scripts/check-helm-values-paths.py` now runs in CI and fails on any values key the target chart cannot read.
```
vinothrallapalli-eGov: `.github/scripts/check-helm-values-paths.py` does not exist in the repo at this PR's head commit — only `check-flyway-dump-alignment.py`, `check-gateway-whitelist-parity.py`, and `check-gatus-coverage.py` are present under `.github/scripts/`, and no workflow references a helm-values-path checker. This section (unlike `docs/observability/enabling-monitoring.md`) carries no "describes an unmerged branch" caveat, so as written it tells an operator a CI safety net exists that doesn't. Either land the script in this PR/a companion PR, or caveat the claim as forthcoming.

local-setup/ansible/README.md
```diff
@@ -650,7 +650,11 @@ shared belongs in `group_vars/digit.yml`.
 4. **digit-ui** — render `globalConfigs.js`, ship nginx config, optional `npm install` + esbuild rebuild for HMR
 5. **OpenBao bootstrap** (first run) + secret-pull for every run
 6. **Compose pull + start** (with profiles)
-7. **Health gates** — wait for kong / persister / hrms / ui / mcp / loki / grafana
+7. **Health gates** — wait for kong / persister / hrms / ui / mcp (fatal), then the
```
vinothrallapalli-eGov: Checked `playbook-deploy.yml` for the described gates: there's a real "Wait for Loki health" task (`ignore_errors: yes`, ~line 4136), but I could not find any equivalent wait/health task for grafana, prometheus, tempo, or otel-collector — only a best-effort Prometheus *config reload* (line 1850) and a Grafana domain templating step, neither of which is a health gate. As written this claims stage 7 waits (even non-fatally) on four services it doesn't currently wait on. If this lands with the `monitoring-fix` branch's other not-yet-merged pieces, worth saying so here the way `enabling-monitoring.md` does.

local-setup/README.md
```diff
@@ -337,12 +337,12 @@ Between them these include:
 
 | Category | Services |
 |----------|----------|
-| **Tracing** | OpenTelemetry Collector, Tempo, Grafana |
+| **Observability** | Prometheus + node-exporter (metrics), Loki + Promtail (logs), OpenTelemetry Collector + Tempo (traces), Grafana (dashboards), Gatus (uptime). How much of this deploys is set by `observability_level` — see [Enabling monitoring](../docs/observability/enabling-monitoring.md). |
```
vinothrallapalli-eGov: `observability_level` doesn't exist anywhere in `local-setup/` today (no hits for the string in any host_vars, group_vars, or compose file) — it's one of the knobs `enabling-monitoring.md` explicitly flags as arriving with the `monitoring-fix` branch. This table row states it as current behavior with no such flag, so an operator following this README today would look for a variable that isn't there yet.

docs/observability/enabling-monitoring.md
```diff
@@ -0,0 +1,137 @@
...
+
+---
+
+## Ansible tier: `observability_level`
+
+Set it in `host_vars/<tenant>.yml`. Levels are **cumulative** — each includes the one above.
+
+| Level | Adds | Footprint |
+|---|---|---|
+| `metrics` | Gatus, Prometheus, node-exporter, Grafana, collector | ~1.2 GB |
```
vinothrallapalli-eGov: The tracking issue (#1601) cites footprint-by-tier as uptime ~64 MB, +metrics ~900 MB, +logs ~1.5 GB, +traces ~2.2 GB. This table's cumulative numbers only match at the endpoints (uptime ~64 MB is repeated correctly at line 62, and traces ~2.2 GB matches), but `metrics` here is ~1.2 GB (vs. ~900 MB) and `logs` is ~1.8 GB (vs. ~1.5 GB) — roughly 300 MB higher at each middle tier with no note explaining the revision. Worth either reconciling with the issue's numbers or adding a one-line "revised from the issue's estimate because X" so the two sources don't quietly disagree.

devops/deploy-as-code/README.md
singhalkarun: Confirmed and fixed in 4a478d3c.

You are right on both counts. `.github/scripts/` at this head holds exactly
`check-flyway-dump-alignment.py`, `check-gateway-whitelist-parity.py` and
`check-gatus-coverage.py`, and the only workflows touching `devops/deploy-as-code/`
are `gateway-whitelist-parity.yml` and `digit_install.yml`. The checker lands with
**#1652**, which also targets `monitoring-fix` and has not merged.

Landing it here was not an option — it is a whole PR with its own workflow — so I
caveated it, and made the caveat say the part that actually matters to an operator
rather than just "forthcoming":

> **Not merged yet.** That guard and its `deploy-as-code validation` workflow arrive
> with **#1652** […] Until #1652 lands, **nothing** in CI validates this directory
> beyond `gateway-whitelist-parity`, which reads exactly two list keys out of
> `env.yaml`. Assume a mis-pathed key here reaches production unremarked.

Also took the more general fix you were pointing at underneath all four of your
threads: `enabling-monitoring.md` now opens with a table of every described-but-
unmerged piece against the PR that brings it (#1657/#1687, #1652, #1675), instead of
one prose sentence that named two examples and left the rest to be discovered one
review comment at a time.

local-setup/ansible/README.md
singhalkarun: Confirmed and fixed in 4a478d3c. Your reading of the playbook is exactly right, and
I checked it on all three relevant branches rather than just this one:

```
$ git show <ref>:local-setup/ansible/playbook-deploy.yml | grep -n 'name: .*Wait for'
```

`monitoring-fix`, `feat/1657-modular-observability` and `fix/1666-profile-defects`
all return the same set — kong, persister, hrms, mcp, ui, openbao, novu, keycloak,
and **Loki** — and none of them has a wait of any kind for grafana, prometheus,
tempo or otel-collector. The line invented four gates. This was not a
"lands-with-the-branch" case; the branch does not add them either.

Worse, the parenthetical consoled the reader with *"the path that reaches a human is
the Gatus check"* — and there is no Gatus check for any observability service.
`grafana`, `prometheus`, `loki`, `tempo`, `otel-collector` and `node-exporter` are all
exempted in `check-gatus-coverage.py` as *"observability plumbing … not a serving
dependency"*, and `promtail` has no HTTP listener. So the sentence pointed at a safety
net that is a tracked gap (#1613, no PR).

Now:

> 7. **Health gates** — wait for kong / persister / hrms / ui / mcp. The only
>    observability service waited on is **loki**, and non-fatally (`ignore_errors`)
>    […] #1657 gates that wait on the logs tier actually being active […]
>    **grafana, prometheus, tempo, otel-collector and node-exporter are not waited on
>    at all** — and are not Gatus-checked either, so nothing reports them either way
>    (#1613).

The tier-gating half is real and does land with this branch chain: #1687 gates the
Loki wait on `'obs-logs' in compose_profiles`, so a `metrics` tenant no longer burns
12×10s retries on a container it deliberately did not deploy and then prints
"UNHEALTHY (non-critical)" for the configured state.

local-setup/README.md
singhalkarun: Confirmed and fixed in 4a478d3c. `git grep observability_level` over `local-setup/`
on `monitoring-fix` returns nothing — it exists only on
`feat/1657-modular-observability` (`group_vars/digit.yml:82`, the validate/compute
tasks in `playbook-deploy.yml`, and `digit.env.j2`). An operator reading this table
today would go looking for a variable that is not there.

Row now reads:

> All of it deploys today; **#1657 adds `observability_level`** to deploy less than
> all of it […] Until that lands the variable does not exist and there is nothing to set.

Phrased that way round on purpose: the current behaviour (everything deploys) is the
fact an operator needs, and the knob is the forthcoming part — the previous wording
had it backwards, presenting the knob as the fact.

docs/observability/enabling-monitoring.md
singhalkarun: Confirmed, reconciled in 4a478d3c — and the discrepancy turned out to have a single
clean cause rather than being a quiet revision, so I wrote up the *why* instead of
just adding a note.

Both sets sum the same proposed caps from #1612 (loki 512, prometheus 512, tempo 384,
otel-collector 320, grafana 256, promtail 128, gatus 64, node-exporter 64). They
disagree about which tier the **OTEL collector** belongs to:

| | #1601 | here |
|---|---|---|
| uptime | 64 | 64 |
| metrics | 64+64+512+256 = **896** (~900 MB) | +320 collector = **1216** (~1.2 GB) |
| logs | +512+128 = **1536** (~1.5 GB) | **1856** (~1.8 GB) |
| traces | +384+320 = **2240** (~2.2 GB) | **2240** (~2.2 GB) |

#1601 counted the collector under *traces*, which is the natural assumption. #1657
moved it to *metrics*, because JVM metrics travel OTLP → collector → `:8889` and
Prometheus scrapes that, so a metrics-only site needs it. That one reclassification is
your ~300 MB, at both middle tiers, and it is exactly why the endpoints still agree:
the total is unchanged and *uptime* contains neither. It is in a collapsed block so it
does not crowd the table.

Two further things came out of chasing this, neither of which you had visibility into:

- **#1687 has since moved the collector out of the tiers entirely** (12 unprofiled
  services `depends_on` it, and Compose rejects the whole project when a selected
  service depends on a profile-excluded one, which broke every ad-hoc
  `docker compose` command). The three table numbers are unaffected — the collector is
  in all of them either way — but the `uptime` figure is a casualty: a Gatus-only level
  would still run the collector, so its floor is **~384 MB, not ~64 MB**. Both the
  table and the "no Gatus-only level" paragraph now say so.
- **"They come from the per-container `mem_limit` values" was false.** #1612 *proposes*
  those caps and is still open; there is no `mem_limit` on any observability container
  in `local-setup/` today — they all run uncapped. Reworded to a budget on paper,
  explicitly not enforced, which is a materially different thing to tell someone sizing
  a 16 GB box.

vinothrallapalli-eGov (approved):

PR #1668, opened 2026-08-07
feat(otel): collector + opt-in javaagent for application metrics

singhalkarun:
Fixes #1646

> **Part 2 is DEFAULT OFF.** Merging this changes nothing about running services. Read "Rollout" before enabling.

## What

This tier produces **no application metrics**. Not "they are not collected" — they are not produced: no service exposes a micrometer or actuator Prometheus endpoint, and 0 of 41 charts define a `ServiceMonitor`, so even a correctly configured Prometheus has nothing to scrape.

Adopts the Ansible tier's approach rather than inventing a second one.

## The design point that shrank this

The collector **re-exposes every OTLP metric on a single Prometheus endpoint**, with per-service labels preserved via the `service.name` resource attribute.

So collection needs **one ServiceMonitor for the collector**, not 41 across 41 charts. The "0 of 41 charts define a ServiceMonitor" problem disappears rather than being solved 41 times — and the shared `common.servicemonitor` template, which is broken in three ways, does not need repairing at all.

## Two parts, one active

**1. The collector — deployed.** Added to the monitoring helmfile with values mirroring `local-setup/otel/otel-collector-config.yaml`. One ServiceMonitor, carrying `release=kube-prometheus-stack` — without that label Prometheus's default selector ignores it, the trap already hit in #1662 and #1663. Inert until part 2 is on.

**2. The agent — wired, gated `otel.enabled`, default off.** Enabling it adds an init container and a `JAVA_TOOL_OPTIONS` javaagent to **every** java-spring service. That is the widest blast radius of any value in this chart and I cannot test it from here, so it lands off: reviewable now, switchable per environment with verification.

## Decisions worth reviewing

**Init container, not baked into images.** The service images are built upstream (`egovio/*`), so baking the agent would need 41 coordinated rebuilds and would pin the agent version per image. The init container copies the jar from the OTel Operator's auto-instrumentation image into an `emptyDir`. `agentJarPath` is configurable because that path differs between the operator image and a hand-built one.

**A dedicated `otel:` block, not `extraVolumes`.** The obvious route was to reuse the chart's existing extension points — but **6 app charts already set `extraVolumes`/`extraVolumeMounts`**, and their values win over the library defaults. The agent would have silently failed to attach to exactly those six. That is a partial, invisible failure, so it needed its own gate.

**The statefulset template is wired identically**, even though no chart uses it today. `extraEnv.java` is **shared** between both templates, so a future statefulset service would otherwise inherit `JAVA_TOOL_OPTIONS` without the jar being staged — and a `-javaagent` path that does not exist makes the **JVM fail at startup**, not degrade.

**Traces stay off.** `otel.tracesExporter: none`, and the collector has no traces pipeline. This tier has no trace backend — jaeger is `installed: false` and is a different architecture — and exporting to an absent backend logs failures on a loop, the failure mode #1664 documents. Both flip on together when a trace store lands, which is also when the dead Jaeger env this chart still injects into 25 live services should be removed.

## Rollout

1. Merge — nothing changes.
2. Set `otel.enabled: true` for **one** environment in `env.yaml`.
3. Verify pods still start (the agent adds JVM startup time and ~100 MB heap overhead — the Ansible tier omits it from `egov-localization` for exactly that reason).
4. Confirm `jvm_*` series appear in Prometheus with a `job` label per service.
5. The Ansible tier's JVM dashboard should render unchanged — same agent, same metric names.

## Verified / not verified

- [x] All five touched files parse; both templates gate identically (5 `otel.enabled` guards each)
- [x] Default off — no otel artifacts render, `OTEL_TRACES_EXPORTER` falls through to the existing `"none"`
- [x] Audited `extraVolumes` usage across all app charts — that is how the 6-chart clash surfaced
- [x] Values-path guard from #1652 still clean
- [ ] **Not verified on a cluster.** In particular: the jar path inside the operator image, the collector's Service DNS name in `otel.endpoint`, and whether the chart's `serviceMonitor.extraLabels` key is spelled that way in this chart version — all three are the kind of detail that has been wrong twice today.

### DIFF
--- a/devops/deploy-as-code/charts/common/values.yaml
+++ b/devops/deploy-as-code/charts/common/values.yaml
@@ -326,6 +326,14 @@ otel:
   # /javaagent.jar. Pinned to match the Ansible tier's agent version
   # (local-setup/otel/download-agent.sh pins 2.11.0) so metric names and the
   # JVM dashboard line up across both tiers.
+  #
+  # The 2.11.0 tag is published and pullable — GHCR's package page only shows
+  # recent tags by default, so check the registry, not the page. Verified
+  # 2026-08-10 against ghcr.io/v2/.../manifests/2.11.0: an OCI image index at
+  # sha256:b32b2153611bb11a98006d8d34c3ee91d307f491fa754ab66f5b0aff31a93a1b
+  # covering linux/amd64, arm64, s390x and ppc64le. Re-check before bumping:
+  # otel.enabled adds this as an init container to every service, so a bad tag
+  # is ImagePullBackOff across the whole tier at once.
   agentImage: "ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-java:2.11.0"
   # Path to the jar INSIDE agentImage. Configurable because it differs between
   # the operator image (/javaagent.jar) and a hand-built one.
--- a/devops/deploy-as-code/charts/common/values.yaml
+++ b/devops/deploy-as-code/charts/common/values.yaml
@@ -160,8 +160,30 @@ extraEnv:
           configMapKeyRef:
             name: egov-config
             key: spring-datasource-tomcat-initialSize
+      {{- if .Values.otel.enabled }}
+      # OTEL javaagent (#1646). JAVA_TOOL_OPTIONS is how the JVM picks the
+      # agent up without changing any service's entrypoint. Metrics AND traces
+      # travel one OTLP pipeline to the collector, which re-exposes metrics for
+      # Prometheus — so this single block is what gives 41 services JVM metrics.
+      - name: JAVA_TOOL_OPTIONS
+        value: "-javaagent:/otel/opentelemetry-javaagent.jar"
+      - name: OTEL_SERVICE_NAME
+        value: {{ template "common.name" . }}
+      - name: OTEL_EXPORTER_OTLP_ENDPOINT
+        value: {{ .Values.otel.endpoint | quote }}
+      - name: OTEL_EXPORTER_OTLP_PROTOCOL
+        value: "grpc"
+      - name: OTEL_METRICS_EXPORTER
+        value: "otlp"
+      - name: OTEL_TRACES_EXPORTER
+        value: {{ .Values.otel.tracesExporter | quote }}
+      # Logs stay on the promtail -> loki path this tier already uses.
+      - name: OTEL_LOGS_EXPORTER
+        value: "none"
+      {{- else }}
       - name: OTEL_TRACES_EXPORTER
         value: "none"
+      {{- end }}
       - name: SPRING_DATASOURCE_DRIVER_CLASS_NAME
         value: "org.postgresql.Driver"
       - name: SERVER_TOMCAT_MAX_THREADS
@@ -271,6 +293,43 @@ securityContext: {}
 
 service:
   additionalAnnotations: {}
+
+# ── Application telemetry (#1646) ──────────────────────────────────────
+# OFF BY DEFAULT, deliberately. Switching it on adds an init container and a
+# JAVA_TOOL_OPTIONS javaagent to EVERY java-spring service in the deployment —
+# the widest blast radius of any value in this chart. Enable it in env.yaml
+# for one environment, verify pods still start and metrics arrive, then roll
+# it wider.
+#
+# What it buys: JVM metrics for all 41 services, which this tier has none of
+# today (0 of 41 charts define a ServiceMonitor, and no service exposes a
+# micrometer/actuator endpoint for one to scrape). The agent exports OTLP to
+# the collector, which re-exposes everything on a SINGLE Prometheus endpoint —
+# so collection needs one ServiceMonitor, not 41.
+#
+# It also supersedes the Jaeger env this chart still injects (extraEnv.jaeger,
+# below): 25 live services point at a node-local jaeger agent that is not
+# deployed and, with useHostPort: false in the jaeger values, could not receive
+# them if it were. Remove that block once this is on everywhere.
+otel:
+  enabled: false
+  # Where the agent sends OTLP. Service name follows the collector release in
+  # charts/monitoring/monitoring-helmfile.yaml.
+  endpoint: "http://otel-collector-opentelemetry-collector.monitoring:4317"
+  # Traces need a backend. This tier has none yet (jaeger is installed: false
+  # and is a different architecture), so traces stay off while metrics flow.
+  # Flip to "otlp" when a trace store lands, and add the traces pipeline to
+  # the collector values at the same time.
+  tracesExporter: "none"
+  # Image carrying the javaagent jar. This is the image the OpenTelemetry
+  # Operator uses for Java auto-instrumentation; it ships the jar at
+  # /javaagent.jar. Pinned to match the Ansible tier's agent version
+  # (local-setup/otel/download-agent.sh pins 2.11.0) so metric names and the
+  # JVM dashboard line up across both tiers.
+  agentImage: "ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-java:2.11.0"
+  # Path to the jar INSIDE agentImage. Configurable because it differs between
+  # the operator image (/javaagent.jar) and a hand-built one.
+  agentJarPath: "/javaagent.jar"
 
 serviceMonitor:
   honorLabels: true
--- a/devops/deploy-as-code/charts/monitoring/values/otel-collector.yaml
+++ b/devops/deploy-as-code/charts/monitoring/values/otel-collector.yaml
@@ -8,9 +8,14 @@
 # receives on a single Prometheus endpoint, with per-service labels preserved
 # via the `service.name` resource attribute. That is what turns "add a
 # ServiceMonitor to each of 41 charts" into "add one ServiceMonitor for the
-# collector" — see #1646. It also carries traces, so it closes the tracing gap
-# in the same move: 25 live services currently emit Jaeger spans to an agent
-# that is not deployed and could not receive them if it were.
+# collector" — see #1646.
+#
+# Metrics only for now. The OTLP receiver below already accepts traces, but the
+# pipeline that would forward them is deliberately absent until a trace backend
+# lands on this tier — see the NOTE under `service.pipelines`, and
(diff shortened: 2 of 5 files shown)


devops/deploy-as-code/charts/monitoring/values/otel-collector.yaml
```diff
@@ -0,0 +1,114 @@
...
+#
+# Mirrors local-setup/otel/otel-collector-config.yaml, which is the working
+# arrangement on the Ansible tier, so both tiers converge on one telemetry
+# architecture rather than two.
+#
+# WHY THIS EXISTS AT ALL: the collector re-exposes every OTLP metric it
+# receives on a single Prometheus endpoint, with per-service labels preserved
+# via the `service.name` resource attribute. That is what turns "add a
+# ServiceMonitor to each of 41 charts" into "add one ServiceMonitor for the
+# collector" — see #1646. It also carries traces, so it closes the tracing gap
```
vinothrallapalli-eGov: This line says the collector "also carries traces, so it closes the tracing gap in the same move," but that's not what's configured: the `service.pipelines` block below (lines 54-58) defines only a `metrics` pipeline, and the comment at lines 59-66 explicitly documents "no traces pipeline yet" — matching `common/values.yaml`'s `otel.tracesExporter: "none"`. This header comment reads like a leftover from an earlier draft (or a copy from the compose-tier description, where Tempo *is* wired) and contradicts the file's own design a few lines down. Worth fixing so a future reader doesn't believe traces are already flowing through this pipeline. Suggest something like: "The receiver already accepts traces; a traces pipeline is added once a trace backend lands (see note below) — see #1646."

vinothrallapalli-eGov (commented):
Static review only (no cluster in this tier). The design is sound and matches the compose tier's javaagent -> collector -> Prometheus pipeline rather than inventing a second architecture. Verified against upstream: (1) `serviceMonitor.extraLabels` is genuinely the correct key for the `open-telemetry/opentelemetry-collector` chart at v0.108.0 (confirmed against upstream values.yaml, which even shows `release: kube-prometheus-stack` as its own example) — the PR's self-flagged uncertainty on this resolves clean; (2) the collector Service DNS name `otel-collector-opentelemetry-collector.monitoring:4317` matches Helm's fullname convention for release `otel-collector` + chart `opentelemetry-collector`, and the `monitoring` namespace matches the helmfile's `env.namespace: monitoring`, so the endpoint should resolve; (3) the javaagent injection is genuinely scoped — the otel env/volume/initContainer blocks live inside `extraEnv.java`, itself gated by `appType == "java-spring"` in `_deployment.yaml`/`_statefulset.yaml`, so non-Java services (postgres, kafka, nginx, UI, etc.) are structurally unaffected even with `otel.enabled: true`; (4) collector resources (512Mi limit / 256Mi request, `memory_limiter` at 400 MiB leaving ~22% GC headroom) look like a deliberate, reasonable fix for the class of issues flagged in #1612/#1635. One inline issue found: a stale/contradictory comment claiming the collector "also carries traces" when the pipeline config two sections later explicitly has no traces pipeline (and `common/values.yaml` sets `tracesExporter: "none"` for that exact reason). Also flagging the unpinned-tag risk on the javaagent image as worth a quick registry check before flipping `otel.enabled` anywhere, though that's already covered by the PR's own "not verified on a cluster" checklist.

devops/deploy-as-code/charts/common/values.yaml
```diff
@@ -272,6 +294,43 @@ securityContext: {}
...
+  # and is a different architecture), so traces stay off while metrics flow.
+  # Flip to "otlp" when a trace store lands, and add the traces pipeline to
+  # the collector values at the same time.
+  tracesExporter: "none"
+  # Image carrying the javaagent jar. This is the image the OpenTelemetry
+  # Operator uses for Java auto-instrumentation; it ships the jar at
+  # /javaagent.jar. Pinned to match the Ansible tier's agent version
+  # (local-setup/otel/download-agent.sh pins 2.11.0) so metric names and the
+  # JVM dashboard line up across both tiers.
+  agentImage: "ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-java:2.11.0"
```
vinothrallapalli-eGov, line 337: Minor/low-confidence: `ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-java:2.11.0` — I could only confirm recent tags (2.27.0+) are published for this image via the public package page; I couldn't independently verify a `2.11.0` tag still exists (older tags are just not shown on the default view, so this may well be fine). Since this is only pulled when `otel.enabled: true` (default off), it won't break anything until someone flips the switch, but worth a quick `docker manifest inspect` / registry check before doing that in any environment — an init-container `ImagePullBackOff` on 41 services at once would be a rough way to find out.

devops/deploy-as-code/charts/monitoring/values/otel-collector.yaml
singhalkarun: Correct, and it is exactly the leftover you guessed — that sentence was written against the compose tier, where Tempo is wired. Fixed in e1c78762.

The contradiction is real on all three counts: `service.pipelines` defines `metrics` only, the NOTE at lines 59-66 spells out why there is no traces pipeline, and `common/values.yaml:323` has `otel.tracesExporter: "none"`. Nothing forwards a span today.

Rewrote the header along the lines you suggested, keeping the Jaeger observation as the open problem it still is rather than deleting it:

```
# Metrics only for now. The OTLP receiver below already accepts traces, but the
# pipeline that would forward them is deliberately absent until a trace backend
# lands on this tier — see the NOTE under `service.pipelines`, and
# `otel.tracesExporter: "none"` in charts/common/values.yaml. So this does not
# yet close the tracing gap: 25 live services still emit Jaeger spans to an
# agent that is not deployed and could not receive them if it were.
```

Now the header, the NOTE and the values key say the same thing, and the "when this changes, change both" pointer is explicit in each direction.

devops/deploy-as-code/charts/common/values.yaml
singhalkarun: Checked the registry directly rather than the package page — the tag is fine, so no version change, but the evidence is now recorded next to the pin (e1c78762) so nobody has to redo this.

```
$ TOKEN=$(curl -s "https://ghcr.io/token?scope=repository:open-telemetry/opentelemetry-operator/autoinstrumentation-java:pull" | jq -r .token)
$ curl -sI -H "Authorization: Bearer $TOKEN" \
    -H 'Accept: application/vnd.oci.image.index.v1+json' \
    https://ghcr.io/v2/open-telemetry/opentelemetry-operator/autoinstrumentation-java/manifests/2.11.0
HTTP/1.1 200 OK
content-type: application/vnd.oci.image.index.v1+json
docker-content-digest: sha256:b32b2153611bb11a98006d8d34c3ee91d307f491fa754ab66f5b0aff31a93a1b
```

An OCI image index covering `linux/amd64`, `linux/arm64`, `linux/s390x`, `linux/ppc64le`. The full tags list has 73 entries and includes `2.11.0` alongside everything from `2.1.0` to `2.19.0` — your read of the package page was right, it just truncates the default view.

Your instinct about *why* to check was the right one and I have kept it in the file: `otel.enabled` adds this as an init container to every service at once, so a bad tag is `ImagePullBackOff` across the whole tier rather than one pod. Digest, platform list and that warning now sit beside the pin, which also stays deliberate: it matches `local-setup/otel/download-agent.sh` (`AGENT_VERSION="${OTEL_AGENT_VERSION:-2.11.0}"`) so metric names line up across both tiers, and bumping it is a two-tier change.

singhalkarun:
## Blocker: this release cannot install — the chart has a values schema

Tried to deploy the collector on a real cluster (kind, via this branch's monitoring helmfile). `helmfile sync` fails:

```
Error: values don't meet the specifications of the schema(s) in the following chart(s):
opentelemetry-collector:
- at '': additional properties 'pgadmin', 'configmaps', 'egov-persister',
  'egov-workflow-v2', 'pgr-services', 'novu-bridge', 'prometheus',
  'egov-localization', 'egov-indexer', 'grafana', 'digit-ui', ... not allowed
```

**Cause.** `opentelemetry-collector` 0.108.0 ships a `values.schema.json` with `additionalProperties: false`. The release uses the same whole-values passthrough as every other release here —

```yaml
values:
  - ./values/otel-collector.yaml
  - {{ .Values | toYaml | nindent 8 }}
```

— which injects **every top-level key of `env.yaml`** (~30 service blocks) into the chart. Charts without a schema silently drop them; this one rejects them.

That is the sharp irony of this initiative: the reason mis-pathed values have gone unnoticed for so long is that `kube-prometheus-stack` ships **no** schema. The one chart that does have one refuses to install at all.

**Attribution:** the `otel-collector` release does not exist on `monitoring-fix`; it is added by `2a220985` in this PR, so this is new, not inherited.

**It is statically detectable.** `helmfile -l name=otel-collector template` reproduces it exactly — no cluster required. Nothing in CI runs `helmfile template` over `devops/deploy-as-code/`, which is why it got this far. Strong argument for the guard in #1651/#1652 growing a template step.

**Fix options**, roughly in order of preference:

1. Scope the values for this release instead of passing the whole tree — pass only `./values/otel-collector.yaml` plus the handful of keys it needs.
2. Pass `--skip-schema-validation` (helm ≥3.16) via `helmDefaults`/release args — suppresses the symptom, keeps the pollution.
3. Filter the passthrough (e.g. `{{ .Values.otel | toYaml }}`) if a scoped sub-tree is introduced.

Option 1 is the only one that leaves the release with values it actually understands.

singhalkarun:

`a25abd6a` — dropped the whole-values passthrough from this release. `values/otel-collector.yaml` references no `.Values`, so nothing was lost.

Re-verified: `otel-collector-opentelemetry-collector … 1/1 Running`. It could not install at all before — helm rejected ~30 `env.yaml` keys against the chart's `values.schema.json`.

One aside worth recording: the first attempt at this fix broke the file in a different way. I wrote the passthrough expression literally inside a YAML comment to say it was absent, and helmfile templated it before parsing — dumping the values tree into the comment and failing with a parse error ~1500 lines away. Exactly the trap #1648 documents. Fixed in the follow-up commit, with the reason left in place.

Lokendra-egov (approved):

vinothrallapalli-eGov (approved):

Issue #1643, opened 2026-08-06
Fixed OTP is enabled by default in the egov-user chart, not just the env template


singhalkarun:
**P0-security · effort: S** · found while reviewing monitoring in `devops/deploy-as-code/` (#1619); not a monitoring issue.

## Problem

The fixed-OTP test switch is **enabled by default**, in the chart itself — not merely in a template an operator is expected to edit.

An OTP is how a citizen proves they control a phone number. With the fixed-OTP switch on, the platform stops generating a random code and accepts one constant value for **every** account. That value is `123456`, and it is in a public repository.

Two places set it, and the second is the one that matters:

**Environment template** — `charts/environments/env.yaml:111-112`:
```yaml
  citizen-otp-fixed: "123456"
  citizen-otp-fixed-enabled: "true"
```

**The chart's own defaults** — `charts/core-services/egov-user/values.yaml:58-59`, the same two keys with the same values.

That second location is why "an operator is expected to override it" is not a sufficient answer: **the values are the chart's defaults**. Removing the env.yaml block does not disable the feature; the chart still ships it on.

They reach the running container unconditionally when set (`egov-user/values.yaml:173-181`), as `CITIZEN_LOGIN_PASSWORD_OTP_FIXED_VALUE` and `CITIZEN_LOGIN_PASSWORD_OTP_FIXED_ENABLED`. The template guards on the key being *present*, not on it being *true* — so setting `citizen-otp-fixed-enabled: "false"` still injects the variable, and the behaviour then depends on how the service interprets the string rather than on the chart.

`egov-user` is `installed: true` (`charts/core-services/coreservices-helmfile.yaml:68-69`), so this is on the default deploy path.

## Related settings in the same file

- `charts/environments/env.yaml:23-24` —
  ```yaml
      ### make it false in production
      dev-enabled: "true"
  ```
  The comment shows the risk was known; the template still ships the unsafe value, with a comment as the only guard.
- `charts/environments/env.yaml:234` — `hrms-dev-mode: "true"`, consumed at `charts/common-services/egov-hrms/values.yaml:155`.
- `charts/environments/env.yaml:108` — `mobile-number-validation: "false"`.

## Why this ranks above the other exposures

A misconfigured admin console exposes data to whoever finds it. A fixed OTP undermines **identity for every citizen account**, and leaves no unusual trace — the resulting logins are indistinguishable from legitimate ones. There is nothing to detect after the fact.

## What I cannot see

[...]

PR #1669, opened 2026-08-07
fix(security): stop shipping fixed-OTP and dev-mode defaults in production charts (#1643)

singhalkarun:
Closes #1643

## What was wrong

`charts/core-services/egov-user/values.yaml` shipped these as **chart defaults**:

```yaml
citizen-otp-fixed: "123456"
citizen-otp-fixed-enabled: "true"
```

Fixed OTP means every citizen login accepts `123456`. Knowing any citizen's mobile number is then enough to log in as them — the OTP stops being a second factor at all.

The guard that was supposed to gate it tested for **key presence**, not truthiness:

```
{{- if index .Values "citizen-otp-fixed-enabled" }}
```

So setting it to `"false"` did not turn it off. The only way to disable it was to delete the key entirely — which is not what an operator reading `false` in a values file would expect. An operator who explicitly wrote `false` got fixed OTP anyway.

`env.yaml` also carried `dev-enabled: "true"` and `hrms-dev-mode: "true"`.

## Changes

- Remove the two fixed-OTP keys from the chart defaults so nothing is enabled unless a deployment opts in
- Fix the guard to test the value:
  ```
  {{- if eq (index .Values "citizen-otp-fixed-enabled" | default "false" | toString) "true" }}
  ```
  `| toString` matters — YAML `false` is a boolean, and comparing a boolean to the string `"true"` would fail the template rather than evaluate cleanly
- `env.yaml`: `dev-enabled: "false"`, `hrms-dev-mode: "false"`, fixed-OTP keys commented out with an explanation

## Exposure

**Nothing live is affected.** No cluster currently runs the `deploy-as-code` tier, so this default never reached a deployment — confirmed with the repo owner rather than assumed. This is a preventive fix: it stops the first real deployment from inheriting fixed OTP.

If that ever changes and you need to check an existing cluster:

```
kubectl get deploy egov-user -o yaml | grep -i CITIZEN_LOGIN_PASSWORD_OTP_FIXED
```

### DIFF
--- a/devops/deploy-as-code/charts/core-services/egov-user/values.yaml
+++ b/devops/deploy-as-code/charts/core-services/egov-user/values.yaml
@@ -55,8 +55,21 @@ default-password-expiry: "90"
 mobile-number-validation: "false"
 roles-state-level: "true"
 citizen-registration-withlogin: "true"
-citizen-otp-fixed: "123456"
-citizen-otp-fixed-enabled: "true"
+# Fixed OTP — a TEST convenience, off by default (#1643).
+#
+# When enabled the platform stops generating a per-login code and accepts one
+# constant for EVERY citizen account, so anyone who knows a phone number can
+# pass verification for it. These keys previously defaulted to "123456"/"true"
+# HERE, in the chart, which is why "operators are expected to override it in
+# env.yaml" was not a sufficient answer: deleting the env.yaml block left the
+# feature on.
+#
+# Absent = off: the injection below is conditional on the key existing, so a
+# deployment that does not set them ships neither env var and the service uses
+# its normal OTP path. Set both in a non-production environment only.
+#
+#   citizen-otp-fixed: "123456"
+#   citizen-otp-fixed-enabled: "true"
 heap: "-Xmx384m -Xms256m"
 java-args: ""
 
@@ -176,9 +189,13 @@ env: |
   - name: CITIZEN_LOGIN_PASSWORD_OTP_FIXED_VALUE
     value: {{ index .Values "citizen-otp-fixed" | quote }}
   {{- end }}
-  {{- if index .Values "citizen-otp-fixed-enabled" }}
+  {{- /* eq "true" rather than a bare truthiness test: every non-empty string
+         is truthy in Go templates, so `citizen-otp-fixed-enabled: "false"`
+         previously still injected the variable and left the outcome to however
+         the service parses it. Now "false" genuinely disables it here. */}}
+  {{- if eq (index .Values "citizen-otp-fixed-enabled" | default "false" | toString) "true" }}
   - name: CITIZEN_LOGIN_PASSWORD_OTP_FIXED_ENABLED
-    value: {{ index .Values "citizen-otp-fixed-enabled" | quote }}
+    value: "true"
   {{- end }}
   - name: JAVA_OPTS
     value: {{ index .Values "heap" | quote }}
--- a/devops/deploy-as-code/charts/core-services/egov-user/values.yaml
+++ b/devops/deploy-as-code/charts/core-services/egov-user/values.yaml
@@ -185,15 +185,22 @@ env: |
   - name: CITIZEN_REGISTRATION_WITHLOGIN_ENABLED
     value: {{ index .Values "citizen-registration-withlogin" | quote }}
   {{- end }}
+  {{- /* eq "true" rather than a bare truthiness test: every non-empty string
+         is truthy in Go templates, so `citizen-otp-fixed-enabled: "false"`
+         previously still injected the variable and left the outcome to however
+         the service parses it. Now "false" genuinely disables it here.
+
+         Both env vars hang off this one switch. The fixed value used to be
+         injected on its own presence check, so a config that set
+         `citizen-otp-fixed` and forgot `-enabled` still shipped the constant
+         into the pod's environment — readable by anyone with exec/describe,
+         and live for any consumer that keys off the value rather than the
+         flag. Off means neither variable is emitted. */}}
+  {{- if eq (index .Values "citizen-otp-fixed-enabled" | default "false" | toString) "true" }}
   {{- if index .Values "citizen-otp-fixed" }}
   - name: CITIZEN_LOGIN_PASSWORD_OTP_FIXED_VALUE
     value: {{ index .Values "citizen-otp-fixed" | quote }}
   {{- end }}
-  {{- /* eq "true" rather than a bare truthiness test: every non-empty string
-         is truthy in Go templates, so `citizen-otp-fixed-enabled: "false"`
-         previously still injected the variable and left the outcome to however
-         the service parses it. Now "false" genuinely disables it here. */}}
-  {{- if eq (index .Values "citizen-otp-fixed-enabled" | default "false" | toString) "true" }}
   - name: CITIZEN_LOGIN_PASSWORD_OTP_FIXED_ENABLED
     value: "true"
   {{- end }}
--- a/devops/deploy-as-code/charts/environments/env.yaml
+++ b/devops/deploy-as-code/charts/environments/env.yaml
@@ -20,8 +20,10 @@ configmaps:
       state-level-tenant-id: <tenant_id> ###### Replace with your own tenant ID if you are not using the default tenant ID (pg)
       host-map: "{'<tenant_id>':'https://<domain_name>/'}" # update state tenant-id & domain
 
-      ### make it false in production
-      dev-enabled: "true"
+      # Was "true" with a `### make it false in production` comment above it —
+      # a comment is not a control (#1643). Defaults to false; flip it in a
+      # non-production environment's own env.yaml.
+      dev-enabled: "false"
       egov-mdms-search-endpoint: /mdms-v2/v1/_search
 
   egov-service-host:
@@ -108,8 +110,10 @@ egov-user:
   mobile-number-validation: "false"
   roles-state-level: "true"
   citizen-registration-withlogin: "true"
-  citizen-otp-fixed: "123456"
-  citizen-otp-fixed-enabled: "true"
+  # Fixed OTP is a TEST convenience and is now off by default (#1643). Enable it
+  # ONLY in a non-production environment, by uncommenting both:
+  #   citizen-otp-fixed: "123456"
+  #   citizen-otp-fixed-enabled: "true"
   egov-state-level-tenant-id: <tenant_id>
 
 
@@ -231,7 +235,8 @@ egov-hrms:
   java-args: -Dspring.profiles.active=monitoring
   heap: "-Xmx192m -Xms192m"
   memory_limits: 386Mi
-  hrms-dev-mode: "true"
+  # Dev mode off by default (#1643) — same reasoning as dev-enabled above.
+  hrms-dev-mode: "false"
 
 
 boundary-service:


devops/deploy-as-code/charts/core-services/egov-user/values.yaml
```diff
@@ -176,9 +189,13 @@ env: |
   - name: CITIZEN_LOGIN_PASSWORD_OTP_FIXED_VALUE
     value: {{ index .Values "citizen-otp-fixed" | quote }}
   {{- end }}
-  {{- if index .Values "citizen-otp-fixed-enabled" }}
+  {{- /* eq "true" rather than a bare truthiness test: every non-empty string
+         is truthy in Go templates, so `citizen-otp-fixed-enabled: "false"`
+         previously still injected the variable and left the outcome to however
+         the service parses it. Now "false" genuinely disables it here. */}}
+  {{- if eq (index .Values "citizen-otp-fixed-enabled" | default "false" | toString) "true" }}
```
vinothrallapalli-eGov, line 199: Confirmed: this now checks the actual value (`eq ... "true"`) rather than mere key presence, defaults to `"false"` when the key is absent/empty, and `toString`s first so a YAML boolean doesn't error against the string comparison. This is the correct fix for the reported bug (old code: `{{- if index .Values "citizen-otp-fixed-enabled" }}`, which is truthy even for the literal string `"false"`).

Bonus defense-in-depth: because the resulting env var is only ever emitted as `"true"` and is otherwise omitted entirely (not set to `"false"`), this also protects against a downstream consumer that only checks env-var *presence* rather than parsing its value — which is plausibly where the original account-takeover bug actually lived (in the egov-user app, outside this repo).

devops/deploy-as-code/charts/core-services/egov-user/values.yaml
```diff
@@ -176,9 +189,13 @@ env: |
   - name: CITIZEN_LOGIN_PASSWORD_OTP_FIXED_VALUE
     value: {{ index .Values "citizen-otp-fixed" | quote }}
```
vinothrallapalli-eGov: Minor/non-blocking: this `CITIZEN_LOGIN_PASSWORD_OTP_FIXED_VALUE` injection (guarded above by `{{- if index .Values "citizen-otp-fixed" }}`, unchanged by this PR) is still a bare presence check, decoupled from `citizen-otp-fixed-enabled`. Not exploitable today since both keys default to absent, but if a future config sets only `citizen-otp-fixed` without also setting `-enabled: "true"`, the raw fixed-OTP value still lands in the pod's env. Consider gating this on the same `-enabled` check for defense-in-depth.

devops/deploy-as-code/charts/environments/env.yaml
```diff
@@ -20,8 +20,10 @@ configmaps:
       state-level-tenant-id: <tenant_id> ###### Replace with your own tenant ID if you are not using the default tenant ID (pg)
       host-map: "{'<tenant_id>':'https://<domain_name>/'}" # update state tenant-id & domain
 
-      ### make it false in production
-      dev-enabled: "true"
+      # Was "true" with a `### make it false in production` comment above it —
+      # a comment is not a control (#1643). Defaults to false; flip it in a
+      # non-production environment's own env.yaml.
+      dev-enabled: "false"
```
vinothrallapalli-eGov, line 26: Confirmed default flipped from `"true"` to `"false"`. This one isn't behind a Helm truthiness guard (it's a plain configmap value), so there's no presence-vs-value bug class here — flipping the default is a complete fix for this key.

devops/deploy-as-code/charts/environments/env.yaml
```diff
@@ -231,7 +235,8 @@ egov-hrms:
   java-args: -Dspring.profiles.active=monitoring
   heap: "-Xmx192m -Xms192m"
   memory_limits: 386Mi
-  hrms-dev-mode: "true"
+  # Dev mode off by default (#1643) — same reasoning as dev-enabled above.
+  hrms-dev-mode: "false"
```
vinothrallapalli-eGov, line 239: Same as `dev-enabled`: plain value flip from `"true"` to `"false"`, no guard involved, no residual bug class here.

vinothrallapalli-eGov (commented):
Verified the value-vs-presence claim against the current diff: the fix is real and correctly implemented.

The guard for `CITIZEN_LOGIN_PASSWORD_OTP_FIXED_ENABLED` now reads `{{- if eq (index .Values "citizen-otp-fixed-enabled" | default "false" | toString) "true" }}` (egov-user/values.yaml:196) instead of the old `{{- if index .Values "citizen-otp-fixed-enabled" }}`. That old form is truthy for ANY non-empty string, including the literal string `"false"` — exactly the presence-not-value bug described. The new form does a real equality check against the string `"true"`, defaults missing/empty to `"false"` (fail-closed), and `toString`s the value first so a YAML boolean `false` doesn't blow up the `eq` (Go templates error comparing bool to string). This is correctly reasoned and correctly written.

It also goes a step further than a pure value check: when disabled, the env var is omitted entirely (`value: "true"` is hardcoded inside the now-value-checked branch) rather than being emitted as the string `"false"`. That's a good defensive choice — it protects against a hypothetical downstream consumer (the egov-user app itself, which lives outside this repo) that checks env-var *presence* rather than parsing its value, which is plausibly the actual origin of the original account-takeover bug.

Defaults are also fixed at the source: both `citizen-otp-fixed` and `citizen-otp-fixed-enabled` are removed from chart defaults (values.yaml) and commented out in the environments template (env.yaml), so nothing ships them unless an operator explicitly opts in. `dev-enabled` and `hrms-dev-mode` are flipped from `"true"` to `"false"` at their definitions — no guard involved there, so no equivalent bug class applies to those two.

Two secondary observations, both non-blocking and out of this PR's stated scope:
1. The `CITIZEN_LOGIN_PASSWORD_OTP_FIXED_VALUE` env var (values.yaml ~188-191) is still gated purely on key presence (`{{- if index .Values "citizen-otp-fixed" }}`), decoupled from the `-enabled` flag. Harmless today since both defaults are absent, but if an operator ever sets only the value without the enabled flag, the raw OTP value still gets injected into the pod env.
2. Several sibling boolean flags in the same values.yaml `env:` block (`mobile-number-validation`, `roles-state-level`, `citizen-registration-withlogin`, `citizen-otp-enabled`, `employee-otp-enabled`, `otp-validation`) still use the old bare `{{- if index .Values "X" }}` presence guard. They aren't touched by this PR's diff so I didn't inline-comment them, but they're the same code shape that caused this bug and worth a follow-up pass if the egov-user app has the same presence-vs-value habit for any of them.

Overall: the specific value-vs-presence bug reported is fixed correctly and defensively. No blocking issues found in this diff.

devops/deploy-as-code/charts/core-services/egov-user/values.yaml
singhalkarun: Thanks — and your read of where the original bug lived is worth acting on, so I extended it.

Your point about presence-vs-value checking applies one block up too. `CITIZEN_LOGIN_PASSWORD_OTP_FIXED_VALUE` was gated on a bare `{{- if index .Values "citizen-otp-fixed" }}`, independent of the flag. See the thread on line 190 — both variables now hang off this single `eq "true"` check, pushed as 0cdc8f89.

devops/deploy-as-code/charts/core-services/egov-user/values.yaml
singhalkarun: Valid, and it reproduces. Confirmed with `helm template` on the egov-user chart before the fix:

| values set | rendered env |
|---|---|
| none | neither var |
| `citizen-otp-fixed: "123456"` only | **`CITIZEN_LOGIN_PASSWORD_OTP_FIXED_VALUE=123456`** |
| value + `-enabled: "false"` | **`CITIZEN_LOGIN_PASSWORD_OTP_FIXED_VALUE=123456`** |
| value + `-enabled: "true"` | both |

So the fixed OTP reached the pod's environment in two configurations where the feature is meant to be off — readable via `describe`/`exec`, and live for any consumer keying off the value rather than the flag. That second row is exactly the "future config sets only `citizen-otp-fixed`" case you described; the third is arguably worse, since it looks explicitly disabled.

Fixed in 0cdc8f89 by nesting the value inside the same `eq "true"` guard. Re-rendered, all five cases:

| values set | rendered env |
|---|---|
| none | neither |
| `citizen-otp-fixed` only | neither |
| value + `-enabled: "false"` | neither |
| value + `-enabled: "true"` | both |
| `-enabled: "true"`, no value | flag only |

Off now means neither variable is emitted.

Not verifiable here: no cluster runs the `deploy-as-code` tier, so this is render-level evidence only — I have not confirmed how egov-user itself behaves when it receives the flag.

devops/deploy-as-code/charts/environments/env.yaml
singhalkarun: Agreed, thanks for checking it separately from the guarded keys — `dev-enabled` is consumed as a plain configmap value, so there is no truthiness bug class here and the flip is the whole fix.

devops/deploy-as-code/charts/environments/env.yaml
singhalkarun: Agreed — same as `dev-enabled`, a plain value with no guard in front of it, so nothing further to do on this key.

devops/deploy-as-code/charts/environments/env.yaml
vinothrallapalli-eGov, line 26: @singhalkarun, please keep the default value as true, as this setup can be used for both production and non-production environments. When a partner initially sets up and starts using the system, they need the default credentials to access it. Until the SMS integration is completed, both employee and citizen login will remain blocked. Hence, we are keeping it true by default. Once the complete setup and SMS integration are done, they can modify the value as required.

devops/deploy-as-code/charts/environments/env.yaml
```diff
@@ -108,8 +110,10 @@ egov-user:
   mobile-number-validation: "false"
   roles-state-level: "true"
   citizen-registration-withlogin: "true"
-  citizen-otp-fixed: "123456"
```
vinothrallapalli-eGov, line 111: @singhalkarun please keep the default value as true, as this setup can be used for both production and non-production environments. When a partner initially sets up and starts using the system, they need the default credentials to access it. Until the SMS integration is completed, both employee and citizen login will remain blocked. Hence, we are keeping it true by default. Once the complete setup and SMS integration are done, they can modify the value as required.

devops/deploy-as-code/charts/core-services/egov-user/values.yaml
```diff
@@ -172,13 +185,24 @@ env: |
...
+         flag. Off means neither variable is emitted. */}}
+  {{- if eq (index .Values "citizen-otp-fixed-enabled" | default "false" | toString) "true" }}
   {{- if index .Values "citizen-otp-fixed" }}
   - name: CITIZEN_LOGIN_PASSWORD_OTP_FIXED_VALUE
     value: {{ index .Values "citizen-otp-fixed" | quote }}
   {{- end }}
-  {{- if index .Values "citizen-otp-fixed-enabled" }}
   - name: CITIZEN_LOGIN_PASSWORD_OTP_FIXED_ENABLED
-    value: {{ index .Values "citizen-otp-fixed-enabled" | quote }}
+    value: "true"
```
vinothrallapalli-eGov, line 205: @singhalkarun Do not hard code, this should able to be overridded from env.yml

vinothrallapalli-eGov (changes requested):

Issue #1642, opened 2026-08-06
pgAdmin is deployed publicly by default with demo credentials and all protections commented out


singhalkarun:
**P0-security · effort: S** · found while reviewing monitoring in `devops/deploy-as-code/` (#1619), but this is not a monitoring issue.

## Problem

pgAdmin — a full web console for browsing and editing Postgres — is **deployed by default, published on the public domain, with no gate in front of it, and with demo credentials shipped in the repo.**

Each part verified:

**1. It deploys by default.** Unlike the monitoring stack, this is not switched off:
- `digit-helmfile.yaml:17` includes `./charts/auxiliary-services/auxiliary-helmfile.yaml`
- `charts/auxiliary-services/auxiliary-helmfile.yaml:22-23` — `- name: pgadmin` / `installed: true`

**2. It is published on the internet.** `charts/auxiliary-services/pgadmin/values.yaml:7-9` — `ingress.enabled: true`, `context: "pgadmin"`, so it is served at `/pgadmin` on the deployment's main domain.

**3. Every protection is commented out.** `pgadmin/values.yaml:10-17` carries an `additionalAnnotations` block containing an IP allow-list (`whitelist-source-range`) and an SSO redirect (`auth-signin` / `auth-url`). All of it is commented. So there is no network restriction and no single sign-on — the only barrier is pgAdmin's own login form.

Worth noting the SSO annotations could not work even if uncommented: they point at `/oauth2/auth`, served by the `oauth2-proxy` chart, which **exists in the tree but is in no helmfile and is never deployed**.

**4. The credentials for that form ship in the repo.** `charts/environments/env-secrets.yaml:75-79` sets an admin account and a read account to demo values. `charts/environments/env.yaml` additionally sets `cookie-protection: "False"`.

## The part that makes this a real defect rather than a bad demo default

The database URL beside those credentials ends with the comment `### replace DB URL`. These are clearly **template defaults an operator is expected to override** — and I cannot see a live cluster, so I cannot say whether any deployment still has them.

That is precisely the issue: **nothing forces the override.**

- no validation that the credentials changed
- no failure if they did not
- **no CI runs against `devops/deploy-as-code/` at all** — a PR touching this directory triggers zero workflows

So the failure mode is silent. An operator who misses one line in a large values file gets an internet-facing database console protected by the word "demo", with nothing anywhere to tell them.

## Proposed fix

[...]

PR #1670, opened 2026-08-07
fix(security): do not install pgAdmin by default; drop demo credentials (#1642)

singhalkarun:
Closes #1642

## What was wrong

`auxiliary-helmfile.yaml` had pgAdmin at `installed: true`, and `env-secrets.yaml` carried demo credentials for it.

pgAdmin is a full database administration console with direct access to the DIGIT database — citizen names, mobile numbers, complaint history. Shipping it installed-by-default with published credentials means anyone who can reach the endpoint gets an interactive SQL console against production data. It is a debugging convenience, not something a deployment needs to run.

## Changes

- `installed: false`, with a comment block explaining what pgAdmin is, why it is off, and how to turn it on deliberately
- Demo passwords in `env-secrets.yaml` replaced with `change-me-strong`, so an operator who does enable it has to make a choice rather than inherit a known password

## Exposure

**Nothing live is affected.** No cluster currently runs the `deploy-as-code` tier, so pgAdmin was never actually deployed with these credentials — confirmed with the repo owner rather than assumed. This is a preventive fix.

If that ever changes, `kubectl get deploy -A | grep -i pgadmin` is the check, and demo credentials would need treating as compromised.

### DIFF
--- a/devops/deploy-as-code/charts/auxiliary-services/auxiliary-helmfile.yaml
+++ b/devops/deploy-as-code/charts/auxiliary-services/auxiliary-helmfile.yaml
@@ -25,13 +25,16 @@ releases:
   # SSO redirect are commented out in pgadmin/values.yaml, and the SSO they
   # point at (`/oauth2/auth`, served by the oauth2-proxy chart) is in NO
   # helmfile and never deployed — so uncommenting them as-is would not protect
-  # it either. The only barrier is pgAdmin's own login, whose credentials ship
-  # in env-secrets.yaml as demo@demo.com / demo.
+  # it either. The only barrier is pgAdmin's own login, and that login used to
+  # ship in env-secrets.yaml as demo@demo.com / demo.
   #
   # Default OFF (#1642). It is an operator convenience, not a serving
   # dependency, so it should be a deliberate choice rather than something every
   # deployment inherits. Before setting this true:
-  #   1. change secrets.pgadmin.admin-password and read-password
+  #   1. fill in all four secrets.pgadmin fields in env-secrets.yaml. They now
+  #      ship blank, and blank means the `pgadmin` Secret is never created, so
+  #      enabling this release without setting them leaves the pod unable to
+  #      start rather than reachable with a known password.
   #   2. reach it through a tunnel or bastion, or deploy oauth2-proxy and
   #      uncomment the annotations in pgadmin/values.yaml
   - name: pgadmin
--- a/devops/deploy-as-code/charts/auxiliary-services/auxiliary-helmfile.yaml
+++ b/devops/deploy-as-code/charts/auxiliary-services/auxiliary-helmfile.yaml
@@ -19,8 +19,23 @@ templates:
       - {{ .Values | toYaml | nindent 8 }}
 
 releases:
+  # pgAdmin is a full read/write console onto the tenant database, and this
+  # chart publishes it on the deployment's own domain at /pgadmin with NO gate
+  # in front of it: the ingress annotations carrying an IP allow-list and an
+  # SSO redirect are commented out in pgadmin/values.yaml, and the SSO they
+  # point at (`/oauth2/auth`, served by the oauth2-proxy chart) is in NO
+  # helmfile and never deployed — so uncommenting them as-is would not protect
+  # it either. The only barrier is pgAdmin's own login, whose credentials ship
+  # in env-secrets.yaml as demo@demo.com / demo.
+  #
+  # Default OFF (#1642). It is an operator convenience, not a serving
+  # dependency, so it should be a deliberate choice rather than something every
+  # deployment inherits. Before setting this true:
+  #   1. change secrets.pgadmin.admin-password and read-password
+  #   2. reach it through a tunnel or bastion, or deploy oauth2-proxy and
+  #      uncomment the annotations in pgadmin/values.yaml
   - name: pgadmin
-    installed: true
+    installed: false
     <<: *default
 
   - name: kafka-connect
--- a/devops/deploy-as-code/charts/auxiliary-services/auxiliary-helmfile.yaml
+++ b/devops/deploy-as-code/charts/auxiliary-services/auxiliary-helmfile.yaml
@@ -28,17 +28,32 @@ releases:
   # it either. The only barrier is pgAdmin's own login, and that login used to
   # ship in env-secrets.yaml as demo@demo.com / demo.
   #
-  # Default OFF (#1642). It is an operator convenience, not a serving
-  # dependency, so it should be a deliberate choice rather than something every
-  # deployment inherits. Before setting this true:
-  #   1. fill in all four secrets.pgadmin fields in env-secrets.yaml. They now
-  #      ship blank, and blank means the `pgadmin` Secret is never created, so
-  #      enabling this release without setting them leaves the pod unable to
-  #      start rather than reachable with a known password.
-  #   2. reach it through a tunnel or bastion, or deploy oauth2-proxy and
-  #      uncomment the annotations in pgadmin/values.yaml
+  # Left ON deliberately: operators substitute secrets and configuration before
+  # deploying, so pgAdmin stays available by default. What #1642 changes is the
+  # credentials, not the install flag -- and that is where the actual exposure
+  # was. `installed: false` never gated it anyway: the `pgadmin` Secret is
+  # rendered by core-services/configmaps, which every deployment applies, so a
+  # default deploy created that Secret from the values shipped in git whether or
+  # not pgAdmin was installed.
+  #
+  # Two things are required of whoever deploys this:
+  #   1. fill in all four secrets.pgadmin fields in env-secrets.yaml. They ship
+  #      BLANK, and blank means the `pgadmin` Secret is never created -- so
+  #      leaving them unset gives a pod that cannot start (it reads
+  #      admin-email/admin-password by secretKeyRef) rather than one reachable
+  #      with a password published in this repo. Placeholders are refused at
+  #      render time, so a half-substituted file fails loudly instead of
+  #      deploying a known credential.
+  #   2. put something in front of it. This is a full read/write console onto
+  #      the tenant database, published on the deployment's own domain at
+  #      /pgadmin, and its own login is currently the only barrier: the ingress
+  #      annotations carrying an IP allow-list and an SSO redirect are commented
+  #      out in pgadmin/values.yaml, and the SSO they point at (`/oauth2/auth`,
+  #      served by the oauth2-proxy chart) is in NO helmfile and never deployed,
+  #      so uncommenting them as-is would not protect it either. Reach it
+  #      through a tunnel or bastion, or deploy oauth2-proxy first.
   - name: pgadmin
-    installed: false
+    installed: true
     <<: *default
 
   - name: kafka-connect
--- a/devops/deploy-as-code/charts/environments/env-secrets.yaml
+++ b/devops/deploy-as-code/charts/environments/env-secrets.yaml
@@ -73,10 +73,14 @@ secrets:
         clientID: QVbnq0L8npoyfxZs96wtBg==
     # To work pgadmin service add details
     pgadmin:
+        # Only used when the pgadmin release is enabled — it ships
+        # installed: false (#1642). These are placeholders, not credentials:
+        # change them before enabling, because the chart publishes pgAdmin on
+        # the public domain with no gate in front of it.
         admin-email: demo@demo.com
-        admin-password: demo
+        admin-password: change-me-strong
         read-email: demo@demo.com
-        read-password: demo
+        read-password: change-me-strong
     # To clone the config and mdms repos, Create github user and add your ssh private key below. This private should have access to config and mdms repo.
     git-sync:
         ssh: |-
--- a/devops/deploy-as-code/charts/environments/env-secrets.yaml
+++ b/devops/deploy-as-code/charts/environments/env-secrets.yaml
@@ -71,16 +71,29 @@ secrets:
         cookieSecret: qwgethjymnbv
         clientSecret: 3a08079easd9d8055470475696fd3baad5292
         clientID: QVbnq0L8npoyfxZs96wtBg==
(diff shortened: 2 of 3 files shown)


devops/deploy-as-code/charts/auxiliary-services/auxiliary-helmfile.yaml
```diff
@@ -19,8 +19,23 @@ templates:
...
 
 releases:
+  # pgAdmin is a full read/write console onto the tenant database, and this
+  # chart publishes it on the deployment's own domain at /pgadmin with NO gate
+  # in front of it: the ingress annotations carrying an IP allow-list and an
+  # SSO redirect are commented out in pgadmin/values.yaml, and the SSO they
+  # point at (`/oauth2/auth`, served by the oauth2-proxy chart) is in NO
+  # helmfile and never deployed — so uncommenting them as-is would not protect
+  # it either. The only barrier is pgAdmin's own login, whose credentials ship
+  # in env-secrets.yaml as demo@demo.com / demo.
```
vinothrallapalli-eGov, line 29: Nit: this comment says the shipped credentials are "demo@demo.com / demo", but this same PR changes the password to `change-me-strong` in env-secrets.yaml (line 81 there). Worth updating so a future operator reading this doesn't assume the password is still `demo`.

devops/deploy-as-code/charts/environments/env-secrets.yaml
```diff
@@ -73,10 +73,14 @@ secrets:
         clientID: QVbnq0L8npoyfxZs96wtBg==
     # To work pgadmin service add details
     pgadmin:
+        # Only used when the pgadmin release is enabled — it ships
+        # installed: false (#1642). These are placeholders, not credentials:
+        # change them before enabling, because the chart publishes pgAdmin on
+        # the public domain with no gate in front of it.
         admin-email: demo@demo.com
-        admin-password: demo
+        admin-password: change-me-strong
```
vinothrallapalli-eGov, line 81: This swaps one hardcoded default for another rather than removing the fallback risk outright. I checked the only consumer (`core-services/configmaps/templates/secrets/pgadmin-secret.yaml`) — it just checks the value is non-empty and renders it, no format/placeholder check. So nothing stops an operator from setting `installed: true` upstream and deploying with this exact literal string, which is now public in this repo. Low severity given `installed: false` is the real gate, but the PR description's claim that enabling pgadmin "forces a choice" isn't technically enforced — worth a `required`/placeholder-detection guard if that guarantee matters, or just soften the claim in the description.

vinothrallapalli-eGov (commented):
Verified statically (no live cluster runs this tier, per the PR). The core fix is solid: `pgadmin.installed` is flipped to `false` in `auxiliary-helmfile.yaml` and this is the only `releases:` entry for pgadmin in the repo (no other helmfile overrides it back to `true`), so the default genuinely gates deployment. `env-secrets.yaml`'s demo passwords are replaced with `change-me-strong`, and the only consumer of those values (`core-services/configmaps/templates/secrets/pgadmin-secret.yaml`) reads them straight through with no `demo`/`admin` fallback baked into the template, so the literal string is really gone from the render path. Two minor, non-blocking issues: (1) the new explanatory comment in `auxiliary-helmfile.yaml` still says the shipped credentials are "demo@demo.com / demo", which this same PR makes untrue; (2) the credential change is a placeholder swap, not an enforced one — nothing (e.g. a Helm `required` guard) stops an operator from setting `installed: true` and deploying with the literal `change-me-strong` still in place, so the PR's claim that enabling it "forces a choice" isn't technically backed. Neither undermines the primary control (`installed: false`), which is the part that actually matters here.

devops/deploy-as-code/charts/environments/env-secrets.yaml
vinothrallapalli-eGov, line 81: This is a placeholder swap, not a removal of demo credentials — `change-me-strong` is just as much a known, hardcoded value in git history as `demo` was. If an operator sets `installed: true` on the pgadmin release without reading the comment above (or without noticing this line), pgAdmin comes up with a publicly-known password and, per the PR's own analysis, no gate in front of it (SSO/allow-list annotations are commented out and oauth2-proxy isn't deployed). Consider making this fail loudly instead of silently working: e.g. a pre-install hook/values check that errors when `admin-password` still equals `change-me-strong`, or templating a randomly-generated secret when the value is left at the placeholder. As-is, the safety here is purely documentation, which is the same failure mode this PR is trying to close.

devops/deploy-as-code/charts/environments/env-secrets.yaml
```diff
@@ -73,10 +73,14 @@ secrets:
         clientID: QVbnq0L8npoyfxZs96wtBg==
     # To work pgadmin service add details
     pgadmin:
+        # Only used when the pgadmin release is enabled — it ships
+        # installed: false (#1642). These are placeholders, not credentials:
+        # change them before enabling, because the chart publishes pgAdmin on
+        # the public domain with no gate in front of it.
         admin-email: demo@demo.com
```
vinothrallapalli-eGov, line 80: Minor: `admin-email`/`read-email` are left as `demo@demo.com` (only the passwords were changed). Not a secret by itself, but since the PR title says "drop demo credentials," worth confirming this is intentional (username isn't sensitive) rather than an oversight.

vinothrallapalli-eGov (commented):
Verified the core fix: `installed: false` is genuinely set on the pgadmin release (auxiliary-helmfile.yaml) with no other helmfile/values override reinstating `true`, and no fallback default elsewhere in the chart tree (checked core-services/configmaps and env.yaml) resurrects the old creds. The explanatory comment block accurately reflects the chart's actual state (ingress SSO/allow-list annotations are commented out in pgadmin/values.yaml, and oauth2-proxy isn't in any helmfile) — good that this was verified against the code rather than just asserted. One gap: the "demo credentials" aren't removed, just swapped for another static placeholder (`change-me-strong`) committed in plaintext to env-secrets.yaml. Nothing in the chart enforces that an operator who flips `installed: true` actually changes it — see inline comment. Low severity given pgAdmin defaults to off and this is explicitly scoped as a preventive fix with no live cluster affected, but worth a follow-up (e.g. a startup check that fails when the password still equals the placeholder, or generating a random secret if unset).

devops/deploy-as-code/charts/auxiliary-services/auxiliary-helmfile.yaml
singhalkarun: Correct, the comment was stale — and it is now wrong in the other direction too, because 46c942a7 drops the placeholder password entirely rather than replacing it (see the env-secrets.yaml threads).

Rewritten: the credentials are described in the past tense as what *used to* ship, and step 1 now says to fill in all four `secrets.pgadmin` fields, noting they ship blank and that blank means the Secret is never created — so enabling the release without setting them leaves the pod unable to start rather than reachable with a known password.

While confirming the rest of this comment I verified the two claims it makes: the `oauth2-proxy` chart exists at `auxiliary-services/oauth2-proxy` but appears in no helmfile, and the allow-list/SSO annotations are still commented out in `pgadmin/values.yaml`. Both accurate, left as written.

devops/deploy-as-code/charts/environments/env-secrets.yaml
singhalkarun: You are right, and digging into it turned up something worse than the placeholder swap.

**`installed: false` never kept these credentials out of the cluster.** The Secret is rendered by `core-services/configmaps`, not by the pgadmin chart, and that chart is applied by every deployment. `helm template` on the branch as it stood:

```
$ helm template cm ./core-services/configmaps -f environments/env-secrets.yaml -f environments/env.yaml
# Source: configmaps/templates/secrets/pgadmin-secret.yaml
kind: Secret
metadata:
  name: pgadmin
  namespace: backbone
data:
  admin-email: "ZGVtb0BkZW1vLmNvbQ=="       # demo@demo.com
  admin-password: "Y2hhbmdlLW1lLXN0cm9uZw==" # change-me-strong
```

18 objects, including that one, with the pgadmin release disabled. So the gate you correctly identified as "the real gate" was not gating this at all.

And your core point stands: `env-secrets.yaml` is **not** SOPS-encrypted (no `sops` metadata in the file), so anything written there is published in this repo regardless of spelling.

Fixed in 46c942a7 by removing the fallback rather than renaming it. All four fields ship blank, and blank is the switch — no `admin-password` means no Secret. Once set, all four are validated, following `db-seed/templates/secret.yaml`, where guarding only some fields left exactly the gap the check existed to close. Placeholders are refused as well as blanks, since a placeholder is a non-empty string that renders straight through:

| config | result |
|---|---|
| default (blank) | **17 objects, no pgadmin Secret** (was 18) |
| `admin-password: <foo>` | `fail` — names the field and the file |
| `admin-password: change-me-strong` | `fail` — notes the value is published in this repo |
| password set, `read-email` blank | `fail` — names `read-email` |
| all four real | Secret renders |

On the PR description's "forces a choice" claim, which you were right to push on: it is now enforced rather than asserted. Enabling the release without filling these in leaves the pod unable to start — the deployment reads `admin-email`/`admin-password` by `secretKeyRef` from a Secret that will not exist.

Not verifiable here: no cluster runs the `deploy-as-code` tier, so the "pod cannot start" step is reasoned from the `secretKeyRef` in `pgadmin/values.yaml` plus the absent Secret, not observed.

singhalkarun: Agreed on all of it — "a placeholder swap, not a removal" and "the safety here is purely documentation, which is the same failure mode this PR is trying to close" are both fair, and the second is the one that made me redo the approach rather than patch the comment.

Taking your two suggested options: I went with neither a random secret nor a pre-install hook, but with removing the fallback outright — the fields ship blank, blank means no Secret is created, and a placeholder that *does* reach the template is a hard `fail` at render time. Detail and the `helm template` evidence in the sibling thread on this line, including the finding that `installed: false` was not keeping the Secret out of the cluster in the first place.

The `fail` refuses both `<...>` and any `change-me*` value, so the specific string this PR had introduced is rejected by name:

```
pgadmin: secrets.pgadmin.admin-password is still the placeholder "change-me-strong",
which is published in this repo. Substitute a real value in environments/env-secrets.yaml.
```

devops/deploy-as-code/charts/environments/env-secrets.yaml
singhalkarun: Good catch, and it was an oversight rather than a deliberate call — the first pass only looked at fields it thought of as secret.

It is moot now: 46c942a7 blanks all four fields, `admin-email` and `read-email` included, so no `demo@demo.com` remains. Worth doing beyond tidiness, because the email is the *username* for that login — shipping a known username alongside a known password made the pair complete, and a known username on its own still removes a guess from anyone probing the endpoint.

The template also rejects a blank email once `admin-password` is set, so filling in only the passwords now fails the render by name rather than deploying pgAdmin with an empty username:

```
pgadmin: secrets.pgadmin.read-email is empty. Set all four pgadmin credential fields
in environments/env-secrets.yaml, or leave admin-password blank to not deploy pgAdmin at all.
```

vinothrallapalli-eGov (approved):

devops/deploy-as-code/charts/auxiliary-services/auxiliary-helmfile.yaml
```diff
@@ -19,8 +19,23 @@ templates:
...
+  #
+  # Default OFF (#1642). It is an operator convenience, not a serving
+  # dependency, so it should be a deliberate choice rather than something every
+  # deployment inherits. Before setting this true:
+  #   1. change secrets.pgadmin.admin-password and read-password
+  #   2. reach it through a tunnel or bastion, or deploy oauth2-proxy and
+  #      uncomment the annotations in pgadmin/values.yaml
   - name: pgadmin
-    installed: true
+    installed: false
```
vinothrallapalli-eGov: @singhalkarun Do not disable, the secrets and configuration will be modified by end user before deploying so no need to disable it.

vinothrallapalli-eGov (changes requested):

singhalkarun:
Done — `installed: true` in `4263fa80`.

Agreed this does not weaken the fix, because the install flag was never what was
holding the line. The `pgadmin` Secret is rendered by `core-services/configmaps`,
which every deployment applies, so a default deploy created that Secret from the
values shipped in git whether or not pgAdmin was installed. What closes the hole
is the credentials shipping blank plus render-time validation, and both are
untouched.

Re-verified all four paths **with `installed: true`**, against the shipped
`env-secrets.yaml`:

| config | result |
|---|---|
| all four blank (as shipped) | **0** pgadmin Secrets rendered |
| only `admin-password` set | `fail` — names `admin-email` |
| `admin-password: change-me-strong` | `fail` — names it as published in this repo |
| all four real | Secret renders |

Two things a deployer now has to do, both stated in the helmfile comment:

1. **Fill in all four fields.** Blank means no Secret, and the Deployment reads
   `admin-email` / `admin-password` by `secretKeyRef` — so leaving them unset
   gives a pod that cannot start rather than one reachable with a published
   password. Worth being explicit that this is the out-of-the-box state: with
   the release installed and the fields blank, pgAdmin will not come up until
   someone sets them. That is deliberate, but it is a change in default
   behaviour, so say the word if you would rather it were something else.

2. **Put something in front of it.** This is the part I would not want lost with
   the flag: pgAdmin is a full read/write console onto the tenant database,
   published on the deployment's own domain at `/pgadmin`, and its own login is
   the only barrier. The ingress annotations carrying an IP allow-list and an
   SSO redirect are commented out in `pgadmin/values.yaml`, and the SSO they
   point at (`/oauth2/auth`, served by the `oauth2-proxy` chart) is in no
   helmfile and never deployed — so uncommenting them as-is would not protect it
   either. A strong operator-chosen password is now the whole defence.

Not verifiable here: no cluster runs this tier, so "the pod cannot start" is
reasoned from the `secretKeyRef` plus the absent Secret, not observed. Everything
in the table above is `helm template` output.

vinothrallapalli-eGov (approved):

PR #1672, opened 2026-08-07
feat(gatus): send outage alerts to Slack, opt-in via webhook env var (#1609)

singhalkarun:
Closes #1609

## Problem

Gatus detects outages and tells nobody. A dashboard only helps someone already looking at it, which is never at 3am. `egov-url-shortening` sat DOWN for days for exactly this reason — a red tile nobody loads looks identical to a green one.

## Why this is safe to merge even though nobody has picked a channel yet

Alerting ships **OFF**, and off is a genuinely inert state rather than a half-configured one. I verified this against the Gatus source rather than assuming it, because the failure mode mattered:

- Gatus runs `os.ExpandEnv` over the whole config before parsing, so an unset `${GATUS_SLACK_WEBHOOK_URL}` becomes an **empty string** (not a literal, not an error)
- `slack.Config.Validate()` returns `ErrWebhookURLNotSet` on an empty `webhook-url`
- `ValidateAlertingConfig` **logs a warning and nils the provider** — it has no error return path to the caller

So a deployment that never sets the variable behaves exactly as it does today. Had Gatus instead failed closed on an invalid provider, this PR would have turned every existing deployment into a crash loop.

**There is no channel name to configure.** The channel is chosen by whoever mints the webhook, so this does not block on the open "which Slack channel / who triages" question from the issue — that decision is made at webhook-creation time, outside the repo.

## Why every endpoint has a two-line `alerts:` block

This is required, not stylistic. Gatus's `default-alert` is a *template merged into alerts an endpoint already declared* — the README is explicit:

> ⚠ You must still specify the `type` of the alert in the endpoint configuration even if you set the default alert of a provider.

Confirmed in `MergeProviderDefaultAlertIntoEndpointAlert`, which returns immediately when the endpoint has no alert object. A provider block on its own attaches nothing to anything. Hence 51 × `alerts: - type: slack` in each tier.

## Thresholds

| Setting | Value | Why |
|---|---|---|
| `failure-threshold` | 3 | ~90s of sustained failure at the 30s interval — a single dropped check or a rolling restart should not page anyone |
| `success-threshold` | 2 | symmetric, avoids flapping |
| `send-on-resolved` | true | an alert with no all-clear leaves people checking manually to see if it is still broken |

## Secret handling

The webhook URL is a credential — anyone holding it can post into the channel — so it is read from the environment in every tier and never committed:

- **compose**: `GATUS_SLACK_WEBHOOK_URL`, defaulted empty across all five compose files
- **ansible**: `gatus_slack_webhook_url`, empty in `group_vars/digit.yml`, set per tenant in `host_vars` (ideally sourced from OpenBao)
- **k3s**: an **optional** `secretKeyRef`, so clusters without the secret still start the pod:

```
kubectl -n digit create secret generic gatus-alerting \
  --from-literal=slack-webhook-url='https://hooks.slack.com/services/...'
```

## Verification

- `check-gatus-coverage.py` → `OK: 88 compose services, 23 k3s Services, 51 endpoints per tier, no drift` (both tiers still identical, as that guard requires)
- static contract tests: **35/35**
- both Gatus configs, all five compose files, and the k3s ConfigMap's *embedded* config all parse

Not verified: an actual Slack delivery, which needs a real webhook.

---

## Update: runtime-verified

Previously only source-verified. Now run against the real `egovio/gatus:latest` image with this exact config:

**Webhook unset (the default every existing deployment gets):**
```
[config.ValidateAlertingConfig] Ignoring provider=slack due to error=webhook-url not set
[config.ValidateAlertingConfig] configuredProviders=[]; ignoredProviders=[... slack ...]
[store.Initialize] Creating storage provider of type=memory
```
Container `status=running`, endpoints monitored normally. Fail-open confirmed — merging this cannot disturb a deployment that never configures Slack.

**Webhook set:**
```
[config.ValidateAlertingConfig] configuredProviders=[slack]
```
and after the 3-failure threshold elapsed (~90s at the 30s interval), with a deliberately fake webhook URL:
```
[watchdog.handleAlertsToTrigger] Sending initial slack alert because alert for endpoint=PostgreSQL ... has been TRIGGERED
[watchdog.handleAlertsToTrigger] Sending slack alert because alert for endpoint with key=infrastructure_pgbouncer ... has been TRIGGERED
```

That last part is the one that mattered: it proves the per-endpoint `alerts:` blocks are actually attached and firing, which is the whole reason all 51 endpoints carry one. The threshold behaved as configured rather than alerting on the first failure.

Still not verified: delivery to a real Slack channel, which needs a real webhook.


vinothrallapalli-eGov (commented):
Reviewed secret handling, opt-in default, and alert routing. Secret handling is correct throughout: the webhook URL is never committed (empty default in group_vars, `${VAR:-}` in every compose file, an *optional* k8s `secretKeyRef` so clusters without the secret still start), and it's never logged. I independently verified the fail-safe claim against upstream Gatus source (`slack.Config.Validate()` returns `ErrWebhookURLNotSet` on empty, and `ValidateAlertingConfig` only warns + nils the provider, never propagating an error) — this matches the PR's runtime-tested log lines, so "opt-in defaults to genuinely off" checks out rather than being an empty/placeholder webhook that silently enables. Thresholds (3-failure / 2-success, `send-on-resolved: true`, ~90s at a 30s interval) are reasonable and avoid single-blip paging. Naming is structurally consistent with #1673: that PR's `grafana_slack_webhook_url` falls back to this PR's `gatus_slack_webhook_url`, so one webhook drives both alert streams. One gap, non-blocking per the PR's own reasoning: the tracking issue's "which channel / who triages per tenant" question has no encoded owner anywhere in this change — no host_vars example sets `gatus_slack_webhook_url`, no doc names a channel or responder. That's a deliberate punt to whoever mints the webhook, which is defensible, but it's exactly the gap the issue warned could mute the alert stream within a week if nobody claims it operationally.

local-setup/ansible/inventory/group_vars/digit.yml
```diff
@@ -215,3 +215,12 @@ digit_ui_esbuild_branch: main
...
 # digit_ui_bundle_image: "ghcr.io/subhashini-egov/digit-ui-esbuild:nightly-develop"
+
+# Slack webhook for Gatus outage alerts (#1609). Empty = alerting OFF, which is
+# the default for every tenant: gatus warns about the unusable provider and
+# carries on, so nothing changes for deployments that do not want alerts.
+#
+# Set it per tenant in host_vars to switch alerting on. Treat it as a SECRET --
+# anyone holding the URL can post into the channel -- so prefer pulling it from
+# OpenBao via secrets_path rather than committing it here.
+gatus_slack_webhook_url: ""
```
vinothrallapalli-eGov, line 226: Confirmed: empty string here is genuinely inert (verified against upstream Gatus `slack.Config.Validate()` / `ValidateAlertingConfig`, matches the PR's own runtime test). One non-blocking gap: no host_vars example in this repo actually sets this var, so there's still no named owner for "who mints the webhook / which channel / who triages" per tenant — the exact question flagged in the tracking issue. Worth a follow-up (e.g. a line in `host_vars/README.md` or the example files) before rollout, since an unclaimed alert stream is how it gets muted within a week.

singhalkarun: Confirmed on both halves.

**On inertness** — your verification is right, and it is worth recording *why* it holds, because it is load-bearing rather than incidental. `slack.Config.Validate()` returns `ErrWebhookURLNotSet` on an empty URL, and `ValidateAlertingConfig` has **no error return at all**: on a failed provider validation it logs `Ignoring provider=%s due to error=%s` and calls `SetAlertingProviderToNil`. That is exactly the `Ignoring provider=slack` / `configuredProviders=[]` the runtime test saw.

The contrast with `ValidateEndpointsConfig`, which *does* return an error, is the whole story of these two PRs: same-shaped empty value, opposite blast radius. It is why an empty webhook here is genuinely safe and an empty `url` in #1622 is a P0. Nothing to change.

**On the ownership gap** — real, and closed here rather than deferred, since it is documentation and carries no merge risk: 95cd462e.

`host_vars/README.md` names `_example.yml` as the field reference, so that is where the guidance went — a new "Monitoring alerts (Gatus → Slack)" section that puts your three questions on the page explicitly:

1. **which channel** — dedicated per environment, not a busy team channel
2. **who triages** — a named person or rota, not "the team"
3. **who mints and rotates it** — recorded, so it survives that person leaving

plus the instruction to store the URL in OpenBao under the tenant's `secrets_path` rather than in the file, and a commented-out example line. Left commented on purpose: it keeps the example inert, and a real webhook URL must never land in this repo. I also added the var to `README.md`'s sensitive-fields table — the list that justifies `host_vars` being gitignored in the first place, which had no entry for it.

Your framing that an unclaimed alert stream gets muted within a week is quoted almost directly in the section, because it is the reason the questions matter and it is the part a future reader will skip if it is not spelled out.

Docs only; no behaviour change. `_example.yml` still parses and the key stays absent from the parsed result.

vinothrallapalli-eGov (approved):

vinothrallapalli-eGov (approved):

PR #1673, opened 2026-08-07
feat(grafana): provision capacity and error-trend alert rules (#1611)

singhalkarun:
Closes #1611

> **Stacked on #1672** (`feat/1609-gatus-slack-alerting`). Review that one first — the diff here will shrink once it merges.

> **Also depends on #1628 for data.** The Kafka consumer-lag rule reads `redpanda_*` series that only exist once #1628 adds the Redpanda scrape job to `otel/prometheus.yml` (this branch scrapes only `otel-collector` and `node`). It is a *data* dependency, not a merge one — the two are separate stacks off `monitoring-fix`. If this merges first, that rule evaluates an empty vector and reports a healthy "ok" until #1628 lands. It is left unpaused deliberately so it starts working the moment the scrape job arrives, and the dependency is recorded on the rule itself. The "DIGIT Kafka Consumer Lag" dashboard its description links to arrives with #1678, stacked on #1628.

## Problem

Gatus answers *"is it up?"*. Nothing watched the gradual failures that actually take these boxes down: a disk filling, memory running out, a JVM creeping toward its heap ceiling, or a service answering `/health` perfectly while failing every real request.

All of that data was **already being collected** — Prometheus has host and JVM metrics, Loki has every container's logs, and the JVM dashboard already charts heap against limits. There was simply no alerting engine of any kind: no Alertmanager, no `rule_files`, and no `provisioning/alerting/` directory beside the `dashboards/` and `datasources/` that already exist.

## Why Grafana's built-in alerting, not Alertmanager

- No extra always-on container on a memory-tight box
- Grafana already has **both** datasources provisioned. The error-flood rule needs **Loki**, which Alertmanager cannot see at all
- Rules ship as provisioned files: reviewable in git, deployed automatically, consistent with how dashboards and datasources are already handled here

## The rules

Metric names are taken from the queries in the dashboards provisioned beside them (`jvm-services.json`, `node-exporter-full.json`), not invented — including the `mountpoint="/"` matcher, which is what the node-exporter dashboard already uses.

| Rule | Source | Threshold | For | Severity |
|---|---|---|---|---|
| Host disk low | Prometheus | < 15% free on `/` | 10m | warning |
| Host memory low | Prometheus | < 10% available | 5m | **critical** |
| Host CPU saturated | Prometheus | > 90% | 15m | warning |
| JVM heap headroom low | Prometheus | > 90% of `jvm_memory_limit_bytes` | 10m | warning |
| Kafka consumer lag high | Prometheus | > 10k messages behind | 15m | warning |
| Log error spike | **Loki** | placeholder | 5m | **paused** |

Memory is critical rather than warning because these boxes have **no swap** — the kernel does not degrade gracefully, it goes straight to OOM-killing whichever JVM is largest. There is no slow middle ground.

**There is deliberately no service-down rule.** That is Gatus's job (#1609); duplicating it means two messages per incident, which teaches people to ignore both. A test enforces this rather than trusting the comment.

## Two decisions worth your attention

**1. `noDataState: OK` on every rule.**
node-exporter ships in the `docker-compose.monitoring.yml` *overlay*, and the JVM metrics need the OTEL pipeline running. So "no data" here usually means *"this deployment does not run that component"* — a valid configuration, not an incident. Alerting on it would put a permanent, unfixable alert in the channel, and a channel with a permanent alert gets muted, which silently disables every **other** rule in this file for that person. A monitoring component that has genuinely died is caught by #1613.

`execErrState` stays at `Alerting`: a query that *errors* is a broken rule, which is worth surfacing.

**2. Log error spike ships `isPaused: true`.**
This is the rule that catches the failure Gatus structurally cannot see — green dashboard, nothing works. It is also the only rule whose threshold cannot be derived from first principles: DIGIT services log exception stack traces during entirely normal operation, so the number needs about a week of observed baseline per tenant.

A wrong threshold is not neutral. Too low and the channel cries wolf until someone mutes it; too high and it never fires, which looks identical to "no errors". To enable: observe `sum by (service_name) (count_over_time(...))` on a real tenant, take the p95, set above it, unpause.

## The crash-loop this avoids

Worth calling out because it is the opposite of how #1609 behaves, and I nearly got it wrong:

- **Gatus** drops a misconfigured alert provider with a warning and keeps running — so an empty webhook there is genuinely inert
- **Grafana** treats a provisioning file it cannot validate as **fatal at startup**
- Grafana expands `$VAR` with `os.ExpandEnv` semantics: **no `${VAR:-default}` support**, and unset becomes an **empty string**

So an empty webhook in the contact point would risk crash-looping Grafana on every deployment that has not configured Slack — taking down the dashboard to add alerting to it.

The default is therefore supplied one level up by **compose** (which does support `:-`), as an unreachable URL under the RFC 2606 `.invalid` TLD. Provisioning always validates, Grafana always starts, and an unconfigured deployment logs a clean DNS failure if a rule ever fires — a visible symptom rather than a silent one.

## One channel, not two

`grafana_slack_webhook_url` falls back to `gatus_slack_webhook_url` in the Ansible tier, so an operator sets **one** value and both engines post to the same place. Override it only to split the two streams.

## New CI guard

`static/grafana-alerting.test.ts` (13 tests). Justified by the failure mode: a typo in these files does not degrade alerting, it takes the dashboard down, on the deployed box rather than in CI — and there is no `grafana --check-config` to catch it. It asserts required fields, unique uids, that each `condition` resolves to a refId in its own rule, that every `datasourceUid` matches an actually-provisioned datasource, the two decisions above, and that no webhook is ever committed as a literal.

`js-yaml` was only present *transitively* via jest, so it is now an explicit devDependency rather than something a jest upgrade could silently remove.

## Verification

- static tests **48/48** (13 new)
- the new guard was **mutation-tested** — breaking a `condition` refId, unpausing the error-spike rule, and hardcoding a webhook each fail the expected test and only that test, so it is not passing vacuously
- `check-gatus-coverage.py` still passes
- the Ansible fallback chain was rendered through jinja2 for all three cases (neither set → placeholder, only gatus set → gatus value, both set → grafana value)

**Not verified:** actual Slack delivery, and a live Grafana startup against these files. Both need a running box — Docker was not available in this environment. Given that bad provisioning is fatal to startup, someone should bring Grafana up once on a scratch box before this reaches a tenant.

---

## Update: runtime-verified, and the crash-loop is real

Docker became available, so this is no longer a prediction. Both cases run against the pinned `egovio/grafana:11.4.0` image with this exact provisioning directory.

**With the compose default (webhook not configured) — the case that matters most:**

Grafana reached `/api/health` 200 in 6s, `status=running, exitcode=0, restarts=0`, and the API confirms provisioning actually took effect rather than being silently skipped:

```
5 rules provisioned
  Host disk low           uid=digit-host-disk-low         paused=False noData=OK for=10m
  Host memory low         uid=digit-host-memory-low       paused=False noData=OK for=5m
  Host CPU saturated      uid=digit-host-cpu-saturated    paused=False noData=OK for=15m
  JVM heap headroom low   uid=digit-jvm-heap-headroom-low paused=False noData=OK for=10m
  Log error spike         uid=digit-log-error-spike       paused=True  noData=OK for=5m
```

Contact point `slack` created, URL redacted by Grafana as a secure setting. The error-spike rule is confirmed paused *by Grafana itself*, not just by the file.

**With the webhook genuinely unset — Grafana dies:**

```
level=error msg="Failed to provision alerting" error="failure to map file contactpoints.yaml:
  failure parsing contact points: slack: failed to validate integration \"slack\"
  (UID slack-webhook) of type \"slack\": recipient must be specified when using the Slack chat API"
```
`status=exited, exitcode=1`.

So the unreachable `.invalid` placeholder is **load-bearing, not defensive over-engineering**. Without it this PR would have taken Grafana down on every deployment that has not configured Slack — the exact opposite of the intent, and it would have been discovered on a tenant rather than here. Note the failure is not "empty URL" as I'd assumed but Grafana falling back to the Slack *chat API* path, which then demands a `recipient` — same outcome, different reason.

`docker compose config` confirms the default is actually delivered, with the variable unset in the environment:
```
GRAFANA_SLACK_WEBHOOK_URL: https://example.invalid/grafana-slack-webhook-not-configured
GATUS_SLACK_WEBHOOK_URL: ""
```
(The Gatus one is correctly empty — that engine fails open, and #1672 now has runtime evidence for it.)

**One pre-existing, unrelated warning** appears in both runs and is not caused by this PR:
```
logger=provisioning.plugins level=error msg="Failed to read plugin provisioning files from directory"
  path=/etc/grafana/provisioning/plugins error="...no such file or directory"
```
Grafana continues past it. There is simply no `plugins/` subdirectory in the provisioning tree.

Still not verified: delivery to a real Slack channel, and whether the rules fire against real data — both need a running stack with Prometheus and Loki populated.

### DIFF
--- a/local-setup/otel/grafana/provisioning/alerting/rules.yaml
+++ b/local-setup/otel/grafana/provisioning/alerting/rules.yaml
@@ -1,0 +1,268 @@
+# Grafana alert rules (#1611).
+#
+# Gatus answers "is it up?". These rules answer "is it about to stop being up?"
+# -- the gradual failures that actually take these boxes down: a disk filling,
+# memory running out, a JVM creeping toward its heap ceiling.
+#
+# THERE IS DELIBERATELY NO SERVICE-DOWN RULE HERE. That is Gatus's job (#1609).
+# Duplicating it would mean two Slack messages for one incident, which teaches
+# people to ignore both.
+#
+# ── noDataState: OK, everywhere ──
+# Every rule below reports OK rather than alerting when its series are missing.
+# That is deliberate. node-exporter lives in the docker-compose.monitoring.yml
+# OVERLAY, and the JVM metrics only exist once the OTEL pipeline is running, so
+# "no data" here usually means "this deployment does not run that component" --
+# a valid configuration, not an incident. Alerting on it would fire a permanent,
+# unfixable alert on every slim deployment, and a channel with a permanent red
+# alert in it is a muted channel. A monitoring component that has genuinely
+# died is caught by Gatus watching the monitoring stack (#1613), which is the
+# right place for it.
+#
+# execErrState is left at Alerting: a query that ERRORS is a broken rule, which
+# is worth surfacing, unlike an absent exporter.
+#
+# ── metric names ──
+# Taken from the queries in the dashboards provisioned beside this file
+# (jvm-services.json, node-exporter-full.json), not invented -- including the
+# mountpoint="/" matcher, which is what node-exporter-full.json uses and which
+# reflects node-exporter stripping its --path.rootfs prefix from the label.
+apiVersion: 1
+
+groups:
+  - orgId: 1
+    name: capacity
+    folder: DIGIT Alerts
+    interval: 1m
+    rules:
+      # ────────────────────────────────────────────────────────────────
+      - uid: digit-host-disk-low
+        title: Host disk low
+        condition: C
+        for: 10m
+        labels:
+          severity: warning
+        annotations:
+          summary: 'Root filesystem is below 15% free on {{ $labels.instance }}'
+          description: >-
+            Logs, database growth and image layers fill this disk. It is fine
+            until it abruptly is not: a full root filesystem takes down every
+            service on the box at once, and Postgres in particular fails in ways
+            that need manual recovery. Reclaim space now rather than at 3am.
+            Usual suspects: docker image/volume churn, and container logs if
+            log rotation is not capped.
+        data:
+          - refId: A
+            relativeTimeRange: {from: 600, to: 0}
+            datasourceUid: prometheus
+            model:
+              refId: A
+              instant: true
+              expr: >-
+                node_filesystem_avail_bytes{mountpoint="/", fstype!="rootfs"}
+                / node_filesystem_size_bytes{mountpoint="/", fstype!="rootfs"}
+          - refId: B
+            relativeTimeRange: {from: 600, to: 0}
+            datasourceUid: __expr__
+            model: {refId: B, type: reduce, expression: A, reducer: last}
+          - refId: C
+            relativeTimeRange: {from: 600, to: 0}
+            datasourceUid: __expr__
+            model:
+              refId: C
+              type: threshold
+              expression: B
+              conditions:
+                - evaluator: {type: lt, params: [0.15]}
+        noDataState: OK
+        execErrState: Alerting
+        isPaused: false
+
+      # ────────────────────────────────────────────────────────────────
+      - uid: digit-host-memory-low
+        title: Host memory low
+        condition: C
+        for: 5m
+        labels:
+          severity: critical
+        annotations:
+          summary: 'Less than 10% memory available on {{ $labels.instance }}'
+          description: >-
+            Sharper here than on a typical box: these tenant hosts have NO SWAP,
+            so the kernel does not degrade gracefully under memory pressure --
+            it goes straight to OOM-killing processes, and what it kills is
+            whichever JVM happens to be largest, not whichever is least
+            important. Severity is critical for that reason: there is no slow
+            middle ground between "tight" and "a service died".
+        data:
+          - refId: A
+            relativeTimeRange: {from: 600, to: 0}
+            datasourceUid: prometheus
+            model:
+              refId: A
+              instant: true
+              expr: node_memory_MemAvailable_bytes / node_memory_MemTotal_bytes
+          - refId: B
+            relativeTimeRange: {from: 600, to: 0}
+            datasourceUid: __expr__
+            model: {refId: B, type: reduce, expression: A, reducer: last}
+          - refId: C
+            relativeTimeRange: {from: 600, to: 0}
+            datasourceUid: __expr__
+            model:
+              refId: C
+              type: threshold
+              expression: B
+              conditions:
+                - evaluator: {type: lt, params: [0.10]}
+        noDataState: OK
+        execErrState: Alerting
+        isPaused: false
+
+      # ────────────────────────────────────────────────────────────────
+      - uid: digit-host-cpu-saturated
+        title: Host CPU saturated
+        condition: C
+        for: 15m
+        labels:
+          severity: warning
+        annotations:
+          summary: 'CPU above 90% for 15 minutes on {{ $labels.instance }}'
+          description: >-
+            Sustained saturation, not a spike -- 15 minutes is long enough to
+            rule out a deploy, a batch index or a JVM warming up. Past this
+            point request latency is already degraded even though every health
+            check still answers, so nothing else will tell you.
+        data:
+          - refId: A
+            relativeTimeRange: {from: 900, to: 0}
+            datasourceUid: prometheus
+            model:
+              refId: A
+              instant: true
+              expr: '1 - avg by (instance) (rate(node_cpu_seconds_total{mode="idle"}[5m]))'
+          - refId: B
+            relativeTimeRange: {from: 900, to: 0}
+            datasourceUid: __expr__
+            model: {refId: B, type: reduce, expression: A, reducer: last}
+          - refId: C
+            relativeTimeRange: {from: 900, to: 0}
+            datasourceUid: __expr__
+            model:
+              refId: C
+              type: threshold
+              expression: B
+              conditions:
+                - evaluator: {type: gt, params: [0.90]}
+        noDataState: OK
+        execErrState: Alerting
+        isPaused: false
+
+      # ────────────────────────────────────────────────────────────────
+      - uid: digit-jvm-heap-headroom-low
+        title: JVM heap headroom low
+        condition: C
+        for: 10m
+        labels:
+          severity: warning
+        annotations:
+          summary: '{{ $labels.service_name }} is using over 90% of its heap limit'
+          description: >-
+            A service climbing toward its heap ceiling will eventually die and
+            restart, dropping in-flight requests. Ten minutes above 90% means
+            steady pressure rather than a GC sawtooth.
+            Caveat when reading this: the ratio is against jvm_memory_limit_bytes,
+            so a service that never publishes a heap limit produces NO series and
+            therefore cannot trigger this rule. Absence of this alert is not
+            proof of headroom -- check the DIGIT JVM Services dashboard.
+        data:
+          - refId: A
+            relativeTimeRange: {from: 600, to: 0}
+            datasourceUid: prometheus
+            model:
+              refId: A
+              instant: true
+              expr: >-
+                sum by (service_name) (jvm_memory_used_bytes{jvm_memory_type="heap"})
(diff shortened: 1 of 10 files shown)


vinothrallapalli-eGov (commented):
Reviewed the 6 provisioned rules against what the stack actually exposes: node_filesystem_avail_bytes, node_memory_MemAvailable_bytes, node_cpu_seconds_total and the jvm_memory_*_bytes pair all match the metrics the already-provisioned node-exporter-full.json/jvm-services.json dashboards use, and the `prometheus`/`loki`/`__expr__` datasourceUids all resolve against what's actually provisioned in `datasources/`. Thresholds (disk <15%, mem <10% with critical severity justified by no-swap OOM behavior, CPU >90%/15m, heap >90%/10m) are reasonable and well-reasoned, not tight enough to flap. None of the 6 rules is a disguised service-down check -- confirmed by reading each condition, not just the CI guard (which only regexes for literal `up == 0` and would miss a rephrased duplicate, worth broadening later but not a blocker since none exists today). One real gap: the `digit-kafka-consumer-lag` rule (added by a second commit for #1623, not mentioned in this PR's description/table) queries `redpanda_kafka_max_offset`/`redpanda_kafka_consumer_group_committed_offset`, but this PR does not touch `local-setup/otel/prometheus.yml`, which still only scrapes `otel-collector` and `node-exporter` -- the Redpanda scrape job lives in a separate, still-open, unmerged PR (#1628), and the dashboard the rule's comment points readers to is in another open PR (#1678). Neither is declared as a stacking dependency the way #1672 is. Because the rule ships `noDataState: OK` and unpaused, until #1628 merges it will sit at a silent, indistinguishable-from-healthy "ok" with an empty result set -- which also means the devbox verification showing this rule `health: ok` doesn't actually demonstrate it's receiving real data, since an unscraped metric returns an empty vector rather than an error either way.

local-setup/otel/grafana/provisioning/alerting/rules.yaml
```diff
@@ -0,0 +1,331 @@
...
+      # rule ships paused and this one does not.
+      #
+      # This is the failure Gatus structurally cannot see: egov-persister falls
+      # behind, the API keeps returning 200, every check stays green, and
+      # complaints are never written to the database.
+      #
+      # 10k for 15m, not a tighter bound: bulk seeding and migrations move real
+      # backlogs through legitimately, and they drain. Sustained 15m above 10k
+      # is not a burst.
+      - uid: digit-kafka-consumer-lag
```
vinothrallapalli-eGov, line 233: This rule depends on Prometheus scraping Redpanda (`redpanda_kafka_max_offset` / `redpanda_kafka_consumer_group_committed_offset`), but this PR doesn't touch `local-setup/otel/prometheus.yml`, and that branch's current scrape_configs only cover `otel-collector` and `node-exporter`. The Redpanda scrape job is added by #1628, which is still open/unmerged, and the "DIGIT Kafka Consumer Lag dashboard" referenced below (line 231) is added by #1678, also open/unmerged. Neither is declared as a dependency here the way #1672 is called out at the top of the PR description.

Since this rule ships `noDataState: OK` and `isPaused: false` (unlike the error-spike rule), until #1628 merges it will evaluate against an empty vector and report a silent, healthy-looking "ok" -- functionally identical to "lag is fine" when really it's "nothing is being measured." Worth either stacking this commit on #1628, shipping it paused until the scrape job lands, or at minimum flagging the dependency explicitly in the PR description so reviewers/mergers sequence it correctly.

local-setup/tests/static/grafana-alerting.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  /**
+   * Gatus owns up/down (#1609). A second engine alerting on the same outage
+   * means two messages per incident, which teaches people to ignore both.
+   */
+  it('does not duplicate Gatus by alerting on service availability', () => {
+    for (const r of allRules) {
+      for (const d of r.data) {
+        if (d.datasourceUid === '__expr__') continue;
+        expect(d.model.expr ?? '').not.toMatch(/\bup\s*==\s*0/);
```
vinothrallapalli-eGov: Minor: this guard only catches the literal pattern `up == 0`. A rule phrased as `count(up) == 0`, `absent(up)`, or a Blackbox-style `probe_success == 0` would slip past it while still being a disguised service-down duplicate. Not a blocker today (manually checked all 6 current rules -- none is a duplicate), but worth broadening the regex or matching on the metric name `up` more generally if this file grows.

local-setup/otel/grafana/provisioning/alerting/rules.yaml
singhalkarun: Confirmed on all counts. This branch's `otel/prometheus.yml` has exactly two scrape jobs:

```
job_name: otel-collector
job_name: node
```

No Redpanda. So `redpanda_kafka_max_offset` / `..._consumer_group_committed_offset` are never collected, the rule evaluates an empty vector, and the file-wide `noDataState: OK` turns that into a healthy-looking "ok" that is indistinguishable from "lag is fine" — which is worse than no rule, because it looks like coverage. You are also right that the PR description does not declare it; the rule table there predates the consumer-lag commit and never got updated.

Of your three options I have taken the third plus a bit more, in d6a42ee9. Stacking on #1628 would mean a base change on an open PR, and this branch is already stacked on #1672 — restacking it onto a second open branch trades one sequencing hazard for a worse one. So the dependency is now recorded **on the rule itself** rather than only in the PR description, since the rule is where the next reader will be:

> DEPENDS ON #1628 FOR ITS DATA. […] Until #1628 is in, this rule evaluates an empty vector and — by the file-wide `noDataState: OK` above — reports a healthy "ok" that is indistinguishable from "lag is fine" when it really means "nothing is being measured".

Left **unpaused** deliberately, and this is the one place I have gone against your suggestion, so it is worth stating the reasoning: both stacks are bound for `monitoring-fix`, so the exposure is only the window where this merges first. Shipping it paused would close that window at the cost of requiring someone to remember to unpause it afterwards — and an alert that is paused-and-forgotten fails exactly the same silent way as one with no data, except permanently. Unpaused, it starts working the moment the scrape job lands. Happy to flip it if you would rather have the window closed; it is a one-line change.

The `#1678` dashboard reference in the annotation is noted in the same comment block.

local-setup/tests/static/grafana-alerting.test.ts
singhalkarun: Good catch, and broadened in d6a42ee9. You were right that it was passing on a technicality — of the five phrasings you named plus the obvious one, `/\bup\s*==\s*0/` caught exactly one.

Rather than widen the regex to enumerate comparisons (which just moves the goalposts — `!= 1`, `< 1`, `<= 0`, `absent_over_time`, …), the guard is now on the **metric**: no capacity or error-trend rule has any business reading an availability signal, so referencing one at all is the failure. That is both stricter and simpler than trying to cover every spelling.

Two refinements to keep it from being annoying: label *values* are stripped before matching, so a legitimate `{job="up-sync"}` is not mistaken for the `up` metric; and word boundaries handle `group_right()` and `sum by (redpanda_group)` for free, since the "up" inside them is preceded by a word character.

Mutation-tested each one individually against the real suite:

| expression | old guard | new guard |
|---|---|---|
| `up == 0` | fail | **fail** |
| `count(up) == 0` | *pass* | **fail** |
| `absent(up{job="x"})` | *pass* | **fail** |
| `min by (job) (up) < 1` | *pass* | **fail** |
| `probe_success == 0` | *pass* | **fail** |
| `sum(up) != 1` | *pass* | **fail** |
| `node_memory_MemAvailable_bytes{job="up-sync"}` | pass | pass |
| `sum by (redpanda_group) (group_right_thing)` | pass | pass |
| `rate(node_cpu_seconds_total{mode="idle"}[5m])` | pass | pass |

Static suite still 48/48. And confirming your manual check independently: none of the 6 current rules references an availability metric, so this went in green rather than by adjusting anything to fit.

vinothrallapalli-eGov (approved):

PR #1674, opened 2026-08-07
fix(alertmanager): read the SMTP password from a mounted secret (#1640)

singhalkarun:
Closes #1640

## What was committed

`charts/monitoring/values/prometheus.yaml` carried a working Gmail app password as a plaintext literal. This repository is **public**, so it was a live credential for anyone who cloned it: authenticated SMTP as `unified.alerts@egovernments.org` — the ability to send mail that genuinely originates from the organisation's domain, which is a phishing primitive, not just a spam one.

Per the issue discussion this password has **already been rotated**, so this PR is about removing the pattern that produced it rather than closing an open exposure.

## Why it went unnoticed — the more useful half

`env-secrets.yaml`, the SOPS-covered file, contained an `alertmanager:` block holding a **full copy** of the Alertmanager routing config, complete with an `<smtp_password>` placeholder. It looked exactly like the credential had a safe, encrypted home.

It did not. That block sat at `secrets.alertmanager.config`, while kube-prometheus-stack reads `alertmanager.config` at the **top level**:

```
env-secrets.yaml  →  secrets.alertmanager.config.receivers[].email_configs[].auth_password
prometheus.yaml   →          alertmanager.config.receivers[].email_configs[].auth_password
```

The two key sets have **zero overlap**, and `grep` confirms zero consumers of `.Values.secrets.alertmanager` anywhere in the tree. The reassuring copy was inert; the plaintext one was what actually shipped.

So the dead block is **removed rather than corrected**. Fixing up the placeholder in place would recreate exactly the false reassurance that let the real credential sit unnoticed. In its place is a comment saying where the secret actually lives now.

This is the same failure mode #1651's CI guard exists for — values that look authoritative and are silently ignored — except that guard only scans `env.yaml`, so it would not have caught this one.

## The fix

The password moves to a Kubernetes Secret mounted into the Alertmanager pods and read via `auth_password_file`. Alertmanager reads the file at send time, so the value never enters git, the Helm release, or `helm get values` output.

Two changes that **must stay together**, and each is useless without the other:

1. the `email-notification` receiver reads `/etc/alertmanager/secrets/alertmanager-smtp/password`
2. `alertmanagerSpec.secrets` mounts `alertmanager-smtp`, which is what makes that path exist

## Operator action

Create the secret once per cluster, in the release's namespace:

```
kubectl -n monitoring create secret generic alertmanager-smtp \
  --from-literal=password='<smtp-app-password>'
```

No cluster currently runs this tier, and the monitoring releases default to `installed: false` even after #1675 wires the helmfile in — so nothing here is live. Create the secret **before** setting `monitoring.metrics: true`, or Alertmanager email will silently fail to send.

## Verification

- the literal is gone from the working tree (`grep` across everything but `.git`)
- both files still parse; the receiver has no `auth_password` key and does have `auth_password_file`
- `alertmanagerSpec.secrets` is `['alertmanager-smtp']`
- the #1651 values-path guard still reports `OK`, with its existing 6 known mis-paths and no new ones

**Not verified:** an actual email send, which needs a live cluster and a real SMTP credential.

Note the credential remains in git **history** — rotation, which has already happened, is the only real remedy for that.


devops/deploy-as-code/charts/environments/env-secrets.yaml
```diff
@@ -112,81 +112,18 @@ secrets:
...
+    # top level, not under `secrets.`, so that block was inert: it looked like
+    # the secret had a safe home in the SOPS-covered file while the credential
+    # that actually shipped was a plaintext literal in
+    # charts/monitoring/values/prometheus.yaml. That appearance is why the real
+    # one went unnoticed, so the dead copy is removed rather than corrected.
+    #
+    # The SMTP password now comes from a Kubernetes Secret mounted into the
+    # Alertmanager pods and read via auth_password_file. See the comment beside
+    # auth_password_file in charts/monitoring/values/prometheus.yaml for the
+    # exact `kubectl create secret` command.
```
vinothrallapalli-eGov: Nit: this file now ends without a trailing newline (`\ No newline at end of file` in the diff). Harmless for YAML parsing but worth a quick fix for consistency with the rest of the tree.

vinothrallapalli-eGov (commented):
Verified against the current diff. The plaintext Gmail app password (`mujp cgjj fwxa whyd`) is gone from both changed files and does not reappear anywhere else in either file (checked full file contents at this commit, not just the diff context). The dead `secrets.alertmanager.config` block in env-secrets.yaml is fully deleted (not commented out) and replaced with an explanatory NOTE comment -- no partial or duplicate remnant remains, and there's only one `alertmanager:` receiver config left in the tree (prometheus.yaml). `auth_password_file: /etc/alertmanager/secrets/alertmanager-smtp/password` and `alertmanagerSpec.secrets: [alertmanager-smtp]` are added together as required for the mount to actually work, and the manual `kubectl create secret` provisioning step matches the existing convention used elsewhere in this chart tree (kafka-kraft, minio, postgresql READMEs all document the same out-of-band secret creation rather than something wired into Ansible/SOPS), so this isn't a dangling reference relative to repo norms -- it does mean the secret must be created before `monitoring.metrics: true` is flipped, which the PR description calls out. One trivial nit flagged inline: the new env-secrets.yaml comment block leaves the file without a trailing newline. No blocking issues found.

devops/deploy-as-code/charts/environments/env-secrets.yaml
singhalkarun: Confirmed and fixed in 59e17c62 — the file's last byte was `0x2e` (`.`), no `0x0a`. Removing the dead `secrets.alertmanager` block took the trailing newline with it.

Newline only; the diff is one line changed with no content difference.

Since a naive conflict resolution on this file could quietly bring the deleted block back — the thing this PR exists to remove — I verified the deletion survived by parsing the result rather than reading the diff:

```
$ python3 -c "import yaml; d=yaml.safe_load(open('env-secrets.yaml')); \
    print('alertmanager' in d['secrets'], len(d['secrets']))"
False 20
```

`secrets.alertmanager` absent, 20 sibling keys intact, and `grep` finds no `slack_api_url` or `auth_password:` anywhere in the file. Also re-checked that nothing reads the removed path (no `secrets.alertmanager` reference in any template), that the plaintext Gmail app-password is gone from the tree, and that `auth_password_file` agrees with the mount: `secrets: [alertmanager-smtp]` mounts at `/etc/alertmanager/secrets/alertmanager-smtp/password`, matching the `--from-literal=password=` in the documented `kubectl` command.

Not verifiable here: no cluster runs the `deploy-as-code` tier, so the mount path and Alertmanager's reading of `auth_password_file` are checked against kube-prometheus-stack's documented behaviour, not observed.

singhalkarun:
## Cluster-verified — works, and it introduces a hard deploy prerequisite

Deployed on a real cluster (kind, this branch's monitoring helmfile).

**The fix itself is correct.** Generated Alertmanager config:

```
smarthost: smtp.gmail.com:587
auth_username: unified.alerts@egovernments.org
auth_password_file: /etc/alertmanager/secrets/alertmanager-smtp/password
```

The file is present in the container (`password -> ..data/password`), and no plaintext `auth_password:` remains anywhere in `values/prometheus.yaml`. That closes the published-credential problem.

**But if the operator does not create the secret first, Alertmanager never starts.** I deliberately deployed without it:

```
MountVolume.SetUp failed for volume "secret-alertmanager-smtp":
  secret "alertmanager-smtp" not found

alertmanager-kube-prometheus-stack-alertmanager-0   0/2   Init:0/1   99s
```

It stays there indefinitely — no crash-loop, no obvious error in the pod's own logs, just a pod that never becomes ready. Creating the secret recovered it to `2/2 Running` within 45 seconds, with no other intervention.

The instruction currently lives only in a comment inside `values/prometheus.yaml` (~line 466). Given the failure mode is a silent init hang rather than a loud error, it deserves the same prominence #1682 gave its Grafana secret — a **"Before enabling"** block in the PR body and the runbook:

```
kubectl -n monitoring create secret generic alertmanager-smtp \
  --from-literal=password='<smtp-app-password>'
```

Worth deciding explicitly, too, whether this should apply when alerting is disabled — as written, the mount is unconditional, so the secret is required even for a deployment that never sends email.

vinothrallapalli-eGov (approved):

PR #1675, opened 2026-08-07
feat(monitoring): wire the monitoring helmfile in, gated per component (#1619)

singhalkarun:
Part of #1619.

## The problem

The entire production monitoring stack was commented out in `digit-helmfile.yaml`:

```
#  - path: ./charts/monitoring/monitoring-helmfile.yaml
```

I traced its provenance: it arrived **already commented** in the wholesale import from upstream DIGIT-DevOps (`2f5649c9`), sitting beside the equally-commented sanitation helmfile. Nobody on this repo disabled it deliberately — it was inherited.

The consequence is that **every monitoring fix in this tree edits files that nothing deploys**. #1659, #1662, #1663, #1664, #1665, #1668 and #1674 are all currently inert. "k8s parity with the Ansible tier" is parity in files only until this line exists.

## Why not just uncomment it

Because the next `helmfile apply` on any cluster would then deploy a full Prometheus / Grafana / Loki stack that nobody asked for — a large, silent change to every deployment.

So each of the six releases is gated on a boolean in `charts/environments/env.yaml`, and **every one defaults to `false`**. Enabling the path deploys nothing. An operator turns on what they want, one line each:

```yaml
monitoring:
  metrics: false     # kube-prometheus-stack — Prometheus, Alertmanager, node-exporter, kube-state-metrics
  logs: false        # loki-stack — Loki + Promtail
  dashboards: false  # Grafana
  probes: false      # blackbox-exporter — external HTTP/TLS probing (#1649)
  traces: false      # Jaeger — needs an Elasticsearch backend; much the most expensive
  kafkaUi: false     # Kafka admin console — not monitoring; an access-control decision
```

That also delivers the modular requirement from #1657 on this tier: a site with limited capacity can run metrics alone and still get alerting; a full stack is six booleans.

## Three details that aren't stylistic

**1. The toggles are read defensively.**
```
installed: {{ default false (index $mon "metrics") }}
```
behind a `hasKey` guard — not `.Values.monitoring.metrics`. helmfile runs templates with `missingkey=error`, so a direct reference is a **hard parse failure** on any operator `env.yaml` written before this block existed. And because a helmfile is one template document, that failure takes out **every other release in the file**, not just monitoring.

**2. Not `dig`.** sprig's `dig` would express this in one call, but it does not exist in helmfile **v0.140.0** — the version `.github/workflows/digit_install.yml` actually pins. I checked rather than assumed:
```
error: function "dig" not defined
```

**3. Defaults live in the sub-helmfile's own `environments:` block**, listed *before* `env.yaml` so an operator's file overrides them by deep merge, and so the keys always exist.

## Verification

Run against the **pinned v0.140.0 binary**, downloaded from the same URL CI uses — not a newer helmfile, whose v1 semantics differ enough to be misleading here.

| Case | Result |
|---|---|
| Top-level build, shipped defaults | 47 → **53** releases; delta is exactly the 6 monitoring releases, **all `installed=false`**, and **no pre-existing release changes its installed state** |
| Partial override (`metrics` + `dashboards` only) | exactly those two on, other four off — deep merge confirmed |
| `env.yaml` with the `monitoring:` block deleted entirely | builds fine, all six off, no error |
| values-path guard / gatus-coverage / static tests | all pass (35/35) |

That first row is the important one: it is direct evidence this PR is a no-op until someone opts in.

## Pre-existing, not a regression

The monitoring helmfile cannot be built without `-e env` — it fails on `.Values.namespace`. I confirmed the **original** file fails identically, so this PR does not introduce it.

## Interaction with other PRs

#1664 enables blackbox and #1668 adds the OTEL collector; both touch this helmfile. Whoever merges second will need a trivial conflict resolution on the `installed:` line. Worth merging this one **first**, since it is what makes the others testable.


vinothrallapalli-eGov (commented):
Verified independently against the raw files at this SHA (not just the diff). Gating is genuinely per-component: six distinct `index $mon "<key>"` lookups (metrics/logs/dashboards/traces/probes/kafkaUi), each feeding a different release's `installed:` field — not one shared boolean. Defaults are false in two independent places (env.yaml's new `monitoring:` block and the sub-helmfile's own pre-env.yaml defaults block), consistent with the PR's build-delta claim. `digit-helmfile.yaml` only appends the monitoring path at the end of the `helmfiles:` list — the five existing entries (backbone-services, core-services, urban, common-services, analytics, auxiliary-services) are untouched and not reordered, so no unrelated release is at risk of a redeploy/restart from this change. Confirmed the `helmfile v0.140.0` pin in `.github/workflows/digit_install.yml` (line 313) matches the version the PR says it tested `dig` availability against. `env.yaml`'s insertion sits cleanly between `global:` and `root-ingress:` with no disruption to surrounding keys. Two low-severity, non-blocking observations left inline for awareness, not requesting changes on either.

devops/deploy-as-code/charts/monitoring/monitoring-helmfile.yaml
```diff
@@ -30,7 +64,9 @@ releases:
   - {{ .Values | toYaml | nindent 8 }}
```
vinothrallapalli-eGov, line 85: Minor/FYI, not blocking: `{{ .Values | toYaml | nindent 8 }}` (pre-existing pattern, reused here for loki-stack/kube-prometheus-stack/blackbox/kafka-ui) now also serializes the new `monitoring:` toggle map into every one of those charts' Helm values, since it dumps the entire merged `.Values` tree rather than a scoped subset. Charts generally ignore unrecognized top-level keys so this is likely harmless, but worth knowing that the toggle state (and any future keys added under it) becomes visible to unrelated charts' rendered values.

devops/deploy-as-code/charts/monitoring/monitoring-helmfile.yaml
```diff
@@ -1,10 +1,43 @@
...
+       no entry for key") in two ordinary situations: running without `-e env`,
+       and an operator env.yaml written before this block existed. Both used to
+       render fine, and a helmfile that will not parse breaks every other
+       release in this file, not just monitoring.
+
+       So: resolve the block once, tolerate its absence, and default each key to
+       false individually. sprig's `dig` would express this in one call but does
+       not exist in helmfile v0.140.0, the version CI pins. */}}
+{{- $mon := dict }}
+{{- if hasKey .Values "monitoring" }}{{ $mon = .Values.monitoring }}{{ end }}
```
vinothrallapalli-eGov: Minor edge case, not blocking: `index $mon "logs"` assumes `$mon` is always a map once `hasKey .Values "monitoring"` is true. If an operator's env.yaml ever set `monitoring` to a non-map value (e.g. a bare `monitoring: true` typo instead of the documented nested block), `index` on a non-map would be a hard render error again, defeating the defensive intent one level up. Unlikely given how clearly both defaults blocks document the expected shape, but a `kindIs "map"` check alongside `hasKey` would close that gap entirely if it's worth the extra defensiveness.

devops/deploy-as-code/charts/monitoring/monitoring-helmfile.yaml
singhalkarun: Confirmed, and now written down in the file header (535a9821) so the next reader does not have to rediscover it.

`helmfile v0.140.0 build` shows `monitoring:` inside the rendered values of every release that uses the `.Values | toYaml` dump — loki-stack, kube-prometheus-stack, jaeger, blackbox and kafka-ui — five copies, one per release, plus the environment block itself. Harmless as you say: Helm ignores unrecognised top-level keys, and none of those charts has a `monitoring` key of its own to collide with (checked).

Not changing the dump itself. Scoping it per release is the right end state but it is a pre-existing pattern across the whole file, it would need each chart's actual key needs worked out, and doing it inside a PR about toggles would bury a behaviour change in an unrelated diff. The header note now says to assume anything added under `monitoring:` becomes visible to those charts, which is the part that could actually bite later.

devops/deploy-as-code/charts/monitoring/monitoring-helmfile.yaml
singhalkarun: Real, and reproducible — thanks. Fixed in 535a9821, though not quite the way you suggested; the reason is worth stating.

Reproduced by replacing the block in `env.yaml` with a bare `monitoring: true`:

```
$ helmfile -f mon-only.yaml -e env build
error during monitoring-helmfile.yaml.part.1 parsing:
  executing "stringTemplate" at <index $mon "logs">: error calling index: can't index item of type bool
```

Exactly the whole-file parse failure the `hasKey` guard exists to prevent, so the gap is genuine.

Where I went a different way: adding `kindIs "map"` to the `hasKey` condition and falling through to the empty dict would make it render, but it would render *wrong*. The operator wrote `monitoring: true` meaning "turn it on" and would get all six toggles false — and on an existing install `installed: false` is not "nothing happens", it is helmfile **uninstalling** the stack. A silent teardown is a worse outcome than the crash.

So the shape check is there, but it fails loudly and names the expected keys:

```
{{- if hasKey .Values "monitoring" }}
{{-   if not (kindIs "map" (index .Values "monitoring")) }}
{{-     fail "`monitoring:` in charts/environments/env.yaml must be a map of component toggles (metrics/logs/dashboards/traces/probes/kafkaUi), not a single value" }}
{{-   end }}
{{-   $mon = .Values.monitoring }}
{{- end }}
```

That keeps the distinction the original guard was drawing: *absent* is a legitimate older config and is tolerated; *present but the wrong shape* is a typo in a config the operator just wrote, and deserves to be told.

Verified with helmfile v0.140.0 `build`: the normal map still renders all six `installed:` lines; `monitoring: true`, `monitoring: ""` and `monitoring: 42` each fail with the message above. Absence is unchanged (all six false), and the pre-existing failure when run without `-e env` is also unchanged — it still stops later at `.Values.namespace`, same as on `monitoring-fix`, so this block is not what makes that case fail.

(Confirmed `dig` is genuinely unavailable on the pinned version while testing, in case it comes up again: `template: stringTemplate:10: function "dig" not defined`.)

vinothrallapalli-eGov (approved):

Issue #1676, opened 2026-08-07
Novu signing secrets render as plaintext env on the Deployments


singhalkarun:
Part of #1601 (monitoring/security umbrella). Follow-up to #1644. **P2 · effort: S**

## Problem

#1644 removed Novu's published example signing secrets, but left the delivery mechanism unchanged: the Novu chart renders `JWT_SECRET`, `STORE_ENCRYPTION_KEY` and `NOVU_SECRET_KEY` as **plaintext `env:` entries on the Deployments**.

```yaml
- name: JWT_SECRET
  value: "..."          # readable in the Deployment spec
```

That means anyone with namespace **read** access can recover them:

```
kubectl -n novu get deploy novu-api -o yaml | grep -A1 JWT_SECRET
```

Deployment-read is a far more widely granted RBAC permission than secret-read — it is what you hand someone to let them look at workloads. So the key that signs every session, and the key that encrypts the stored SMS/email provider credentials, are effectively readable by anyone who can inspect the cluster's workloads.

`JWT_SECRET` is the sharp one: knowing it means a session can be **forged outright**, with no password to guess and nothing to intercept.

## Affected components

| Component | Secrets in its env block |
|---|---|
| `novu-api` | `JWT_SECRET`, `STORE_ENCRYPTION_KEY`, `NOVU_SECRET_KEY` |
| `novu-worker` | `STORE_ENCRYPTION_KEY` |
| `novu-ws` | `JWT_SECRET` |

## Proposed fix

Move them into a Kubernetes Secret referenced via `secretKeyRef`, with two supply routes:

1. **`novu.secrets.existingSecret`** — name a Secret created out of band. The values then never enter `values.yaml`, the Helm release, or `helm get values` at all. Correct choice for anything real.
2. **Chart-managed** (default) — the chart creates the Secret from the values. Better than plaintext env, since the value moves behind secret-read RBAC. Worth being explicit that Helm still stores the rendered manifest in its own release Secret, so this route does not hide the value from someone who can read Secrets in the namespace.

## Acceptance

- [ ] No `JWT_SECRET` / `STORE_ENCRYPTION_KEY` / `NOVU_SECRET_KEY` appears as a literal `value:` in any rendered Deployment
- [ ] `existingSecret` suppresses chart Secret creation and retargets every `secretKeyRef`
- [ ] A blank secret with no `existingSecret` fails rendering loudly rather than deploying an empty key
[...]

Issue #1644, opened 2026-08-06
Novu dashboard deploys with placeholder JWT/encryption secrets and no TLS


singhalkarun:
**P0-security · effort: S** · found while reviewing monitoring in `devops/deploy-as-code/` (#1619); not a monitoring issue.

## Problem

The Novu dashboard — which manages notification templates, subscribers, and the credentials for the SMS/email providers that message citizens — deploys by default with **placeholder signing secrets** and **no TLS**.

`charts/backbone-services/backboneservices-helmfile.yaml:131-138` — `- name: novu` / `installed: true`, in an active helmfile.

### 1. Signing and encryption secrets are upstream placeholders

`charts/backbone-services/novu/values.yaml:25-29`:
```yaml
novu:
  secrets:
    jwtSecret: "your-secret"
    storeEncryptionKey: "change-this-key-32-characters!!!"
    novuSecretKey: "change-this-secret-key"
```

Neither `env.yaml` nor `env-secrets.yaml` overrides them (verified — the only `novu` keys in `env.yaml` are `novu-bridge` config at `:249-255` and two service URLs at `:55-56`).

**`jwtSecret` is the most serious of the three.** It signs and verifies session tokens. A known signing secret means a session token can be *manufactured* — no password required and nothing to intercept. `storeEncryptionKey` is what encrypts stored provider credentials, i.e. the SMS/email API keys; a known key makes them readable.

### 2. The ingress has no TLS at all

`charts/backbone-services/novu/templates/ingress.yaml` defines `rules:` and a `host:` and **no `tls:` block** — unlike every other ingress in the tree, which either carries `<domain>-tls-certs` or its own secret. The dashboard is therefore served over plain HTTP: login and session cookie in cleartext.

### 3. The hostname is another environment's, hardcoded

`charts/backbone-services/novu/values.yaml:15` — `host: "pgroneclick.digit.org"`, not derived from `global.domain` like every other ingress, and not overridden anywhere.

This cuts both ways and is worth stating precisely: on a normal deployment the dashboard is **not** reachable at the deployment's own domain, so it is partly broken for legitimate use — but that is not a security control. An ingress matches on the `Host` header, which a client sets freely.

## What I cannot see

[...]

PR #1677, opened 2026-08-07
fix(novu): move signing secrets into a Secret via secretKeyRef (#1676)

singhalkarun:
Closes #1676

> **Stacked on #1671** (`fix/1644-novu-tls-secrets`), which removed the published example values. This PR fixes how they are *delivered*.

## Problem

#1644 stopped the chart shipping the upstream project's example secrets, but left the delivery mechanism unchanged: `JWT_SECRET`, `STORE_ENCRYPTION_KEY` and `NOVU_SECRET_KEY` rendered as plaintext `env:` entries on the Deployments.

```
kubectl -n novu get deploy novu-api -o yaml | grep -A1 JWT_SECRET
```

Deployment-read is a far more widely granted RBAC permission than secret-read — it is what you hand someone so they can look at workloads. So the key signing every session, and the key encrypting the stored SMS/email provider credentials, were readable by anyone who could inspect the cluster's workloads. `JWT_SECRET` is the sharp one: knowing it means a session can be **forged outright**.

## Fix

A Secret plus `secretKeyRef`, with two supply routes:

1. **`novu.secrets.existingSecret`** — name a Secret created out of band. The values never enter `values.yaml`, the Helm release, or `helm get values`.
2. **Chart-managed** (default) — the chart creates the Secret from the values.

I have been deliberately explicit in the template comment that route 2 is *not* a complete hiding: Helm stores the rendered manifest in its own release Secret, so the value is still recoverable by anyone who can read Secrets in that namespace. The real gain is moving it out from behind deployment-read. Overstating that would be worse than not fixing it.

The Secret name comes from a single helper used by both the Secret and every `secretKeyRef`, so they cannot drift. `required()` on each value means a blank secret fails rendering loudly rather than silently deploying an **empty** signing key — worse than a weak one.

## Verification

Rendered all three paths with `helm template`:

| Case | Result |
|---|---|
| Default | one Secret created; all 5 refs across `novu-api`, `novu-worker`, `novu-ws` resolve to `secretKeyRef`; **zero plaintext leaks** |
| `existingSecret` set | chart creates **no** Secret; every ref retargets to the external name |
| Blank value, no `existingSecret` | fails with `novu.secrets.jwtSecret must be set (or set novu.secrets.existingSecret)` |

No live cluster runs this tier, so nothing needs rotating today.

### DIFF
--- a/devops/deploy-as-code/charts/backbone-services/novu/templates/secret.yaml
+++ b/devops/deploy-as-code/charts/backbone-services/novu/templates/secret.yaml
@@ -1,0 +1,41 @@
+{{- /*
+Signing and encryption secrets for Novu.
+
+These used to render as plaintext `env:` entries on the Deployments, which meant
+anyone with namespace READ access could recover them with
+`kubectl get deploy -o yaml`. Deployment read is a very common RBAC grant --
+far more widely handed out than secret read -- so the keys that sign every
+session and encrypt the stored SMS/email provider credentials were effectively
+readable by anyone who could look at the workloads.
+
+Two ways to supply them, in order of preference:
+
+  1. novu.secrets.existingSecret -- name a Secret you created out of band. The
+     values then never enter values.yaml, the Helm release, or `helm get values`
+     at all. This is the right choice for anything real.
+
+  2. Leave it empty and this template creates the Secret from
+     novu.secrets.{jwtSecret,storeEncryptionKey,novuSecretKey}. Better than
+     plaintext env -- the value moves behind secret-read RBAC -- but be aware
+     Helm still stores the rendered manifest in its own release Secret, so it is
+     recoverable by anyone who can read Secrets in this namespace.
+*/}}
+{{- if not .Values.novu.secrets.existingSecret }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "novu.secretName" . }}
+  namespace: {{ include "novu.namespace" . }}
+  labels:
+    app.kubernetes.io/name: novu
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+type: Opaque
+stringData:
+  # Signs session tokens. A known value lets a session be forged outright --
+  # no password to guess and nothing to intercept.
+  jwt-secret: {{ required "novu.secrets.jwtSecret must be set (or set novu.secrets.existingSecret)" .Values.novu.secrets.jwtSecret | quote }}
+  # Encrypts the stored SMS/email provider credentials. Must be EXACTLY 32
+  # characters -- Novu fails to start otherwise.
+  store-encryption-key: {{ required "novu.secrets.storeEncryptionKey must be set (or set novu.secrets.existingSecret)" .Values.novu.secrets.storeEncryptionKey | quote }}
+  novu-secret-key: {{ required "novu.secrets.novuSecretKey must be set (or set novu.secrets.existingSecret)" .Values.novu.secrets.novuSecretKey | quote }}
+{{- end }}
--- a/devops/deploy-as-code/charts/backbone-services/novu/templates/secret.yaml
+++ b/devops/deploy-as-code/charts/backbone-services/novu/templates/secret.yaml
@@ -19,8 +19,41 @@ Two ways to supply them, in order of preference:
      plaintext env -- the value moves behind secret-read RBAC -- but be aware
      Helm still stores the rendered manifest in its own release Secret, so it is
      recoverable by anyone who can read Secrets in this namespace.
+
+`required` below rejects only a blank or nil value. A forgotten override is not
+blank -- it is the example string this chart ships, which is published in this
+repository and would render straight through into a real Secret. So the
+placeholders are rejected explicitly, following the same idea as
+backbone-services/db-seed/templates/secret.yaml, which refuses the `<...>`
+placeholders env.yaml ships rather than letting them reach a cluster.
+
+This is the single place all three values pass through, which is why the check
+lives here and not in the helmfile: the deployments read them by secretKeyRef,
+so nothing can reach a pod without being written here first.
+
+Both placeholder shapes are refused -- the repo-wide `<...>` convention and any
+`change-me` value. A published JWT signing key is a forged session for anyone
+who reads this repo, so failing the render is the right outcome; this tier
+already requires substitution before it can deploy (db-seed does the same for
+`<db_host>` and friends in env.yaml).
 */}}
 {{- if not .Values.novu.secrets.existingSecret }}
+{{- $jwt := required "novu.secrets.jwtSecret must be set (or set novu.secrets.existingSecret)" .Values.novu.secrets.jwtSecret | toString }}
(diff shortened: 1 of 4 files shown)


vinothrallapalli-eGov (commented):
Verified the core claim: all 5 secretKeyRef sites (JWT_SECRET x2 in api/ws, STORE_ENCRYPTION_KEY x2 in api/worker, NOVU_SECRET_KEY x1 in api) now resolve via secretKeyRef with zero remaining plaintext `value:` entries for these three secrets, and the `existingSecret` path correctly suppresses Secret creation via the `if not .Values.novu.secrets.existingSecret` guard in secret.yaml. Cross-check with #1671: the helmfile there sets `novu.secrets.jwtSecret/storeEncryptionKey/novuSecretKey` from `secrets.novu.*` in env-secrets.yaml, and those field names line up exactly with what #1677's values.yaml/secret.yaml expect — no naming drift, and no other environment file overrides `novu.secrets.*` that could go stale. One nuance on the 'fails loudly' claim: `required()` only fires on a genuinely blank/nil value, not on the chart's own non-empty placeholder defaults (`jwtSecret: "change-me"`, etc., values.yaml lines 51-53) — so a `helm install` with zero overrides will render and deploy a real Secret containing the well-known placeholder value without any warning. That's not a regression (the same placeholder was already deployed in plaintext before this PR) but it means the loud-failure guard only protects against an *empty* override, not a forgotten one. Minor/informational, not blocking.

devops/deploy-as-code/charts/backbone-services/novu/values.yaml
```diff
@@ -39,6 +39,15 @@ novu:
...
+    # release, or `helm get values`. It must carry the keys `jwt-secret`,
+    # `store-encryption-key` and `novu-secret-key`:
+    #   kubectl -n novu create secret generic novu-secrets \
+    #     --from-literal=jwt-secret='...' \
+    #     --from-literal=store-encryption-key='<exactly 32 chars>' \
+    #     --from-literal=novu-secret-key='...'
+    existingSecret: ""
     jwtSecret: "change-me"
     storeEncryptionKey: "change-me-exactly-32-characters!"
     novuSecretKey: "change-me"
```
vinothrallapalli-eGov, line 53: Nit / informational: `required()` in secret.yaml only rejects a truly blank/nil value — it will happily accept these `"change-me"` placeholder defaults. So a `helm install` with no overrides at all renders successfully and bakes the known placeholder into a real Secret object, with no warning that the value is still an example. Not a regression from this PR (same placeholder was previously shipped in plaintext), but worth knowing the 'fails loudly on blank' guarantee doesn't extend to 'fails loudly on forgotten override'.

devops/deploy-as-code/charts/backbone-services/novu/templates/secret.yaml
```diff
@@ -0,0 +1,41 @@
...
+  name: {{ include "novu.secretName" . }}
+  namespace: {{ include "novu.namespace" . }}
+  labels:
+    app.kubernetes.io/name: novu
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+type: Opaque
+stringData:
+  # Signs session tokens. A known value lets a session be forged outright --
+  # no password to guess and nothing to intercept.
+  jwt-secret: {{ required "novu.secrets.jwtSecret must be set (or set novu.secrets.existingSecret)" .Values.novu.secrets.jwtSecret | quote }}
```
vinothrallapalli-eGov: Confirmed this ties back cleanly to #1671: the helmfile there sets `novu.secrets.jwtSecret` (and storeEncryptionKey/novuSecretKey) from the SOPS-covered `secrets.novu.*` values in env-secrets.yaml, and those field names match exactly what this template reads. No naming drift between the two PRs.

devops/deploy-as-code/charts/backbone-services/novu/values.yaml
singhalkarun: Correct, and I have closed the gap — `required()` rejects only blank or nil, so the shipped example strings rendered straight through into a real Secret with nothing to flag them.

cbfcf5ae adds explicit placeholder rejection here, at the one point all three values pass through (the deployments read them by `secretKeyRef`, so nothing reaches a pod without being written by this template). Both shapes are refused — the repo-wide `<...>` convention and any `change-me` value — following `db-seed/templates/secret.yaml`:

| values | result |
|---|---|
| chart defaults (`change-me`) | `fail`, names `jwtSecret` and the value |
| `env-secrets.yaml` as shipped (`change-me-strong`) | `fail`, names `jwtSecret` |
| `jwtSecret: <novu_jwt>` | `fail`, names the placeholder |
| all three real, 32-char key | Secret renders |
| `existingSecret` set | no Secret, no guards, all 5 `secretKeyRef`s point at the out-of-band name |

Your framing of the two guarantees was the useful part — "fails loudly on blank" and "fails loudly on forgotten override" are genuinely different, and only the first was covered. It now fails on the second by name:

```
novu: novu.secrets.jwtSecret is still the example value "change-me", which is published
in this repository. Generate a real value and set secrets.novu.jwtSecret in
environments/env-secrets.yaml, or point novu.secrets.existingSecret at a Secret you
created out of band.
```

Since `novu` ships `installed: true`, this means the backbone tier will not deploy until an operator substitutes them. Deliberate: a published JWT signing key is a forged session for anyone who reads this repo, and the tier already requires substitution before it can deploy (db-seed refuses `<db_host>` from `env.yaml` identically). `env-secrets.yaml`'s comment now says so explicitly.

One addition beyond your comment, flagged so you can object: the same commit enforces the **exactly-32-characters** rule on `storeEncryptionKey`. This file already documented the invariant but did not check it, and Novu refuses to start otherwise — so a 31-character key was a crash-loop instead of `novu: novu.secrets.storeEncryptionKey must be EXACTLY 32 characters, got 10`. Happy to drop it if you would rather keep this PR to the placeholder question.

Not verifiable here: no cluster runs the `deploy-as-code` tier. The 32-character rule is taken from the pre-existing comment in this file, not confirmed against a running Novu.

devops/deploy-as-code/charts/backbone-services/novu/templates/secret.yaml
singhalkarun: Thanks for checking the cross-PR naming — I re-verified it after cbfcf5ae, since that commit touches this template, and it still holds. The helmfile in #1671 sets `novu.secrets.{jwtSecret,storeEncryptionKey,novuSecretKey}` and those are exactly the three fields read here.

Also confirmed the other half of the wiring, which is where drift would actually bite:

```
$ helm template novu . --set-string novu.secrets.jwtSecret=... (all three real)
1  name: novu-secrets          # the Secret this template creates
5  name: novu-secrets          # secretKeyRefs resolving to it
0  plaintext secret values in any Deployment env
```

Five references: `jwt-secret` in `api` and `ws`, `store-encryption-key` in `api` and `worker`, `novu-secret-key` in `api`. All resolve through the `novu.secretName` helper, and with `existingSecret` set all five follow it to the out-of-band name instead — so the Secret and its readers cannot drift apart, which was the point of defining the helper once.

vinothrallapalli-eGov (approved):

PR #1678, opened 2026-08-07
feat(kafka): add the consumer-lag dashboard (#1623)

singhalkarun:
Closes #1623

> **Stacked on #1628** (`feat/1623-redpanda-metrics`), which made the broker visible but deliberately stopped short of this panel.

## What was blocking it

Consumer lag is the headline reason #1623 exists — the failure Gatus structurally cannot see. `egov-persister` falls behind, the API keeps returning 200, every health check stays green, and complaints are simply never written to the database.

#1628 could not ship the panel honestly, for a good reason:

- Redpanda does not export lag directly; it must be derived as `max_offset − committed_offset`, joined on topic and partition
- Neither official Redpanda dashboard (18134/18135) contains such a panel, so neither revealed **the label names the join needs**
- Guessing them would ship panels that render blank while looking complete — exactly the trap #1616 describes

## How it's resolved

I read the labels off a real broker instead of guessing — the pinned `egovio/redpanda:v24.1.1`, with two consumer groups deliberately left behind:

```
redpanda_kafka_consumer_group_committed_offset{redpanda_group,redpanda_topic,redpanda_partition}
redpanda_kafka_max_offset{redpanda_namespace,redpanda_topic,redpanda_partition}
```

**The asymmetry is the entire difficulty**, and it is why guessing would have failed: `committed_offset` carries `redpanda_group` and *no* namespace; `max_offset` carries `redpanda_namespace` and *no* group. So:

```promql
redpanda_kafka_max_offset{redpanda_namespace="kafka"}
  - on(redpanda_topic, redpanda_partition) group_right()
    redpanda_kafka_consumer_group_committed_offset
```

`group_right()` so the result keeps the group label, and the `redpanda_namespace="kafka"` filter because otherwise `max_offset` also matches the `controller` and `id_allocator` internal topics.

## Verification against a live broker

Prometheus scraping that Redpanda, cross-checked against `rpk group describe`:

| Group | Dashboard query | `rpk` TOTAL-LAG |
|---|---|---|
| `egov-persister` | 25 | 25 |
| `egov-indexer` | 18 (10 + 8 across two topics) | 18 |

Also confirmed: **every panel query returns data** (no blank panels), the `group` template variable resolves to both groups, and Grafana 11.4.0 provisions the file with no errors.

## The panels

| Panel | Why |
|---|---|
| Total lag by group (timeseries) | the number that matters: climbing and staying up means records accepted and never processed |
| Current lag by group (stat) | thresholds at 1k/10k, deliberately **wide** — a healthy consumer often sits a few hundred behind during a burst, so tighter bands would mostly measure normal operation |
| Lag by topic/partition (table) | one hot partition behind while the rest keep up usually means a poison record or uneven partition key, not an undersized consumer — opposite responses |
| Consumers per group | lag climbing while this reads **0** means the consumer is *gone*, not slow |

Shipped as its own dashboard rather than edited into `redpanda-broker.json`, which is vendored upstream and should stay re-vendorable.

A matching lag **alert** belongs with the alerting rules in #1673 rather than here; now that the expression is verified, that is a small follow-on.

### DIFF
--- a/local-setup/otel/grafana/provisioning/dashboards/kafka-consumer-lag.json
+++ b/local-setup/otel/grafana/provisioning/dashboards/kafka-consumer-lag.json
@@ -47,7 +47,7 @@
     {
       "id": 1,
       "title": "Total lag by consumer group",
-      "description": "Messages produced but not yet committed by each group. This is the number that matters: when it climbs and stays up, records are being accepted and never processed.\n\nRedpanda does not export lag directly -- it is derived as max_offset - committed_offset, joined on topic and partition. The join uses group_right() because committed_offset carries redpanda_group and max_offset does not.",
+      "description": "Messages produced but not yet committed by each group. This is the number that matters: when it climbs and stays up, records are being accepted and never processed.\n\nRedpanda does not export lag directly -- it is derived as max_offset - committed_offset, joined on topic and partition. The join uses group_right() because committed_offset carries redpanda_group and max_offset does not.\n\nNegative values are possible and are diagnostic, not noise: they are left un-clamped on purpose. Both series come from the SAME Prometheus target and scrape (job \"redpanda\", redpanda:9644/public_metrics), so they are mutually consistent within a scrape and cannot skew against each other. A committed offset AHEAD of the partition high-water mark therefore means something real -- an offset reset, a topic truncation/recreate, or a series that has gone stale on one side. clamp_min(..., 0) would render exactly that condition as 0, i.e. as \"caught up\", which is the misreading it would be meant to prevent.\n\nmax_offset is collapsed with max by (namespace, topic, partition) before the join. On this single-broker, --smp 1 stack it is a no-op (one series per partition), but on a multi-broker cluster the same partition is exported by every replica and the on(topic, partition) join would fail with \"multiple matches for labels: many-to-one matching must be explicit\". max, not sum: the high-water mark of a partition is not the sum of its replicas' views of it.",
       "type": "timeseries",
       "datasource": {
         "type": "prometheus",
@@ -62,7 +62,7 @@
       "targets": [
         {
           "refId": "A",
-          "expr": "sum by (redpanda_group) (redpanda_kafka_max_offset{redpanda_namespace=\"kafka\"} - on(redpanda_topic, redpanda_partition) group_right() redpanda_kafka_consumer_group_committed_offset{redpanda_group=~\"$group\"})",
+          "expr": "sum by (redpanda_group) (max by (redpanda_namespace, redpanda_topic, redpanda_partition) (redpanda_kafka_max_offset{redpanda_namespace=\"kafka\"}) - on(redpanda_topic, redpanda_partition) group_right() redpanda_kafka_consumer_group_committed_offset{redpanda_group=~\"$group\"})",
           "legendFormat": "{{redpanda_group}}"
         }
       ],
@@ -112,7 +112,7 @@
     {
       "id": 2,
       "title": "Current lag by group",
-      "description": "Same figure, right now. Green under 1k, amber past 1k, red past 10k -- deliberately wide bands: a healthy consumer often sits a few hundred behind during a burst, so a tighter threshold would mostly measure normal operation.",
+      "description": "Same figure, right now. Green under 1k, amber past 1k, red past 10k -- deliberately wide bands: a healthy consumer often sits a few hundred behind during a burst, so a tighter threshold would mostly measure normal operation.\n\nNegative values are possible and are diagnostic, not noise: they are left un-clamped on purpose. Both series come from the SAME Prometheus target and scrape (job \"redpanda\", redpanda:9644/public_metrics), so they are mutually consistent within a scrape and cannot skew against each other. A committed offset AHEAD of the partition high-water mark therefore means something real -- an offset reset, a topic truncation/recreate, or a series that has gone stale on one side. clamp_min(..., 0) would render exactly that condition as 0, i.e. as \"caught up\", which is the misreading it would be meant to prevent.",
       "type": "stat",
       "datasource": {
         "type": "prometheus",
@@ -128,7 +128,7 @@
         {
           "refId": "A",
           "instant": true,
-          "expr": "sum by (redpanda_group) (redpanda_kafka_max_offset{redpanda_namespace=\"kafka\"} - on(redpanda_topic, redpanda_partition) group_right() redpanda_kafka_consumer_group_committed_offset)",
+          "expr": "sum by (redpanda_group) (max by (redpanda_namespace, redpanda_topic, redpanda_partition) (redpanda_kafka_max_offset{redpanda_namespace=\"kafka\"}) - on(redpanda_topic, redpanda_partition) group_right() redpanda_kafka_consumer_group_committed_offset{redpanda_group=~\"$group\"})",
           "legendFormat": "{{redpanda_group}}"
         }
       ],
@@ -174,7 +174,7 @@
     {
       "id": 3,
       "title": "Lag by topic and partition",
-      "description": "Where the lag actually is. A single hot partition behind while the rest keep up usually means a poison record or an uneven partition key, not an undersized consumer -- and the two need opposite responses.",
+      "description": "Where the lag actually is. A single hot partition behind while the rest keep up usually means a poison record or an uneven partition key, not an undersized consumer -- and the two need opposite responses.\n\nmax_offset is collapsed with max by (namespace, topic, partition) before the join. On this single-broker, --smp 1 stack it is a no-op (one series per partition), but on a multi-broker cluster the same partition is exported by every replica and the on(topic, partition) join would fail with \"multiple matches for labels: many-to-one matching must be explicit\". max, not sum: the high-water mark of a partition is not the sum of its replicas' views of it.",
       "type": "table",
       "datasource": {
         "type": "prometheus",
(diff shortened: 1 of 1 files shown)


local-setup/otel/grafana/provisioning/dashboards/kafka-consumer-lag.json
```diff
@@ -0,0 +1,290 @@
...
+        "h": 7,
+        "w": 10,
+        "x": 0,
+        "y": 9
+      },
+      "targets": [
+        {
+          "refId": "A",
+          "instant": true,
+          "expr": "sum by (redpanda_group) (redpanda_kafka_max_offset{redpanda_namespace=\"kafka\"} - on(redpanda_topic, redpanda_partition) group_right() redpanda_kafka_consumer_group_committed_offset)",
```
vinothrallapalli-eGov: This panel's query is missing the `{redpanda_group=~"$group"}` filter that panels 1 and 3 apply (see lines 65 and 194). As written, the "Current lag by group" stat always aggregates every consumer group regardless of the dashboard's `group` template-variable selection — likely an oversight rather than intentional, since the panel is titled/scoped as per-group.

local-setup/otel/grafana/provisioning/dashboards/kafka-consumer-lag.json
```diff
@@ -0,0 +1,290 @@
...
+      "gridPos": {
+        "h": 9,
+        "w": 24,
+        "x": 0,
+        "y": 0
+      },
+      "targets": [
+        {
+          "refId": "A",
+          "expr": "sum by (redpanda_group) (redpanda_kafka_max_offset{redpanda_namespace=\"kafka\"} - on(redpanda_topic, redpanda_partition) group_right() redpanda_kafka_consumer_group_committed_offset{redpanda_group=~\"$group\"})",
```
vinothrallapalli-eGov: No `clamp_min(..., 0)` around the `max_offset - committed_offset` subtraction. Since these are two independently-scraped gauges (not read from a single atomic snapshot), a broker restart, leader election, or ordinary scrape-timing skew can transiently let `committed_offset` exceed a stale `max_offset` reading, producing a negative lag dip on the timeseries/stat panels. Worth wrapping in `clamp_min(...)` (or a deliberate note that negative values are expected and diagnostic) so on-call doesn't misread a dip as "caught up".

local-setup/otel/grafana/provisioning/dashboards/kafka-consumer-lag.json
```diff
@@ -0,0 +1,290 @@
...
+        "w": 14,
+        "x": 10,
+        "y": 9
+      },
+      "targets": [
+        {
+          "refId": "A",
+          "instant": true,
+          "format": "table",
+          "expr": "redpanda_kafka_max_offset{redpanda_namespace=\"kafka\"} - on(redpanda_topic, redpanda_partition) group_right() redpanda_kafka_consumer_group_committed_offset{redpanda_group=~\"$group\"}"
```
vinothrallapalli-eGov: Per Redpanda's public-metrics reference, `redpanda_kafka_max_offset` carries a `shard` label in addition to namespace/topic/partition — it's emitted per-shard, not guaranteed unique per (topic, partition) across a cluster. This local-setup's Redpanda runs `--smp 1` (single shard, single broker), which masks the case where a partition handoff between shards/brokers could transiently produce two matching series and trip Prometheus's "many-to-many matching not allowed" on this un-aggregated table query. Not a bug for the verified single-node topology, but worth a defensive `sum by(redpanda_topic, redpanda_partition, redpanda_namespace)` on the max_offset side (or an explicit comment) if this dashboard is expected to survive a multi-broker/multi-shard Redpanda deployment.

vinothrallapalli-eGov (commented):
Confirmed genuinely dashboard-only (single new file, no changes to `otel/prometheus.yml`), and it correctly targets `feat/1623-redpanda-metrics` (base of #1628), so no scrape-config duplication risk. The `group_right()` join is structurally correct for the committed_offset/max_offset label asymmetry described in the PR body. Three non-blocking issues found: the 'Current lag by group' stat panel silently ignores the `$group` template variable (inconsistent with the other two panels that filter on it), the subtraction has no `clamp_min(...,0)` guard against transient negative lag from scrape-timing skew or leader election, and `redpanda_kafka_max_offset` is emitted per-shard per Redpanda's public-metrics reference, so the 'exactly one series per (topic,partition)' assumption is untested beyond this repo's single-shard (`--smp 1`) broker. Also noting for the record: the PR body's own verification table only names `egov-persister`/`egov-indexer`, not the four groups (`egov-pgr-services`, `egov-infra-persist`, `egov-wf-services`, `audit-service`) referenced elsewhere as 'verified' — worth reconciling which groups were actually checked.

local-setup/otel/grafana/provisioning/dashboards/kafka-consumer-lag.json
singhalkarun: Confirmed — straight oversight, fixed in 7ebcb826. Panel 2 now carries the same `{redpanda_group=~"$group"}` filter as panels 1, 3 and 4, so all four honour the template variable:

```promql
sum by (redpanda_group) (
  max by (redpanda_namespace, redpanda_topic, redpanda_partition) (redpanda_kafka_max_offset{redpanda_namespace="kafka"})
  - on(redpanda_topic, redpanda_partition) group_right()
    redpanda_kafka_consumer_group_committed_offset{redpanda_group=~"$group"}
)
```

(The `max by (...)` is from the separate fix on your other comment.) Placed on the right-hand side to match panels 1 and 3 — `committed_offset` is the series that carries `redpanda_group`, so that is the only side it can go on. Expression parse-checked with `promtool check rules` after substituting the variable.

local-setup/otel/grafana/provisioning/dashboards/kafka-consumer-lag.json
singhalkarun: Half of this does not apply, and the half that does argues the other way — so I have taken the second option you offered (a deliberate note) rather than clamping. 7ebcb826 adds it to the panel descriptions.

**"two independently-scraped gauges" is not the case here.** Both series come from the same Prometheus job, the same target and the same scrape: `job_name: redpanda`, `metrics_path: /public_metrics`, `targets: [redpanda:9644]` (`otel/prometheus.yml:34-37`, from #1628). Redpanda exports both `redpanda_kafka_max_offset` and `redpanda_kafka_consumer_group_committed_offset` from that one endpoint, so within a scrape they are one snapshot at one timestamp and share the broker's own view of the log. There is no inter-target scrape skew to protect against, and on `--smp 1 --node-id 0` there is no leader election either.

**Where a negative can still come from, it should be visible.** A committed offset genuinely ahead of the partition high-water mark means an offset reset, a topic truncate/recreate, or one side of the join having gone stale. Those are real conditions an on-call wants to see.

**And clamping would cause the exact misreading it is meant to prevent.** `clamp_min(..., 0)` renders every one of those cases as `0` — which on a lag panel reads as "caught up". A negative reads as "this number is wrong, go look"; a clamped zero reads as "all good". The un-clamped version is the safer of the two for the scenario in your comment.

Worth noting the verification in the PR description also argues the arithmetic is not off by one: the query returned 25 and 18 against `rpk group describe`'s TOTAL-LAG of 25 and 18 for `egov-persister` and `egov-indexer` (the latter as 10 + 8 across two topics). A caught-up consumer therefore reads 0, not -1, so a negative on this dashboard is genuinely anomalous rather than a permanent cosmetic wart.

local-setup/otel/grafana/provisioning/dashboards/kafka-consumer-lag.json
singhalkarun: Defensive aggregation added in 7ebcb826, but with `max` rather than `sum`, and the `shard` premise does not hold — the real multi-node hazard is `instance`.

**On `shard`:** `redpanda_kafka_max_offset` as scraped here carries no `shard` label. The labels were read off the pinned `egovio/redpanda:v24.1.1` broker for this PR (recorded in the description) and are `{redpanda_namespace, redpanda_topic, redpanda_partition}` — the asymmetry with `committed_offset` is the whole reason the join looks the way it does. The scrape uses `/public_metrics`, not `/metrics`; `shard` lives on the internal endpoint. I think the reference you are recalling is the vendored upstream board: `redpanda-broker.json` offers a `$aggr_criteria` custom variable with options `cluster` / `instance` / `instance,shard`, applied uniformly to all 19 panels including ones whose metrics have no `shard`. That is a dashboard-wide dropdown, not evidence of the label on this series.

**The concern is still right, for a different label.** On a multi-broker cluster every replica exports `redpanda_kafka_max_offset` for the same `(topic, partition)`, differing only by `instance` — which `on(redpanda_topic, redpanda_partition)` ignores. That is a genuine `multiple matches for labels: many-to-one matching must be explicit` waiting to happen, and it needs no shard handoff to trigger; three brokers is enough.

**`max`, not `sum`.** This matters: summing high-water marks across replicas would add three views of the same offset and report roughly 3x the true value as lag — a large fake lag spike exactly when a replica set is being read. A partition's high-water mark is not the sum of its replicas' views of it, it is the maximum. So all three lag panels now read:

```promql
max by (redpanda_namespace, redpanda_topic, redpanda_partition) (redpanda_kafka_max_offset{redpanda_namespace="kafka"})
  - on(redpanda_topic, redpanda_partition) group_right()
    redpanda_kafka_consumer_group_committed_offset{redpanda_group=~"$group"}
```

A strict no-op on this topology (one series per partition, so `max` of one), and the rationale is in the panel descriptions. Output labels are unchanged — `group_right()` already took its labels from the right-hand side and dropped left-hand labels outside `on()` — so the table's `organize` transformation still applies cleanly. All four expressions parse-checked with `promtool check rules`.

Not claimed as fixed: full multi-broker correctness. `sum by (redpanda_group)` on panels 1 and 2 would still double-count if `committed_offset` were itself exported by more than one broker, which I cannot verify without a real cluster. This change removes the hard query error; it does not make the dashboard multi-broker-certified.

vinothrallapalli-eGov (approved):

PR #1679, opened 2026-08-07
fix(ingress-nginx): apply the security config that was silently ignored (#1648)

singhalkarun:
Closes #1648

> **Stacked on #1663** (`feat/1648-gateway-metrics`), which fixed the metrics half. This is the rest.

> **Narrative revised after review** ([thread](https://github.com/egovernments/Citizen-Complaint-Resolution-System/pull/1679#discussion_r3748099261)). The original version of this description said the whole `ingress-nginx:` block was inert. That is wrong, and the corrected version below is smaller in scope but sharper — and it turned up a real defect in #1663, now fixed there.

## What was actually wrong

The vendored chart carries a **local modification** to `templates/_helpers.tpl` (lines 6-9), absent from upstream ingress-nginx 4.10.0:

```gotemplate
```

`mergeOverwrite` mutates its destination in place and the destination is the root context, so the first template that renders a label merges `.Values["ingress-nginx"]` over `.Values` for the whole render. Someone added this deliberately so `env.yaml`'s block would be honoured. It means the `ingress-nginx.controller.*` subtree **is** read — but only that subtree, and only where the key names match.

So the block splits three ways:

| Setting | Status before this PR |
|---|---|
| `controller.addHeaders` (X-Frame-Options, nosniff, X-XSS-Protection, CORS) | **live** — hoisted to `.Values.controller.addHeaders` |
| `ssl-protocols`, `ssl-ciphers`, `ssl-ecdh-curve` | **inert** — siblings of `controller:`, so they hoist to root `.Values.ssl-*`, not `.Values.controller.config.*` |
| `controller.replicas` | **inert** — the chart reads `replicaCount` |

The middle row is the one that matters: `ssl-protocols: TLSv1.2 TLSv1.3` was never applied, so the controller kept its build default, which on some builds still permits TLSv1.0/1.1. The cipher suite and ECDH curve were not pinned either.

## What this PR does

Re-paths all of it explicitly through the helmfile, at the release's root `controller:` key. `addHeaders` is included as belt-and-braces even though the hoist already delivers it — stating it beats depending on a hand-patched helper that a chart bump could drop.

Not by moving the block to a top-level `controller:` key in `env.yaml`: **kafka-kraft reads `.Values.controller` too** (13 templates, its own controller block), and the whole-values passthrough is shared — a root-level move would leak ingress settings into the broker.

An inline values block rather than `set:`, because `Access-Control-Allow-Headers` contains **commas**, which helm's `--set` parses as separate assignments. (And, as #1663 discovered, `--set` loses to the hoist anyway for any key `env.yaml` names.)

## A foreign hostname, removed

`Access-Control-Allow-Origin` named `urban-lts.digit.org` — a *different* environment, inherited from the upstream import. Because `addHeaders` is one of the keys that **was** live, this was being advertised on every response as the only allowed origin, breaking CORS for this deployment's own SPA. It is now `https://<domain_name>`.

To be precise about that, since the first draft of this description was not: `<domain_name>` is a **placeholder, not a template**. This file is plain YAML at the point it is loaded; nothing interpolates it. It is the same manual-substitution convention as `global.domain` on line 2, so one sed keeps them in sync — which is the improvement over a hardcoded foreign host.

`setSslTls` named the same foreign host and is commented out: nothing reads it. The default certificate comes from `--default-ssl-certificate=egov/{{ .Values.domain }}-tls-certs` in `_params.tpl` (itself a local addition), which the helmfile feeds from `global.domain`.

## A trap worth recording

This cost real time and is invisible on review: **a Go-template action inside a YAML comment here is still rendered.**

helmfile templates the file *before* parsing it as YAML, so `#` protects nothing. An interpolation written inside an explanatory comment expanded the entire values tree into the middle of that comment line and corrupted the document — with a parse error pointing thousands of lines away in the rendered output, nowhere near the cause. There is now a note in the file saying so.

## Verification

Pinned **helmfile v0.140.0** (CI's version; v1 rejects these `.yaml` files) plus `helm template`. No cluster — this tier has none, so nothing here is a runtime claim.

`helmfile -f backboneservices-helmfile.yaml -e env -l name=ingress-nginx template` on the merged result:

```
ConfigMap ingress-nginx-custom-add-headers:
    all six headers present, Access-Control-Allow-Origin: https://<domain_name>
ConfigMap ingress-nginx-controller:
    add-headers: backbone/ingress-nginx-custom-add-headers
    ssl-ciphers:     EECDH+CHACHA20:EECDH+AES
    ssl-ecdh-curve:  X25519:prime256v1:secp521r1:secp384r1
    ssl-protocols:   TLSv1.2 TLSv1.3
Deployment ingress-nginx-controller: replicas 1
Service    ingress-nginx-controller-metrics          (from #1663)
ServiceMonitor ingress-nginx-controller: release=kube-prometheus-stack   (from #1663)
no PrometheusRule                                    (from #1663)
```

That last step is the one that matters: it shows the settings reach the controller's own ConfigMaps, not merely that the values merged.

`.github/scripts/check-helm-values-paths.py` from #1652, run against this branch: `OK: 2 external chart(s) checked, no new unread key paths.`

## Follow-on for #1652

The values-path guard lists `ingress-nginx` under `KNOWN_UNREAD`. The entry does not break CI, but its reason text is stale twice over: the chart *does* reach the `controller.*` subtree via the patched helper, and the helmfile re-paths the rest deliberately. It should move to `ALLOW` with a reason naming both when #1652 merges. Not edited here — that script does not exist on this branch.

### DIFF
--- a/devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
+++ b/devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
@@ -161,6 +161,22 @@ releases:
     # and corrupts the document.
     - controller:
         replicaCount: {{ index .Values "ingress-nginx" "controller" "replicas" | default 1 }}
+        # Roll the controller whenever the headers change.
+        #
+        # The `add-headers` ConfigMap is read by nginx when the controller
+        # builds its configuration, and the running process does not pick up
+        # later edits to it. Verified on a cluster: the ConfigMap was updated at
+        # 11:39:14 and the controller was still serving the OLD
+        # Access-Control-Allow-Origin at 11:41 -- it had even logged a RELOAD,
+        # for the main config, while the custom headers stayed stale. A restart
+        # applied them immediately.
+        #
+        # Without this annotation a `helmfile apply` that changes only these
+        # headers reports success and changes nothing that a browser can see --
+        # the same silent no-op this whole block exists to fix, one level up.
+        # The hash makes the pod template differ, so helm rolls it for us.
+        podAnnotations:
+          checksum/add-headers: {{ index .Values "ingress-nginx" "controller" "addHeaders" | toYaml | sha256sum }}
         # Rendered into its own ConfigMap and referenced by the controller's
         # `add-headers` key -- this is what actually applies X-Frame-Options,
         # nosniff and the CORS policy to every response.
--- a/devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
+++ b/devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
@@ -127,6 +127,39 @@ releases:
     - ./ingress-nginx/values.yaml
     - {{ .Values | toYaml | nindent 8 }}
     - domain: {{ .Values.global.domain }}
+    # Security config (#1648). env.yaml's `ingress-nginx:` block is read by
+    # NOTHING -- this vendored chart reads .Values.controller.* at the root, so
+    # everything under `ingress-nginx.controller.*` lands one level too deep.
+    # The consequence was not cosmetic: the security headers, the TLS cipher
+    # suite and the protocol floor were all silently absent while the file
+    # stated them plainly.
+    #
+    # Re-pathed HERE rather than by moving the block to a top-level
+    # `controller:` key, because kafka-kraft reads .Values.controller too (13
+    # templates, its own controller block) and the whole-values passthrough on
+    # the line above is shared -- a root-level move would leak these into the
+    # broker.
+    #
+    # An inline values block rather than `set:`: Access-Control-Allow-Headers
+    # contains COMMAS, which helm's --set parses as separate assignments.
+    #
+    # NB: do not write a Go-template action in these comments. helmfile renders
+    # the file as a template BEFORE parsing it as YAML, so `#` does not protect
+    # anything -- an interpolation here expands into the middle of the comment
+    # and corrupts the document.
+    - controller:
+        replicaCount: {{ index .Values "ingress-nginx" "controller" "replicas" | default 1 }}
+        # Rendered into its own ConfigMap and referenced by the controller's
+        # `add-headers` key -- this is what actually applies X-Frame-Options,
+        # nosniff and the CORS policy to every response.
+        addHeaders: {{ index .Values "ingress-nginx" "controller" "addHeaders" | toYaml | nindent 10 }}
+        # nginx ConfigMap entries. ssl-protocols is the one that matters most:
+        # without it the controller keeps its build default, which on some
+        # builds still permits TLSv1.0/1.1.
+        config:
+          ssl-ciphers: {{ index .Values "ingress-nginx" "ssl-ciphers" | quote }}
+          ssl-protocols: {{ index .Values "ingress-nginx" "ssl-protocols" | quote }}
+          ssl-ecdh-curve: {{ index .Values "ingress-nginx" "ssl-ecdh-curve" | quote }}
   # Gateway metrics (#1648). Every citizen request crosses this controller and
   # it contributes nothing to Prometheus, so "are we returning errors, and on
   # which route?" can only be answered by reading logs.
--- a/devops/deploy-as-code/charts/environments/env.yaml
+++ b/devops/deploy-as-code/charts/environments/env.yaml
@@ -178,10 +178,19 @@ ingress-nginx:
       X-Content-Type-Options: "nosniff"
       X-Frame-Options: "SAMEORIGIN"
       X-XSS-Protection: "1; mode=block"
-      Access-Control-Allow-Origin: "https://urban-lts.digit.org"
+      # Derived from global.domain by the helmfile (#1648). This used to name
+      # urban-lts.digit.org -- a DIFFERENT environment's hostname, inherited
+      # from the upstream import. It was inert (the whole block was), but had it
+      # ever taken effect it would have advertised a foreign origin as the only
+      # allowed one, breaking CORS for this deployment's own SPA.
+      Access-Control-Allow-Origin: "https://<domain_name>"
       Access-Control-Allow-Methods: "GET, POST, OPTIONS"
       Access-Control-Allow-Headers: "DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range"
-    setSslTls: "urban-lts.digit.org-tls-certs"
+    # Not read by anything (#1648). The default TLS certificate is set by
+    # --default-ssl-certificate in _params.tpl, which derives it from
+    # `domain` -- passed in by the helmfile from global.domain. Kept only to
+    # record that the key is inert; the value named another environment.
+    # setSslTls: "<domain_name>-tls-certs"
 # <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
 
   


devops/deploy-as-code/charts/environments/env.yaml
```diff
@@ -178,10 +178,19 @@ ingress-nginx:
       X-Content-Type-Options: "nosniff"
       X-Frame-Options: "SAMEORIGIN"
       X-XSS-Protection: "1; mode=block"
-      Access-Control-Allow-Origin: "https://urban-lts.digit.org"
+      # Derived from global.domain by the helmfile (#1648). This used to name
+      # urban-lts.digit.org -- a DIFFERENT environment's hostname, inherited
+      # from the upstream import. It was inert (the whole block was), but had it
+      # ever taken effect it would have advertised a foreign origin as the only
+      # allowed one, breaking CORS for this deployment's own SPA.
+      Access-Control-Allow-Origin: "https://<domain_name>"
```
vinothrallapalli-eGov, line 198: Minor: the PR description says this "now derives from global.domain", but this file isn't templated at this point — it's a literal `<domain_name>` placeholder string, same convention as `global.domain` itself (line 2) and the other `<domain_name>` placeholders in this file. Functionally fine (a single manual/sed replace of the placeholder keeps both in sync, which is a real improvement over the old hardcoded foreign host), but worth correcting the PR body's wording so a future reader doesn't assume this is dynamically templated from `global.domain` at render time — it isn't.

devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
```diff
@@ -127,6 +127,39 @@ releases:
...
+    # NB: do not write a Go-template action in these comments. helmfile renders
+    # the file as a template BEFORE parsing it as YAML, so `#` does not protect
+    # anything -- an interpolation here expands into the middle of the comment
+    # and corrupts the document.
+    - controller:
+        replicaCount: {{ index .Values "ingress-nginx" "controller" "replicas" | default 1 }}
+        # Rendered into its own ConfigMap and referenced by the controller's
+        # `add-headers` key -- this is what actually applies X-Frame-Options,
+        # nosniff and the CORS policy to every response.
+        addHeaders: {{ index .Values "ingress-nginx" "controller" "addHeaders" | toYaml | nindent 10 }}
```
vinothrallapalli-eGov, line 167: Informational (verified via helm template before/after, not a blocking issue): this `addHeaders` re-path is good practice, but I could reproduce the same rendered `add-headers` ConfigMap content using the *old* (pre-fix) values shape too — Helm's values-coalescing merges a top-level values key matching the chart's own name (`ingress-nginx:`) into that chart's root values, so `ingress-nginx.controller.addHeaders` was likely already reaching `.Values.controller.addHeaders` before this change, independent of helmfile's re-pathing. That quirk does *not* rescue `ssl-ciphers`/`ssl-protocols`/`ssl-ecdh-curve` (siblings of `controller:`, so they'd land at root `Values.ssl-ciphers` rather than `Values.controller.config.ssl-ciphers`) or `replicaCount` (env.yaml key was `replicas`), so those two pieces of this fix are the ones that were genuinely inert and are now genuinely fixed. Worth tempering the PR narrative accordingly, but no change requested to the code itself — explicit re-pathing here is more robust than relying on the coalescing behavior regardless.

vinothrallapalli-eGov (commented):
Verified independently with `helm`/`helmfile` (v3.20.2 / 0.163.1) against a scratch copy of this repo's charts (no live cluster, static only, matching the prior devbox pass). Findings:

1. **Path translation is correct and the object graph is real.** `helmfile write-values` + `helm template` confirm `controller.replicaCount`, `controller.addHeaders`, and `controller.config.{ssl-ciphers,ssl-protocols,ssl-ecdh-curve}` are exactly the paths `templates/controller-configmap.yaml` / `controller-configmap-addheaders.yaml` read (checked against the vendored chart, `ingress-nginx-4.10.0`). Rendered ConfigMaps carry all four headers and all three ssl-* entries.
2. **Important nuance on the severity claim, reproduced with a before/after diff:** I rendered the release with the values structure as it existed *before* this PR's `values:` addition (env.yaml's `ingress-nginx.controller.addHeaders`, one level "too deep" per the PR body) and the security headers (X-Frame-Options, nosniff, XSS-Protection, CORS) **were already landing in the real `add-headers` ConfigMap** — because Helm's own values-coalescing merges a top-level key that matches the chart's own name (`ingress-nginx:`, here) into that chart's root values, independent of helmfile. I reproduced this with a minimal 4-line values file and a negative control (a differently-named top-level key does nothing). So the header/CORS half of the "silently ignored" claim looks overstated — it was very likely already active. What genuinely *was* broken and is genuinely fixed here: `ssl-ciphers`/`ssl-protocols`/`ssl-ecdh-curve` (siblings of `controller:`, so the coalescing quirk lands them at root `Values.ssl-ciphers` rather than `Values.controller.config.ssl-ciphers`) and `replicaCount` (env.yaml used `replicas`, a different key the chart never reads regardless of nesting). Re-pathing everything explicitly is still the right, more robust fix regardless of this quirk (it shouldn't depend on an incidental, version-fragile Helm behavior) — just flagging that the PR's risk narrative for the headers specifically needs a caveat.
3. **Cipher/protocol choices are sound**: `TLSv1.2 TLSv1.3` only, ECDHE-only cipher list (`EECDH+CHACHA20:EECDH+AES`), standard curves. No `Access-Control-Allow-Credentials` alongside the origin header, so no wildcard+credentials issue (origin is a single value, never `*`).
4. **No regression to #1663**: the metrics `set:` block (lines below this PR's insertion) targets `controller.metrics.*`, disjoint from this PR's `controller.{replicaCount,addHeaders,config}`. Confirmed via full `helmfile template` that both merge without collision. Separately (pre-existing, not caused by this PR — reproduced identically with this PR's change backed out): the metrics `set:` overrides don't appear to take effect in the rendered manifest (`--enable-metrics=false` still renders, no ServiceMonitor object appears) — worth a look under #1663, out of scope here.

Two inline notes below (one minor doc-accuracy nit, one informational).

devops/deploy-as-code/charts/environments/env.yaml
singhalkarun: Correct — the PR body was wrong, and so was the comment I put in `env.yaml` next to the value, which is the worse of the two since it ships. Both fixed in 73d1ca1.

This file is plain YAML at the point it is loaded; nothing interpolates it. `<domain_name>` here is the same manual-substitution convention as `global.domain` on line 2 and the other placeholders in the file — one sed keeps them in sync, which is the actual improvement over a hardcoded foreign host. The comment now says exactly that instead of claiming render-time derivation:

```yaml
      # A placeholder, NOT a template (#1648). This file is plain YAML at this
      # point -- nothing interpolates it -- so `<domain_name>` is the same
      # manual-substitution convention as global.domain on line 2 [...]
```

Worth noting what your other comment forced me to correct in the same breath: the old value was **not** inert. `addHeaders` is one of the keys the chart's patched `_helpers.tpl` hoists to the root, so `Access-Control-Allow-Origin: https://urban-lts.digit.org` was live on every response, not merely sitting in the file. The PR body has been rewritten accordingly.

devops/deploy-as-code/charts/backbone-services/backboneservices-helmfile.yaml
singhalkarun: This is the most valuable comment on the stack — your observation is right, your explanation of *why* is not, and the real mechanism turns out to break the sibling PR. Narrative corrected in 73d1ca1; #1663 has a code fix.

**The mechanism isn't Helm's coalescing.** Helm hoists a top-level key matching a *subchart's* name; for the chart being installed directly it does not, and this chart has no subcharts. What actually does it is a **local modification to the vendored chart** — `templates/_helpers.tpl:6-9`, which I diffed against upstream ingress-nginx 4.10.0:

```
$ diff upstream/templates/_helpers.tpl vendored/templates/_helpers.tpl
5a6,9
> {{- $envOverrides := index .Values (tpl (default .Chart.Name .Values.name) .) -}}
> {{- $baseValues := .Values | deepCopy -}}
> {{- $values := dict "Values" (mustMergeOverwrite $baseValues $envOverrides) -}}
> {{- with mustMergeOverwrite . $values -}}
```

`mergeOverwrite` mutates its destination and the destination is the root context, so the first template to render a label merges `.Values["ingress-nginx"]` over `.Values` for the entire render. Someone added this deliberately so the `ingress-nginx:` block would be honoured. Renaming the chart to `zzz-nginx` makes the behaviour disappear, which pins it to `.Chart.Name` rather than anything Helm does natively.

**Your split is exactly right, for that reason.** `controller.addHeaders` was already arriving. `ssl-ciphers` / `ssl-protocols` / `ssl-ecdh-curve` are siblings of `controller:`, so the hoist lands them at root `.Values.ssl-ciphers`, not `.Values.controller.config.*`; and `replicas` is simply the wrong name for `replicaCount`. Those two are the genuinely-inert pieces this PR repairs, and the comment in the helmfile now says so instead of "read by NOTHING".

**Where it stops being a narrative issue.** The hoist happens at *template* time — after values files and after `--set` — so anything env.yaml names beats a release-level override. #1663 pinned `controller.metrics.enabled=true` via `set:` while env.yaml said `false` one level down, and env.yaml won: through helmfile, that branch rendered **zero** metrics objects. #1663 now writes its pins into the `ingress-nginx:` subtree instead. It also has to pin `prometheusRule.enabled: false`, because env.yaml's `true` was live all along and only gated off by `metrics.enabled` — un-gating it renders a PrometheusRule with `spec: null`, which the CRD rejects.

No change to the code shape here, as you said: the explicit re-path stays, and it stays at the release's root `controller:` rather than moving into env.yaml, since `kafka-kraft` reads `.Values.controller` off the same shared passthrough. Merged result renders the add-headers ConfigMap, all three `ssl-*` keys in the controller ConfigMap, `replicas: 1`, the metrics Service and a correctly-labelled ServiceMonitor.

singhalkarun:
## Cluster-verified — with one operational catch

Deployed on a real cluster (kind on the devbox, via the actual backbone helmfile). The previously-inert config genuinely applies now.

**Headers land in a live response** (`HTTP/1.1 200`, through the controller):
```
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-XSS-Protection: 1; mode=block
Access-Control-Allow-Origin / -Methods / -Headers: all present
```

**TLS settings reach the generated nginx.conf** — the deepest available proof:
```
ssl_protocols TLSv1.2 TLSv1.3;
ssl_ciphers 'EECDH+CHACHA20:EECDH+AES';
ssl_ecdh_curve X25519:prime256v1:secp521r1:secp384r1;
```
`replicas: 1` from `replicaCount`, as intended.

The `<domain_name>` placeholder behaves exactly as the (corrected) description says: a **global** `sed` resolves it. A partial substitution that only touches `global.domain` leaves `Access-Control-Allow-Origin: https://<domain_name>` — confirmed by doing precisely that.

### Catch worth documenting: addHeaders changes need a controller restart

Timeline from the cluster:

| Time | Event |
|---|---|
| 11:39:14 | ConfigMap updated to the new origin |
| 11:41:13 | controller **still serving the old value** |
| after `rollout restart` | new value served |

The controller logged `RELOAD` for the main config, but the `add-headers` ConfigMap content is fixed at pod start. So a `helmfile apply` that changes only CORS/headers **reports success while the old headers keep serving** — the same silent-no-op shape this issue exists to fix, one level up. Worth a line in the runbook: restart the controller after changing `addHeaders`.

singhalkarun:

`0581892a` — added a `checksum/add-headers` pod annotation over the rendered headers, so helm rolls the controller whenever they change.

Re-verified end to end: changed `X-Frame-Options` from `SAMEORIGIN` to `DENY` in `env.yaml`, re-ran `helmfile sync`, and with **no manual restart**:

```
pod BEFORE: ingress-nginx-controller-5f656c6f-jkr4b     annotation 9bf024e3a1554ea0…
pod AFTER:  ingress-nginx-controller-6cbb447845-dr56g   annotation 0ce23e01f8f2f04c…
served:     X-Frame-Options: DENY
```

Previously the controller kept serving the old value indefinitely — it even logged a RELOAD for the main config while the custom headers stayed stale.

vinothrallapalli-eGov (approved):

PR #1680, opened 2026-08-07
feat(grafana): vendor dashboards in-repo instead of fetching from master (#1650)

singhalkarun:
Closes #1650

> **Stacked on #1665** (`fix/1650-grafana-auth`), which fixed the auth half. This is the dashboards half.

## What was wrong

`values/grafana.yaml` defined 8 dashboards by URL, every one pinned to `master`:

- **Not reproducible.** `master` is mutable. Two deployments a week apart could get different dashboards, and neither could be rebuilt as it was.
- **A deploy-time network dependency, and a supply-chain one.** An init container fetched and applied-as-config whatever those URLs returned at pod start.
- **Three of the eight came from an account outside this organisation** (`dotdc/grafana-dashboards-kubernetes`).
- **No review path.** Dashboard changes reached production without passing through this repo.

## What this does

All 8 are vendored under `charts/monitoring/dashboards/dashboards/`, captured from pinned upstream **commits** rather than `master`:

| Source | Commit |
|---|---|
| `dotdc/grafana-dashboards-kubernetes` | `7fc1dc8ccd23eeb8c58ddd8be2a122922d424937` |
| `egovernments/configs` | `6b8e739d26288ce60c77d9efe7985fda47c5b5d3` |

Per-file provenance is recorded in a README so they can be re-fetched or diffed against a newer upstream later. Content is **byte-for-byte upstream, deliberately not reformatted**, so a future diff shows only real changes rather than whitespace noise.

They render into ConfigMaps via a small local chart and are picked up by Grafana's dashboard sidecar, which this enables.

**One ConfigMap per dashboard, not one combined:** a ConfigMap caps at 1 MiB and these total ~480 KB. A single map would work today and fail on whoever adds the ninth — the kind of limit you want to be nowhere near.

## The orphan, resolved

The issue flagged `values/ingress.json` (59 KB, referenced by nothing) and asked whether it was the intended source or dead. It is **neither, quite**: it is the *same board* as the fetched ingress dashboard — identical `uid` (`k8s-nginx-ingress-prometheus-ng`) and all 16 panels, differing only in `title` and `description`.

So it was a duplicate rather than a lost original, and it is deleted. Git records it as a rename onto the vendored copy.

## Verification

- the local chart renders **8 ConfigMaps**, each carrying the `grafana_dashboard` label the sidecar selects on
- **Grafana 11.4.0 loads all 8** and rejects none — checked via `/api/search`, so this is Grafana's own verdict on the JSON, not mine
- the monitoring helmfile still builds under the pinned **helmfile v0.140.0**

## Deliberately not done

**No dashboard covers a DIGIT application service** — all eight are cluster/infra views. That depends on #1646 landing, since application metrics do not exist in this tier yet. When it does, the Ansible tier's JVM dashboard should be *reused* rather than a new one authored, because the metric names will match.

## Merge note

The new `dashboards` release is ungated. When **#1675** (per-component monitoring toggles) merges, it should be gated alongside `grafana` on `monitoring.dashboards`. Left ungated here only because that toggle block does not exist on this branch; deploying it unconditionally is harmless meanwhile, since they are inert ConfigMaps until a Grafana with the sidecar reads them.


vinothrallapalli-eGov (commented):
Verified the three headline claims from the devbox pass. (1) No runtime fetch remains: `values/grafana.yaml`'s `dashboards:` block is now `{}`, `extraInitContainers: []` is empty, and a repo-wide tree check at this SHA turns up no curl/wget/init-container pulling dashboards from a URL anywhere in `charts/monitoring/` — the sidecar (`sidecar.dashboards.enabled: true`) reads ConfigMaps instead. (2) ConfigMap/etcd size: measured all 8 vendored JSON files via the GitHub contents API — total is ~478 KB (largest single file, `k8s-views-nodes.json`, is ~102 KB), matching the PR's ~480 KB claim and comfortably under the 1 MiB etcd object limit per-ConfigMap and even combined. One-ConfigMap-per-dashboard is a reasonable margin choice. (3) Provenance/licensing: the new `dashboards/README.md` records source repo + pinned commit SHA + upstream path per file, so the three `dotdc/grafana-dashboards-kubernetes` boards (Apache-2.0) are attributed, not silently absent — though the actual Apache-2.0 license/notice text isn't vendored alongside them, which is a minor compliance nit worth a look, not a blocker. Also checked that the new `dashboards` helmfile release shares the same namespace expression as `grafana`, which matters because the sidecar's `searchNamespace` is `null` (same-namespace-only) — wiring is correct. No issues found that would block merge; flagging the license-text gap as a low-severity nit.

devops/deploy-as-code/charts/monitoring/dashboards/README.md
```diff
@@ -0,0 +1,50 @@
...
+`../values/grafana.yaml`). No network fetch at pod start.
+
+## Provenance
+
+Captured **2026-08-07** from these upstream commits, not from `master`, so the
+snapshot is attributable and re-fetchable:
+
+| Source repo | Commit |
+|---|---|
+| `dotdc/grafana-dashboards-kubernetes` | `7fc1dc8ccd23eeb8c58ddd8be2a122922d424937` |
```
vinothrallapalli-eGov, line 22: Minor/non-blocking: this table attributes the three `dotdc/grafana-dashboards-kubernetes` boards to their source repo and commit, which is good — but Apache-2.0 (dotdc's license) technically calls for retaining a copy of the license/notice when redistributing, not just a source link. Might be worth vendoring the upstream `LICENSE` alongside these three files (or linking to it explicitly here) so the attribution is self-contained rather than requiring a trip to GitHub. Not a blocker either way.

singhalkarun: Right on the licence terms, and cheap enough that "not a blocker" was too generous. Done in 9a8616d3.

Apache-2.0 §4(a) asks a redistributor to *give* recipients a copy of the licence, not to point at where one lives, so a source link left the attribution incomplete. Vendored the upstream `LICENSE` verbatim from the same pinned commit the boards came from (`7fc1dc8ccd23eeb8c58ddd8be2a122922d424937`) at:

```
devops/deploy-as-code/charts/monitoring/dashboards/third-party/LICENSE.dotdc-grafana-dashboards-kubernetes
```

Deliberately not named plain `LICENSE` at the chart root — that would read as *this* chart being Apache-2.0/David Calvert, which it is not.

Walked the rest of §4 while there so the README records the answers rather than leaving them to be re-derived:

- §4(b) — nothing to mark. The three files are byte-for-byte upstream.
- §4(c) — copyright holder named in the README: Copyright 2020 David Calvert.
- §4(d) — does not apply. Upstream ships no `NOTICE` file (`contents/NOTICE?ref=<pin>` → 404).

Also checked the other source, since the same question applies to it: `egovernments/configs` is MIT, same org and same licence as this repo, so the five boards from there carry no separate redistribution obligation. That is in the README too.

Rendering is untouched — the chart's `.Files.Glob "dashboards/*.json"` does not reach `third-party/`, so `helm template ./dashboards` still emits the same 8 ConfigMaps and `helm lint` passes.

singhalkarun:
## Cluster-verified: all 8 dashboards load — plus an orphaned provider that errors forever

On a real cluster the vendoring works. All 8 ConfigMaps render, the sidecar writes them, and Grafana's API confirms them loaded:

```
dashboards loaded: 8
  Blackbox Exporter (HTTP prober) · Kubernetes / Views / Global · ...Namespaces
  ...Nodes · ...Pods · Kubernetes Logs from Loki
  Kubernetes Nginx Ingress Dashboard · kubernetes-persistent-volumes
```

The sidecar's provider path (`/tmp/dashboards`) matches where it writes, so this does **not** repeat #1682's mount-path defect.

### But: a stale `dashboardProviders` entry survives

This PR removes the `dashboards:` URL block and leaves `dashboardProviders:` untouched (0 lines of the diff touch it). That provider points at a directory only the deleted URL-fetch mechanism ever created:

```yaml
- name: 'grafana-dashboards-kubernetes'
  folder: 'Kubernetes'
  options:
    path: /var/lib/grafana/dashboards/grafana-dashboards-kubernetes
```

Result — Grafana logs this **every 10 seconds, indefinitely**:

```
level=error msg="Cannot read directory" error="stat /var/lib/grafana/dashboards/grafana-dashboards-kubernetes: no such file or directory"
level=error msg="failed to search for dashboards"
```

Two consequences:

1. A permanent error loop in Grafana's logs, which is exactly the noise that trains people to ignore Grafana's logs.
2. **All 8 boards land in `General`, not a `Kubernetes` folder** — the folder was that provider's, and the sidecar provider does not set one. The chart's own comment says the two are a pair: *"To use dashboards you must also enable/configure dashboardProviders"*. Half the pair was removed.

Suggested fix: drop the `dashboardProviders` block along with the `dashboards` block, and if the `Kubernetes` folder grouping is wanted, set it on the sidecar provider (`sidecar.dashboards.folder` / the `grafana_folder` annotation) instead.

singhalkarun:

`d02fcb95` — removed the orphaned `dashboardProviders` entry and moved the folder onto the sidecar provider, which is what actually loads the ConfigMaps.

Re-verified on the same kind cluster the finding came from:

- **0** `Cannot read directory` lines in Grafana's log (was one every 10 seconds, indefinitely)
- all 8 boards now load into the **`Kubernetes`** folder instead of `General`

vinothrallapalli-eGov (approved):

PR #1681, opened 2026-08-07
docs(observability): add an alerting runbook (#1609, #1611)

singhalkarun:
Part of #1609 / #1611. Stacked on #1667.

## Why

The alerting PRs (#1672, #1673) deliver rules. Nothing told anyone what to **do** when one fires — which is the difference between alerting and a noisier dashboard.

## What it covers

Each of the six rules: what it means, why its threshold is what it is, and the first commands to run. Plus how to turn alerting on in each tier, and why every rule reports OK on missing data.

## Three things it records that were otherwise easy to lose

**1. The owner is undecided.** No team or channel has been named. The runbook says so plainly rather than implying coverage exists — every rule assumes a human eventually looks, and until one does, enabling alerting is incomplete.

**2. The Gatus/Grafana asymmetry**, because it bites. Gatus fails **open** on an unset webhook (warns, drops the provider, keeps serving); Grafana fails **closed** (exits 1, does not start). That is why the Compose default is an unreachable `.invalid` URL rather than empty — and the runbook says not to "tidy" it to empty, with the exact error to recognise if someone does.

**3. The tuning procedure for the paused error-spike rule** — observe p95 for ~a week on a real tenant, then set above it. Left implicit, someone would eventually unpause it with the placeholder threshold, and it would either cry wolf until the channel is muted (silently disabling every *other* rule for that person) or never fire.

## Also updates `enabling-monitoring.md`

Its Kubernetes section said the monitoring helmfile is commented out. #1675 changes that — the path is now wired in with each component gated on `monitoring.*`. The section now documents the toggles, a suggested order for picking a subset, and that `kafkaUi` is a **write-capable admin console with no authentication**, not observability.

## Verification

Checked mechanically rather than by eye: every rule title in `rules.yaml` appears in the runbook, and the only rule shipping paused is the one the runbook describes as paused.

### DIFF
--- a/docs/observability/alerting-runbook.md
+++ b/docs/observability/alerting-runbook.md
@@ -1,0 +1,159 @@
+# Alerting runbook
+
+What alerts exist, what each one means, and what to do when one fires.
+
+Alerting ships **off**. Nothing here sends anything until someone sets a webhook — see [Turning it on](#turning-it-on). That is deliberate: an alert channel nobody reads is worse than no alerting, because it looks like coverage.
+
+---
+
+## Who receives these
+
+**Undecided at the time of writing.** No team or channel has been named.
+
+That is a real gap, not a formality: every rule below assumes a human eventually looks. Until an owner exists, treat enabling alerting as incomplete, and record the decision here when it is made — otherwise the next person rediscovers the question rather than the answer.
+
+There is no channel name anywhere in the repo, and there should not be. The channel is chosen by whoever mints the webhook.
+
+---
+
+## Two engines, deliberately
+
+| | Gatus | Grafana |
+|---|---|---|
+| Answers | *is it up?* | *is it about to stop being up?* |
+| Source | HTTP/TCP checks | Prometheus + Loki |
+| Fires on | a service not responding | disk, memory, CPU, JVM heap, Kafka lag, error floods |
+
+**Gatus owns up/down. Grafana deliberately has no service-down rule.** Duplicating it would mean two messages per incident, which teaches people to ignore both. A test enforces this rather than trusting the comment.
+
+---
+
+## Turning it on
+
+The webhook URL is a **credential** — anyone holding it can post into the channel. It is read from the environment in every tier and never committed.
+
+**Compose**
+```bash
+# in .env
+GATUS_SLACK_WEBHOOK_URL=https://hooks.slack.com/services/...
+GRAFANA_SLACK_WEBHOOK_URL=https://hooks.slack.com/services/...
+```
+
+**Ansible** — set one value; Grafana falls back to the Gatus webhook, so a tenant picks a channel once:
+```yaml
+# host_vars/<tenant>.yml — prefer sourcing from OpenBao
+gatus_slack_webhook_url: "https://hooks.slack.com/services/..."
+```
+
+**k3s** — an optional Secret; create it to switch alerting on, no manifest edit:
+```bash
+kubectl -n digit create secret generic gatus-alerting \
+  --from-literal=slack-webhook-url='https://hooks.slack.com/services/...'
+```
+
+### Two behaviours that differ, and matter
+
+**Gatus fails open.** An unset webhook becomes an empty string, Gatus logs `Ignoring provider=slack due to error=webhook-url not set`, drops the provider and keeps serving. Verified at runtime.
+
+**Grafana fails closed.** It expands `$VAR` with no `${VAR:-default}` support, and treats provisioning it cannot validate as **fatal at startup**. An empty webhook makes it exit 1 with `recipient must be specified when using the Slack chat API` — it does not start.
+
+That is why the Compose default is an unreachable `https://example.invalid/...` URL rather than an empty string. **Do not "tidy" it to empty.** If you see Grafana crash-looping after touching alerting config, this is the first thing to check.
+
+---
+
+## Gatus alerts
+
+One alert per monitored endpoint, after **3 consecutive failures** (~90s at the 30s interval) — long enough that a single dropped check or a rolling restart does not page anyone. A recovery message follows.
+
+**When one fires:** the alert names the endpoint and group. Check the Gatus board first (`/status/`) to see whether it is one service or many — many at once usually means the box, not the service.
+
+---
+
+## Grafana alerts
+
+### Host disk low · `< 15% free on /` for 10m · warning
+
+Fine until it abruptly is not: a full root filesystem takes down every service at once, and Postgres in particular fails in ways needing manual recovery.
+
+Usual causes: Docker image/volume churn, and container logs if rotation is not capped.
+
+```bash
+df -h /
+docker system df
+du -sh /var/lib/docker/containers/* | sort -h | tail
+```
+
+### Host memory low · `< 10% available` for 5m · **critical**
+
+Critical rather than warning because **these boxes have no swap**. The kernel does not degrade gracefully under pressure — it goes straight to OOM-killing, and it kills whichever JVM is largest, not whichever is least important. There is no slow middle ground between "tight" and "a service died".
+
+```bash
+free -m
+docker stats --no-stream
+dmesg -T | grep -i 'killed process'   # confirm whether it has already happened
+```
+
+### Host CPU saturated · `> 90%` for 15m · warning
+
+Sustained, not a spike — 15 minutes rules out a deploy, a batch index or a JVM warming up. Past this point request latency is already degraded even though every health check still answers, so nothing else will tell you.
+
+### JVM heap headroom low · `> 90% of limit` for 10m · warning
+
+A service climbing toward its ceiling will eventually die and restart, dropping in-flight requests. Ten minutes above 90% means steady pressure rather than a GC sawtooth.
+
+**Reading this one carefully:** the ratio is against `jvm_memory_limit_bytes`, so a service that never publishes a heap limit produces **no series and cannot trigger the rule**. Absence of this alert is not proof of headroom — check the *DIGIT JVM Services* dashboard.
+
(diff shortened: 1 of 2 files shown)


docs/observability/alerting-runbook.md
```diff
@@ -0,0 +1,159 @@
...
+
+What alerts exist, what each one means, and what to do when one fires.
+
+Alerting ships **off**. Nothing here sends anything until someone sets a webhook — see [Turning it on](#turning-it-on). That is deliberate: an alert channel nobody reads is worse than no alerting, because it looks like coverage.
+
+---
+
+## Who receives these
+
+**Undecided at the time of writing.** No team or channel has been named.
```
vinothrallapalli-eGov, line 18: This accurately reflects reality (#1609/#1611 shipped without naming a channel or owner), but it's worth flagging explicitly: the tracking issue (#1601) calls the owner decision a hard blocker -- "an alert stream with no named owner gets muted within a week." This runbook is the artifact that was supposed to encode that decision once made, but as written it only documents the gap rather than closing it. Consider adding a concrete next step here (e.g. "file/link the issue that tracks naming an owner") so this doesn't quietly become the permanent state, especially once alerting actually gets turned on for a tenant.

docs/observability/alerting-runbook.md
```diff
@@ -0,0 +1,159 @@
...
+
+A service climbing toward its ceiling will eventually die and restart, dropping in-flight requests. Ten minutes above 90% means steady pressure rather than a GC sawtooth.
+
+**Reading this one carefully:** the ratio is against `jvm_memory_limit_bytes`, so a service that never publishes a heap limit produces **no series and cannot trigger the rule**. Absence of this alert is not proof of headroom — check the *DIGIT JVM Services* dashboard.
+
+### Kafka consumer lag high · `> 10k` for 15m · warning
+
+The failure Gatus structurally cannot see: `egov-persister` falls behind, the API keeps returning 200, every check stays green, and complaints are **never written to the database**.
+
+First question is *slow or gone*: the **Consumers per group** panel on the *DIGIT Kafka Consumer Lag* dashboard reading **0** means the consumer has died, which needs the opposite fix from a slow one. Lag concentrated on a single partition usually means a poison record, not an undersized consumer.
```
vinothrallapalli-eGov: The *DIGIT Kafka Consumer Lag* dashboard and its "Consumers per group" panel don't exist in the repo yet -- `local-setup/otel/grafana/provisioning/dashboards/` only has `jvm-services.json`, `node-exporter-full.json`, `loki-logs.json`, and `tempo-traces.json` (registered via `dashboards.yaml`). This matches the Kafka lag rule's own annotation in `rules.yaml` (#1673), so it's not a new inaccuracy introduced here, but issue #1623 ("Kafka/Redpanda has no metrics") is still open -- the dashboard this runbook sends an on-call responder to look at doesn't exist. Worth a note that the dashboard is pending #1623, or removing the specific panel reference until it ships, so the first documented triage step doesn't lead to a 404.

docs/observability/enabling-monitoring.md
```diff
@@ -72,51 +72,62 @@ At `metrics` and `logs`, `OTEL_TRACES_EXPORTER` is set to `none` — otherwise e
 
 ---
 
-## Kubernetes tier: it is off by default
+## Kubernetes tier: wired in, every component off by default
 
-`devops/deploy-as-code/digit-helmfile.yaml` lists the tiers to install. The monitoring line is commented out:
+`devops/deploy-as-code/digit-helmfile.yaml` lists the tiers to install. The monitoring line **used to be commented out entirely**; as of #1675 it is present, and each component is gated instead:
```
vinothrallapalli-eGov: This is written in present tense ("as of #1675 it is present") but #1675 is still an open PR, not merged -- on current `master` the helmfile line is still commented out and `charts/environments/env.yaml` has no `monitoring:` toggle block (verified: the `monitoring:`/`kafkaUi` keys described below at lines 96-105 don't exist in `env.yaml` yet, though #1675's diff does add them in the shape described). If #1681 merges before #1675, this section will describe a Kubernetes config that doesn't actually exist in the tree yet. Worth either noting the dependency explicitly (similar to how the PR description calls out "stacked on #1667") or holding this section until #1675 lands.

vinothrallapalli-eGov (commented):
Cross-checked the six alert descriptions (names, thresholds, durations, severities) against `rules.yaml` in #1673 and the Gatus/Grafana webhook plumbing against #1672 -- all match exactly, including the `for:` windows, the 15%/10%/90%/90%/10k thresholds, and the Compose `.invalid` fallback rationale. Two things worth a look before merge: the runbook faithfully records that no triage owner/channel has been named yet (per the tracking issue's own warning about unowned alert streams getting muted), rather than resolving it -- which is honest but means the artifact doesn't yet close the gap #1609 flagged as a blocker. Also flagged a dashboard reference that doesn't exist yet, and a Kubernetes docs section written in present tense about a helmfile/env.yaml state that depends on #1675, which is still open.

docs/observability/alerting-runbook.md
singhalkarun: Agreed, and closed in d1b09e51. You are right that documenting the gap was not the
job this page was supposed to do.

It turns out no new issue is needed — #1601 already designates where the decision
gets made, and I had not read it closely enough to notice:

> **#1609 needs two decisions before it ships:** which Slack channel and webhook, and
> who triages per tenant. An alert stream with no named owner gets muted within a
> week. #1611 reuses the same channel — decide once, in #1609.

So the section now quotes that, names **#1609** as the concrete next step, and gives
the answer somewhere to land rather than leaving it as prose:

```
- [ ] Owner / triage rota decided in #1609, then written into the table below.

| Tier                 | Channel | Owner  | Decided in |
| Compose              | unset   | unset  | #1609      |
| Ansible (per tenant) | unset   | unset  | #1609      |
| k3s                  | unset   | unset  | #1609      |
```

Per-tier because the three wire up independently and a tenant box may well be someone
else's problem than the cluster.

And the sentence that makes it a gate rather than a note, which is your "especially
once alerting actually gets turned on" point: **do not enable alerting for a tenant
while those rows are empty** — an unowned stream gets muted, and a muted channel
disables every rule on this page, not just the one that was noisy.

Only the *owner* column belongs in the repo; the channel stays out, chosen by whoever
mints the webhook. That part of the original text was right and is kept.

docs/observability/alerting-runbook.md
singhalkarun: Confirmed, and it is worse than a dead link — fixed in d1b09e51.

Your inventory is right (`dashboards/` has `jvm-services`, `node-exporter-full`,
`loki-logs`, `tempo-traces`; `pgr-analytics` arrives with #1634). Chasing the
dependency in #1673's own `rules.yaml` comments turned up two unmerged PRs, not one:

- the **DIGIT Kafka Consumer Lag** dashboard is **#1678**
- the `redpanda_*` series the rule queries come from the Redpanda scrape job in
  **#1628** (#1678 is stacked on it)

both under issue #1623. So until **#1628** lands, the rule evaluates an empty vector
and — via the file-wide `noDataState: OK` — reports a healthy **ok** that is
indistinguishable from "lag is fine" when it means *nothing is being measured*. A
green Kafka rule is currently not evidence the persister is keeping up. That seemed
more important to say than the missing panel, so the section leads with it, and
explains why the rule ships unpaused anyway (so it starts working the moment the
scrape job appears, rather than waiting on someone remembering to unpause it).

Rather than delete the panel reference I gave the responder something that works
today — the same call `digit-mcp`'s `kafka_lag` tool already makes:

```bash
docker exec digit-redpanda rpk group describe egov-persister
```

`LAG` per partition, empty member list means gone. That answers *slow or gone*
without the board.

Generalised it too, since the Kafka rule is only the loudest instance: the
"Why every rule reports OK on missing data" section now states the cost of that
choice — an `ok` means healthy **or** not measured, and the channel cannot tell you
which — so confirm a rule has series behind it before reading its silence as
reassurance.

docs/observability/enabling-monitoring.md
singhalkarun: Confirmed and fixed in d1b09e51. Verified as you describe: on `monitoring-fix` the
helmfile line is still commented and `charts/environments/env.yaml` has no
`monitoring:` block; #1675's diff adds both in exactly the shape documented here.

Holding the section back would leave the page describing the *pre*-#1675 world at the
moment #1675 lands, which is the same failure in the other direction — and since
#1675 and #1681 both target `monitoring-fix` and neither depends on the other, there
is no merge order that avoids a window. So I noted the dependency at the section head
and said which way the window cuts:

> **This section describes #1675, which has not merged.** On `monitoring-fix` today the
> helmfile line is still commented out and `charts/environments/env.yaml` has no
> `monitoring:` block […] whichever lands first, expect a window where the tree and
> this page disagree. Nothing here is wrong about the intended shape — it is the timing
> that is unsettled.

Three related things while I was in there:

1. The **branch-dependency table** #1667 gained at the top of `enabling-monitoring.md`
   now carries #1675 and #1673 explicitly, so this is not the only place a reader
   meets it.
2. The runbook needed the same treatment and had none: `rules.yaml`,
   `contactpoints.yaml`, `policies.yaml` and `grafana-alerting.test.ts` live on
   **#1673**, stacked on **#1609** — a different branch chain from this one, so they
   are in no tree this PR touches. Every threshold, `for`, `isPaused` and
   `noDataState` quoted on the page was read off #1673's `rules.yaml`, and the page
   now says so.
3. The two-tier table's "off by default — one commented line" was left over from
   #1667's version and would have been stale the moment #1675 landed; now "six
   per-component booleans, all `false`".

vinothrallapalli-eGov (approved):

PR #1682, opened 2026-08-07
feat(k3s): add the monitoring stack, phases 1 and 2 (#1618)

singhalkarun:
Part of #1618 — phases 1 and 2 of 4.

## The problem

Until now the **only** observability manifest in `local-setup/k8s/` was `tools/gatus.yaml`. No Prometheus, Grafana, Loki, Promtail, Tempo, otel-collector or node-exporter.

That was confusing rather than merely incomplete: the Gatus config is kept byte-identical across tiers *with CI enforcing it* — signalling "these tiers should match" — while every other observability component silently did not exist here.

## What this adds

| Phase | State |
|---|---|
| 1 · Metrics — Prometheus, node-exporter DaemonSet, Grafana | **done** |
| 2 · Logs — Loki, Promtail DaemonSet | **done** |
| 3 · Traces | not done — see below |
| 4 · Gatus wire-up | blocked — see below |

Off by default behind `MONITORING=true`, and never in CI: that run asserts a fixed set of resources come up healthy, and five more would slow it and broaden what can flake without testing anything CI exists to test.

Config is generated from `local-setup/otel/` — the same files the compose tier uses — so the tiers cannot drift. Two deliberate exceptions, both annotated inline: **Promtail's discovery half is rewritten** (k3s runs containerd; there is no Docker socket), and **Prometheus adds a kubelet/cadvisor job** because container CPU/memory comes from the kubelet here, not a Docker daemon. Job *names* are preserved so existing dashboards resolve unchanged.

## Validated by running it, not by dry-run

I stood up a real k3s cluster and deployed against it. That is the only reason the following were caught — every one of them is silent or misleading on paper:

**1. `--web.enable-lifecycle=false` crash-loops Prometheus.** The flag parser rejects the explicit `false` with `unexpected false`. Omission is the only correct spelling — and the default is already off, which is what #1608 wants anyway.

**2. `mountPropagation: HostToContainer` on `/` makes the kubelet refuse to create the node-exporter container** wherever `/` is not a shared or slave mount — notably k3s inside Docker, which is how this tier is developed locally:
```
path "/" is mounted on "/" but it is not a shared or slave mount
```
Dropped. A node-exporter that cannot start reports nothing; one that misses a later-mounted filesystem reports slightly less.

**3. Grafana loaded ZERO dashboards while appearing perfectly healthy** — `Running`, `1/1`, health endpoint 200 — because the JSONs were mounted outside the provider's path. They are now mounted as *subdirectories of* that path, which the file provider walks recursively, so `dashboards.yaml` is reused verbatim rather than forked for this tier.

**4. The 456 KB node-exporter dashboard cannot be applied client-side at all.** This one the issue predicted but under-diagnosed. `kubectl apply` copies the whole object into the `last-applied-configuration` **annotation**, capped at **256 KiB**:
```
ConfigMap "grafana-dash-node" is invalid: metadata.annotations:
Too long: must have at most 262144 bytes
```
It fails well below the ~1 MiB *object* limit — so one-ConfigMap-per-dashboard (#1618 item 4) does **not** avoid it: the largest single dashboard already exceeds the annotation cap on its own. `Tiltfile.k8s` applies that one **server-side**, which does not use the annotation.

## Evidence from the cluster

- all **5 pods Running, 0 restarts**, from a clean from-scratch apply
- Prometheus targets `node` and `kubelet-cadvisor` **up**, with `instance` relabelled to the **node name** (not the ephemeral pod, which would break dashboard continuity on every reschedule)
- node-exporter serving **73** host metric series
- Loki receiving logs with **`service_name` populated** — `[coredns, grafana, local-path-provisioner, loki, node-exporter, prometheus, promtail]`. That label is the *only* one `loki-logs.json` selects on, so getting it wrong would leave the dashboard silently empty on this tier
- Grafana loading **all 4 dashboards**, provisioning **all 3 datasources**, and **refusing anonymous access with 401** — this tier does not reintroduce what #1602 closed
- `check-gatus-coverage.py` still passes; static tests 35/35

## Deliberately not done

**Phase 3 (traces).** Deploying a collector alone yields empty dashboards: `grep OTEL_ k8s/{core,app}-services/*.yaml` returns nothing, so no service in this tier emits telemetry. Making it real means per-service OTEL env across ~20 Deployments, including the compose tier's deliberate exceptions (`egov-localization` metrics-only to avoid ~200 MB of agent heap; OTP/SMS off entirely). That is the same work as #1646 and should follow the same decisions rather than be guessed at here. The `otel-collector` scrape target correspondingly shows **down** — honest rather than silently absent.

**Phase 4 (Gatus wire-up).** `GATUS_OBSERVABILITY` is introduced by #1613 (PR #1636), not merged. Flagged for whoever does it: **the endpoints will need namespaced DNS** — Gatus runs in `digit`, this stack in `monitoring`, so `http://grafana:3000` does not resolve; it needs `grafana.monitoring.svc.cluster.local`.

## Before enabling

```
kubectl -n monitoring create secret generic grafana-admin \
  --from-literal=password='<choose one>'
```

### DIFF
--- a/local-setup/Tiltfile.k8s
+++ b/local-setup/Tiltfile.k8s
@@ -15,6 +15,15 @@ PGR_PATH = CCRS_PATH + '/backend/pgr-services'
 # CI mode: use pre-built Docker Hub images, no live reload
 CI_MODE = os.getenv('CI', '') != '' or os.getenv('TILT_CI', '') != ''
 
+# Monitoring stack (#1618). OFF by default and opt-in via MONITORING=true, for
+# the same reason the production tier's components default to installed:false --
+# Prometheus, Grafana, Loki and a Promtail DaemonSet are a real resource cost,
+# and turning them on for every `tilt up` would change what a plain local
+# bring-up means. Never on in CI: the tier's CI run asserts a fixed set of
+# resources come up healthy, and adding five more would slow it and broaden
+# what can flake without testing anything CI is there to test.
+MONITORING = os.getenv('MONITORING', '').lower() in ('1', 'true', 'yes') and not CI_MODE
+
 # ==================== ConfigMaps from files ====================
 # Generated dynamically so edits to source files auto-apply on tilt up
 # NB: the directory is configs/egov-persister, not configs/persister. 09a8ec2e renamed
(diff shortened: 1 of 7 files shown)


vinothrallapalli-eGov (commented):
Reviewed against the three focus points. (1) Auth bar: grafana.yaml sets GF_AUTH_ANONYMOUS_ENABLED=false, GF_AUTH_DISABLE_LOGIN_FORM=false, and pulls GF_SECURITY_ADMIN_PASSWORD from a required Secret (no literal default) -- matches the bar #1605 set on compose and #1665 set on the production k8s tier; the pod fails closed (CreateContainerConfigError) rather than falling back to an open default if the secret is missing. Good parity, not a weaker link. (2) Tracing scope: confirmed no Jaeger config exists anywhere the diff touches -- the only Jaeger file in the repo is devops/deploy-as-code/charts/monitoring/values/jaeger.yaml, which #1619 owns and this PR does not go near. The Grafana Tempo datasource is provisioned only because it's carried over verbatim from the shared local-setup/otel/grafana/provisioning/ directory (also used by compose); no Tempo or otel-collector workload is deployed, and the PR/README are explicit that the otel-collector scrape target will show 'down' until phase 3. This does not prejudge #1619's OTLP/Tempo-vs-Jaeger decision. (3) Resources: all five new workloads (prometheus, node-exporter, grafana, loki, promtail) set both CPU/memory requests and a memory limit; nothing is unbounded. Total footprint is modest (~600Mi requested / ~2.4Gi memory limit ceiling across all five), reasonable for an opt-in (MONITORING=true, never-in-CI) local dev stack on a constrained k3s node -- no OOM regression of the kind seen in #1612/#1635/#1666. No concrete defects found; nothing inline to anchor.

vinothrallapalli-eGov (approved):

Issue #1683, opened 2026-08-07
Config changes do not reach running containers: rsync's atomic write orphans single-file bind mounts (Kong runs stale policy)


singhalkarun:
Found while verifying #1607 on the eGov devbox. **P1 · effort: S**

## Symptom

After a successful deploy, Kong was still serving routes that had been removed from `kong.yml`:

```
host      /opt/digit/kong/kong.yml   Aug  7 13:21   md5 2c8578b7…   grafana|tempo services: 0
container /kong/kong.yml             Jul 27 10:02   md5 4311e54f…   grafana|tempo services: 2
```

The container was reading an **11-day-old copy**. Kong's admin API confirmed `grafana-service` and `tempo-service` were still loaded and proxying (200), while the file on disk no longer declared them.

`kong reload` did **not** help — it reported `Kong reloaded` and kept serving the old routes, because it re-read the same stale in-container file. Only `docker compose up -d --force-recreate kong` fixed it.

## Cause

`playbook-deploy.yml`'s **Synchronize config directories** task uses `synchronize` (rsync), and **rsync writes atomically by default**: temp file, then rename. That replaces the inode.

A **single-file** bind mount (`/opt/digit/kong/kong.yml → /kong/kong.yml`) is bound to the *original inode*. Once it is replaced, the container keeps reading the old one — permanently, across reloads, until the container is recreated.

Proven in isolation on the devbox:

| Write method | Host | Container sees |
|---|---|---|
| in-place (`>` redirect) | `v2-inplace` | `v2-inplace` ✅ |
| **atomic replace (temp + rename)** | `v3-atomic` | **`v2-inplace`** ❌ |
| after container recreate | `v3-atomic` | `v3-atomic` ✅ |

And with rsync specifically, from a clean mount:

| rsync mode | Host inode | Container sees update |
|---|---|---|
| default | changed | **no** |
| `--inplace` | **unchanged** (6665 → 6665) | **yes** |

## Why this is P1 rather than cosmetic

It is a **silent-success** failure. The file on disk is correct, the playbook is green, and `kong reload` reports success — while the gateway runs the old policy indefinitely. Nothing surfaces the mismatch.

That matters most for `kong.yml`, because it carries the **auth and RBAC policy**. A tightened rule can appear deployed and simply not be in force. The RBAC work in #1255 depends on being able to trust that a `kong.yml` change reaches the gateway.

## Scope

[...]

PR #1684, opened 2026-08-07
fix(ansible): make config changes actually reach running containers (#1683)

singhalkarun:
Closes #1683

Found while verifying #1607 on the devbox — the fix was correct but had no effect on a running deployment.

## The bug

After a green deploy, Kong was still serving routes deleted from `kong.yml`. The container was reading an **11-day-old copy**, and `kong reload` reported success while changing nothing, because it re-read that same stale file.

**Cause:** the config-dir sync uses rsync, which writes a temp file and renames it over the destination — **replacing the inode**. Several of these files are bind-mounted into containers *individually* (`kong.yml`, `gatus/config.yaml`, `otel-collector-config.yaml`, `prometheus.yml`, the nginx `*-proxy.conf` files), and a single-file bind mount is bound to the **inode, not the path**. Once rsync swaps it, the container keeps reading the old file forever.

Proven in isolation rather than inferred:

| Write method | Container sees |
|---|---|
| in-place (`>` redirect) | new content ✅ |
| **atomic (temp + rename)** | **old content** ❌ |
| after container recreate | new content ✅ |

| rsync mode | Host inode | Container sees update |
|---|---|---|
| default | changed | **no** |
| `--inplace` | **unchanged** (6665 → 6665) | **yes** |

## Why it matters

It's a **silent-success** failure — file on disk correct, playbook green, reload reporting OK, gateway running the old policy. Worst for `kong.yml`, which carries the **auth and RBAC policy**: a tightened rule can appear deployed and simply not be in force. That's precisely what the RBAC work in #1255 needs to be able to trust.

Most services self-heal only because they happen to be recreated by the "Recreate services with the new env" task. **Nothing recreates Kong when only its config changed.**

## The change

1. **`--inplace`** on the config sync, so updates keep the inode and reach running containers.
   **`delay_updates: false` is required with it** — ansible's `synchronize` adds `--delay-updates` by default and rsync rejects the combination outright (`--inplace cannot be used with --delay-updates`), failing every item and aborting the deploy at that task. I hit exactly that on the first run; it's why this needs a real deploy to validate, not just review.
2. **Reload Kong when `kong.yml` changed**, so the policy is applied rather than merely present. `kong reload` rather than a recreate — it's Kong's documented zero-downtime path, and it only works because `--inplace` keeps the mounted file current.

## Trade-off, stated plainly

`--inplace` writes directly into the destination, so an interrupted transfer can leave a **partially-written config** where the default would have left the previous file intact. For small text configs written by a deploy that health-checks afterwards, a config that is actually applied beats one that is atomically stale. But it is a real trade, not a free win.

## Verified end to end on the devbox

Renamed a live Kong route in source and deployed:

- in-container file updated with **no recreate** — Kong stayed `Up 22 minutes`
- host and container copies **identical** afterwards
- old path → **404**, new path → served
- real API routes (`/egov-mdms-service/health`) → **200** across the reload
- the new task fires **only on change** (`changed: [karun]`)
- full run **ok=126, changed=20, failed=0**; 43 containers, 0 unhealthy
- #1605 / #1606 / #1607 / #1625 all still holding afterwards

## Note for anyone auditing this themselves

Several images are distroless, so `docker exec md5sum` fails and its error text can be mistaken for a hash mismatch — that produced a false positive on my first sweep. Compare with `docker cp` instead.

### DIFF
--- a/local-setup/ansible/playbook-deploy.yml
+++ b/local-setup/ansible/playbook-deploy.yml
@@ -1882,6 +1882,51 @@
         | list | length > 0
       changed_when: true
 
+    # Same failure mode as Kong, one hop further down (#1683). The OpenTelemetry
+    # Collector reads --config once at startup and does not watch the file, and
+    # nothing else recreates it on a routine redeploy: unlike gatus/grafana it
+    # declares no `environment:` at all, so the ".env changed" recreate above
+    # never selects it. Result: otel-collector-config.yaml now correctly lands
+    # on disk and inside the container -- and is still not applied.
+    #
+    # Gatus is deliberately NOT restarted here. It watches its config file and
+    # reloads itself, which the --inplace sync above is what makes possible: the
+    # watcher follows the mounted inode, so the old rename-over-the-top swap was
+    # invisible to it. A restart would also drop its uptime history for no gain.
+    # prometheus.yml is handled separately by its own SIGHUP reload (#1608).
+    #
+    # A restart, not a reload: the collector has no reload signal or admin
+    # endpoint. It is a few seconds of dropped telemetry, which is why the gate
+    # below is on the ONE file that matters rather than on the otel/ directory
+    # as a whole -- that directory also holds prometheus.yml, the Loki/Tempo/
+    # Promtail configs, the javaagent jar and every Grafana dashboard JSON, so a
+    # directory-level gate would bounce the collector on any dashboard tweak.
+    #
+    # synchronize's per-file itemized output is what makes the narrow gate
+    # possible: stdout_lines carries one rsync `>f.st...... <name>` line per
+    # changed file, and is empty when the dir is unchanged. Verified locally
+    # against ansible.posix.synchronize 2.2.2 + rsync 3.4.4: editing only
+    # prometheus.yml leaves this gate false while the item still reports
+    # changed=true; editing otel-collector-config.yaml flips it true.
+    - name: "OTEL Collector — restart when its config changed (#1683)"
+      ansible.builtin.shell: |
+        set -e
+        if [ -z "$(docker ps -q -f name='^digit-otel-collector$')" ]; then
+          echo "NOCHANGE collector not running"
+          exit 0
+        fi
+        docker restart digit-otel-collector
+      args:
+        executable: /bin/bash
+      register: otel_collector_restart
+      changed_when: "'NOCHANGE' not in otel_collector_restart.stdout"
+      when: >-
+        config_sync.results | default([])
+        | selectattr('item.key', 'equalto', 'otel')
+        | map(attribute='stdout_lines') | flatten
+        | select('search', 'otel-collector-config')
+        | list | length > 0
+
     # ==================== Wait for Persister (Kafka consumer) ====================
     - name: Wait for egov-persister to be healthy
       shell: docker inspect --format '{{ "{{" }}.State.Health.Status{{ "}}" }}' egov-persister
--- a/local-setup/ansible/playbook-deploy.yml
+++ b/local-setup/ansible/playbook-deploy.yml
@@ -1200,6 +1200,44 @@
       synchronize:
         src: "{{ item.value }}/"
         dest: "{{ digit_dir }}/{{ item.key }}/"
+        # --inplace is load-bearing (#1683), not a performance tweak.
+        #
+        # rsync's default is to write a temp file and rename it over the
+        # destination. That REPLACES THE INODE. Several of these files are
+        # bind-mounted into containers individually (kong.yml, gatus/config.yaml,
+        # otel-collector-config.yaml, prometheus.yml, the nginx *-proxy.conf
+        # files), and a single-file bind mount is bound to the inode it was
+        # created with -- not to the path. Once rsync swaps the inode, the
+        # running container keeps reading the OLD file forever.
+        #
+        # This is a silent-success failure: the file on disk is correct, the
+        # playbook is green, and `kong reload` even reports success, while the
+        # gateway carries on serving the previous policy. It was found with Kong
+        # running an 11-day-old kong.yml -- still proxying routes that had been
+        # deleted -- because nothing recreates Kong when only kong.yml changes.
+        # Most other services self-heal only because they happen to be recreated
+        # by the "Recreate services with the new env" task below.
+        #
+        # --inplace writes into the existing file, so the inode survives and the
+        # container sees the update. Verified on a real box: default rsync
+        # changed the inode and the container never saw the new content;
+        # --inplace kept it (6665 -> 6665) and the container saw it immediately.
+        #
+        # The trade: an interrupted transfer can leave a partially-written
+        # config where the default would have left the previous file intact. For
+        # small text configs written by a deploy that health-checks afterwards,
+        # a config that is actually applied beats one that is atomically stale.
+        #
+        # delay_updates MUST be false alongside it. ansible's synchronize adds
+        # --delay-updates by default, and rsync refuses the combination outright:
+        #   "rsync: --inplace cannot be used with --delay-updates"
+        # which fails every item in this loop and aborts the deploy at this task.
+        # (--delay-updates is the very batch-then-rename behaviour we are
+        # removing, so turning it off is the point, not a workaround.)
+        delay_updates: false
+        rsync_opts:
+          - "--inplace"
+      register: config_sync
       loop: "{{ digit_config_dirs | dict2items }}"
       vars:
         digit_config_dirs:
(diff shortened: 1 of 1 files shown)


local-setup/ansible/playbook-deploy.yml
```diff
@@ -1200,6 +1200,44 @@
...
+        # a config that is actually applied beats one that is atomically stale.
+        #
+        # delay_updates MUST be false alongside it. ansible's synchronize adds
+        # --delay-updates by default, and rsync refuses the combination outright:
+        #   "rsync: --inplace cannot be used with --delay-updates"
+        # which fails every item in this loop and aborts the deploy at this task.
+        # (--delay-updates is the very batch-then-rename behaviour we are
+        # removing, so turning it off is the point, not a workaround.)
+        delay_updates: false
+        rsync_opts:
```
vinothrallapalli-eGov, line 1238: `--inplace` (and the loss of the atomic-write safety net) is applied to every dir in `digit_config_dirs`, including `db/` (full-dump.sql, ~833KB, not bind-mounted into any running container per docker-compose.egov-digit.yaml) and `postman/`. For those, there's no bind-mount-inode reason to take on the partial-write risk described above — an interrupted transfer there just risks a corrupt SQL dump with no upside. Worth considering scoping `--inplace` to only the dirs that actually back single-file bind mounts (kong, gatus, otel, nginx), or confirming that whatever consumes db/full-dump.sql tolerates/detects a truncated file.

local-setup/ansible/playbook-deploy.yml
```diff
@@ -1820,6 +1858,30 @@
...
+    # Without this, a tightened auth or RBAC rule silently does not take effect
+    # -- the file is right, the deploy is green, and the gateway keeps serving
+    # the previous policy (#1683).
+    #
+    # `kong reload` is used rather than a container recreate because it is
+    # Kong's documented zero-downtime path: workers are replaced gracefully, so
+    # in-flight API calls are not dropped. It only works because the --inplace
+    # sync above keeps the mounted file current; with the default rsync the
+    # reload re-reads a stale inode and reports success while changing nothing.
+    - name: "Kong — reload declarative config when kong.yml changed (#1683)"
```
vinothrallapalli-eGov, line 1875: The PR's own diagnosis (and the comment block above at 1207-1209) names `gatus/config.yaml` and `otel-collector-config.yaml` as being affected by the same inode-swap bug as `kong.yml`, but only Kong gets an explicit reload task here. Gatus has built-in file-watch/auto-reload so it likely self-heals now that `--inplace` keeps the inode stable, but the OpenTelemetry Collector (`egovio/opentelemetry-collector-contrib`) does not hot-reload its config file by default — and nothing else recreates it on a routine redeploy (no `.env`-driven env vars, so the generic `up -d` recreate never touches it). If that's right, `otel-collector-config.yaml` changes will now correctly land on disk/in the container but still silently fail to apply until the container is restarted — the same failure mode this PR fixes for Kong, just one hop further down. Worth confirming, and adding a `docker restart digit-otel-collector` (or equivalent) gated on `config_sync` changes for the otel item if so.

vinothrallapalli-eGov (commented):
Verified the core claim: `--inplace` + `delay_updates: false` are set once on the looped `synchronize` task (playbook-deploy.yml:1199-1240), so they apply uniformly to every dir in `digit_config_dirs` (otel, configs, nginx, kong, postman, db, seeds, gatus, jupyter, scripts, keycloak) — that covers all the individually bind-mounted single files I could find in docker-compose.egov-digit.yaml (kong.yml, gatus/config.yaml, otel-collector-config.yaml, prometheus.yml, the nginx *-proxy.conf files, realm-template.json, DataLoader_v2.ipynb, user-seed.sh). No other rsync task in the playbook targets a live single-file bind mount, so the `--inplace` fix isn't missing anywhere it's needed on the rsync side. Two gaps worth a look (neither blocking, both medium-ish): (1) the new Kong reload has no counterpart for otel-collector, which the PR's own description names as equally affected but which AFAIK has no built-in hot-reload of its config file — the fix now lands correct bytes at the right inode, but nothing tells the collector process to re-read them, so it may stay silently stale exactly like Kong did. (2) `--inplace` is applied to the whole loop including db/ (an 833KB full-dump.sql, not bind-mounted at all) and postman/, trading the atomic-write safety net for zero benefit there. Also flagging, outside this diff, that `local-setup/ansible/playbook-deploy.yml`'s "Add API proxy to digit-ui nginx config" task writes `nginx/digit-ui.conf` via `copy:` (same atomic-rename semantics as default rsync) straight into a path that's individually bind-mounted into the digit-ui container — dormant today only because that content block has no Jinja variables, but the identical bug class the moment it does.

local-setup/ansible/playbook-deploy.yml
singhalkarun: Checked this one carefully and the premise does not hold for `db/`, so I have left the loop as it is — but the reason is worth spelling out, because it is invisible from `docker-compose.egov-digit.yaml` alone.

Ansible does not deploy that file on its own. `compose_files` (L1476) always layers `egov-digit.yaml + migrations.yml + monitoring.yml`, plus `fast-path.yml` when `db_fast_path` and a per-tenant overlay if present. Across that stack `db/` **does** back single-file bind mounts:

- `docker-compose.migrations.yml:34` → `./db/flyway-history-map.yml:/map.yml:ro` — and migrations.yml is layered *unconditionally* ("it is the migration model, not an opt-in extra")
- `docker-compose.fast-path.yml:49,53` → `./db/full-dump.sql`, `./db/keycloak-init.sql`

So `db/` is in the same category as kong/gatus/otel, not outside it. Sweeping the whole stack, the dirs with genuinely **no** compose reference at all are `postman/` and `scripts/`; `configs/` is mounted as *directories* (`./configs/egov-persister:/persister-config:ro`), which resolve by path and are immune to the inode swap either way. Scoping the loop would therefore mean splitting it to exclude two unreferenced dirs — a second task and a second convention for no behavioural difference.

On the truncated-dump risk itself: it self-heals on the next sync rather than persisting. rsync sets the destination's mtime only *after* the transfer completes, so an interrupted `--inplace` leaves a size/mtime that fails the next run's quick-check. Verified locally against ansible.posix.synchronize 2.2.2 + rsync 3.4.4, by truncating the destination and restoring an old mtime to simulate the interruption:

```
dest before : 15 bytes (truncated)
next run    : changed=True   '>f.st...... full-dump.sql'
dest after  : dump v2 much longer content here      (complete)
```

And the residual exposure is narrower than "a corrupt SQL dump": `full-dump.sql` is only read by Postgres' `/docker-entrypoint-initdb.d/`, which runs **on a fresh volume only**. It would take an interrupted deploy *followed by* a from-scratch Postgres init *before* the next deploy for it to bite.

Worth adding that the same self-healing is what bounds the risk you describe in the comment block for every other dir too, so I have not tried to argue the trade-off is free — just that it is smaller than one deploy's window.

local-setup/ansible/playbook-deploy.yml
singhalkarun: You are right, and it is fixed in e186ea5d. This was the real gap in the PR — the fix stopped one hop short of its own diagnosis.

Confirmed on the mechanism: the collector reads `--config` once at startup and has no reload signal or admin endpoint, and it is the one config-mounted service in this file with **no `environment:` block at all**, so the ".env changed → recreate" task never selects it either. Gatus, by contrast, does have `${GATUS_PROFILE_*}` env, so it gets recreated whenever those move — on top of its own file-watch. Net: `otel-collector-config.yaml` was landing on disk and inside the container correctly, and still never being applied.

Added a `docker restart digit-otel-collector`, gated on **that one file** rather than on the `otel/` directory. That distinction matters here: `otel/` also holds `prometheus.yml`, the Loki/Tempo/Promtail configs, the javaagent jar and every Grafana dashboard JSON, so a directory-level gate would bounce the collector on any dashboard tweak. `synchronize`'s per-file itemized `stdout_lines` makes the narrow gate possible.

Since a Jinja expression evaluating correctly is not the same as the task firing correctly, I exercised the task itself — real ansible.posix.synchronize 2.2.2 + rsync 3.4.4, stubbed docker:

| scenario | result |
|---|---|
| nothing changed | **skipped** |
| only `prometheus.yml` changed | **skipped** (item still reports `changed=true`) |
| `otel-collector-config.yaml` changed | **changed** — restart issued |
| config changed, container not running | **ok** — no-op, no failure |

Same run also confirms the `--inplace` premise directly: inode `109227952` survived a full content rewrite.

Two deliberate omissions, both matching your read:

- **Gatus** is left alone. It watches its config and reloads itself, and `--inplace` is precisely what makes that work — the watcher follows the mounted inode, so the old rename-over-the-top swap was invisible to it. Restarting it would also drop its uptime history for no gain. *Flagging this as the one thing I have not verified at runtime* — I have not deployed to the devbox (other agents are on it), so "Gatus self-heals" rests on its documented file-watch rather than an observed reload.
- **`prometheus.yml`** has the same shape but is already handled by its own SIGHUP reload in #1608 / #1625, so adding anything here would collide with it.

vinothrallapalli-eGov (approved):

PR #1687, opened 2026-08-10
fix(observability): three defects found deploying #1666 (teardown, profiles, Loki wait)

singhalkarun:
Closes the three defects found deploying #1666 to the devbox. Stacked on `feat/1657-modular-observability`.

## 1 · Lowering the level freed no memory

`docker compose up` only declines to **start** services outside the active profiles — it never **stops** running ones. So `observability_level: metrics` flipped the exporter vars and left tempo, loki and promtail up:

```
COMPOSE_PROFILES=mcp,obs-metrics docker compose <files> up -d
  tempo before: Up 9 minutes (healthy)
  tempo after:  Up 9 minutes (healthy)
```

`--remove-orphans` does not help — they are declared and merely unselected, which compose treats as "leave alone".

On any existing box the downgrade released **nothing** — ~1 GB stayed held — while the config said metrics-only and the deploy reported success.

**Fixed:** a teardown task that runs after the stack is up, names the services of inactive tiers, and only acts when those containers exist.

## 2 · Every ad-hoc `docker compose` command was broken

Twelve unprofiled core services `depends_on: otel-collector`, so profiling it invalidated the whole project:

```
service "egov-filestore" depends on undefined service "otel-collector": invalid compose project
```

| Command (no `COMPOSE_PROFILES`) | before | after |
|---|---|---|
| `config -q` | ❌ | ✅ |
| `ps` | ❌ | ✅ |
| `logs --tail=1 kong` | ❌ | ✅ |
| `up -d --no-deps kong` | ❌ | ✅ |

Deploys were unaffected — the playbook passes the profiles. What broke was everything a human types, including six runbooks under `docs/`.

**Fixed** by unprofiling **only** `otel-collector`. I checked every obs-profiled service for unprofiled dependents: the collector has **12**, all five others have **zero**. It is also the right line conceptually — the tiers gate what *consumes* telemetry, not the pipeline everything emits into. A site on `metrics` still needs the collector; what it does not need is Tempo and Loki.

## 3 · A metrics-only tenant waited two minutes for Loki

Ungated, the Loki health-wait burned 12 retries × 10s on a container the tenant deliberately did not deploy, then printed `Loki: UNHEALTHY (non-critical)`. `ignore_errors` kept it from failing, so it degraded quietly rather than announcing itself.

**Fixed** with a `when:` gate on the logs tier — and on the status line too, since `loki_health` is undefined when the wait is skipped.

## Verified on the devbox at `observability_level: metrics`

| | Result |
|---|---|
| tempo / loki / promtail | **RUNNING → GONE** (44 → 41 containers) |
| teardown task | `changed` when there was work, **`ok` on rerun** |
| Loki wait | **skipping**, 0 retries (was 12), `ignored=0` (was 1) |
| metrics tier + collector | prometheus, grafana, node-exporter, otel-collector all up |
| deploy | `ok=120 changed=19 failed=0`, 0 unhealthy |

## Two mistakes worth recording

Three deploys failed getting here, both placement bugs rather than logic bugs:

- `_obs_profiles` is a task-scoped `vars:` entry on the `set_fact`, **not a fact** — undefined by the time the teardown runs. It derives from `compose_profiles` instead.
- The task was initially placed **before `compose_files` was computed**.

A Jinja unit test passed throughout, because it exercised the expression and placement was what broke. Worth knowing: for ansible, testing the expression is not testing the task.


local-setup/ansible/playbook-deploy.yml
```diff
@@ -2398,6 +2398,63 @@
...
+      when:
+        - ansible_system != "Darwin"
+        - _obs_teardown | length > 0
+      vars:
+        # Services belonging to each tier, so an inactive tier can be named.
+        # otel-collector is deliberately absent: it is unprofiled and always
+        # runs (every instrumented service depends on it) -- see the comment on
+        # the service in docker-compose.egov-digit.yaml.
+        _obs_tier_services:
+          obs-metrics: [prometheus, grafana, node-exporter]
```
vinothrallapalli-eGov, line 2456: `node-exporter` shouldn't be in this list, for the same reason `otel-collector` was deliberately excluded (per your comment two tasks below): it has no `profiles:` entry in `docker-compose.monitoring.yml`, and that overlay is unconditionally added to `compose_files` ("Always layered", per that file's header and the `Compute compose -f flags` task). So at any level below `metrics`, this task will remove `digit-node-exporter`, but the next `up -d` — including "Recreate services with the new env" a few tasks above this one, and every subsequent redeploy — recreates it unconditionally since nothing gates its start. The teardown then fires again next run, so this one service flaps `changed` on every deploy instead of settling, and the memory you're trying to free at lower tiers doesn't actually stay freed. The devbox verification didn't catch this because it only tested dropping to `metrics` (where obs-metrics/node-exporter stays active) — tempo/loki/promtail are fine because they're genuinely `profiles:`-gated in the base compose file. Suggest dropping `node-exporter` from this tier's list (mirroring the otel-collector exclusion) or giving it a real `profiles: [obs-metrics]` gate in the monitoring overlay so `up -d` actually respects the tier.

vinothrallapalli-eGov (commented):
Verified fixes #2 (otel-collector un-profiling) and #3 (Loki-wait gate) against the actual compose file/facts and they hold up: I independently counted otel-collector's unprofiled dependents at exactly 12 (excluding the 2 `search`-profiled and 2 `notifications`-profiled ones that also depend on it), matching the PR's claim, and confirmed the 5 obs-profiled services have zero unprofiled dependents. The Loki-wait `when:` correctly keys off `'obs-logs' in compose_profiles.split(',')`, matching how `loki`'s own `profiles: [obs-logs]` is declared, and there's no other ungated reference to `loki_health` later in the file.

However, fix #1 (the teardown task) has a real gap: `_obs_tier_services.obs-metrics` includes `node-exporter`, but `node-exporter` (defined in `docker-compose.monitoring.yml`) has **no `profiles:` key** and that overlay is unconditionally layered into `compose_files` ("Always layered" per the file's own header and the `Compute compose -f flags` task) — so it starts on every `docker compose up -d` regardless of `observability_level`, exactly like `otel-collector`, which this same PR deliberately excludes from the teardown list for that reason ("otel-collector is deliberately absent: it is unprofiled and always runs"). Left as written, dropping below `metrics` will tear node-exporter down once, but the next `up -d` (recreate-with-new-env step later in this same play, or any subsequent redeploy) recreates it unconditionally, and the teardown task removes it again — a non-idempotent flap that also means the resource savings from disabling the metrics tier don't actually include node-exporter. This scenario wasn't exercised by the devbox verification, which only went down to `observability_level: metrics` (where obs-metrics, and thus node-exporter, stays active) — the traces/logs teardown paths that were exercised (tempo/loki/promtail) are correctly profiled and don't have this issue. Medium-severity: doesn't break anything, but silently defeats part of the fix's stated goal and the "steady-state redeploy is a no-op" invariant for this one service. Base branch targeting (`feat/1657-modular-observability`, matching #1666's own head) is correct and the diff is clean/scoped to the three described fixes.

singhalkarun:
Heads-up from fixing the docs on #1667 — the `observability_level` comment block in `local-setup/ansible/inventory/group_vars/digit.yml` (lines ~52-72) still describes the pre-#1687 world, so this branch now contradicts itself. Three claims:

1. `#   metrics  gatus, prometheus, node-exporter, grafana, collector  ~1.2 GB` — gatus and the collector are both unprofiled on this branch, so neither is in the `metrics` tier; they run at every level. (The ~1.2 GB is still right, since the collector is included either way.)

2. > A gatus-only `uptime` level (~64 MB) is deliberately NOT offered yet. 16 application services declare `depends_on: otel-collector`, so a level that omits the collector risks failing the whole deploy…

   That is exactly the reasoning `7e032928` overturns. The deploy risk is gone; what remains is that an `uptime` level would still run the collector, so its floor is **~384 MB, not ~64 MB** (gatus 64 + collector 320).

3. `# (Footprints are the mem_limit ceilings from #1612, not measured usage.)` — #1612 *proposes* those caps and is open; no observability container carries a `mem_limit` today. They all run uncapped, so these are budgets on paper rather than ceilings.

`4a478d3c` on `docs/1620-1658-monitoring-docs` fixes the same three in `docs/observability/enabling-monitoring.md` and the phrasing there should port over directly. Not touching this branch since it is your PR — flagging so the code comment and the docs do not ship disagreeing.

local-setup/ansible/playbook-deploy.yml
singhalkarun: The reasoning is right but the premise is not: node-exporter **does** have a `profiles:` entry. It has carried one since 9ad09bb0 (#1657), the base commit this PR is stacked on:

```yaml
# local-setup/docker-compose.monitoring.yml
  node-exporter:
    image: prom/node-exporter:v1.8.2
    container_name: digit-node-exporter
    # host metrics — part of the metrics tier (#1657)
    profiles: [obs-metrics]
```

Verified against the real compose stack, no cluster needed:

```
$ COMPOSE_PROFILES= docker compose -f docker-compose.egov-digit.yaml \
    -f docker-compose.monitoring.yml config --services   # filtered to the obs services
otel-collector
$ COMPOSE_PROFILES=obs-metrics ...
grafana node-exporter otel-collector prometheus
$ COMPOSE_PROFILES=obs-metrics,obs-logs,obs-traces ...
grafana loki node-exporter otel-collector prometheus promtail tempo
```

So below `metrics` the next `up -d` does not select node-exporter, the teardown reports `NOCHANGE` on the following run, and the memory stays freed. No flap. (The Linux "Recreate services with the new env" task is also *above* the teardown at line 2393, not below it, and the macOS converge loop you may have been looking at is `when: ansible_system == "Darwin"` while the teardown is `!= "Darwin"` — they never both run.)

Two things your comment did catch, though.

**The file said the opposite of what it does.** The header still read "node metrics are wanted on every deployed box, not gated behind an opt-in flag" — true when #1335 wrote it, stale since #1657. "Always layered" is about the `-f` flag; it says nothing about the service starting, and the header conflated the two. Fixed in 887deca8, along with the teardown task's `vars:` comment, which now states the actual membership rule — **gated vs ungated**, not overlay vs base file — and why otel-collector is the one exclusion. Your instinct that the otel-collector reasoning should apply symmetrically was correct; it just lands on the other side of the line here. For the record all six listed services are gated: tempo `obs-traces`, loki/promtail `obs-logs`, prometheus/grafana `obs-metrics` in the base file, node-exporter `obs-metrics` in the overlay.

**Where the stale text you quoted probably came from.** There is a *second*, root-level `docker-compose.monitoring.yml`, tracked in git, and it genuinely lacks the profiles entry:

```
$ diff docker-compose.monitoring.yml local-setup/docker-compose.monitoring.yml
24a25,26
>     # host metrics — part of the metrics tier (#1657)
>     profiles: [obs-metrics]
```

Nothing reads it. The playbook copies `../docker-compose.monitoring.yml` relative to `local-setup/ansible/` (playbook-deploy.yml:1134), and `check-gatus-coverage.py` reads the `local-setup` copy. Root `docker-compose.egov-digit.yaml` has drifted the same way — 1655 diff lines, last touched at `3cf87989` vs `7e032928` for the live one. That is a pre-existing repo-wide duplicate-tree problem rather than this PR's, so I have not folded a fix into a defect PR, but it is almost certainly what a `grep` lands on first and it is worth its own issue.

Also worth stating plainly, since you flagged the coverage gap: you are right that the devbox run only exercised dropping to `metrics`. The claims above are static rather than live — `docker compose config --services` resolves profile selection exactly as `up -d` does, but no lower-tier teardown has been observed on a box. Comment-only change, no behaviour touched.

vinothrallapalli-eGov (approved):

PR #1695, opened 2026-08-10
fix(dashboard): make analytics calendar tenant-configurable

KDwevedi:
## Context

This is the first implementation slice from the dashboard current-state review:

- Reference issue: https://github.com/KDwevedi/Citizen-Complaint-Resolution-System/issues/29
- Timezone addendum: https://github.com/KDwevedi/Citizen-Complaint-Resolution-System/issues/29#issuecomment-5236519283

The review found three competing calendar authorities: fixed EAT literals in the backend and database, UTC parsing in parts of the query path, and browser-local date math in the dashboard. This PR makes the backend authoritative for the analytics business calendar while keeping the esbuild frontend responsible for presentation.

## What changes

- Adds optional `dss.DashboardConfig.timeZone` as an IANA zone, resolved at the tenant state root with the existing analytics config TTL.
- Resolves one immutable `BusinessCalendar` per analytics request and threads the same zone and `asOf` through named windows, explicit date ranges, pinned-window decisions, runtime buckets, composed KPIs, and response metadata.
- Adds additive response metadata: `calendar: { timeZone, businessDate }`; the existing `asOf` and all calculations use that same captured calendar.
- Moves elapsed daily/hourly average authority to `pgr-services`; the esbuild FE no longer recomputes those values using browser-local dates.
- Uses the configured zone for default/reset filters and Last Updated rendering in `digit-ui-esbuild/products/dashboard`.
- Adds a forward Flyway migration that resolves per-tenant zones, recreates the active grain materialized views, and makes daily snapshots tenant-calendar-aware. Existing applied migrations are unchanged.
- Aligns deterministic `DashboardConfig` record selection across Java, SQL, and FE.
- Seeds `Africa/Nairobi` for `ke` and documents fallback, cache, refresh, and historical-snapshot behavior.

## Compatibility and failure behavior

- Missing, blank, invalid, or unavailable configuration falls back to `Africa/Nairobi`, preserving the previous effective behavior without consulting JVM or browser local time.
- Fallback warnings are bounded to once per state-root cache generation.
- The API change is additive.
- Existing historical `complaint_open_state_daily` rows are not rewritten; only snapshots captured after a timezone change use the new zone.

## Validation

- `pgr-services`: full Maven suite, 241 tests passed.
- `digit-ui-esbuild/products/dashboard`: full Node suite, 208 tests passed.
- `digit-ui-esbuild`: production build passed. Existing unrelated duplicate-key/direct-eval warnings remain.
- New Flyway migration parsed successfully as 19 PostgreSQL statements with `pglast`.
- `git diff --check` passed after rebasing on current `master`.

A live PostgreSQL/Flyway apply and materialized-view refresh could not be run locally because no PostgreSQL service or Docker daemon was available. CI or a deployment test environment should retain that as the final database integration gate.

## Follow-up boundary

This PR fixes the calendar authority first. The broader dashboard cleanup from issue #29 remains separate: a server-provided render/view model, removal of duplicated FE KPI/filter semantics, consolidation of the batch contract, and retirement or isolation of the legacy dashboard route.

### DIFF
--- a/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsPlanner.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsPlanner.java
@@ -19,7 +19,6 @@ import java.util.stream.Collectors;
 @Component
 public class AnalyticsPlanner {
 
-    private static final ZoneId EAT = ZoneId.of("Africa/Nairobi");           // UTC+3
     private static final Pattern ALIAS = Pattern.compile("^[a-zA-Z_][a-zA-Z0-9_]{0,63}$");
     private static final Set<String> BUCKETS = new HashSet<>(Arrays.asList("day","week","month","quarter","year"));
     private static final Pattern LAST_N_DAYS = Pattern.compile("^last_(\\d+)d$");
@@ -36,7 +35,7 @@ public class AnalyticsPlanner {
         }
     }
 
-    public Planned plan(JsonNode q, AnalyticsScope scope){
+    public Planned plan(JsonNode q, AnalyticsScope scope, BusinessCalendar calendar){
         String grainName = q.hasNonNull("grain") ? q.get("grain").asText() : inferGrain(q);
         Grain g = catalog.grain(grainName);
         if (g == null) throw new IllegalArgumentException("unknown_grain: " + grainName);
@@ -77,9 +76,13 @@ public class AnalyticsPlanner {
         if (window != null && window.hasNonNull("timeBucket")) {
             String unit = window.get("timeBucket").asText();
             if (!BUCKETS.contains(unit)) throw new IllegalArgumentException("invalid_param: timeBucket '" + unit + "'");
+            // The resolved tenant zone rides a JDBC-bound param (never string-concatenated) — the
+            // value is already ZoneId.of-validated by KpiCatalogService.resolveTimeZone, but binding
+            // it keeps this expression safe even if that validation is ever bypassed.
             String expr = g.isEpochMs(timeCol)
-                ? "date_trunc('" + unit + "', to_timestamp(" + timeCol + "/1000) AT TIME ZONE 'Etc/GMT-3')::date"
+                ? "date_trunc('" + unit + "', to_timestamp(" + timeCol + "/1000) AT TIME ZONE ?::text)::date"
                 : "date_trunc('" + unit + "', " + timeCol + ")::date";
+            if (g.isEpochMs(timeCol)) selectParams.add(calendar.zoneId.getId());
             String alias = "bucket";
             selectExprs.add(expr + " AS " + alias);
             groupExprs.add(expr);
@@ -107,7 +110,7 @@ public class AnalyticsPlanner {
                 conj.add(predicate(g, e.getKey(), e.getValue(), whereParams));
             }
         }
-        applyWindow(window, g, timeCol, conj, whereParams);
+        applyWindow(window, g, timeCol, conj, whereParams, calendar);
         applyScope(scope, g, conj, whereParams);
 
         // ---- assemble ----
@@ -285,8 +288,8 @@ public class AnalyticsPlanner {
     // ---------- window ----------
 
     /**
-     * Resolve a named window to its inclusive start instant (epoch-ms) in EAT. Every window ends at
-     * {@code now}, so the name alone fixes the interval {@code [start, now)}.
+     * Resolve a named window to its inclusive start instant (epoch-ms) in {@code zone}. Every window
+     * ends at {@code now}, so the name alone fixes the interval {@code [start, now)}.
      *
      * <p>Returns {@code null} for the boundless names ({@code all}) and for {@code live}, which is a
      * state predicate rather than a time interval — callers handle those before asking.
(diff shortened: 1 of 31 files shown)


backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsService.java
```diff
@@ -114,8 +114,15 @@ private Map<String,Object> doQuery(JsonNode body, RequestInfo requestInfo, Strin
         Set<String> callerRoles = extractRoles(requestInfo);
         boolean publicFloor = isPublicFloor(callerRoles);
 
+        // ONE resolved zone + ONE captured asOf for the whole batch (#29): composer, planner,
+        // compose ops and the response itself all read this same BusinessCalendar, so a request can
+        // never straddle two clock readings or disagree with its own echoed asOf/calendar.
+        long asOfMs = asOf();
```
vinothrallapalli-eGov: **Correctness — NPE on empty tenants.** `long asOfMs = asOf();` unconditionally auto-unboxes the nullable `Long` that `asOf()` returns. `asOf()` runs `SELECT max(facts_built_at) FROM complaint_facts`, which returns SQL `NULL` (not an exception) whenever `complaint_facts` has zero rows — true for any brand-new tenant with no complaints yet, or any tenant right after a fresh deploy before `DashboardRefreshScheduler`'s first run populates the MV.

Before this PR that nullable `Long` was only unboxed inside `computeCompose`'s `dailyAvgFromWeekly`/`hourlyAvgFromDaily` branches (which no currently-seeded `KpiDefinition` uses), so the null was otherwise stored harmlessly in a `Map`. Widening the unboxing to the top of `doQuery` — which runs on every request — turns a dormant NPE into a guaranteed one: every `POST /v2/analytics/_query` call for a fresh/empty tenant now 500s via the controller's catch-all, instead of returning the old working `{"asOf": null, ...}` response.

None of the new tests mock `jdbc.queryForObject` to return `null`, so this regression isn't caught by the added coverage.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/KpiCatalogService.java
```diff
@@ -155,6 +167,163 @@ public boolean isDepartmentScopingDisabled(String tenantId) {
...
+        for (Map<String, Object> r : records) {
+            if ("default".equals(trimmedId(r))) return r;
+        }
+
+        Map<String, Object> smallest = null;
+        String smallestId = null;
+        for (Map<String, Object> r : records) {
+            String id = trimmedId(r);
+            if (id == null || id.isEmpty()) continue;
+            if (smallestId == null || id.compareTo(smallestId) < 0) {
```
vinothrallapalli-eGov: **Correctness — silent department-scoping regression.** The tie-break for picking a `dss.DashboardConfig` record changed from "first record in MDMS API response order" to "record whose trimmed `id` is lexicographically smallest."

For a state root with two active `dss.DashboardConfig` records and no record explicitly `id:"default"`, this can select a *different* record than before. If the record the API previously returned first happened to carry `departmentScoping:"disabled"`, and a different record's `id` now sorts lexicographically smaller without that flag, employee department scoping silently flips back to *enforced* after this deploys — with no config change from the operator and no warning logged (this path only warns on the timeZone axis, not departmentScoping).

backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsPlanner.java
```diff
@@ -295,39 +298,40 @@ private Object value(JsonNode v){
...
         if (window == null || !window.hasNonNull("name")) return;
         String name = window.get("name").asText();
         if (name.equals("all")) return;
         if (name.equals("live")) {
             if (g.filterable.contains("is_open")) conj.add("is_open = ?"); else return;
             params.add(true); return;
         }
-        long now = System.currentTimeMillis();
-        Long fromMs = windowStartMs(name, now);
+        long now = calendar.asOfMs;
```
vinothrallapalli-eGov: **Correctness — 'today' window now anchors to a stale clock.** `long now = calendar.asOfMs` replaces the previous `System.currentTimeMillis()`. `asOfMs` is the last successful `complaint_facts` materialized-view refresh time, not the real current instant, so named/pinned windows (`dtd`, `wtd`, `last_Nd`, etc.) are now computed relative to whenever the MV last refreshed.

If `DashboardRefreshScheduler` is disabled or stuck/failing for, say, 2 days, `'today'`/`'this week'` resolve relative to that stale timestamp — and because the MV data is frozen at the same stale point, the query returns a real, non-zero total for that old day, rendered under a `'today'` label. Previously, a stale MV would correctly return 0 rows for 'today' against the real wall-clock boundary, revealing the staleness. Now it silently reports an old day's numbers as current, which is worse than the failure it presumably was trying to avoid.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/KpiCatalogService.java
```diff
@@ -130,22 +151,13 @@ public Optional<KpiDefinition> getDef(String kpiId, String tenantId) {
...
-            if (!records.isEmpty()) {
-                Object v = records.get(0).get("departmentScoping");
-                disabled = v instanceof String && DEPT_SCOPING_DISABLED.equalsIgnoreCase(((String) v).trim());
-            }
+            Map<String, Object> record = dashboardConfig(tenantId);
+            if (record == null) return false;
+            Object v = record.get("departmentScoping");
+            boolean disabled = v instanceof String && DEPT_SCOPING_DISABLED.equalsIgnoreCase(((String) v).trim());
             if (disabled)
                 log.info("dss.DashboardConfig.departmentScoping=disabled at {} — employee department scoping OFF",
```
vinothrallapalli-eGov, line 165: **Correctness — log volume regression.** This INFO log used to be gated behind `deptScopingCache`, so it fired once per state root per ~5-minute TTL window. After refactoring onto the shared `dashboardConfig(tenantId)` cache, this call sits outside any cache-miss check and now runs on every call to `isDepartmentScopingDisabled`.

Any tenant with `departmentScoping=disabled` now generates one INFO line per analytics request instead of one per 5-minute cache window — inflating log volume/cost for busy tenants. Notably, the sibling `resolveTimeZone` fallback logging added in this same PR *was* given a TTL-bounded `shouldWarnTimeZoneFallback` gate to avoid exactly this; this log call seems to have been missed.

vinothrallapalli-eGov (commented):
Automated high-effort review (4 confirmed findings, most-severe first). All flagged inline.

vinothrallapalli-eGov (approved):

PR #1696, opened 2026-08-10
fix(pgr): video fullscreen controls mid-screen + timeline thumbnail (CCSD-2153)

Hari-egov:
Two FE defects from the CCSD-2027 video-playback work (#1593), both confirmed in code.

## 1. Fullscreen controls appear mid-screen
The `<video>` has inline `width`/`max-height` (ComplaintPhotos ~13.75rem/10rem; TimeLineWrapper 160/110). Chrome's fullscreen UA rule sizes a fullscreen video 100%×100% but **without `!important`**, so the inline `max-height` wins — the video stays a small letterboxed box in the black viewport and the native control bar pins to *its* bottom (screen middle).

**Fix:** a `video:fullscreen` rule (with `!important`, plus `-webkit-`/`-moz-` variants in **separate** rules so an unknown prefixed pseudo can't drop the whole group) that releases the size constraints in fullscreen. Injected once at the root of each component that renders a player.

## 2. Timeline thumbnail blank/stretched
The timeline `<video>` had no `poster` and no `object-fit`, so a remote presigned src with `preload="metadata"` painted only its black background (or stretched the frame).

**Fix:** append `#t=0.1` to seek+paint the first frame as the thumbnail (safe after the S3 `?X-Amz-…` query — S3 ignores fragments) and set `object-fit:contain`. Added `object-fit:contain` to the ComplaintPhotos player too, for aspect-ratio parity.

## Feasibility / risk
FE-only (`ComplaintPhotos.js`, `TimeLineWrapper.js`); no backend/MDMS/config. Low risk — fullscreen rules are inert outside fullscreen; `#t=0.1` and `object-fit:contain` are no-ops where a frame already paints. Build clean; bundle carries the fullscreen rules, `#t=0.1`, and object-fit.

## Note
If the filestore/S3 presign disallows range requests, some browsers may still show a black first frame — the fully robust alternative is a server-generated poster (backend, out of FE scope).

### DIFF
--- a/digit-ui-esbuild/products/pgr/src/components/ComplaintPhotos.js
+++ b/digit-ui-esbuild/products/pgr/src/components/ComplaintPhotos.js
@@ -69,7 +69,7 @@ const MediaAttachment = ({ url, kind, label, downloadLabel }) => {
       src={url}
       aria-label={label}
       onError={() => setFailed(true)}
-      style={{ display: "block", width: "13.75rem", maxWidth: "100%", maxHeight: "10rem", borderRadius: 6, background: "#000" }}
+      style={{ display: "block", width: "13.75rem", maxWidth: "100%", maxHeight: "10rem", objectFit: "contain", borderRadius: 6, background: "#000" }}
     />
   );
 };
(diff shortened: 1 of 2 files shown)


Hari-egov:
**Root-cause correction (second commit `230cf374`) — the first commit's theory was wrong.**

Pixel-level testing (real element-fullscreen via user-gesture clicks + screenshots + a live CSS bisect) showed Chrome's fullscreen sizing already beats the inline styles — a clean harness page renders controls at the bottom **with or without** the `video:fullscreen` override. The real culprit is a global rule in the **vendored platform CSS** (origin `packages/css/src/components/topbar.scss`, orphaned, uncommented):

```css
video::-webkit-media-controls-panel { top: 55%; position: absolute; width: 100%; }
```

It pins Chrome's control panel at 55% of **every** video — mid-screen bar in fullscreen, and half-hidden controls on small timeline tiles (why some tiles looked control-less). Disabling all app CSS in-fullscreen snapped the bar to the bottom; removing this one rule reproduces that. Removed from the scss source and both vendored copies.

Also per QA: timeline tile enlarged **160×110 → 280×180**.

**Verified locally before pushing** (screenshots in the verification run):
- fullscreen → control bar at the true bottom, video fills the screen
- tiles → full control row (incl. fullscreen button) visible inside the tile

PR #1703, opened 2026-08-10
feat(dashboard): add curated public dashboard

### LINKED ISSUE #1540: [Enhancement] Publicly Accessible UI for Dashboard
We want to build a mechanism for enabling a curated public dashboard available on a endpoint accessible without credentials.

Timelines:

- Product to give go ahead for dev beginning by 6th Aug 2026 with a POV on the configuration UI @rahuljose-egov 
- @KDwevedi to validate be capability to enable this by 5th Aug 2026
- Assuming go ahead, reviewable instance to be available by first half 9th Aug 2026

CC @subhamegov

KDwevedi:
## Summary

Closes #1540.

- serves the existing esbuild dashboard without credentials at the canonical `/digit-ui/public-dashboard` URL (`/dashboard` remains a compatibility redirect)
- adds a small authenticated configurator page at `/configurator/manage/public-dashboard` with the public URL, copy/open controls, and enable/disable
- persists the fail-closed switch as `dss.DashboardConfig.publicDashboardEnabled`; Kenya is explicitly seeded enabled, while absent/false/malformed config remains disabled
- applies configurator changes immediately through a protected analytics config-refresh route, with the existing five-minute cache TTL as fallback
- returns a data-free disabled envelope so the shared dashboard shell shows a deliberate unavailable state; public query calls are rejected while disabled
- uses the same esbuild dashboard source as the employee UI; there is no Lovable/prototype integration or copied dashboard
- exposes only the exact Kong public pack/query routes, with per-IP rate limiting and request-size bounds
- restricts anonymous queries to bare KPI IDs selected by the curated aggregate-only `public-default` pack; arbitrary params, record counts, and employee credentials are excluded

## Configuration and access

1. Sign in to the configurator and open **Public Dashboard** in the sidebar.
2. Enable the state-root switch.
3. Share the URL shown by the page: `/digit-ui/public-dashboard`.

The MDMS write preserves existing DashboardConfig timezone, number-format, department-scoping, and allowed-role fields. The public frontend is read-only and does not read employee auth, refresh tokens, filter storage, or layout storage.

## Dependency / dashboard PR compatibility

This branch still contains #1695 because the public and employee dashboard paths share the same calendar-aware engine. It now also contains the complete latest #1717 behavior, including the `a9ebc869a` add-KPI geometry correction (charts `4x6`, full-width tables `12x6`).

Clean merge-tree checks pass against current upstream `master`, #1695, and #1717. After those focused PRs land, this PR can be rebased so their commits disappear from this diff.

## Security boundary

Only these two exact Kong routes are auth-optional:

- `POST /pgr-services/v2/analytics/public/packs`
- `POST /pgr-services/v2/analytics/public/_query`

The refresh route and configurator remain on the normal protected PGR/MDMS paths. Disabled public pack discovery exposes no KPI descriptors or layout; disabled public queries return `404 public_dashboard_disabled`. Enabled queries force the synthetic `PUBLIC` role and can reference only KPIs in the selected PUBLIC pack.

## Verification

- pgr-services focused config/public/cache suite: 50 tests passed on JDK 17
- dashboard: 216 Node tests passed
- configurator: 9 focused tests passed; changed files pass ESLint
- configurator production build passed
- digit-ui-esbuild production build passed (employee + public entries)
- JSON/schema validation and `git diff --check` passed
- clean merge trees against current `master`, #1695, and #1717

A live composed-environment smoke test remains the deployment gate for the anonymous route and Kong behavior. Per project workflow, #1540 remains **In progress** until deployment, then can move to review.

## Summary by CodeRabbit

* **New Features**
  * Added credential-free public dashboards with curated KPI packs, responsive layouts, read-only controls, and shareable links.
  * Added configuration tools to enable or disable public access, refresh settings, and copy or open dashboard URLs.
  * Added tenant-specific dashboard time zones with Nairobi fallback.
* **Bug Fixes**
  * Improved date calculations, snapshots, DST handling, and dashboard “Last updated” timestamps across time zones.
  * Added request limits and safer validation for public analytics queries.
* **Documentation**
  * Documented public dashboard exceptions and time-zone configuration.


KDwevedi:
Integration checkpoint after the #1407/#1717 review:

- this branch contains #1717's original persistence commit (`3a77304a6`)
- #1717 has since advanced to `a9ebc869a`, preserving the existing Add-KPI defaults (charts `4x6`, tables `12x6`) with regression assertions
- before #1703 merges, sync the landed/latest #1717 change and rerun the combined employee/public dashboard tests and both bundles

The latest `test`/`tilt-test` failures should also remain an explicit deployment gate: the compose run shows `kong-gateway` starting and then exiting, while the Tilt run captures no telemetry events. The static config, gateway parity, and whitelist checks pass, but the Kong exit still needs diagnosis before the anonymous route is considered deployable.

KDwevedi:
Implementation checkpoint (`d611f1e28`):

- canonical public URL is now `/digit-ui/public-dashboard`; `/dashboard` is retained only as a compatibility redirect
- added `/configurator/manage/public-dashboard`, backed by `dss.DashboardConfig.publicDashboardEnabled`
- enable/disable is enforced server-side and applied immediately through a protected cache refresh; failures fall back to the normal five-minute TTL with a visible warning
- disabled pack bootstrap returns no descriptors/layout, disabled queries return `404 public_dashboard_disabled`, and the public shell shows a clean unavailable state
- no Lovable/prototype integration was added; this remains the same esbuild dashboard engine
- synced the complete latest #1717 geometry fix into this branch

Validation: 50 focused Java tests, 216 dashboard tests, 9 configurator tests, both production UI builds, ESLint, JSON validation, and clean merge trees against current `master`, #1695, and #1717.

The remaining gate is deployment: diagnose/rerun the composed Kong smoke, then test the URL anonymously before moving #1540 from In progress to review.

KDwevedi:
CI follow-up (`5cc6ef26b`): both failed integration jobs had the same concrete startup error, not an application-test failure. Kong 3 rejected the two public route regex strings because declarative regex paths must begin `~/`; the branch used the older `~^/` form.

Fixed both routes to the supported, end-anchored form:

- `~/pgr-services/v2/analytics/public/packs$`
- `~/pgr-services/v2/analytics/public/_query$`

Added an infra contract that rejects any future Kong regex path not beginning `~/` and pins both public routes as end-anchored. The focused infra suite passes (8/8) and `git diff --check` passes. Fresh CI is now running on this commit.

KDwevedi:
Public-dashboard shell fixes are now on this branch and verified against the 8c24g sandbox.

Verified at 1440x1000 and 390x844:
- shared broad `DashboardHeader` retained
- public sidebar absent
- 5 curated KPI tiles render
- body/root/dashboard start at x=0, y=0 with no horizontal overflow
- mobile tiles reflow to one full-width column
- mobile Export control remains inside the header
- localization uses tenant `pg`; no raw `DASHBOARD_*` keys remain

Local checks: 214/214 dashboard tests pass, localization generation is in sync, and the production UI build succeeds (with the existing repository warnings). The PR remains draft while the fresh CI run completes.

KDwevedi:
Expanded and re-verified the public dashboard on the sandbox.

- Public pack now contains 8 aggregate, non-PII tiles: the 5 headline KPIs plus complaint types, 30-day complaint trends, and department resolution rates.
- The backend pack owns tile selection, ordering, default windows, and layout; the frontend only renders and responsively reflows it.
- Regenerated the embedded DIGIT-MCP dashboard catalog so it stays in parity with the canonical catalog.
- Fixed bounded mobile chart sizing after live verification.
- Live verification: one shared header, no public sidebar, no page gutters or horizontal overflow, no raw dashboard localization keys, and all 8 tiles populated on desktop and mobile.
- Validation: 215/215 dashboard tests pass; production UI build and catalog generation checks pass.

Keeping this PR in draft while the refreshed long-running CI jobs complete.

digit-mcp/src/tools/dashboard-catalog-seed.ts
```diff
@@ -368,6 +368,15 @@ export const DASHBOARD_CATALOG_SCHEMAS: Record<string, Record<string, unknown>>
...
+      },
+      "timeZone": {
+        "type": "string",
+        "description": "IANA time zone id (e.g. \"Africa/Nairobi\", \"Africa/Maputo\", \"Asia/Kolkata\") the analytics engine resolves ONE BusinessCalendar from per request — every named window (dtd/wtd/mtd/qtd/ytd), pinned-window suppression decision, explicit dateFrom/dateTo bound, timeBucket, and compose elapsed-time calculation is judged against this zone's calendar day, not UTC or the server's JVM zone (#29). Optional: absent, empty, or not a valid IANA zone id falls back to Africa/Nairobi (the historical hardcoded default every unconfigured tenant already behaved as — a migration-compatibility default, not a real default timezone). The request path sees a configuration change after the shared analytics config-cache TTL expires. The underlying complaint_facts/complaint_events materialized views are refreshed by DashboardRefreshScheduler on its own cadence (pgr.dashboard.refresh.interval.ms), so a DB-derived tile's data reflects a timeZone change only from its NEXT refresh onward, not instantaneously."
```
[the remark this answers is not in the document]
KDwevedi: Addressed in 8b9d23cd7: corrected the canonical DashboardConfig propagation wording and regenerated the embedded DIGIT-MCP catalog. The catalog generator check passes.

digit-ui-esbuild/products/dashboard/src/services/publicRuntimeIsolation.test.js
```diff
@@ -0,0 +1,159 @@
...
+  for (const kpiId of publicPack.tiles) {
+    const definition = definitionsById.get(kpiId);
+    assert.ok(definition, `${kpiId} must exist in the KPI catalog`);
+    assert.equal(definition.viz?.pii, false, `${kpiId} must remain aggregate-only`);
+    assert.ok(definition.rbac?.visibleTo?.includes("PUBLIC"), `${kpiId} must remain PUBLIC`);
+    if (expectedInsights.includes(kpiId)) {
+      const windowParam = definition.params?.find(({ name }) => name === "window");
+      assert.equal(definition.query?.window?.name, "last_30d", `${kpiId} query must cover 30 days`);
+      assert.equal(windowParam?.default, "last_30d", `${kpiId} default must cover 30 days`);
+    }
```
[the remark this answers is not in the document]
KDwevedi: Addressed in 8b9d23cd7: the regression now asserts the complete ordered eight-tile public allowlist while retaining uniqueness, layout parity, aggregate-only, PUBLIC visibility, and 30-day checks. All 216 dashboard tests pass.

digit-ui-esbuild/products/dashboard/src/styles/dashboard.css
```diff
@@ -3994,6 +4125,21 @@
...
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  margin: -1px;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border: 0;
+}
```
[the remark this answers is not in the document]
KDwevedi: Addressed in 8b9d23cd7: replaced the deprecated clip declaration with clip-path inset 50 percent in the source and generated dashboard CSS. Regeneration is byte-stable and the production UI build passes.

local-setup/tests/static/infra-contracts.test.ts
```diff
@@ -12,6 +12,24 @@ import * as path from 'node:path';
...
+   */
+  test('every regex path uses the Kong 3 declarative `~/` prefix', () => {
+    const regexPaths = [...KONG.matchAll(/^\s*-\s+(~\S+)\s*$/gm)].map((match) => match[1]);
+    expect(regexPaths.filter((routePath) => !routePath.startsWith('~/'))).toEqual([]);
+  });
+
+  test('public analytics regex routes stay end-anchored and Kong-parseable', () => {
+    expect(KONG).toContain('- ~/pgr-services/v2/analytics/public/packs$');
+    expect(KONG).toContain('- ~/pgr-services/v2/analytics/public/_query$');
+  });
```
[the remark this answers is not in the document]
KDwevedi: Addressed in 8b9d23cd7: the contract now filters public analytics regex paths and asserts the exact two-route allowlist, so extra and wildcard anonymous routes fail. The infra suite passes 8 of 8.

KDwevedi:
CodeRabbit follow-up completed in `8b9d23cd7`.

Addressed all seven inline findings and the outside-diff null-body finding:
- anonymous public 500 responses no longer expose exception details
- null bodies return the existing `400 invalid_param` envelope across pack, query, and refresh
- timezone joins select the deepest configured dot-boundary tenant root in both grains and daily snapshots
- scoped dashboard layouts never read or write the shared legacy slot
- the complete ordered eight-tile public allowlist is pinned by regression coverage
- DashboardConfig refresh wording is corrected and the MCP catalog regenerated
- screen-reader CSS uses `clip-path` in source and generated output
- anonymous Kong routes are asserted as an exact two-route allowlist

Validation: PGR 260/260, dashboard 216/216, infra contracts 8/8; production UI build, CSS regeneration parity, catalog generation parity, and `git diff --check` all pass. All seven review threads are resolved. Fresh CI is running.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsController.java
```diff
@@ -71,6 +72,138 @@ public ResponseEntity<Map<String,Object>> schema(){
...
+            return ResponseEntity.badRequest().body(error(e));
+        } catch (Exception e) {
+            log.error("public analytics query failed", e);
+            return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(publicUnavailable());
+        }
+    }
+
+    /** Refresh DashboardConfig after an authenticated configurator MDMS write. */
+    @PostMapping("/config/_refresh")
+    public ResponseEntity<Map<String,Object>> refreshDashboardConfig(
```
vinothrallapalli-eGov, line 196: **Correctness:** `POST /v2/analytics/config/_refresh` performs no role/ownership check on the caller-supplied `tenantId`, despite the comment above documenting it as an "authenticated configurator write" hook.

Any authenticated caller with a valid token — not just a configurator/admin role — can repeatedly POST an arbitrary `tenantId` here. `refreshPublicDashboardConfig` evicts and immediately re-fetches that tenant's `dss.DashboardConfig` from MDMS on every call, letting a low-privileged authenticated user defeat the analytics config-cache TTL for any tenant and multiply MDMS load well beyond the normal 5-minute refresh cadence.

digit-ui-esbuild/products/dashboard/src/utils/composeKpi.js
```diff
@@ -28,43 +28,17 @@ export function evaluateCompose(compose, results) {
...
+    // 'dailyAvgFromWeekly' / 'hourlyAvgFromDaily': the backend's D1a composition (calendar-
+    // aware, tenant-timeZone-correct — see BusinessCalendar / #29) is now the sole authority
+    // for these averages. This engine must NEVER recompute an "elapsed periods since asOf"
+    // average with browser-local Date math (setDate/setHours/new Date() are all local-clock
+    // reads that silently disagree with the tenant's configured calendar). Returning null
+    // here is intentional: KpiTile's resolveScalar falls through to the backend's
+    // result.value whenever evaluateCompose yields null.
+    case 'dailyAvgFromWeekly':
+    case 'hourlyAvgFromDaily':
+      return null;
```
vinothrallapalli-eGov, line 40: **Correctness:** `evaluateCompose` now unconditionally returns `null` for `dailyAvgFromWeekly`/`hourlyAvgFromDaily`, relying on the backend's D1a composition (`AnalyticsService`) as the sole source of these averages. But the backend only performs that computation when `isComposeDef()` is true, which requires the KPI definition's `query` field to be `null`. The `dss.KpiDefinition` MDMS schema documents that pairing only as a convention ("null for compose-only defs") — it isn't enforced.

If any tenant's KPI catalog has (or a future configurator edit creates) a def whose `viz.compose.type` is one of these two AND whose `query` is non-null, the backend's compose gate skips composition entirely, and this function returns `null` too — so `KpiTile.resolveScalar` falls through to the raw, un-divided total (e.g. a full weekly count shown where a daily average was intended) with no error or suppressed-state indicator.

configurator/src/api/publicDashboardConfig.ts
```diff
@@ -0,0 +1,34 @@
...
+  );
+  if (own.length === 0) return null;
+
+  return own.find((record) => String(record.data?.id ?? '').trim() === 'default')
+    ?? [...own].sort((left, right) => {
+      const leftId = String(left.data?.id ?? '').trim();
+      const rightId = String(right.data?.id ?? '').trim();
+      if (!leftId) return rightId ? 1 : 0;
+      if (!rightId) return -1;
+      return leftId.localeCompare(rightId);
```
vinothrallapalli-eGov: **Correctness:** This tiebreak uses `localeCompare`, but the equivalent dedupe logic in `KpiCatalogService.selectDashboardConfigRecord` (Java, `String.compareTo`) and `digit-ui-esbuild`'s `useDashboardConfig.js` (`selectDashboardConfigRecord`, JS `<`) both use ordinal/code-point comparison — not locale-aware.

If a state root ever has more than one active `dss.DashboardConfig` record with non-"default" ids that differ in case or contain non-ASCII characters (the surrounding comments here explicitly anticipate hand-seeded/duplicate MDMS data), `localeCompare`'s locale-aware ordering can pick a different "winning" record than the ordinal comparison pgr-services and the live dashboard use. The Configurator's Public Dashboard toggle would then read/write a different `DashboardConfig` record than the one actually served — an admin's Enable/Disable action could silently not affect the live config, or appear to save while nothing changes. Suggest switching to `leftId < rightId ? -1 : leftId > rightId ? 1 : 0` (or `localeCompare(rightId, undefined, ...)`'s ordinal equivalent) to match the other two implementations.

digit-ui-esbuild/products/dashboard/src/AdminDashboard.jsx
```diff
@@ -865,6 +933,19 @@ const AdminDashboardInner = ({ onSignOut, embedded = false }) => {
           {t("DASHBOARD_COMMON_CATALOG_UNAVAILABLE", "Catalog unavailable")}: {catalogError}
         </div>
       )}
+      {showPublicDisabled && (
+        <div className="tw-mx-auto tw-mt-12 tw-max-w-xl tw-rounded-lg tw-border tw-border-border tw-bg-surface tw-p-8 tw-text-center tw-shadow-sm">
+          <h1 className="tw-text-xl tw-font-semibold tw-text-foreground">
+            {t("DASHBOARD_PUBLIC_NOT_AVAILABLE_TITLE", "Public dashboard is not available")}
```
vinothrallapalli-eGov, line 939: **Cleanup:** `DASHBOARD_PUBLIC_NOT_AVAILABLE_TITLE` and `DASHBOARD_PUBLIC_NOT_AVAILABLE_DESCRIPTION` (used here and just below) aren't added to any localization source. The sibling key `DASHBOARD_HEADER_PUBLIC_SUBTITLE` added elsewhere in this PR was seeded into `digit-mcp/src/tools/dashboard-l10n-seed.ts` plus `local-setup/db/dss-mdms-seed/l10n/en_IN.json` and `pt_PT.json`, but these two new keys were not. A pt_PT (e.g. Mozambique) tenant visiting the public dashboard while it's disabled will see hardcoded English text instead of a translation.

backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsPlanner.java
```diff
@@ -362,6 +369,11 @@ private void applyScope(AnalyticsScope scope, Grain g, List<String> conj, List<O
         }
     }
 
+    /** Escape caller-controlled text before appending a SQL LIKE wildcard. */
+    static String escapeLikeLiteral(String value) {
```
vinothrallapalli-eGov, line 373: **Cleanup:** This new `escapeLikeLiteral()` helper is used for the `scope.tenantId` LIKE clause a few lines up (line 343), but the `boundaryPrefix` LIKE clause a bit further down (line 359, unchanged by this PR) still inline-duplicates the identical escaping chain (`.replace("\\","\\\\").replace("%","\\%").replace("_","\\_")`) instead of calling this helper. If a future fix to LIKE-escaping is applied only here, the `boundaryPrefix` path silently keeps the old, unfixed behavior.

vinothrallapalli-eGov (commented):
Automated review (high effort, 4 finders + independent verification per finding).

KDwevedi:
Addressed Vinoth’s five review findings in `1bfeb8d1b`: protected the config-refresh hook by configurator role and tenant ownership, made server-composed averages fail closed, aligned ordinal DashboardConfig selection, added missing en_IN/pt_PT unavailable-state localisation in the dashboard module, and reused the LIKE-literal escape helper. Added a public-runtime regression assertion for the localisation keys/module. Focused backend tests (JDK 17), dashboard tests, configurator tests, and both frontend production builds pass.

vinothrallapalli-eGov (approved):

vinothrallapalli-eGov (approved):

PR #1717, opened 2026-08-11
fix(dashboard): restore scoped KPI layout persistence

### LINKED ISSUE #1276: [Bug] Employee dashboard: unable to add KPIs

KDwevedi:
Fixes #1276.

## What changed

- reconnects `useCatalogLayout` to the existing tested `layoutStore` helpers instead of duplicating persistence and geometry logic
- hydrates saved Add-KPI layouts when the role-visible catalog is ready, even when the pack returns `defaultLayout: []`
- scopes layout storage by tenant and employee UUID, with the legacy global key retained as a read-only migration fallback
- adds hook-level regression coverage for empty-pack hydration and cross-user isolation

## Scope

This is the focused extraction from #1407. It deliberately excludes Portuguese localization, header/CSS changes, seed tooling, catalog fallback behavior, and auth transport rewrites.

The drag/drop attach path and empty-grid drop target from #1287/#1311 are unchanged.

## Validation

- `node --test $(rg --files products/dashboard -g "*.test.js")` — 176/176 passing
- `npm run build` — passing (existing repository warnings only)
- merged with #1703 locally and reran its expanded dashboard suite — 216/216 passing; employee and public bundles build

## Conflict check

`git merge-tree --write-tree` is clean against both active dashboard branches:

- #1695 (`fix/dashboard-business-calendar`)
- #1703 (`feat/1540-public-dashboard`), which now contains the pre-resolved combined hook behavior so either PR can land first

🤖 Scoped with Claude Code Sonnet agents.

### DIFF
--- a/digit-ui-esbuild/products/dashboard/src/hooks/useCatalogLayout.js
+++ b/digit-ui-esbuild/products/dashboard/src/hooks/useCatalogLayout.js
@@ -1,13 +1,18 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
-import {
-  GRID_COLS,
-  UNIFORM_CHART_SIZE_CONSTRAINTS,
-  MAP_SIZE_CONSTRAINTS,
-  FULL_WIDTH_TABLE_GRID,
-  DEFAULT_CHART_GRID,
-  DROPPING_ITEM_ID,
-} from "../constants/layoutConfig";
+import { GRID_COLS, DROPPING_ITEM_ID } from "../constants/layoutConfig";
+import { getEmployeeInfo, getTenantId } from "../services/authService";
 import { createCatalogDragGeometry, isCatalogCard } from "../utils/catalogDragGeometry";
+import {
+  LEGACY_STORAGE_KEY,
+  storageKeyFor,
+  readSavedLayout,
+  persistLayout,
+  buildSeedLayout,
+  normalizeItem,
+  resolveInitialLayout,
+  sizeConstraintsForKpi,
+  defaultSizeForKpi,
+} from "../utils/layoutStore";
 
 /**
  * useCatalogLayout — catalog-world layout hook (kpiId-keyed).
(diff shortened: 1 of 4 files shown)


KDwevedi:
Follow-up from the #1407 scope review: fixed the default-size regression introduced when `useCatalogLayout` was wired to the extracted layout store.

- normal charts remain `4x6`
- full-width tables remain `12x6`
- cards, maps, and ranked lists retain their existing defaults
- the helper now reuses `DEFAULT_CHART_GRID` and `FULL_WIDTH_TABLE_GRID` so it cannot drift from the canonical constants

Validation:

- 176/176 dashboard Node tests pass
- complete `digit-ui-esbuild` production bundle succeeds
- `git diff --check` passes

Commit: a9ebc869a

vinothrallapalli-eGov (approved):

Issue #1700, opened 2026-08-10
[Bug] Configurator UI: Unable to create new complaint types


Gurjeet-eGov:
# Description

Configurator UI: Unable to create new complaint types 
**"hierarchyType": "PGR"** is not sent in payload hence the UNIQUE_IDENTIFIER_EMPTY_ERR

---

# Steps to Reproduce

1. Login to configurator ui
2. Go to Complaint Management > Complaint Types
3. Create new

---

# Expected Behavior

User shall be able to create new complaint types 

---

# Actual Behavior

```
{
    "ResponseInfo": null,
    "Errors": [
        {
            "id": null,
            "parentId": null,
            "code": "UNIQUE_IDENTIFIER_EMPTY_ERR",
            "message": "Values defined against unique fields cannot be empty.",
            "description": null,
            "params": null
        }
    ]
}
```

```
curl 'https://bometfeedbackhub.digit.org/mdms-v2/v2/_create/RAINMAKER-PGR.ComplaintHierarchy' \
  -H 'accept: */*' \
  -H 'accept-language: en-GB,en-US;q=0.9,en;q=0.8' \
  -H 'authorization: Bearer ad80323c-8d42-4c55-9bed-1d534558ca79' \
  -H 'baggage: sentry-environment=production,sentry-public_key=9858dc027b16081569d8e1a491c42efc,sentry-trace_id=6175ecc81348475fa0fbeb69fe167b56,sentry-org_id=4510791908720640,sentry-sampled=true,sentry-sample_rand=0.0013622022704358194,sentry-sample_rate=1' \
  -H 'content-type: application/json' \
[...]

PR #1719, opened 2026-08-11
fix(configurator): stamp hierarchyType/levelCode on complaint type create

Lokendra-egov:
## Summary
- `RAINMAKER-PGR.ComplaintHierarchy` requires `hierarchyType` + `levelCode` and uses `[hierarchyType, code]` as its unique key, but the Configurator's Complaint Types create form never carried either field — create failed with `UNIQUE_IDENTIFIER_EMPTY_ERR`. `serviceDefToLeafWrite()` now stamps `hierarchyType: 'PGR'` and `levelCode: 'SUB_TYPE'` (matching every existing leaf row) when absent, so create fills them in without touching an edit's already-populated values.
- Also fixes two unrelated bugs that were blocking any build of this package: `@digit-mcp/data-provider` listed itself as its own dependency (breaks `npm install` with a registry 404), and `DigitApiClient.ts` uses `Buffer` without `@types/node` declared (breaks `tsc`).

Fixes #1700

## Test plan
- [x] `npm test` in `configurator/packages/data-provider` — 54 passing
- [x] `npx vitest run` in `configurator/` — 97 passing
- [x] `npm run build` (both packages) and `docker build` of the configurator image — succeed end-to-end
- [x] Verified the actual fixed payload (with `hierarchyType`/`levelCode` stamped) against a live MDMS `_create` call — succeeds with a real `uniqueIdentifier`, no `UNIQUE_IDENTIFIER_EMPTY_ERR`; cleaned up the test record afterward

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved data handling by applying default hierarchy and level information when these values are missing.
  * Existing hierarchy and level values continue to be preserved when provided.

### DIFF
--- a/configurator/packages/data-provider/package.json
+++ b/configurator/packages/data-provider/package.json
@@ -24,9 +24,6 @@
     "test:integration": "node --import tsx --test src/providers/dataProvider.integration.test.ts",
     "test:watch": "node --import tsx --test --watch src/**/*.test.ts"
   },
-  "dependencies": {
-    "@digit-mcp/data-provider": "^0.1.0"
-  },
   "peerDependencies": {
     "ra-core": ">=4.0.0"
   },
@@ -36,6 +33,7 @@
     }
   },
   "devDependencies": {
+    "@types/node": "^20.0.0",
     "ra-core": "^5.0.0",
     "tsx": "^4.15.0",
     "typescript": "^5.5.0"
--- a/configurator/packages/data-provider/src/providers/dataProvider.ts
+++ b/configurator/packages/data-provider/src/providers/dataProvider.ts
@@ -178,6 +178,13 @@ function mapLeafToServiceDef(
  *  The metadata strip (id / `_*`) is left to the caller. */
 function serviceDefToLeafWrite(data: Record<string, unknown>): Record<string, unknown> {
   const out: Record<string, unknown> = { ...data };
+  // hierarchyType/levelCode are required + part of the x-unique key
+  // (RAINMAKER-PGR.json) but the Complaint Type create form never carries
+  // them — it only knows the legacy ServiceDefs fields. Stamp the same
+  // constants every leaf row in this hierarchy already uses (CCRS#1700);
+  // guarded so an edit's spread-in existing values are never overwritten.
+  if (!out.hierarchyType) out.hierarchyType = 'PGR';
+  if (!out.levelCode) out.levelCode = 'SUB_TYPE';
   // serviceCode -> code (the leaf's code IS the serviceCode stored on a
   // complaint). Populate `code` from a filled Service Code whenever `code` is
   // absent OR blank — the create form carries `code: ""` (empty string, not
--- a/configurator/packages/data-provider/src/providers/dataProvider.ts
+++ b/configurator/packages/data-provider/src/providers/dataProvider.ts
@@ -171,20 +171,65 @@ function mapLeafToServiceDef(
   };
 }
 
+// Used only when a tenant has no ComplaintHierarchyDefinition yet — a fresh
+// tenant bootstrapping its first leaf row before ever visiting "Complaint
+// Hierarchies" to declare one. Every real tenant's definition overrides these.
+const FALLBACK_HIERARCHY_TYPE = 'PGR';
+const FALLBACK_LEAF_LEVEL_CODE = 'SUB_TYPE';
+
+interface HierarchyDefinitionLevel {
+  levelCode?: string;
+  isLeafServiceCode?: boolean;
+}
+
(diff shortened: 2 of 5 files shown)


configurator/packages/data-provider/package.json
```diff
@@ -24,9 +24,6 @@
     "test:integration": "node --import tsx --test src/providers/dataProvider.integration.test.ts",
     "test:watch": "node --import tsx --test --watch src/**/*.test.ts"
   },
-  "dependencies": {
```
vinothrallapalli-eGov, line 27: @Lokendra-egov Why this has been removed?

configurator/packages/data-provider/package.json
```diff
@@ -36,6 +33,7 @@
     }
   },
   "devDependencies": {
+    "@types/node": "^20.0.0",
```
vinothrallapalli-eGov, line 36: @Lokendra-egov Why new dependency added?

configurator/packages/data-provider/src/providers/dataProvider.ts
```diff
@@ -178,6 +178,13 @@ function mapLeafToServiceDef(
  *  The metadata strip (id / `_*`) is left to the caller. */
 function serviceDefToLeafWrite(data: Record<string, unknown>): Record<string, unknown> {
   const out: Record<string, unknown> = { ...data };
+  // hierarchyType/levelCode are required + part of the x-unique key
+  // (RAINMAKER-PGR.json) but the Complaint Type create form never carries
+  // them — it only knows the legacy ServiceDefs fields. Stamp the same
+  // constants every leaf row in this hierarchy already uses (CCRS#1700);
+  // guarded so an edit's spread-in existing values are never overwritten.
+  if (!out.hierarchyType) out.hierarchyType = 'PGR';
```
vinothrallapalli-eGov: @Lokendra-egov The hierarchy type should be mandatory and user should input in that case why we are setting default? and also we can't hard levelCode to sub type, it will be dynamic?

vinothrallapalli-eGov (changes requested):

configurator/packages/data-provider/package.json
Lokendra-egov: This entry was `@digit-mcp/data-provider` listing itself as its own dependency — present since this file's very first commit. That's why `npm install` always failed with a 404 against the public registry: this package isn't published anywhere, and the parent `configurator` app already consumes it via a `file:` path (not a version range), so it was never meant to resolve externally. Removed it since it has no real runtime dependencies — the only external import anywhere in `src/` is `ra-core`, which is already declared correctly as a peer/dev dependency.

configurator/packages/data-provider/package.json
Lokendra-egov: `DigitApiClient.ts` uses the `Buffer` type (for file uploads) but `@types/node` was never declared as a dependency, so `tsc` failed with `Cannot find name 'Buffer'`. This was blocking any build of the package at all — unrelated to the #1700 fix itself, but needed to actually compile/test it.

configurator/packages/data-provider/src/providers/dataProvider.ts
Lokendra-egov: Good catch — fixed properly in the latest commit (a8587f96), two changes:

1. `serviceDefToLeafWrite` now only stamps hierarchyType/levelCode when an explicit `newLeafDefaults` argument is passed (create only). On update it's called with no defaults at all, so `dataProvider.update()`'s existing `{...existing.data, ...sanitized}` merge naturally preserves whatever the record already has — an edit can no longer silently overwrite these fields, since the edit form never renders them.
2. On create, added `resolveNewLeafDefaults()`, which derives both values from the tenant's actual `RAINMAKER-PGR.ComplaintHierarchyDefinition` (the declared `hierarchyType` + whichever level is flagged `isLeafServiceCode`) instead of a hardcoded literal — so `levelCode` is no longer assumed to always be `SUB_TYPE`. It only falls back to `PGR`/`SUB_TYPE` when a tenant has no definition at all yet (a fresh-tenant bootstrap case with nothing to derive from).

On making hierarchyType a required manual input instead of deriving it: I went with automatic derivation rather than asking the operator to type it on every single sub-type creation, since in practice a tenant already has exactly one (or a small, already-configured) hierarchyType via 'Complaint Hierarchies', and asking someone filling out a routine 'add a sub-type' form to know/type that string exactly is friction and a typo risk for something the system already knows from existing config. Open to making it an explicit field instead if you still think that's better after this — let me know.

Added unit tests covering: dynamic resolution against a non-default hierarchyType/levelCode, the no-definition fallback, and update no longer overwriting an existing record's values.

vinothrallapalli-eGov (approved):

PR #1724, opened 2026-08-11
fix(configurator): expose keywords, order, and multi-department fields on Complaint Type forms

### LINKED ISSUE #1713: [Bug] Configurator UI: The create complaint type UI does not have fields as per the complaint hierarchy
# Description

Creating a new complaint type from configurator ui does not show all the fields of type and sub type as per the hierarchy data 

# Actual Behavior

Image

Image

Lokendra-egov:
## Summary
`RAINMAKER-PGR.ComplaintHierarchy` leaf rows support `keywords`, `order`, and a `departments[]` list (`department` is just "the primary of this list"), but the Complaint Type create/edit forms didn't expose most of it:
- `keywords` was hardcoded to the literal string `"complaint"` on create for every type, not editable at all.
- `order` was hardcoded to `0`, not editable.
- `departments[]` had no UI at all — only the single primary `department` field existed, so a type could never route to more than one department.

Adds a new `DigitFormMultiSelect` (checkbox-list — no combobox/search, since there's no multi-select primitive in this codebase yet and the reference lists here are short) and wires it into both `ComplaintTypeCreate` and `ComplaintTypeEdit` as `departments`, deriving the primary `department` from the first checked entry via each form's `transform`. `keywords` and `order` are now plain editable inputs on both forms.

Out of scope (tracked separately, per discussion on the issue): there's currently no way to create interior "Type"/category-level nodes through the Configurator UI at all — only leaf "Sub-Type" nodes via this form. That's a larger feature, not a field-mapping fix, and is being left for a follow-up.

Fixes #1713

## Test plan
- [x] `npx tsc -b` — clean
- [x] `npx vitest run` — 97 passing, no regressions
- [x] Will verify live on bomet via the actual Configurator UI before/after merge

## Summary by CodeRabbit

* **New Features**
  * Added a reusable multi-select field for choosing multiple departments, with validation, helper text, loading states, and reference data support.
  * Complaint types can now be associated with multiple departments.
  * Added a display-order field when editing complaint types.
* **Improvements**
  * The first selected department is automatically used as the primary department.
  * Existing complaint records retain department selections, including legacy records.
  * Department hierarchy defaults now adapt based on available complaint data.
  * Added guidance for department selection and keyword fields.
  * Removed the default complaint keyword from new complaint types.


configurator/src/resources/complaint-types/ComplaintTypeEdit.tsx
```diff
@@ -1,23 +1,34 @@
...
+    <DigitEdit title="Edit Complaint Type" transform={transform}>
       <FieldSection title="Details">
         <div className="space-y-4">
           {/* Grouping key — parent node code in the ComplaintHierarchy tree
               (replaces the old free-text menuPath). */}
           <DigitFormInput source="parentCode" label="Parent (group code)" />
-          <DigitFormSelect
-            source="department"
-            label="Department"
+          <DigitFormMultiSelect
```
vinothrallapalli-eGov, line 24: **Correctness:** This `DigitFormMultiSelect` (source="departments") is not seeded from the legacy singular `department` field, unlike the existing List/Show views which already fall back to `department` when `departments` is absent.

Pre-existing complaint-type records only ever had `department`; `departments` is a brand-new field these rows never populated. Since `validate={v.required}` treats an empty array as invalid, an admin editing any pre-existing complaint type (even just to fix `slaHours` or toggle Active) sees an empty, required Departments checklist and save is blocked until they manually re-check a department — and because nothing in the UI shows what the record's actual current department is, picking the wrong one silently changes the complaint type's backend routing department (`department: departments[0]` in this file's own `transform`).

configurator/src/admin/DigitFormMultiSelect.tsx
```diff
@@ -0,0 +1,141 @@
...
+          if (typeof parsed === 'string') return parsed;
+          if (parsed && typeof parsed === 'object' && 'message' in parsed)
+            return String((parsed as { message: unknown }).message);
+          return String(parsed);
+        } catch { return rawError.slice(15); }
+      })()
+    : rawError;
+
+  const toggle = (value: string, checked: boolean) => {
+    const next = checked ? [...selected, value] : selected.filter((v) => v !== value);
```
vinothrallapalli-eGov: **Correctness:** `toggle` appends newly-checked values to the end of the array, so re-checking a department after unchecking it changes which one becomes `departments[0]` ("primary"), with no UI indication of order.

Example: an admin has departments A (primary) and B checked. They uncheck A to correct a mis-click, then re-check it — A is re-appended at the end, making B (not A) the new `departments[0]`. `ComplaintTypeEdit`'s `transform` (`department: departments[0]`) then silently makes B the backend routing department on save, even though the admin's intent was only to toggle A off and back on.

configurator/packages/data-provider/src/providers/dataProvider.ts
```diff
@@ -182,27 +182,82 @@ interface HierarchyDefinitionLevel {
...
+    if (definitions.length === 1) {
+      return leafDefaultsFromDefinition(definitions[0])!;
     }
+
+    // Presence alone isn't enough here — a stray one-off leftover row under
+    // the wrong hierarchyType (exactly the live scenario this is fixing)
+    // would tie with a candidate that has real, dominant usage, and array
+    // order would silently decide the winner again. Count occurrences and
+    // pick the candidate with the MOST usage, not just "any at all".
+    const sample = await client.mdmsSearch(tenantId, 'RAINMAKER-PGR.ComplaintHierarchy', { limit: 500 });
```
vinothrallapalli-eGov: **Correctness:** This second `mdmsSearch` call sits inside the same `try` block as the already-successful `definitions` lookup above, so a transient failure here discards the real `definitions` and falls through to the hardcoded `FALLBACK_HIERARCHY_TYPE`/`FALLBACK_LEAF_LEVEL_CODE` instead of degrading to "oldest/first active definition" using data already fetched.

For a tenant with two active `ComplaintHierarchyDefinition` rows (the exact ambiguous state this PR is fixing), if this usage-counting call times out or errors, the new complaint type gets created under a generic fallback `hierarchyType` matching neither real definition — making it invisible in every other view that filters by the real `hierarchyType`. That reproduces, via a new path, the exact bug (CCRS#1713) this PR was written to fix. Consider wrapping only this call in its own try/catch so a failure here falls back to `definitions[0]` rather than the hardcoded constants.

configurator/src/admin/DigitFormMultiSelect.tsx
```diff
@@ -0,0 +1,141 @@
...
+  /** Optional helper text shown below the list (muted) */
+  help?: string;
+}
+
+/** Checkbox-list multi-select bound to an array form value (e.g. `departments`).
+ *  Mirrors DigitFormSelect's reference-fetching, but for a field that can hold
+ *  more than one value. Plain checkboxes rather than a searchable combobox —
+ *  there's no combobox primitive in this codebase yet, and reference lists
+ *  here (e.g. departments) are short enough that a flat list reads fine. */
+export function DigitFormMultiSelect({
```
vinothrallapalli-eGov, line 35: **Cleanup / reuse:** `configurator/src/resources/designations/DepartmentChipInput.tsx` already implements a reference-backed multi-select for departments — it takes a `source` prop, calls `useGetList('departments', ...)`, coerces legacy string values into a one-element array, and writes back a `string[]` of codes. That's the same shape `ComplaintTypeCreate`/`Edit` need for `departments` here.

Rather than lifting `DepartmentChipInput` into a shared location (e.g. `admin/widgets`) and reusing it with `source="departments" reference="departments"`, this PR hand-rolls a second, parallel implementation with its own `useGetList` call, choice-mapping/sorting, and error-message-unwrapping — and without `DepartmentChipInput`'s legacy-string-to-array coercion. A future fix (e.g. to legacy-value handling or an accessibility issue) now has to be applied to both components, and will likely only land in one, leaving the other silently inconsistent. Worth considering generalizing `DepartmentChipInput` instead of adding this new component, if the checkbox-list UI here isn't a deliberate departure from its chip-combobox pattern.

vinothrallapalli-eGov (commented):
Automated review (high effort, 4 finders + independent verification per finding).

One additional confirmed finding falls outside this PR's diff so it can't be anchored inline: **`configurator/src/resources/complaint-types/ComplaintTypeList.tsx:253`** — the pre-existing inline "Department" cell editor in the datagrid still writes only the singular `department` field via `{...record, department: val}`, but list/show rendering (as of this PR) now prefers the non-empty `departments` array whenever present. An admin using that inline editor on a complaint type created via the new Create/Edit forms will have their edit silently swallowed in the UI (the cell keeps showing the old `departments` value) while `department` — which backend routing/validation reads — actually changes underneath. Worth a follow-up to make the inline editor `departments`-aware too.

configurator/src/admin/DigitFormMultiSelect.tsx
```diff
@@ -0,0 +1,119 @@
...
+    if (!data) return [];
+    return data.map((item) => ({
+      value: String(getNestedValue(item, optionValue) ?? item.id),
+      label: String(getNestedValue(item, optionText) ?? getNestedValue(item, optionValue) ?? item.id),
+    }));
+  }, [staticChoices, data, optionValue, optionText]);
+
+  const selected: string[] = Array.isArray(field.value) ? field.value : [];
+  const hasError = fieldState.invalid && fieldState.isTouched;
+  const errorMessage = fieldState.error?.message;
```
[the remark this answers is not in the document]
Lokendra-egov: Already resolved by a later commit in this branch (before this review ran): `hasError` is no longer gated on `isTouched` at all — it's just `fieldState.invalid`. A pre-existing record edited before `departments` existed loads the field empty/invalid, and the error now shows immediately on mount rather than waiting for a touch/dirty event, which is a superset of what this comment asks for. No further change needed here.

configurator/src/resources/complaint-types/ComplaintTypeEdit.tsx
```diff
@@ -1,23 +1,34 @@
...
 import { FieldSection } from '@/admin/fields';
 import { BooleanInput } from '@/admin/widgets';
 
+// `department` (the schema's single "primary" field, used by backend
+// routing/validation) is derived from the first checked entry in
+// `departments` — records saved before `departments` existed only carry
+// `department`, so this keeps working for them once re-saved here.
+function transform(data: Record<string, unknown>) {
+  const departments = Array.isArray(data.departments) ? (data.departments as string[]) : [];
+  return { ...data, department: departments[0] };
```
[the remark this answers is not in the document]
Lokendra-egov: Valid — fixed in d53fef35. Rather than patching the transform, I fixed it at the read side: `mapLeafToServiceDef` in `dataProvider.ts` now seeds `departments` from the legacy singular `department` whenever a record has no `departments` yet, mirroring the fallback `ComplaintTypeList`/`ComplaintTypeShow` already use for rendering. That way every consumer (not just this transform) sees a properly-seeded array, and the transform's `departments[0]` correctly resolves to the record's actual existing department instead of `undefined`. Added a regression test.

configurator/src/resources/complaint-types/ComplaintTypeEdit.tsx
Lokendra-egov: Same root cause as CodeRabbit's comment above — fixed the same way, at the data-provider read side (`mapLeafToServiceDef`) rather than per-form, so List/Show/Edit/Create all get a consistently-seeded `departments` array for legacy rows. Added a regression test covering the read path specifically. Thanks for spelling out the exact failure mode (blind re-pick silently changing the routing department) — that's exactly the scenario the test now locks in.

configurator/src/admin/DigitFormMultiSelect.tsx
Lokendra-egov: Valid, fixed. `toggle` now derives the written array by filtering `choices` (the stable, rendered order) rather than appending the newly-checked value to the end of click history. Re-checking A after an accidental uncheck now puts it back in the same slot it always occupied, so it stays `departments[0]` regardless of toggle order. No new state needed — just reads off the same `choices` list already used to render the checkboxes.

configurator/packages/data-provider/src/providers/dataProvider.ts
Lokendra-egov: Valid, fixed exactly as suggested. Split into two try/catches: the `definitions` fetch keeps its own (failure there has no fallback data to degrade to, so it still returns the hardcoded bootstrap constants), and the usage-count `ComplaintHierarchy` search now has a separate try/catch that, on failure, returns the oldest-created of the already-fetched `definitions` instead of falling through to the hardcoded constants. Added a regression test that throws specifically on the usage-count call and asserts the oldest known definition wins, not `FALLBACK_HIERARCHY_TYPE`/`FALLBACK_LEAF_LEVEL_CODE`.

configurator/src/admin/DigitFormMultiSelect.tsx
Lokendra-egov: Fair point on the duplication, and good catch that `DepartmentChipInput` already had legacy-value coercion this component lacked. I fixed that specific overlap at the root instead of in either component: `mapLeafToServiceDef` (dataProvider.ts) now seeds `departments` from the legacy `department` on read, so neither component needs its own coercion — that removes the actual duplicated logic you flagged (legacy-value handling), and any future fix to it only has to land once, at the data layer, regardless of which UI ends up using it.

On the component itself: I kept `DigitFormMultiSelect` rather than switching to `DepartmentChipInput` because the two are a deliberate UX difference, not an accidental duplicate — a bounded checkbox list you scan at a glance vs. a type-to-search combobox for a longer list. `DepartmentChipInput` is also currently wired to `Designation` specifically (source defaults to `department`, single field, no static-choices mode), so reusing it here as-is would mean either lifting/generalizing it now or forking its search/keyboard-nav logic too. Given the checkbox UI is intentional for this short, bounded departments list, I'd rather not bundle that generalization into this PR — but agree it's worth doing if/when a third consumer needs the same reference-multi-select pattern, at which point unifying both into one shared component in `admin/widgets` makes more sense than it does for just these two.

configurator/packages/data-provider/src/providers/dataProvider.ts
```diff
@@ -182,31 +193,99 @@ interface HierarchyDefinitionLevel {
...
+    const usageCounts = new Map<string, number>();
+    for (const r of sample) {
+      const ht = (r.data as { hierarchyType?: unknown } | undefined)?.hierarchyType;
+      if (typeof ht === 'string') usageCounts.set(ht, (usageCounts.get(ht) ?? 0) + 1);
+    }
+    const byUsage = definitions
+      .map((d) => ({ def: d, count: usageCounts.get((d.data as { hierarchyType?: unknown } | undefined)?.hierarchyType as string) ?? 0 }))
+      .filter((x) => x.count > 0)
+      .sort((a, b) => b.count - a.count);
```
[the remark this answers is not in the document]
Lokendra-egov: Valid, fixed in 7cc0965e. The usage-count lookup now counts only ACTIVE rows whose `{hierarchyType, levelCode}` match each candidate definition's own leaf level (via `leafDefaultsFromDefinition`, the same leaf-level predicate already used elsewhere in this file) — not every row that merely shares the hierarchyType. Interior nodes and soft-deleted leaves no longer contribute to a candidate's score.

Also fixed the existing "MOST existing leaf usage" test's fixtures, which were missing `levelCode` on the sample rows and would have scored 0 under the stricter matching, and added a dedicated regression test with inactive leaves + active interior rows mixed in on the losing candidate — it has more total rows (6 vs 2) but only 1 genuine active leaf, and must still lose to the candidate with 2 genuine active leaves.

configurator/packages/data-provider/src/providers/dataProvider.ts
```diff
@@ -153,12 +153,23 @@ function mapLeafToServiceDef(
...
+  // `department` — fall back to it (same convention ComplaintTypeList/Show
+  // already use for rendering) so editing a legacy row doesn't load an
+  // empty, required multi-select that blocks Save until the operator
+  // re-picks a department blind (CCRS#1724 review).
+  const departments =
+    Array.isArray(data.departments) && data.departments.length
+      ? data.departments
+      : data.department != null && data.department !== ''
+        ? [String(data.department)]
+        : [];
```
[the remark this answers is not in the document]
Lokendra-egov: Valid, fixed in 6429ea3f. The legacy-department fallback now trims the value and rejects it (falls back to an empty `departments`) when it's whitespace-only or, case-insensitively, `NA` — the sentinel `ComplaintHierarchySetup`'s bulk-import path stamps for rows with no real department. A genuine value is trimmed but keeps its original case. Added a regression test covering `na`, whitespace-only, and a genuine value with incidental whitespace.

configurator/packages/data-provider/src/providers/dataProvider.ts
```diff
@@ -182,31 +193,99 @@ interface HierarchyDefinitionLevel {
...
+    return { hierarchyType: FALLBACK_HIERARCHY_TYPE, levelCode: FALLBACK_LEAF_LEVEL_CODE };
+  }
+  if (definitions.length === 1) {
+    return leafDefaultsFromDefinition(definitions[0])!;
+  }
+
+  const oldestOfKnown = () =>
+    leafDefaultsFromDefinition(
+      [...definitions].sort((a, b) => (a.auditDetails?.createdTime ?? 0) - (b.auditDetails?.createdTime ?? 0))[0],
+    )!;
```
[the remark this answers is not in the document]
Lokendra-egov: Valid, fixed in 6429ea3f. `oldestOfKnown` no longer treats a missing/malformed `createdTime` as epoch-0 — definitions with a known timestamp now always sort before ones without, and ties (including multiple unknowns) break on `uniqueIdentifier` for determinism, consistent with why this whole disambiguation exists. Added a regression test with one definition missing `auditDetails` entirely against one with a real, later timestamp — the one with actual evidence must still win.

Lokendra-egov:
@vinothrallapalli-eGov re: the out-of-diff finding on `ComplaintTypeList.tsx:253` (inline Department cell editor writing only `department`) —

Valid and fixed in 6429ea3f, at the data-provider layer rather than the datagrid column config: `dataProvider.update()` now detects when `department` changed but the submitted `departments` is byte-for-byte identical to what the record already had (exactly the signature of that inline cell editor's `{...record, department: newVal}` payload, as opposed to an intentional edit through the dedicated Edit form's multi-select, which always submits a genuinely-changed array) and resyncs `departments` to `[newDepartment]` so both fields move together. Added two regression tests: one confirming the resync fires for the single-column inline edit, and one confirming it does *not* fire (and does not clobber a real multi-department edit) when `department` is unchanged.

I kept the fix off the datagrid column config itself since neither the column-level `editable.transform` nor the grid-level `mutationOptions.transform` hooks have access to the previous value to distinguish "this edit actually changed department" from "department just came along in the spread" — the data-provider's `update()` already re-fetches the existing record right before this point, so it was the natural place with enough context to do this safely.

configurator/src/resources/complaint-types/ComplaintTypeEdit.tsx
```diff
@@ -1,23 +1,34 @@
...
-          <DigitFormSelect
-            source="department"
-            label="Department"
+          <DigitFormMultiSelect
+            source="departments"
+            label="Departments"
             reference="departments"
-            placeholder="Select department..."
+            validate={v.required}
+            help="First one checked becomes the primary department (used for routing)."
```
vinothrallapalli-eGov: @Lokendra-egov The labels, help texts should support multiple locale, so don't hardcode.

vinothrallapalli-eGov (approved):

configurator/src/resources/complaint-types/ComplaintTypeEdit.tsx
Lokendra-egov: @vinothrallapalli-eGov Fixed in f20adb08

PR #1726, opened 2026-08-11
fix(dashboard): surface analytics batch errors

KDwevedi:
## Summary

Part of #1050.

Fixes a backend/frontend analytics contract mismatch that currently lets authorization and query failures render as No data.

The backend has always returned per-entry failures inline:

    results.tile = { error, message }

The dashboard instead reads a top-level errors map that the backend never emitted. As a result:

- kpi_forbidden and pii_forbidden can bypass the existing tile error UI
- companion query failures can remain silent
- dashboard error-widget telemetry reports zero
- composed-KPI denial can leave partial=false

## Backend compatibility

Batch analytics responses now additively emit an errors index while retaining every legacy inline error entry:

    {
      results: {
        ok: { rows: [] },
        denied: { error: kpi_forbidden, message: ... }
      },
      errors: {
        denied: { error: kpi_forbidden, message: ... }
      },
      partial: true
    }

The legacy results shape is unchanged for existing consumers. partial is derived from the canonical error index, including the composed-result fast path.

Single-query responses are unchanged.

## Frontend compatibility

runKpiBatch now normalizes at the transport boundary:

- legacy inline errors
- new top-level string errors
- top-level {error,message} errors
- top-level {code,message} errors
- base and __prior, __series, or __pins companion failures
- partial repair when any error exists
- no mutation of source payloads
- reserved keys such as __proto__ and constructor are handled as ordinary own keys

runBatchQueries remains raw, so filter-option fallback behavior continues to consume the legacy inline result shape.

AdminDashboard passes the canonical {code,message} directly to the existing KpiTile error branch and uses the same shape for whole-batch failures.

## Why this precedes ABAC enforcement

ABAC denials must be distinguishable from valid empty datasets before policy evaluation is connected to analytics. Treating denied as empty is both misleading to users and invisible to observability.

This PR changes error presentation and accounting only. It does not change which KPI is authorized, query planning, SQL, scope calculation, or catalog visibility.

## Tests

Backend contract tests cover:

- mixed success and kpi_forbidden
- caught per-entry exceptions
- composed-KPI denial
- successful empty error index and partial=false
- retention of legacy inline errors

Frontend tests cover:

- legacy and new error shapes
- explicit-error precedence
- valid empty rows
- base and companion resolution
- non-mutation
- reserved object-property keys

## Validation

- full pgr-services suite on JDK 17: 200 passed, 1 existing skip
- full dashboard Node suite: 181 passed
- production npm run build: succeeded
- git diff --check: passed
- build emits only the repository existing duplicate-key and direct-eval warnings

### DIFF
--- a/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsService.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsService.java
@@ -119,9 +119,12 @@ public class AnalyticsService {
         out.put("scope", scopeInfo(scope));
 
         if (body.has("queries") && body.get("queries").isObject()) {
-            // batch dict form: { name -> queryNode } => { results: { name -> result }, partial }
+            // Batch dict form. Error entries stay inline under results for wire compatibility,
+            // and are also indexed under the additive top-level errors map consumed by the
+            // dashboard. This lets old callers keep working while data and errors have an
+            // unambiguous transport seam for new callers.
             Map<String,Object> results = new LinkedHashMap<>();
-            boolean partial = false;
+            Map<String,Object> errors = new LinkedHashMap<>();
             Iterator<Map.Entry<String,JsonNode>> it = body.get("queries").fields();
             while (it.hasNext()) {
                 Map.Entry<String,JsonNode> e = it.next();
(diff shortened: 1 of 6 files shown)


KDwevedi:
Closing as superseded by merged #1781. `master` now normalizes legacy inline and optional top-level analytics errors, chunks oversized batches, preserves partial/error telemetry, and deliberately keeps a successful base tile visible when only a companion query fails. Replaying this draft would reintroduce the older companion-failure behavior and conflict with that reviewed contract. The backend may still add a top-level `errors` index later, but current consumers no longer require it and already accept it additively.

PR #1723, opened 2026-08-11
refactor(pgr): extract role-safe ABAC policy primitives

KDwevedi:
## Summary

Part of #1050.

Extracts a behavior-neutral ABAC policy core from #1441 onto current master. This establishes the policy types and fail-closed invariants that later search, analytics, and dashboard integrations can share without changing any endpoint behavior in this PR.

The core provides:

- exact tenant, HTTP method, URL, and canonical role-set policy identity
- a pure access-policy source boundary whose inputs exactly match the cache key
- explicit resolved, unauthorized, unavailable, invalid-request, and invalid-policy outcomes
- deeply immutable JSON-compatible policy documents
- a parsed condition seam from PolicyAction to JsonLogic evaluation
- fail-closed evaluation requiring an explicit Boolean true verdict
- success-only TTL caching with role isolation, expiry, and a hard concurrent capacity bound
- deterministic regression coverage for cache and failure boundaries

## Why this is a draft

The stale #1441 branch combines policy transport, row scope, field masking, analytics, frontend behavior, seeds, and cleanup across a large diff. This draft separates the reusable policy boundary so its security semantics can be reviewed before those concerns are wired together.

This is the first slice of the cleanup program, not ABAC enforcement and not the final backend/frontend contract.

## Deliberately excluded

- an egov-accesscontrol transport adapter
- a Spring-managed registry or runtime call site
- trusted identity and tenant binding
- search, count, or plain-search enforcement
- department and jurisdiction scope derivation
- field masking and dashboard analytics authorization
- frontend rendering changes
- configurator or seed changes
- retirement of existing dashboard RBAC

The legacy action enabled field is intentionally not interpreted as an authorization kill switch. Existing role-mapped API actions, including the target PGR action, use enabled=false; role-visible source resolution remains authoritative until the platform contract is clarified.

## Backend/frontend direction

The desired split is still the target: the backend owns what data and presentation or masking instructions are permitted, while the frontend renders that contract without independently reconstructing authorization policy. This PR supplies only the backend policy-document primitive needed for that later envelope.

## Follow-up slices

1. Bind the registry to a validated principal and a concrete access-control adapter, initially in shadow mode.
2. Derive one canonical scope object and apply it consistently to search, count, plain search, and analytics.
3. Define the backend-owned field and presentation envelope; keep the frontend renderer declarative.
4. Gate analytics with the same policy boundary, reconcile seeds, then retire duplicate legacy RBAC paths.

## Security notes

[...]

PR #1727, opened 2026-08-11
feat(pgr): add access-control policy source

KDwevedi:
## Summary

Adds the concrete, role-scoped transport behind the policy primitives in #1723.

- introduces a lazy `AccessControlPolicySource` for `POST /access/v1/actions/mdms/_get`
- sends the exact canonical role set, tenant, and `actionMaster`; deliberately sends neither caller `RequestInfo` nor `enabled`
- uses a dedicated client with explicit 2s connect / 5s read timeouts
- filters on exact method + URL while ignoring legacy method-less actions
- preserves the full opaque policy document, including nested resource/condition values and `enabled:false`
- rejects malformed/duplicate-key successful responses as `INVALID_POLICY`
- keeps network/HTTP failures distinct as `SOURCE_UNAVAILABLE` and clean empty results as `NOT_AUTHORIZED`
- preserves duplicate exact matches so the registry rejects ambiguous policies
- prevents upstream response bodies/policy data from leaking through registry error logs

## Runtime boundary

This slice is intentionally dormant. The source bean is lazy, `AccessPolicyRegistry` is still not a Spring bean, and there is no request-path call site or enforcement change.

It does **not** solve trusted identity. No caller-authored `RequestInfo` is used or forwarded.

## Stack and rollout dependencies

- stacked on and depends on #1723; until that merges, this PR shows both commits
- wire response depends on Digit-Core [#1403](https://github.com/egovernments/Digit-Core/pull/1403) projecting `method`, `resource`, and `condition`
- the enriched `actions-test` policy seed must be deployed before enforcement
- the activation PR must wire `EGOV_ACCESSCONTROL_HOST` for PGR Compose/Kubernetes/Helm deployments
- live-probe a multi-role response before enforcement to confirm the endpoint does not duplicate role-visible actions unexpectedly
- trusted-principal resolution and registry/call-site wiring remain later slices

`enabled` is intentionally omitted: the canonical policy action 2008 is `enabled:false` as a navigation action but remains the role-mapped ABAC policy document.

Tracks #1050.

## Verification

- Java 17 focused policy tests: 36 passed
- Java 17 full PGR suite: 232 tests, 0 failures, 0 errors, 1 existing skip
- `git diff --check`
- independent correctness/security and integration re-reviews: no blockers
- Claude read-only review run without a configured budget cap; its startup-inertness and unrelated-row findings were incorporated

### DIFF
--- a/backend/pgr-services/pom.xml
+++ b/backend/pgr-services/pom.xml
@@ -141,6 +141,11 @@
             <version>4.13.2</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.github.jamsesso</groupId>
+            <artifactId>json-logic-java</artifactId>
+            <version>1.1.0</version>
+        </dependency>
     </dependencies>
     <repositories>
      		<repository>
--- a/backend/pgr-services/src/main/java/org/egov/pgr/config/PGRConfiguration.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/config/PGRConfiguration.java
@@ -127,6 +127,22 @@ public class PGRConfiguration {
     @Value("${egov.mdms.search.endpoint}")
     private String mdmsEndPoint;
 
+    // Access-control policy source
+    @Value("${egov.accesscontrol.host}")
+    private String accessControlHost;
+
+    @Value("${egov.accesscontrol.actions.mdms.get.path}")
+    private String accessControlActionsMdmsGetPath;
+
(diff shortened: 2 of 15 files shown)


KDwevedi:
Closing as superseded by merged #1441. The merged contract resolves role-scoped action 2008 through `MDMSUtils`, shares that lookup between scope and field policy, and applies the reviewed strict-mode/fail-closed semantics. This draft’s separate transport/registry wiring is a competing implementation and is intentionally not being rebased onto `master`.

Issue #520, opened 2026-05-04
[Bug] Citizen UI: Swahili language does not reflect for data within the map


Gurjeet-eGov:
Swahili language does not reflect for data within the map.

**To Reproduce**
1. Go to citizen ui
2. login to a valid user
3. create a complaint 
4. change language to swahili
5. Reach the step to select location on map

**Expected behavior**
Map shall reflect the data in swahili language 

https://github.com/user-attachments/assets/5a75feff-ee2e-4bfa-b620-ccfeee495b68

prtkjakhar:
@KDwevedi @ChakshuGautam — partial fix merged in [theflywheel/digit-ui-esbuild#95](https://github.com/theflywheel/digit-ui-esbuild/pull/95):

**What's fixed**: the four Nominatim calls (reverse-geocode + search + suggestion + click-to-resolve) now pass `Accept-Language` derived from the active i18n locale (`sw_KE` → `sw`, falling back to `en`). After deploy, the address tooltip + search results on the Pin-Location step will read Swahili when the citizen switches the UI to Swahili.

**Still out of scope — needs your call**: the place / street labels drawn ON the map tiles are baked into CARTO's `dark_all` raster tiles in English. No header can change them. Two paths:

1. **Frontend swap to a vector-tile provider** (MapTiler / OpenMapTiles + a localized style). Supports runtime language switching, but needs an API key and a small style rework.
2. **Backend / infra**: self-host a localized OSM tile server.

Could you confirm which direction you want to go before we open a follow-up?

KDwevedi:
@KDwevedi to figure out if we can just inject label switches.

KDwevedi:
## Findings — what's actually feasible

Searched the OSM/Leaflet ecosystem and verified each option against our current `react-leaflet` + CARTO `dark_all` setup (`ComplaintLocationMap.js:167`, `GeoLocations.js:393`). The "just inject labels" intuition is half-right — the constraint is that **once a tile is rasterized, the text is image data, not text data**, so you cannot override it with CSS/JS. Any real fix moves the labels off the CARTO raster onto a vector source we can re-render per locale.

### Ranked options

1. **MapTiler + `leaflet-maptilersdk` plugin** — best fit
   - Drop-in Leaflet layer (`L.MaptilerLayer`), `setLanguage(L.MaptilerLanguage.SWAHILI)` at runtime.
   - 70+ languages including Swahili.
[...]

PR #1728, opened 2026-08-11
docs: OpenStreetMap technology and CMS integration guide

KDwevedi:
## What

`docs/openstreetmap/` — a guide to the OpenStreetMap stack that every geospatial feature in CCRS sits on (citizen complaint pin maps, configurator boundary onboarding, dashboard choropleths).

Parts 10–30 describe the OSM ecosystem and stay true regardless of our code. Parts 40–50 describe our implementation and our history, and go stale when we ship.

| Part | Covers |
|---|---|
| 10 — Data | The OSM data model, `admin_level`, ODbL obligations, and where boundary data comes from (OSM and otherwise) |
| 20 — Services | Overpass and Nominatim: query shapes, real quotas, the policies that bind us, and what self-hosting costs |
| 30 — Rendering | Tiles, client libraries, providers, attribution |
| 40 — Our implementation | Every CCRS call site, deploy flags, `boundary-service` quirks, known gaps |
| 50 — What we have onboarded | Maputo, Bomet, Nairobi: what worked and what OSM could not supply |

## Why

The knowledge was spread across five code paths, three READMEs and tribal memory. Things we learned the hard way and had recorded nowhere durable:

- **Nominatim's policy forbids client-side autocomplete** — *"you must not implement such a service on the client side using the API."* Debouncing does not make it compliant; the pattern is barred, not the rate. This is the actual justification for `turbopass/` existing, and its README now says so instead of looking like reinvention.
- **OSM's `name` is in the local language.** A strict `["name"="X"]` returns zero results for any place picked by its translated name. We shipped this bug (#757 → #980); part 50 records it in full.
- **`bounded=1` is a silent killer.** A stale hardcoded viewbox makes Nominatim return nothing for every other tenant, with no error.
- **`admin_level` is not portable**, and levels are non-contiguous — "wards are level 8" is wrong more often than right.
- **`OVERPASS_USE_AREAS` enables area generation, not `OVERPASS_RULES_LOAD`** (which only sets the updater's work/sleep ratio). Without area generation a self-hosted Overpass imports fine, answers `rel[…]` fine, and returns empty for every `area[…]` query. Part 20 gives the triage order: extract coverage → area generation → local-language name match.

## What part 50 adds

Three real onboardings, with figures verified against the onboarding artefacts rather than recalled:

- **Maputo** — OSM supplied 70 polygons (1 city + 6 of 7 distritos + 63 of 63 bairros) against a **1,256-boundary** kit. 68 matched kit codes; 2 carry no `code` property and are silently dropped. All **1,185 quarteirões — 94% of the tenant, and the level closest to service delivery — have no OSM data at all.** The lesson: assume OSM depth runs out one level above where the work happens, and plan the fallback before onboarding.
- **Bomet** — the 25 ward polygons did *not* come from OSM. They came from the HDX *administrative-wards-in-kenya-1450* shapefile (CC BY; American Red Cross/NLC/IEBC/SOK), because OSM lacked ward coverage. Joining needed fuzzy name matching (`Ndaraweta` ↔ `NADARAWETA`), and backfilling an existing tenant is expensive because `boundary-service` has no `/boundary/_update`.
- **Nairobi** — the retired static ward sidecar *half*-worked: ward resolution succeeded, locality autofill silently found no match, and the citizen was left filling a mandatory dropdown by hand. Harder to diagnose than a clean failure. #874 tracks this and is still open.

## Reviewer notes

- **Docs only.** No code, config, schema or behaviour change. The three modified files each gain a short cross-link block and nothing else.
- **Placement** follows the existing multi-part guide convention (`dashboard-configuration/`, `notifications-guide/`): a `README.md` index with a linking table plus `NN-topic.md` parts. No frontmatter, no nav file.
- `docs/map-config.md` remains authoritative for the `RAINMAKER-PGR.MapConfig` record; this guide does not restate it. Cross-links run both ways, and from `overpass/README.md` and `turbopass/README.md`.
- **Part 40 ends with a 6-item gap list.** The headline: our *operator-side* OSM dependencies are hardened behind `enable_overpass`/`enable_turbopass`; the *citizen-side* ones are not — Nominatim and basemap tiles are public keyless services on the complaint-filing hot path, with no self-host option and no identifying `User-Agent`. Gaps 1 and 2 are one architectural decision. **No PRs or issues are raised off that list here.**
- **Context for #520** (Swahili not reflected in map data): raster tile labels cannot follow the locale — they are baked into the CARTO tiles, so the `Accept-Language` fix that solved the address-text path cannot reach them. Parts 20/30 sketch the vector-tile route. Not claiming a fix.
- Source **line numbers were deliberately removed** from part 40 — they decay immediately. File paths, env vars, flags and ports remain.

## Verification

- All referenced repo paths confirmed to exist; all 32 relative links resolve (scripted, 0 broken).
- Part 50's figures computed directly from the onboarding XLSX and GeoJSON, not from notes: 1,256 boundary rows by type, 70 sidecar features by level, and the 68/70 code-match with the two unmatched features named.
- External policy claims fetched from source: [Nominatim usage policy](https://operations.osmfoundation.org/policies/nominatim/), [tile usage policy](https://operations.osmfoundation.org/policies/tiles/), [Overpass API wiki](https://wiki.openstreetmap.org/wiki/Overpass_API).
- Reviewed twice: once for technical accuracy (which corrected the ODbL Produced Works wording, the Overpass area-generation variable, the Nominatim `UPDATE_MODE` requirement, and the Overpass string-escaping claim), then once for verbosity and clarity, which halved it — 1,269 → 635 lines, 10.0k → 4.8k words, six parts consolidated into four.

One incidental finding, left in the gap list rather than fixed to keep this docs-only: the `osm` basemap theme uses the retired `{s}.tile.openstreetmap.org` subdomain form, while the current policy specifies plain `tile.openstreetmap.org`.


docs/openstreetmap/30-rendering.md
```diff
@@ -0,0 +1,67 @@
...
+| Ops complexity | Low | Style JSON, glyphs, sprites |
+
+For a multi-language municipal platform, runtime label localisation is the strongest
+argument for vector: raster tile labels cannot follow the user's locale, so a Swahili or
+Portuguese UI still shows English map labels. That is the one constraint here that has
+actually bitten us — see [part 40](40-ccrs-implementation.md).
+
+## Libraries
+
+CCRS uses **[Leaflet](https://leafletjs.com) 1.9** everywhere, called directly rather than
```
vinothrallapalli-eGov: @KDwevedi Do not use CCRS in any new document let call CMS (Compliant Management System). Please do this change in other documents in the PR

vinothrallapalli-eGov (changes requested):

docs/openstreetmap/30-rendering.md
KDwevedi: Addressed in 3ef92179c. Replaced CCRS with the established DIGIT Complaint Management System (DIGIT CMS) nomenclature across the new OSM guide, renamed Part 40 to `40-cms-implementation.md`, and updated every inbound link. Validation: 32/32 relative Markdown links resolve, no added OSM-guide line retains `CCRS`, no old Part 40 link remains, and `git diff --check` passes.

vinothrallapalli-eGov (approved):

Issue #1707, opened 2026-08-10
[Bug] Configurator UI: Data count on landing page is not the correct as per actual data


Gurjeet-eGov:
# Description

Data count on landing page is not the correct as per actual data 

---

# Steps to Reproduce

1. Login to configurator
2. Notice the department and designation data count on ui
3. Click on each card and check actual count 

---

# Expected Behavior

Count shall be correct

---

# Actual Behavior

https://github.com/user-attachments/assets/69deda67-abe0-47b2-83cd-7bc1827c03c5

Gurjeet-eGov:
Unresolved 

<img width="1366" height="660" alt="Image" src="https://github.com/user-attachments/assets/33e886f3-f771-44aa-96b5-695444aa35f1" />

<img width="1046" height="217" alt="Image" src="https://github.com/user-attachments/assets/5a3c6eb4-8b22-4e49-abf9-c9b91397ca0d" />

KDwevedi:
Same fix as #1817

Gurjeet-eGov:
Resolved 

https://github.com/user-attachments/assets/66e8771a-186b-492a-ac22-80f996d7a0b1

PR #1731, opened 2026-08-11
fix(configurator): landing-page resource counts match list totals (#1707)

soniikajal:
## Summary
Fixes [#1707](https://github.com/egovernments/Citizen-Complaint-Resolution-System/issues/1707): Configurator landing-page badges showed the wrong totals compared to the resource lists.

- Dashboard cards fetched with `perPage=1` and MDMS used a `perPage+1` "has more" heuristic, so tenants / departments / designations often showed **2**. MDMS v2 returns no total, so the provider now pages the active set and counts it.
- Localization counted `en_IN` first, then jumped after a delayed locales filter (e.g. 7902 → 14095). The card and the list now pin the same `locales` filter.
- Boundaries queried leftover `PW_*` test hierarchies and missed `ADMIN`, so the card showed **0**. Hierarchy discovery pages past the `PW_*` stubs.
- Expired tokens on HRMS / PGR / access-control (`InvalidAccessTokenException`, and bodyless gateway 401s) left those cards at 0 while MDMS still loaded; treat that as a dead session and bounce to login.


KDwevedi (commented):
1. **Generic MDMS sorting regression**
The generic no-filter MDMS path server-pages records but neither passes the requested sort to the API nor calls clientSort, so sort.field/order is ignored.
Please restore sorting and add a shuffled-response test. [Relevant code](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/ac693ce08f6bc1e80f06fcbb6f86777ebb62b0cb/configurator/packages/data-provider/src/providers/dataProvider.ts#L1004-L1028)

2. **Bodyless 401 bypasses session-expiry handling**
request() calls response.json() before classifying the HTTP status; an empty or non-JSON 401 throws SyntaxError and never invokes the expiry handler.
Please safely parse the body and add a request-level bodyless-401 test. [Relevant code](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/ac693ce08f6bc1e80f06fcbb6f86777ebb62b0cb/configurator/packages/data-provider/src/client/DigitApiClient.ts#L85-L112)

3. **PR description cleanup**
Please remove the claim that complaints used the current page length as the total.
The PGR branch already used the _count endpoint before this PR, and this change does not modify that behavior.

KDwevedi (approved):
Root cause independently confirmed: the MDMS perPage+1 heuristic caused the landing-card/list mismatch. Follow-up pagination, truncation, sorting, and bodyless-401 review findings are addressed. Offline provider tests and the production build pass.

PR #1753, opened 2026-08-13
fix(migration): register CMS_ADMIN + CMS_DASHBOARD_VIEWER in the --cms phase (CCSD-1937)

Hari-egov:
Jira: [CCSD-1937](https://digit-discuss.atlassian.net/browse/CCSD-1937)

## Problem

QA (Gurjeet) flagged that `CMS_ADMIN` is missing from access-control on the env. Root cause: the DDH seed files already contain the role row **and** its 22 roleaction grants (CMS_DASHBOARD_VIEWER too, 24 grants) — but `ccrs-migrate.cjs`'s `CMS_ROLES` filter only listed the 4 workflow roles + `CMS_VIEWER`, so the `--cms` phase never registered them. Without the role code registered, HRMS cannot assign `CMS_ADMIN` to anyone, so the admin cross-department search (gated on `SUPERUSER + CMS_ADMIN`, PRs #1260–#1262) was reachable by SUPERUSER only.

## Fix

One-line list change (+ explanatory comment): add `CMS_ADMIN` and `CMS_DASHBOARD_VIEWER` to `CMS_ROLES`. No seed-data changes needed — the rows/grants were already in the DDH files. Ensure-semantics unchanged: existing rows untouched, safe to re-run.

## Verification

Ran `--only cms` against **cms-pilot** with this change: `roles: 7 ensured @ mz`, `actions + grants ensured (227 grants)`; MDMS search confirms `CMS_ADMIN` now registered at mz. Dry-run first, and scoped with `--only cms` to avoid the pgr-masters phase (which would have re-added the second ComplaintRelatedToMap authority row on pilot).

## Notes for reviewers

- The `/pgr-services/v2/request/_admin/_search` endpoint has **no actions-test row anywhere** — its role gate is enforced in-service (`AdminComplaintSearchService`), which is why SUPERUSER worked without one. Left as-is; add a catalog row only if we adopt per-URI gateway RBAC.
- The ticket's remaining 5 codes (CENTRAL_USER, DEPARTMENT_USER, COMPLAINTS_VIEWER/EDITOR/CREATOR) are referenced nowhere in FE or backend — dept scoping ships via `pgr.department.scope.roles` + EmployeeDepartmentScopeService instead. Registering them is a product decision tracked on the ticket, deliberately out of scope here.

[CCSD-1937]: https://digit-discuss.atlassian.net/browse/CCSD-1937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/migration/ccrs-migrate.cjs
+++ b/docs/migration/ccrs-migrate.cjs
@@ -760,12 +760,14 @@ async function phaseLanding() {
 
 /* ═════════════════════════════ PHASE: cms ═════════════════════════════ */
 
-// CCSD-1937: CMS_ADMIN and CMS_DASHBOARD_VIEWER were in the DDH seed files
-// (role rows + grants) but absent from this list, so --cms never registered
-// them — CMS_ADMIN could not be assigned via HRMS and the admin
-// cross-department search (SUPERUSER + CMS_ADMIN, PR #1260-#1262) was
-// reachable by SUPERUSER only.
-const CMS_ROLES = ['CMS_RECEPTION_OFFICER', 'CMS_SCREENING_OFFICER', 'CMS_SUPERVISOR', 'CMS_CASE_MANAGER', 'CMS_VIEWER', 'CMS_ADMIN', 'CMS_DASHBOARD_VIEWER'];
+// CCSD-1937: CMS_ADMIN, CMS_DASHBOARD_VIEWER and CONFIDENTIAL_COMPLAINT_VIEWER
+// were in the DDH seed files (role rows + grants) but absent from this list,
+// so --cms never registered them. Consequences: CMS_ADMIN could not be
+// assigned via HRMS (admin cross-department search, PR #1260-#1262, reachable
+// by SUPERUSER only), and employee onboarding that assigns any of these roles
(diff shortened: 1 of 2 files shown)


Hari-egov:
Scope extended per the deploy-completeness requirement ("deploy this and everything should be up — no 'Role not valid' errors"):

- `3ea00efc` — also register `CONFIDENTIAL_COMPLAINT_VIEWER` (same seed-file-present/runner-omitted gap; hit during new-server IGE employee onboarding).
- `6e9f1fe5` — **the deploy now seeds CMS access-control itself**: new post-bootstrap playbook task runs `ccrs-migrate --only cms` against the local gateway (after ADMIN provisioning, before configurator-i18n). Idempotent, 3 retries, and the runner's non-zero exit on FAILED/PARTIAL fails the deploy loudly. `ansible-playbook --syntax-check` passes.

Net effect: a fresh `./deploy.sh <tenant>` comes up with all 8 CMS roles + 243 grants + PGR workflow registered — employee onboarding works with zero manual steps.

PR #1756, opened 2026-08-14
Pull Master into Moz release

Hari-egov:
Pull Master into Moz release


Hari-egov:
Superseded by #1757 — head of this PR is `master` itself, so the conflict could not be resolved here (same situation as the #1653 sync). #1757 carries master + the resolved conflict, with all three touched stacks build-verified.

PR #1763, opened 2026-08-14
feat(dashboard): geography drill-down filter — Província → Distrito → Município (CCSD-2171)

Hari-egov:
Jira: [CCSD-2171](https://digit-discuss.atlassian.net/browse/CCSD-2171) — **PR 2 of 2 (FE), stacked on #1762** (backend `boundaryPath` param; merge #1762 first — GitHub will retarget this to `release-v2.12-moz` automatically).

## What

The dashboard's flat ward `<select>` becomes a **boundary drill-down** — Província → Distrito → Município — with the exact interaction model of the complaint-type filter (chip + traversal panel: trail, descend-in-place, "All in ⟨X⟩", reset):

- **Same panel, not a copy** — `ComplaintTypeTreePanel` gains three optional props (`labelFor`/`allLabel`/`allInLabel`, defaults preserve today's behaviour bit-for-bit); `GeographyTreeFilter` reuses it wholesale.
- **Tree source**: `boundary-relationships/_search` at the tenant's `HIERARCHY_TYPE` (globalConfigs pin; hierarchy-definition discovery fallback), pruned against the same ABAC-scoped `DISTINCT ward_code` the flat list uses — users only see branches they can access. Any failure → **flat select stays** (the complaint-type degrade path verbatim).
- **Wire contract**: leaf ward → `params.ward` (unchanged, works on every backend today); interior node → `params.boundaryPath` (subtree filter from #1762; older backends ignore the unknown param, so interior selections degrade to unfiltered — never an error). Selection persists as the `{ code, path, leaf }` trio with sanitize/repair/hold-through-hiccup rules mirroring complaintType.
- **Path fidelity**: FE pipe-paths are byte-identical to the analytics MV's `boundary_path` — verified live on cms-pilot (`maputo_cidade|kanyaka|municipio_maputo_kanyaka` from the facts grain matches the relationships chain).
- **l10n**: `DASHBOARD_GEO_FILTER_ALL_IN` seeded in en_IN + pt_PT (enforced by the localization contract test, which caught it).

## Verification

- New `boundaryTree.test.js` — 10 tests (MV-path fidelity, pruning + strays, leaf/interior/legacy params, invalid-path degrade, repair walk-up, normalize, humanize).
- **Full dashboard sweep: 238/238 green** (includes the tree-panel render smoke and the localization contract).
- `npm run build` (esbuild bundle): clean.

## Behaviour matrix

| Situation | Result |
|---|---|
| Tree + #1762 backend deployed | full drill-down, filter at any level |
| Tree present, old backend image | leaves filter normally; interior selection = unfiltered (silent, complaintPath precedent) |
| Boundary fetch fails / flat tenant | today's flat ward select, untouched |
| Persisted old string filter | keeps working as a leaf ward (legacy contract) |

[CCSD-2171]: https://digit-discuss.atlassian.net/browse/CCSD-2171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/digit-ui-esbuild/products/dashboard/src/components/ComplaintTypeTreeFilter.jsx
+++ b/digit-ui-esbuild/products/dashboard/src/components/ComplaintTypeTreeFilter.jsx
@@ -61,8 +61,22 @@ const TRAIL_MAX = 4;
  * The panel body. Exported (also for the ReactDOMServer render smoke): pure
  * React against the tree + applied code, owns only the transient browse
  * location. Mounted fresh on every open, so browse state self-resets.
+ *
+ * Hierarchy-agnostic via three optional props (defaults keep the
+ * complaint-type behaviour bit-for-bit; the geography drill-down passes its
+ * own — CCSD-2171): `labelFor(tree, code)` resolves a node's display label,
+ * `allLabel` is the virtual-root label ("All types" / "All wards"),
+ * `allInLabel` prefixes the subtree-apply row ("All in").
  */
-export function ComplaintTypeTreePanel({ tree, appliedCode, onApply, t }) {
+export function ComplaintTypeTreePanel({
+  tree,
+  appliedCode,
+  onApply,
+  t,
+  labelFor = nodeDisplayLabel,
+  allLabel,
+  allInLabel,
+}) {
   const [browseCode, setBrowseCode] = useState(() => browseBaseCode(tree, appliedCode));
   const rootRef = useRef(null);
   const mountedRef = useRef(false);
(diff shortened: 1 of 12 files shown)


Hari-egov:
**Local end-to-end run completed** (full stack on this box: branch pgr-services image + branch digit-ui bundle, mz dataset with 88 boundary-coded complaints):

- **Backend, against live SQL ground truth** — `subtree` on `boundary_path` returns exactly the DB counts: total 88, `tete` subtree **23/23**, `tete|dist_tete` **12/12**, `ward eq municipio_matola` **16/16**. Pipe-guard predicate confirmed in the running image.
- **UI, real browser**: chip renders, panel lists ABAC-pruned humanized provinces, descend into Tete shows the trail + "All in Tete" + mixed-depth children (2-level district-leaves like Tsangano next to 3-level interiors — depth variance handled). Interior apply → chip "Tete", header subtitle "Tete · …", persisted trio `{code:"tete", path:"tete", leaf:false}` → `boundaryPath=tete`. Leaf apply → "Tete › Tsangano", `{leaf:true}` → `ward=tsangano`. Reopen shows the applied node checked at its browse base.
- **One real bug found and fixed by this run** (amended into the commit): `boundaryHierarchyService` called `authFetch` with a plain `body`, but authFetch's contract is `buildBody` (a function, re-invoked with fresh RequestInfo on 401-refresh replays) — the request went out body-less and 400'd. The graceful degrade masked it perfectly (flat select stayed), which is exactly why it needed a live run to surface. Now uses `buildBody` + `sessionCritical:false`, matching `boundaryService.js`.

Environment note for reviewers: the local box originally served the boundary tree only at city tenants; cms-pilot serves it at the state tenant (verified live), which is the layout this feature targets — the local DB was aligned to pilot's layout for the test.

PR #1774, opened 2026-08-17
[Configurator/PGR] Hide legacy dashboard behind default-off flag (#1386)

### LINKED ISSUE #1386: [Bug] Admin unable to download PGR Dashboard report from configurator UI
# Description

Admin unable to download PGR Dashboard report from configurator UI

---

# Steps to Reproduce

1. Login to config ui
2. Go to PGR Dashboard
3. Click on Download Report

---

# Actual Behavior

It says generating report, and then no file is downloaded 

https://github.com/user-attachments/assets/027529d1-4749-4706-a7ac-688199b49e64

KDwevedi:
## Decision implemented

Closes #1386. Implements the direction in https://github.com/egovernments/Citizen-Complaint-Resolution-System/issues/1386#issuecomment-5313867834:

- stop maintaining the legacy report-download behavior;
- hide the legacy PGR dashboard from UI by default;
- retain it behind an explicit feature flag;
- do not delete or deprecate its code yet.

## What changed

- Adds one default-off build flag across both legacy copies: `VITE_ENABLE_LEGACY_PGR_DASHBOARD`.
- Configurator hides the PGR Dashboard sidebar entry and redirects direct `/manage/pgr-dashboard` visits when disabled.
- The citizen PGR SPA hides its copied Citizen Dashboard entry and redirects direct `/dashboard` visits to `/complaints`; password and Keycloak login now use the same flag-selected landing.
- Wires the flag through the supported Ansible builds and both Dockerfiles. Operators can temporarily restore the surface with `enable_legacy_pgr_dashboard: true` (Ansible) or the equivalent Vite build arg.
- Keeps both dashboard implementations, their data hooks, weekly-report code, and `/pgr-services/v2/dashboard` backend API intact.
- Updates E2E expectations for the default-off contract while retaining an opt-in Configurator verification path.

This is isolated legacy-surface work and does not change the catalog-driven employee/public dashboard.

## Verification

- `npm test` in `configurator`: 17 files, 120 tests passed.
- Configurator production build passed with the flag absent/default-off and with `VITE_ENABLE_LEGACY_PGR_DASHBOARD=true`.
- Citizen feature-flag tests: 13 passed.
- `npm run build` in `digit-ui-v2` passed with the flag absent/default-off and explicitly enabled.
- `ansible-playbook --syntax-check` passed for `playbook-deploy.yml` (example inventory filename is not auto-detected, so Ansible reports only the existing inventory warning).
- `git diff --check` passed.

## Rollout

Rebuild/redeploy both Configurator and digit-ui-v2. No backend or database migration is required.


configurator/src/App.tsx
```diff
@@ -161,7 +162,10 @@ function ManagementAdmin() {
         <CustomRoutes>
           <Route path="/notification-configure" element={<NotificationConfigure />} />
           <Route path="/advanced" element={<AdvancedPage />} />
-          <Route path="/pgr-dashboard" element={<PgrDashboard />} />
+          <Route
+            path="/pgr-dashboard"
+            element={LEGACY_PGR_DASHBOARD_ENABLED ? <PgrDashboard /> : <Navigate to="/" replace />}
```
vinothrallapalli-eGov: **Bug:** Disabling the legacy dashboard redirects to `/` instead of directly to `/manage`, chaining into a second, non-`replace` redirect and leaving a stray history entry.

**Scenario:** An operator follows a stale link/bookmark to `/manage/pgr-dashboard` with the flag off (the default): this `<Navigate to="/" replace>` lands on the outer `path="/"` route, which then fires `<Navigate to="/manage" />` (no `replace`) — two silent client-side redirects and an extra un-replaced history entry. Clicking Back from `/manage` lands back on `/`, which immediately auto-forwards to `/manage` again, so Back is effectively a no-op/loop for that hop.

Consider redirecting straight to `/manage` here to avoid the double hop.

configurator/e2e/tests/pgr-dashboard.test.mjs
```diff
@@ -13,14 +13,32 @@ export async function run() {
...
-    // Navigate to PGR Dashboard
     const baseUrl = process.env.E2E_BASE_URL || 'https://crs-mockup.egov.theflywheel.in';
     await page.goto(`${baseUrl}/manage/pgr-dashboard`, {
       waitUntil: 'networkidle2',
       timeout: NAV_TIMEOUT,
     });
 
-    // Verify we're on the right page
+    const legacyEnabled = /^(1|true|yes|on)$/i.test(
+      process.env.ENABLE_LEGACY_PGR_DASHBOARD || '',
```
vinothrallapalli-eGov: **Bug:** This 'is legacy mode on' check reads `process.env.ENABLE_LEGACY_PGR_DASHBOARD`, a runtime test-env var with no relation to `VITE_ENABLE_LEGACY_PGR_DASHBOARD`, the build-time flag actually baked into the bundle under test — nothing cross-checks the two.

**Scenario:** CI/an operator runs this suite against a rollback image built with `VITE_ENABLE_LEGACY_PGR_DASHBOARD=true` but forgets to also export `ENABLE_LEGACY_PGR_DASHBOARD=true` for the test process — the test then asserts 'hidden by default' against a build where the dashboard is actually reachable, producing a confusing failure (or in the opposite mismatch, silently skipping the deeper rollback-mode assertions) with no diagnostic pointing at the real cause.

configurator/src/config/featureFlags.ts
```diff
@@ -0,0 +1,12 @@
+/** Build-time feature flags for intentionally retained legacy surfaces. */
```
vinothrallapalli-eGov: **Cleanup:** `isEnabledFlag` and its Vitest suite are duplicated byte-for-byte between `configurator/` and `digit-ui-v2/` instead of living in one shared module.

**Scenario:** A future change to the accepted truthy values (e.g. adding a new alias string) is made in one copy and not the other, so the two SPAs silently disagree on what 'legacy dashboard enabled' means for the same ansible-driven deploy — a build using both bundles could end up with the dashboard visible in one and hidden in the other for the same flag value.

vinothrallapalli-eGov:
**Bug (code review):** `tests/integration-tests/tests/admin/pgr-dashboard.spec.ts` was left completely untouched and still unconditionally asserts the legacy PGR dashboard renders (H1 heading, KPI tiles, chart canvases, sidebar nav button), even though this PR's own stated intent is that the dashboard is hidden by default.

**Scenario:** Run this suite against any environment built with the new default (`VITE_ENABLE_LEGACY_PGR_DASHBOARD=false`, which is every documented default in this PR): navigating to `/manage/pgr-dashboard` now redirects away, so all 7 tests fail on missing heading/KPIs/charts/sidebar button — the exact regression this PR fixed in its two sibling suites (`configurator/e2e/tests/pgr-dashboard.test.mjs` and `digit-ui-v2/tests/citizen-login.spec.ts`) but missed here.

(Not posted inline — this file isn't part of the diff.)

configurator/src/App.tsx
KDwevedi: Fixed in de1d6b134: the disabled route now redirects directly to `/manage` with `replace`, avoiding the intermediate `/` route and its extra history entry. The integration gate test also asserts the direct `/manage` destination.

configurator/e2e/tests/pgr-dashboard.test.mjs
KDwevedi: Fixed in de1d6b134: Configurator now publishes the actual baked flag value on `window.__CCRS_BUILD_FLAGS__`. The E2E reads that contract from the loaded bundle and, when `VITE_ENABLE_LEGACY_PGR_DASHBOARD` is provided to the test process, fails with an explicit diagnostic if it does not exactly match the baked raw value.

configurator/src/config/featureFlags.ts
KDwevedi: Fixed in de1d6b134: both SPAs now import `isEnabledFlag` from the single repo-level `ui-shared/featureFlags.ts` module, and the duplicate digit-ui-v2 suite was removed. The canonical 13-case parser suite remains and passes.

KDwevedi:
Addressed the integration-suite gap in de1d6b134. `tests/integration-tests/tests/admin/pgr-dashboard.spec.ts` now reads the flag baked into the loaded Configurator bundle, explicitly verifies the default-off redirect/navigation contract, and skips the seven rollback-dashboard UI assertions unless that bundle has the legacy flag enabled. If the test job supplies `VITE_ENABLE_LEGACY_PGR_DASHBOARD`, the suite also cross-checks it against the baked raw value and reports a mismatch directly. Playwright discovery passes for all eight dashboard tests; both SPA production builds and the shared 13-case flag-parser suite pass.

Lokendra-egov:
Reviewed this one closely since a flag-gating change spanning two "copies" is exactly where partial application tends to hide. Two real issues:

**1. The ansible/VM build path never gets this flag at all.** `digit_ui_v2_repo`/`digit_ui_v2_branch` (both untouched by this PR, in `playbook-deploy.yml` and `_example.yml`) still point at `https://github.com/ChakshuGautam/Citizen-Complaint-Resolution-System.git@feat/digit-ui-v2-keycloak` — an external fork. I fetched that fork/branch directly: no `src/config/featureFlags.ts`, and `App.tsx` still has the unconditional legacy dashboard route. So `VITE_ENABLE_LEGACY_PGR_DASHBOARD` gets passed into the `npm run build` for that clone, but nothing in that source reads it. Any deploy using the default vars (i.e. anything that doesn't override `digit_ui_v2_repo`/`digit_ui_v2_branch`) still ships the legacy dashboard visible by default — the opposite of "default-off across both copies."

**2. `npm run dev` is broken for both apps.** Both `configurator/src/config/featureFlags.ts` and `digit-ui-v2/src/config/featureFlags.ts` import from `'../../../ui-shared/featureFlags'`, outside each app's own Vite root. I reproduced it: copied the exact files and ran `npx vite` for `configurator/` — the resolved module 403s ("outside of Vite serving allow list"), since Vite's workspace-root autodetection stops at `configurator/package-lock.json` and neither `vite.config.ts` sets `server.fs.allow`. Production `vite build` is unaffected, which is presumably why this wasn't caught. Fix is small: add `server.fs.allow` to both `vite.config.ts`s.

Everything else checked out — flag name/default consistent everywhere else it's read, both routes actually guard (not just hide from nav), nothing left dangling from the "keep the code, hide the UI" approach.

KDwevedi:
Addressed in 1ba48da88.

**1. The ansible/VM build path now gets the flag.** Confirmed: `digit_ui_v2_repo`/`digit_ui_v2_branch` defaulted to `ChakshuGautam/...@feat/digit-ui-v2-keycloak`, which has no `src/config/featureFlags.ts`, so the default deploy passed `VITE_ENABLE_LEGACY_PGR_DASHBOARD` into a build that never read it. The citizen SPA now builds from the **vendored in-tree** `digit-ui-v2/` + `ui-shared/` by default — the same rule the configurator and turbopass builds already follow — so a deploy ships the revision it is deployed from. Setting `digit_ui_v2_repo` still selects the clone path for transitional/standalone use, and a new preflight assert fails the play with an explicit message if the selected source lacks the two feature-flag modules, instead of silently shipping the legacy dashboard visible. The example host_vars that pinned external branches (`bomet`, `localhost-full`, `localhost-slim`) are commented out so they use the vendored default, and `_example.yml` + the nginx template comment now describe the vendored path.

**2. `server.fs.allow` added to both `vite.config.ts`.** Reproduced on the pinned Vite 7.3.2: a cold request for `/@fs/<repo>/ui-shared/featureFlags.ts` returns 403 "outside of Vite serving allow list" (it only resolved once the importing module had already been transformed into the graph — fragile either way). Both configs now allow the app root and `../ui-shared`; the same cold request returns 200 for both apps after the change. `vite build` was and is unaffected.

Verification: `ansible-playbook --syntax-check playbook-deploy.yml` passed; all changed YAML parses; the new jinja source-comment expression renders correctly with and without `digit_ui_v2_repo` set; `npm test` in configurator — 17 files / 120 tests passed; production builds of both SPAs passed.

Lokendra-egov (approved):

PR #1778, opened 2026-08-17
feat(pgr): make escalation status states configurable

priyanshu-egov:
- Add pgr.escalation.states property to PGRConfiguration with default values PENDINGATLME,PENDINGFORASSIGNMENT
- Update EscalationScheduler to read escalation states from config instead of hardcoded constants, enabling per-deployment customization
- Add pgr.escalation.states property to application.properties with detailed inline documentation explaining state requirements
- Add RUNBOOK.md documenting the complete escalation feature setup, requirements, configuration, and troubleshooting
- Add enable-escalation.sh script for automated local setup of escalation feature with idempotent steps
- Allows deployments to customize which workflow states trigger escalation based on their BusinessService configuration

### DIFF
--- a/backend/pgr-services/src/main/java/org/egov/pgr/service/EscalationScheduler.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/service/EscalationScheduler.java
@@ -71,8 +71,14 @@ public class EscalationScheduler {
         List<Long> defaultSlaByLevel = getDefaultSlaByLevel(escalationConfig);
         Map<String, List<Long>> overrides = getOverrides(escalationConfig);
 
-        // Search for complaints in PENDINGATLME and PENDINGFORASSIGNMENT
-        Set<String> statuses = new HashSet<>(Arrays.asList(PENDINGATLME, PENDINGFORASSIGNMENT));
+        // Search for complaints in the configured applicationStatus values
+        // (pgr.escalation.states) — deployments customize this to match
+        // whichever states in their BusinessService workflow carry a named
+        // assignee and have an ESCALATE action wired on them.
+        Set<String> statuses = Arrays.stream(config.getEscalationStates().split(","))
+                .map(String::trim)
+                .filter(s -> !s.isEmpty())
+                .collect(Collectors.toSet());
 
         // Get all tenants — for now, use the state-level tenant from config
         // In a multi-tenant setup, this would iterate over all tenants
--- a/backend/pgr-services/src/main/resources/application.properties
+++ b/backend/pgr-services/src/main/resources/application.properties
@@ -178,6 +178,12 @@ pgr.escalation.batch.size=100
 pgr.escalation.default.sla.ms=432000000
 pgr.escalation.max.depth=3
 pgr.escalation.kafka.topic=pgr-escalation-events
+# Comma-separated applicationStatus values the escalation scan searches for
+# candidate complaints. Must match states that actually carry a named
+# assignee in the deployed BusinessService workflow (the scheduler skips any
+# complaint with no current assignee) and that have an ESCALATE action wired
+# on them — otherwise the workflow transition is rejected and nothing moves.
+pgr.escalation.states=PENDINGATLME,PENDINGFORASSIGNMENT
 
 # Dashboard MV refresh
 pgr.dashboard.refresh.enabled=true
--- a/local-setup/scripts/enable-escalation.sh
+++ b/local-setup/scripts/enable-escalation.sh
@@ -1,0 +1,506 @@
+#!/usr/bin/env bash
+# =============================================================================
+# enable-escalation.sh
+#
+# Turn on pgr-services auto-escalation (EscalationScheduler) on a *running*
+# DIGIT / CMS deployment. This is the scripted form of the runbook validated
+# live against tenant `mz`. It does NOT build the pgr-services image — you
+# provide an already-built/pushed image tag (PGR_SERVICES_IMAGE) — and it
+# does NOT guess your org chart — you provide the case-manager -> supervisor
+# UUID pair to link in HRMS (EMPLOYEE_UUID / SUPERVISOR_UUID).
+#
+# Each step is independent and non-dependent on the others having run in the
+# same invocation: every step re-derives whatever it needs (a fresh token,
+# a fresh business-service fetch, a fresh env read) rather than relying on
+# in-memory state from an earlier step. Run them in any order, any subset,
+# any number of times.
+#
+# Steps:
+#   1. workflow-action     Add ESCALATE (self-loop, role SYSTEM) to a state
+#                           on the target BusinessService. Idempotent — skips
+#                           if that state already has an ESCALATE action.
+#   2. mdms-check           Report whether RAINMAKER-PGR.EscalationConfig
+#                           exists for TENANT. With ESCALATION_SEED=true,
+#                           creates it (schema + data) from the committed
+#                           JSON in utilities/default-data-handler if missing.
+#   3. hrms-link            Point EMPLOYEE_UUID's current HRMS assignment at
+#                           SUPERVISOR_UUID (reportingTo). Required inputs,
+#                           no defaults — this is org-chart data, not a knob.
+#   4. lookup-system-user   Discover the tenant's INTERNAL_MICROSERVICE_ROLE
+#                           user uuid via egov-user (prints it; does not
+#                           mutate anything). Feeds step 5 unless you already
+#                           know SYSTEM_USER_UUID.
+#   5. deploy               Set PGR_SERVICES_IMAGE / PGR_ESCALATION_STATES /
+#                           EGOV_INTERNAL_MICROSERVICE_USER_UUID in the
+#                           deployed compose + .env, then recreate pgr-services
+#                           (--no-deps, so one-shot migration containers that
+#                           already ran once are never re-triggered).
+#   6. verify               Tail pgr-services logs for an escalation scan and
+#                           report the last scan's scanned/escalated/skipped.
+#
+# Usage:
+#   ./enable-escalation.sh --list                     # print steps + exit
+#   ./enable-escalation.sh --only workflow-action      # just step 1
+#   ./enable-escalation.sh --only hrms-link            # just step 3
+#   ./enable-escalation.sh --from deploy               # steps 5..6
+#   ./enable-escalation.sh --dry-run                   # print, run nothing
+#
+#   PGR_SERVICES_IMAGE=egovio/pgr-services:my-tag \
+#   EMPLOYEE_UUID=eca6a910-... SUPERVISOR_UUID=97430b68-... \
+#     ./enable-escalation.sh
+#
+# NOTE: PGR_SERVICES_IMAGE (step 5), EMPLOYEE_UUID + SUPERVISOR_UUID (step 3)
+# have NO default — they must be supplied. Every other step/variable is
+# tunable via env vars in the CONFIG block below.
+# =============================================================================
+
+set -euo pipefail
+
+# -----------------------------------------------------------------------------
+# CONFIG — every tunable is an env var with a sane default and a "# what it is".
+# -----------------------------------------------------------------------------
+
+# Filesystem + gateway.
+DIGIT_HOME="${DIGIT_HOME:-/opt/digit}"                    # where the compose stack + .env live
+DIGIT_URL="${DIGIT_URL:-http://localhost:18000}"          # in-box Kong gateway origin
+
+# Tenant + admin identity.
+TENANT="${TENANT:-mz}"                                    # tenant to operate on
+ADMIN_USER="${ADMIN_USER:-ADMIN}"                         # admin user
+ADMIN_PASS="${ADMIN_PASS:-eGov@123}"                      # admin password
+
+# Workflow target (step 1).
+BUSINESS_SERVICE="${BUSINESS_SERVICE:-PGR}"               # BusinessService code to patch
+ESCALATE_STATE="${ESCALATE_STATE:-INVESTIGATION}"         # state to add the ESCALATE action to (self-loop)
+ESCALATE_ROLE="${ESCALATE_ROLE:-SYSTEM}"                  # role allowed to fire it (matches the scheduler's RequestInfo)
+
+# MDMS (step 2). No default seed — off unless explicitly requested.
+ESCALATION_SEED="${ESCALATION_SEED:-false}"               # true = create schema+data if missing
+MAX_DEPTH="${MAX_DEPTH:-3}"                               # only used when seeding
+SLA_BY_LEVEL="${SLA_BY_LEVEL:-3600000,14400000,86400000}" # only used when seeding (ms, one per level)
+REPO_ROOT="${REPO_ROOT:-$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)}"
+
+# HRMS link (step 3). REQUIRED — no default, this is org-chart data.
+EMPLOYEE_UUID="${EMPLOYEE_UUID:-}"                        # REQUIRED for step 3 — case manager (or any) employee uuid
+SUPERVISOR_UUID="${SUPERVISOR_UUID:-}"                    # REQUIRED for step 3 — their new reportingTo target
+DB_CONTAINER="${DB_CONTAINER:-docker-postgres}"           # postgres container name
+DB_USER="${DB_USER:-egov}"                                # db user
+DB_NAME="${DB_NAME:-egov}"                                # db name
+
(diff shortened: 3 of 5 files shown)


Hari-egov (commented):
## Review — tested end to end on a local full stack

I built this branch, deployed it against a live DIGIT stack with real data, and exercised **both** escalation paths through the UI. Summary: **the change works and the approach is right — no backend change is needed to make escalation fire.** What follows is one defect I'd fix in this PR, plus prerequisites the PR doesn't yet ship.

### Build

`mvn clean test` on Java 17: **BUILD SUCCESS, 274 tests, 0 failures.** `Collectors` is imported and `@Data` generates `getEscalationStates()`, so no compile risk. Worth noting there are **no escalation tests anywhere in the suite** — before or after this change.

### Automatic escalation — verified

With `PGR_ESCALATION_STATES=REFERRED,INVESTIGATION` the scan reached 14 complaints it could never have seen with the hardcoded values (`REFERRED 12 + INVESTIGATION 2`, matching the DB exactly):

```
Escalated complaint P-2026-000028 from level 0 to 1
  (assignee: [04a71b3b… case manager] -> 14100c4c… supervisor)
Escalation scan complete: scanned=14, escalated=1, skipped=13
```

`additionalDetails` shows `escalationLevel: 1`, `escalatedFrom`, `lastEscalatedAt`; status stays `INVESTIGATION` (self-loop, correct). In the UI the complaint's *Responsável Atual* changed from the case manager to the supervisor. The next scan correctly did nothing — the level-1 SLA is 4h.

### Manual escalation — verified

Adding `ESCALATE` on `REFERRED` with role `CMS_SUPERVISOR` is enough; a supervisor escalates from the UI with no code change:

```
ESCALATE  assignee=879019ea…  "Escalonamento manual pelo supervisor para teste"
ASSIGN    assignee=14100c4c…  "screening -> supervisor"
APPLY
```

The self-loop is what makes this work cleanly: `computeAssigneeRoles` reads the **target** state's forward actions, which on `REFERRED` resolve to `[CMS_SUPERVISOR]` — the correct tier. A self-loop on `INVESTIGATION` would instead offer case managers.

---

## The one defect I'd fix in this PR

**A single uncaught exception abandons the entire status scan.**

The `try` at `EscalationScheduler.java:96` wraps the whole per-status block. Scanning `PENDINGFORASSIGNMENT` on my stack reported `scanned=15` against **61 rows in the database**:

```
ERROR ServiceRequestRepository : 400 "INVALID USER: No user found for the uuids: [84a4d3a4-…]"
ERROR EscalationScheduler      : Error scanning complaints in status PENDINGFORASSIGNMENT for tenant mz
```

That UUID belongs to a **deactivated** employee (`active = f`). egov-user excludes inactive users, the enrichment throws, and the catch abandons the status — **46 of 61 complaints were never evaluated**, while the summary line printed a healthy-looking `scanned=15`.

This is pre-existing, not introduced here. But this PR is what makes people start relying on the scan, and the failure is invisible: one deactivated employee anywhere in a tenant silently disables escalation for every complaint in that state.

```java
for (String status : statuses) {
    List<ServiceWrapper> complaints;
    try {
        complaints = searchComplaintsByStatus(stateLevelTenantId, status);
    } catch (Exception e) {
        log.error("Search failed for status {} — status skipped entirely", status, e);
        continue;
    }
    for (ServiceWrapper wrapper : complaints) {
        scanned++;
        try {
            // ... existing per-complaint body ...
        } catch (Exception e) {
            failed++;                    // distinct from "skipped"
            log.error("Escalation failed for complaint {}",
                      wrapper.getService().getServiceRequestId(), e);
        }
    }
}
```

Counting `failed` separately matters as much as the scoping — otherwise "nothing to do" and "46 complaints blew up" produce identical output.

## Two small additions in the same file

An empty value silently disables the feature, and the resolved states are never logged:

```java
if (statuses.isEmpty()) {
    log.warn("pgr.escalation.states is empty — escalation scan will do nothing");
    return;
}
log.info("Escalation scan started for states {}", statuses);
```

`@Value`'s default only applies when the property is **absent**, so a deployment setting `PGR_ESCALATION_STATES=` gets an empty set and a healthy-looking `scanned=0`.

---

## Prerequisites this PR doesn't ship

Setting `pgr.escalation.states` alone escalates nothing. On a stack where the property was correct, three separate things still had to be true:

| # | Prerequisite | State before | Effect if missing |
|---|---|---|---|
| 1 | `EGOV_INTERNAL_MICROSERVICE_USER_UUID` | **unset** | scheduler has no SYSTEM user; nothing escalates |
| 2 | `ESCALATE` action on the workflow state | **0 of 11 CMS states** | transition rejected, ERROR logged per complaint every cycle, forever |
| 3 | HRMS `reportingTo` | **0 of 506 assignments** | every complaint dies at "No supervisor found" |

(1) is required and currently undocumented in `application.properties` — worth adding beside the new property, since it's the least discoverable of the three. (2) is worth adding to `CmsPgrWorkflowConfig.json` so a fresh install isn't inert; `enable-escalation.sh` fixes it at runtime but not at seed time. (3) is genuine org-chart data and correctly left to the operator.

### Localization — three keys, three different code paths

All three render raw tokens on a Portuguese UI until seeded:

| Key | Renders |
|---|---|
| `ESCALATE` (bare code) | action **menu** item — this is why `ASSIGN`→"Atribuir" worked and ESCALATE showed raw |
| `CS_ACTION_ESCALATE` | modal **heading** (`CS_ACTION_${action}`) |
| `WF_PGR_ESCALATE` | **timeline** row |

Two gotchas for whoever seeds them: a mixed-locale payload is rejected with `DUPLICATE_RECORDS` (one request per locale), and after upserting you need `redis-cli DEL computedMessages messages` **and** a browser refresh — the frontend caches per module in `localStorage`.

### Notifications

`RAINMAKER-PGR.NotificationRouting` has **0 rows** for ESCALATE, and `NotificationService:1083` returns an empty recipient list for both the `AUTO_ESCALATE` and `SYSTEM` audiences. No notification was emitted for either escalation I performed. An overdue complaint lands on a supervisor's desk silently — for a feature whose purpose is "this is late, someone senior must act", that seems worth a routing row.

---

## Two questions on the manual path

1. **The assignee field is optional.** `buildActionFormConfig` makes it mandatory only for `ASSIGN`, so a supervisor can escalate to nobody. Intended?
2. **The dropdown is department-scoped.** On a complaint with no department it offered **179** supervisors nationwide; once a department was set it offered **1**. Combined with `CMS_SUPERVISOR` being the top tier, escalation is peer-to-peer within a department with no route to a higher authority. Is that the intended model, or should escalation follow the HRMS chain the way the automatic path does?

## Script notes

`enable-escalation.sh` is genuinely useful and the runbook is good. Two things:

- `--dry-run` skips the `UPDATE` but still runs its verification, so `hrms-link` reports `[FAIL]` and stops the run.
- The `hrms-link` step writes directly to `eg_hrms_assignment`, bypassing the HRMS service and its async persister. Given we've had HRMS/egov-user desync before, a second opinion on that write seems worthwhile.

---

## Out of scope, raising separately

While testing I found **51 complaints carrying workflow state uuids from the retired `me` / `me.ige` / `me.igsae` tenants**, left behind when the root tenant became `mz`. `getNextActionOptions` matches state by uuid, so it returns `[]` and the employee **action button is disabled for every action** — ASSIGN, REJECT, REASSIGN, not just ESCALATE. Unrelated to this PR; I'll raise it on its own.

Nothing here blocks merge in my view — the defect above is the only item I'd want in before this lands, and it's ~15 lines in a file this PR already touches.

backend/pgr-services/src/main/java/org/egov/pgr/service/EscalationScheduler.java
```diff
@@ -71,8 +71,14 @@ public void scanAndEscalate() {
...
-        // Search for complaints in PENDINGATLME and PENDINGFORASSIGNMENT
-        Set<String> statuses = new HashSet<>(Arrays.asList(PENDINGATLME, PENDINGFORASSIGNMENT));
+        // Search for complaints in the configured applicationStatus values
+        // (pgr.escalation.states) — deployments customize this to match
+        // whichever states in their BusinessService workflow carry a named
+        // assignee and have an ESCALATE action wired on them.
+        Set<String> statuses = Arrays.stream(config.getEscalationStates().split(","))
+                .map(String::trim)
+                .filter(s -> !s.isEmpty())
+                .collect(Collectors.toSet());
```
Hari-egov, line 81: **Suggest logging the resolved states, and warning when the list is empty.**

`@Value`'s default only applies when the property is **absent**. A deployment that sets `PGR_ESCALATION_STATES=` (empty) produces an empty set here, the loop below runs zero times, and the scan logs a healthy-looking `scanned=0, escalated=0, skipped=0` — the feature is silently off.

Since this is a newly configurable knob, the resolved list is also the first thing an operator needs to confirm:

```java
if (statuses.isEmpty()) {
    log.warn("pgr.escalation.states is empty — escalation scan will do nothing");
    return;
}
log.info("Escalation scan started for states {}", statuses);
```

Separately, and more seriously — the `try` a few lines below wraps the **entire** per-status block, so one uncaught exception abandons every remaining complaint in that status. On my test stack a single deactivated employee caused `scanned=15` against 61 rows in the DB; 46 complaints were never evaluated and nothing in the output indicated it. Details and a suggested fix are in the review comment.

backend/pgr-services/src/main/resources/application.properties
```diff
@@ -178,6 +178,12 @@ pgr.escalation.batch.size=100
 pgr.escalation.default.sla.ms=432000000
 pgr.escalation.max.depth=3
 pgr.escalation.kafka.topic=pgr-escalation-events
+# Comma-separated applicationStatus values the escalation scan searches for
+# candidate complaints. Must match states that actually carry a named
+# assignee in the deployed BusinessService workflow (the scheduler skips any
+# complaint with no current assignee) and that have an ESCALATE action wired
+# on them — otherwise the workflow transition is rejected and nothing moves.
+pgr.escalation.states=PENDINGATLME,PENDINGFORASSIGNMENT
```
Hari-egov, line 186: **`EGOV_INTERNAL_MICROSERVICE_USER_UUID` is a hard requirement for this feature and isn't documented here.**

The scheduler builds its SYSTEM `RequestInfo` from `config.getEgovInternalMicroserviceUserUuid()`. On the stack I tested against, that variable was **unset** — the property block below was otherwise correct, and nothing escalated. It's the least discoverable of the three prerequisites, because unlike a missing workflow action it produces no distinctive error.

Since this block is already documenting what the new property needs, worth a line beside it:

```properties
# The escalation scan acts as this user. REQUIRED — with no value the workflow
# transition is rejected and nothing escalates. Discover it per environment with
# `enable-escalation.sh --only lookup-system-user`.
egov.internal.microservice.user.uuid=
```

The inline comment above the property is good and accurate — the two conditions it describes (a named assignee, and an ESCALATE action on the state) both bit me in exactly the way it warns. Worth adding that the CMS workflow config in this repo ships **0 ESCALATE actions across all 11 states**, so a fresh install satisfies neither condition until `enable-escalation.sh` is run.

local-setup/scripts/enable-escalation.sh
```diff
@@ -0,0 +1,506 @@
...
+#   REQUIRED inputs, no defaults — this is org-chart data, not a knob.
+# =============================================================================
+do_hrms_link() {
+  step "hrms-link" "Link EMPLOYEE_UUID -> SUPERVISOR_UUID in HRMS reportingTo"
+
+  require "EMPLOYEE_UUID is set" '[[ -n "$EMPLOYEE_UUID" ]]' || { err "set EMPLOYEE_UUID=<uuid>"; return 1; }
+  require "SUPERVISOR_UUID is set" '[[ -n "$SUPERVISOR_UUID" ]]' || { err "set SUPERVISOR_UUID=<uuid>"; return 1; }
+
+  run "UPDATE eg_hrms_assignment.reportingto for $EMPLOYEE_UUID" \
+    "db_psql \"UPDATE eg_hrms_assignment SET reportingto='$SUPERVISOR_UUID' WHERE employeeid='$EMPLOYEE_UUID' AND iscurrentassignment=true;\""
```
Hari-egov, line 327: **`--dry-run` reports a false failure and stops the run.**

The dry run skips the `UPDATE` but still executes its verification, so the check reads back the unchanged value and fails:

```
[dry-run] db_psql "UPDATE eg_hrms_assignment SET reportingto='…' WHERE …"
[FAIL] reportingTo is '', expected '14100c4c-…'
[FAIL] step 'hrms-link' failed
[FAIL] stopped on failure
```

Since `--dry-run` is what someone reaches for first, this reads as a broken script. Skipping the verify (or short-circuiting it to OK) when `DRY_RUN=true` would fix it.

**Separately, on this step's write path:** it issues a direct `UPDATE` against `eg_hrms_assignment`, bypassing egov-hrms and its async persister. We've been bitten by HRMS/egov-user desync before (employees created in egov-user but absent from `eg_hrms_employee`), so a direct write to the assignment table is worth a second opinion — an HRMS `_update` call would keep the service's own invariants intact, if one is available for `reportingTo`.

Otherwise this script is genuinely useful — the idempotency checks are careful, and the runbook's prerequisites table is accurate. Both conditions it lists caught me in exactly the way it describes.

Hari-egov:
## Verified locally — and ported to Mozambique

Following up on my review above with test results.

**Verified locally ✅** — built this branch and ran it on a full local DIGIT stack against real data. Both escalation paths work end to end through the employee UI:

- **Automatic** — a complaint moved level 0 → 1 and was reassigned to the assignee's HRMS supervisor
- **Manual** — a supervisor fired ESCALATE from the UI on a `REFERRED` complaint and selected an assignee

`mvn clean test`: BUILD SUCCESS, 274 tests, 0 failures.

**No frontend changes needed ✅** — the employee UI is already workflow-driven, and `computeAssigneeRoles` resolves the correct tier from the target state with no code change. What looked like UI defects were missing **localization** rows, not code.

**MDMS change required ⚠️** — `RAINMAKER-PGR.EscalationConfig` at the state-level tenant. On `cms-pilot` the schema was registered with **0 data rows**, so escalation could not have fired there regardless of this property.

I've ported this to the Mozambique repo as **eGov-Global/CMS-MOZAMBIQUE#8**, together with the seed data this PR leaves to `enable-escalation.sh` (the two `ESCALATE` workflow actions and the three localization keys), so a fresh install isn't inert. The two robustness fixes from my review are in that PR as well — happy to raise them here separately if you'd prefer them upstream first.

@priyanshu-egov — to exercise this on `cms-pilot` the backend needs deploying: pilot is still running a `pgr-services` image built before this change, so `pgr.escalation.states` isn't read there yet. The MDMS config and workflow actions are already applied and verified on pilot; only the service image is missing. Could you build and deploy it?

Note two prerequisites that will otherwise make it look like the feature doesn't work — both cost me a while to find:

1. **`EGOV_INTERNAL_MICROSERVICE_USER_UUID` must be set** — unset means the transition is rejected with no distinctive error (this is the inline comment on `application.properties`).
2. **HRMS `reportingTo` must be populated** — it was empty on all 506 assignments on my stack, and every complaint died at *"No supervisor found"* until linked.

cc @pradeepkumarcm-egov

backend/pgr-services/src/main/java/org/egov/pgr/service/EscalationScheduler.java
Hari-egov: changed added

backend/pgr-services/src/main/resources/application.properties
Hari-egov: changed added

local-setup/scripts/enable-escalation.sh
Hari-egov: changed added

KDwevedi:
Closing as superseded.

This PR made the escalation states configurable through a Spring property (`pgr.escalation.states`). #2049 landed the same capability one layer up: the scan states now come from `RAINMAKER-PGR.EscalationConfig.eligibleStatuses` in MDMS v2, resolved city → state, with `PGR_ESCALATION_ELIGIBLE_STATUSES` / `pgr.escalation.eligible.statuses` kept only as the fail-closed fallback when no record is readable.

`EscalationScheduler` on `master` now iterates `policy.getEligibleStatuses()`, and `EscalationService.validateAutomaticThreshold` rejects a complaint whose status is not in that set. A deployment changes the states by editing one MDMS record — no redeploy, and from #2090 onward no JSON either.

That closes the enhancement this PR was written for (#2044), so the property approach here would now be a second, competing source of truth for the same setting.

The `docs/pgr-escalation/RUNBOOK.md` and `local-setup/scripts/enable-escalation.sh` content is not carried over verbatim; `docs/escalation.md` and `docs/migration/pgr-escalation-self-loop.md` cover the same ground against the current design. If anything in the runbook is still uniquely useful, it is worth a small docs-only PR rather than reviving this branch.

Issue #1641, opened 2026-08-06
Vulnerability Assessment for MCP Service


singhalkarun:
Security review of `digit-mcp/` at `815b237`. Scope: the **MCP surface** — tools reachable over `POST /mcp` (JSON-RPC) and stdio. Non-MCP HTTP surfaces in the same process are listed as out of scope at the bottom.

**Status: all 16 findings fixed in #1779** (draft, → `develop`). Two deployment prerequisites must be verified before that merges — see the bottom.

**Root cause for most rows:** tool handlers held ambient admin authority rather than acting on behalf of a caller. Eight copies of a local `ensureAuthenticated()` (`tools/filestore-acl.ts:233`, `hrms.ts:531`, `snapshot.ts:16`, `mdms-tenant.ts`, `validators.ts:1127`, `localization.ts:170`, `idgen-location.ts:144`, `pgr-workflow.ts:944`) logged in with `CRS_USERNAME`/`CRS_PASSWORD` whenever no token was set — so *absent* caller credentials were what triggered *elevation to ADMIN*. Replaced by a single helper plus a transport-derived auth mode: `ambient` for stdio (the process owner supplied the credentials), `token` for HTTP (caller must present a validated token; the server never substitutes its own).

| ID | Sev | Location | Finding | Fix | Status |
|---|---|---|---|---|---|
| C1 | Critical | `src/index.ts:713` | `/mcp` dispatched via `createServer({ enableAllGroups: true })` with no auth check of any kind; tool-level `ensureAuthenticated()` then supplied stored ADMIN credentials. Anonymous caller ⇒ full tool surface. | `/mcp` authenticates before building the server; bearer introspected against egov-user `/user/_details` (30s cache); state tenant derived from the token, not the `X-State-Tenant` header. Anonymous ⇒ 401. | ✅ #1779 |
[...]

PR #1779, opened 2026-08-17
fix(digit-mcp): authenticate /mcp, /v1 and /api; authorize MCP tools

singhalkarun:
Security fixes for the MCP service. Findings and severities are tracked in #1641 — this implements all 13 rows plus 3 authorization items (N1–N3) found while fixing them.

**Six commits.** `3364fd7` fixes the tracked findings; `5d73d21` closes `/v1/*` and `/api/*`; `197771f` and `f64120c5` fix what five independent reviews found, and `bde20475` fixes what a second review round found in those fixes — including four defects `5d73d21` introduced, the reason none of it was reaching the cluster, and an SSRF at the lowest access tier.

**Draft for review.** Deployment prerequisites are listed at the bottom.

## Root cause

Eight near-identical copies of a local `ensureAuthenticated()` each fell back to `CRS_USERNAME`/`CRS_PASSWORD` whenever no token was set. Combined with `/mcp` having no auth check at all, **an anonymous caller was strictly more privileged than a badly-authenticated one** — sending no credentials triggered elevation to the container's stored ADMIN.

The fix is one shared helper plus an **auth mode** derived from the transport:

| Mode | When | Behaviour |
|---|---|---|
| `ambient` | stdio (default) | Env credentials may be used — the process was spawned by the developer whose credentials those are, so nothing is granted that the caller didn't already have. Authorization checks are skipped for the same reason. |
| `token` | HTTP (default) | Caller must present a token validated against egov-user. The server never substitutes its own credentials. |

`MCP_AUTH_MODE` overrides the derivation; setting `ambient` on HTTP logs a loud warning.

### Critical

| ID | Change | Files |
|---|---|---|
| **C1** | **All three routes** (`/mcp`, `/v1/*`, `/api/*`) authenticate before doing any work; bearer token introspected via `/user/_details` with a 30s cache. State tenant derived from the token rather than a caller-supplied header. | `src/index.ts`, `src/services/auth.ts` (new), `src/services/digit-api.ts`, `packages/data-provider/.../endpoints.ts` |
| **C3** | `base_url` host allow-list; ad-hoc targets require explicit credentials; `environment` added to `snapshotAuth`/`restoreAuth` so the redirect can't persist. | `src/tools/mdms-tenant.ts`, `src/services/auth.ts`, `src/services/digit-api.ts` |
| **C4** | Password reset guarded on `!existingUserUuid`; `loginCredentials` reports no password on the uuid-link path. | `src/tools/hrms.ts` |
| **C5** | Existing-citizen password reset removed; `citizenLogin` only returned for accounts `pgr_create` created. | `src/tools/pgr-workflow.ts` |

### High

| ID | Change | Files |
|---|---|---|
| **H1** | `tenant_cleanup` returns a read-only preview unless `confirm=<tenant_id>`; root tenants require `allow_root_tenant: true`, checked before auth or any lookup. | `src/tools/mdms-tenant.ts` |
| **H2** | `Authorization` attached to enc-service calls via new `encHeaders()`; both encryption tools require auth. | `src/services/digit-api.ts`, `src/tools/encryption.ts` |
| **H3** | `resolveArtifactPath()` confines snapshot reads/writes to `MCP_ARTIFACT_DIR`; rejects traversal, unrelated absolutes, and sibling-prefix escapes. | `src/services/auth.ts`, `src/tools/snapshot.ts` |
| **H5** | `digitApi` is now an `AsyncLocalStorage`-backed proxy; `/mcp` requests run in `runWithIsolatedClient()` in token mode. No call sites changed. | `src/services/digit-api.ts`, `src/index.ts` |

### Medium / Low

| ID | Change | Files |
|---|---|---|
| **M5** | Telemetry opt-in — no default endpoint; requires `MATOMO_URL`. Startup line names the endpoint when active. | `src/services/telemetry.ts` |
| **M6** | `configure` + `snapshot_capture` → `risk: 'write'`; SUPERUSER grant behind `provision_roles` (default false), with a response note when skipped. | `src/tools/mdms-tenant.ts`, `src/tools/snapshot.ts` |
| **M1** | Module-level `process.env.PGPASSWORD` removed; passed per-invocation. DB details from `DIGIT_DB_*`. | `src/services/shell.ts` |
| **L3** | Sanitization extended to engram content (search + full read), remote docs content, and both workflow `comment` fields. | `src/tools/docs.ts`, `src/tools/pgr-workflow.ts` |
| **L4** | `USER node`, pre-chowned writable dirs, no baked-in DB password; chart adds pod/container `securityContext`, read-only rootfs + volumes, image tag pinned to `Chart.AppVersion`. | `Dockerfile`, `helm/.../deployment.yaml`, `helm/.../values.yaml` |

### Authorization (N1–N3)

Authentication alone was not a privilege boundary: DIGIT permits citizen self-registration (`pgr_create` provisions citizens with password auth), and several tools — `decrypt_data`, `snapshot_capture`, `db_counts`, `kafka_lag`, `trace_*` — never consult DIGIT's own accesscontrol.

- **N1** New `ToolMetadata.access`: `public` | `employee` | `admin`, enforced in `server.ts` before dispatch. **Default `employee`**, so a new tool is never public by accident. Classified all 70 tools: 9 public, 34 admin, 27 employee-by-default. Keys on **roles, not `user.type`** — `type` would have let a provisioned citizen through.
- **N2** Redaction moved to `src/utils/redact.ts`: case-insensitive substring match at any depth (was 4 exact names, top level only). `sensitiveOutput` on `ToolMetadata` stops `decrypt_data` plaintext reaching the events table.
- **N3** `X-Forwarded-For` honoured only from `MCP_TRUSTED_PROXIES` peers (IPv4/CIDR/`*`); rejected claims logged as `untrusted_forwarded_for`.

## Incidental fix worth flagging

The configmap template enumerated four keys by hand, so **any new value added to `values.yaml` never reached the container** — several of the fixes above would have shipped inert. Replaced with a `range` over `.Values.env`, which also picks up develop's `CRS_API_URL`/`CRS_ENV_NAME` without a template edit.

## Behaviour changes (review these)

1. **Every HTTP caller now needs a token** — `/mcp`, `/v1/*` and `/api/*`. README and `install.sh` are updated (the installer prompts for one, or takes `DIGIT_ACCESS_TOKEN`). `MCP_AUTH_MODE=ambient` restores old behaviour for a loopback-bound socket.
2. **`tenant_cleanup` no longer deletes on first call** — automation must pass `confirm`.
3. **`configure` no longer auto-grants roles** — pass `provision_roles: true` where the tenant-fallback path relied on it.
4. **`configure` doesn't persist across `/mcp` calls in token mode** (per-request client). Tenant now derives from the token; all tools take `tenant_id` explicitly.
5. **`output_path` rejects paths outside `MCP_ARTIFACT_DIR`** — set that var to keep a previous location.
6. **Telemetry stops** where the demo-host default was relied on.
7. **`snapshot_capture`'s encryption canary degrades** when unauthenticated (was already the documented degradation model for that layer).
8. **`/v1/*` ignores `X-State-Tenant`** — the tenant comes from the token, same as `/mcp`.
9. **The session viewer asks for a token.** Its `/api/*` calls need an admin-role account; the token is held in `sessionStorage` for the tab only.
10. **`health_check` moved from `public` to `employee`.** `GET /healthz` still serves orchestrator liveness unauthenticated.
11. **The chart now deploys `egovio/digit-mcp`**, not `ghcr.io/chakshugautam/digit-mcp`. `image.tag` is required, with no fallback — pin a `develop-<sha8>` per environment.

## Verification

- `npm run build` clean; `helm lint` clean; chart renders with both develop's and the new values.
- **`npm run test:security` — 74 checks — and `npm run test:security:http` — 34 checks, both in CI** (`.github/workflows/digit-mcp-ci.yml`): auth-mode derivation, access tiers per identity, the tool classification itself (the public surface is asserted as an exact set), deep redaction, artifact confinement, the `base_url` allow-list, forwarded-for trust incl. IPv6, and interleaved-concurrency isolation for both the API client and the request context. Mutation-checked: defaulting the tier back to `public` fails 2.7; trusting `X-Forwarded-For` unconditionally fails four of section 7.
- `npm test` passes (70 tools registered).
- `npm run test:safety`: 51 pass / **2 fail — both pre-existing**. Verified by stashing this work and re-running on a clean tree: identical failures (`2.3 reject 9 digits`, `2.4 reject 11 digits`).
- Live probe: anonymous `/mcp` → 401, junk bearer → 401, `/healthz` → 200.
- Image built and run with `--read-only --user 1000:1000 --cap-drop ALL`: starts, serves, rootfs confirmed unwritable, gate holds.

## Deployment prerequisites

1. **Confirm `/user/_details` is routable** through the gateway. Validation is fail-closed by design — if introspection can't reach egov-user, every `/mcp` call returns 401.
2. **Confirm the admin role codes.** `MCP_ADMIN_ROLES` defaults to `SUPERUSER, MDMS_ADMIN, SYSTEM_ADMIN, STADMIN`. If the operator `ADMIN` account holds none of these, the 34 admin-tier tools will 403 in token mode. Denials print both required and held roles, so a mismatch is one call to diagnose.
3. **Pin `image.tag`.** `values.yaml` defaults to `nightly-develop`; pin a `develop-<sha8>` per environment. Rotating `CRS_PASSWORD` and closing the published container port remain operational items.

## Review follow-up (second commit)

The first commit closed `/mcp` and left three routes open on the same listener. Review also turned up a chart change that would not have deployed at all. Both are fixed in `5d73d21`.

### Authentication — the rest of C1

| Route | Was | Now |
|---|---|---|
| `/v1/*` | accepted **any** non-empty bearer string; fell back to `CRS_USERNAME`/`CRS_PASSWORD` | same introspection as `/mcp`; env fallback gated on ambient mode |
| `/v1/tenant/:id/export` | unauthenticated whole-tenant MDMS dump | admin tier (stated explicitly — it is not a registered tool) |
| `/v1/tools` | anonymous enumeration of the admin tool surface | bearer required (public tier: no role check) |
| `/api/stats`, `/api/sessions`, `/api/sessions/:id/{events,messages}`, `/api/pgr/dashboard` | fully anonymous, serving recorded tool args, result prefixes and client IPs | one gate over the namespace, admin tier |

`checkToolAccess` now runs on all three REST dispatch paths (single, streamed, bulk). Introspection, its cache and the tier check moved into `auth.ts`, so the routes cannot drift into disagreeing about what a token means. The session viewer gained a sign-in gate, since its data calls now need one.

### Deployment

- **The chart pointed at the wrong image.** `ghcr.io/chakshugautam/digit-mcp` is a personal registry fed by a different repository — nothing in this tree reached the cluster through it, so every fix in this PR would have shipped inert. Repointed to `egovio/digit-mcp`, which `build/build-config.yml` publishes from this Dockerfile.
- **The `Chart.AppVersion` tag fallback resolved to `1.0.0`, which has never existed** in either registry (checked against both). A `helm upgrade` on default values went straight to `ImagePullBackOff`.
- **`SESSION_DB_PASSWORD` plumbed through values/secret.** Dropping the Dockerfile `ENV` had changed nothing on its own — `db.ts` still defaulted to the same published string, one layer down.

### Correctness

- `AuthRequiredError` was documented as surfacing a 401 but **nothing mapped it** — every "not authenticated" error reported as `category: 'internal'`, telling the agent to retry rather than to authenticate.
- The logger and session store kept per-request identity in **instance fields**, so two concurrent `/mcp` requests cross-attributed each other's tool calls. Moved to `AsyncLocalStorage` — the same fix H5 applied to the API client, applied to the audit trail N3 exists to make trustworthy.
- Cached `UserInfo` is **copied per caller** rather than shared by reference across request-scoped clients.
- The `base_url` allow-list derived from the **live client's environment, which `configure` itself writes**; it now comes from startup config, so an accepted call cannot widen the guard that vets the next one.
- `MCP_TRUSTED_PROXIES` handles IPv6 literals and CIDRs — an IPv6 entry silently never matched, collapsing every client to the ingress address on a dual-stack cluster.
- `resolveArtifactPath` compares real paths on both sides: a symlink inside the artifact directory could point out of it, and a symlinked ancestor (`/var` → `/private/var`, a mounted volume) rejected legitimate input.
- `health_check` moved from `public` to `employee` — it reports which services exist, how fast they answer and how they fail, which is the infrastructure detail `ToolAccess` says the public tier must not expose.
- `configure`'s duplicated three-part provisioning condition collapsed to `if/else`.

## Third commit (`197771f`) — review findings

Five reviewers with separate briefs: adversarial security, Node/TS correctness, SRE/deployment, test-suite mutation testing, and API/docs. Everything below was reproduced before being fixed.

### Defects the second commit introduced

| | |
|---|---|
| `resolveArtifactPath` ran `mkdirSync` on the caller's path **before** the confinement check | A rejected path still authored directories anywhere the process could write |
| The symlink guard covered the wrong case | A **dangling** symlink passed: `realpathSync` throws on it, so it read as a not-yet-existing leaf, and `writeFileSync` followed it out. Creating a new file elsewhere is the case that matters |
| Form 1 claimed **any** `Authorization` header, not just `Bearer` | A proxy-injected or browser-cached `Basic` header hard-401'd, shadowing the `body.auth` credentials the caller sent |
| The `/api/*` gate preceded the CORS preflight short-circuit | `OPTIONS /api/stats` → 401. Preflights carry no credentials, so every cross-origin consumer was permanently broken |

### Pre-existing, found by review

- **`docs_get` was an SSRF primitive at the `public` tier.** The egress guard was `url.includes('docs.digit.org')` — a substring test on the whole URL, so `http://10.0.0.1/x?docs.digit.org` passed — then `fetch` followed redirects and returned the body verbatim. Reproduced: a citizen-role token read an internal endpoint through a 302. Now a hostname allow-list, `redirect: 'manual'`, a size cap, `sanitizeUserContent` (the L3 pass missed this path), and `employee` tier.
- **The REST shim wrote no audit record at all.** `mcpLogger`/`sessionStore` are called from `server.ts` only, so every successful `/v1/*` call — `tenant_destroy` included — was unlogged. Only denials were.
- **`enable_tools` was `public` + `risk: 'read'`** while mutating the long-lived REST registry every caller shares.

### The chart was not what deploys

`devops/deploy-as-code/charts/auxiliary-services/auxiliary-helmfile.yaml:46` hard-overrode `image.repository`/`tag` to a personal ghcr registry pinned to a June build. It is the only thing installing this chart in-cluster, so repointing `values.yaml` in the second commit **changed nothing** — and the new env vars were being injected into an image that ignores them. Repointed there, in compose, and in the ansible playbook. Also: `pullPolicy: Always` (the default tag moves), `checksum/config` so a ConfigMap-only change actually rolls, `sizeLimit` on all three emptyDirs, and the `develop-<sha8>` pinning advice removed — the nightly prune keeps only 5, so such a pin vanishes within a week along with the rollback target.

### One thing deliberately **not** done

The local compose stack is **not** set to `MCP_AUTH_MODE=ambient`, despite that being the obvious fix for "the local stack now 401s". `local-setup/ansible/templates/nginx-site.conf.j2:207` proxies `/v1/` on the **public** vhost and defers authentication to this server — so ambient there would publish an anonymous ADMIN surface, which is the hole this PR exists to close. A contract test asserts it stays off. The ansible `/v1` calls use `body.auth` and keep working; the xlsx runbook's calls did not, and now carry a bearer.

### Tests: enforcement, not just policy

Mutation testing found the unit suite proved the tier *decision* correct while proving nothing about whether it is *called* — deleting `checkToolAccess` from `server.ts`, `authorizeRest` from a REST path, or the whole `/api/*` gate each left it 57/57 green. `test-security-http.ts` (27 checks, in CI) drives a real server over a real socket against a stub egov-user: anonymous reachability, all three REST dispatch paths, the MCP path, the `/api/*` gate, Form 2, audit records, redaction. All three mutations now fail it.

Four unit tests were also defective: 3.3 was a tautology that could not fail, 4.1 did not test depth, 7.8 was vacuous, 8.2 left the shared client authenticated for every later test. Fixed, plus regressions for both path bugs, a partial-prefix IPv6 CIDR, and `environment` isolation.

### Docs

`docs/api/tools/decrypt_data.md` still said the tool needs no authentication — the exact claim H2 was filed against. The OAuth recipe added to the README in the second commit was wrong (DIGIT's client secret is empty; `$USER` collides with the shell's own variable). All six new env vars documented, tool counts corrected to 70, `install.sh` no longer writes an empty bearer under `--yes`, and the viewer gate gained a label, `aria-live`, and a token-minting hint.

**Tier counts, asserted by test 3.3b: 6 public / 30 employee / 34 admin of 70.**

### Shipping and operability
- **The Dockerfile never copied `ui/`**, so the session viewer 404s in every container — which made the sign-in gate added in `5d73d21` dead code wherever it actually runs.
- **The chart never set `SESSION_DB_*`**, so the image default (`localhost:15433`) had no listener in a pod: sessions were silently not persisted and `/api/stats` returned "Database not available". The `SESSION_DB_PASSWORD` secret added earlier did nothing on its own.
- **`MCP_BIND`**, defaulting to loopback in ambient+http. The ambient warning told operators to use a loopback-bound socket while the bind address was hardcoded to `0.0.0.0`.

### Failure modes that lied
- **`validateToken` returned `null` for a 5xx, a DNS failure and a timeout alike**, so an egov-user outage surfaced as *"Invalid or expired access token"* on every route — sending operators after credentials while the platform was down, and making the viewer discard a working token. Now a distinct error mapped to **503**, plus single-flight so a cold token costs one round trip rather than N against an already-struggling service.
- **`runStreamed` committed HTTP 200 before authenticating**, then emitted `event: error`. Any client keying on the status code read a denial as success.
- **`tenant_cleanup`'s dry run returned `success: true`** — the field every other tool uses to mean "it happened", so automation would take a preview for a completed cleanup.

### Leaks
- **`employee_create`, `user_create`, `pgr_create` return plaintext provisioning passwords**, and the session store persists a 200-char result prefix — so the credential reached Postgres and the JSONL log. Marked `sensitiveOutput`.
- **`/v1/version` served `gitSha`, `buildTime`, `nodeVersion` anonymously** — which CVEs apply, to anyone. Withheld unless authenticated; service name and uptime stay open so it remains a liveness probe.
- **The REST shim resolved the tool before authenticating**, so 404-for-unknown vs 401-for-known enumerated the tool surface.
- `serveStatic` compared with `startsWith` and no trailing separator.

### Parsing
- **`ipv6Groups` parsed a trailing dotted quad with `parseInt(_, 16)`** — `'1.2.3.4'` became `1` — and `normalizeIp` only recognised the `::ffff:` shorthand, so a fully-spelled v4-mapped peer never matched an IPv4 entry, and `::ffff:10.0.0.0/104` was rejected outright. All three fixed.
- **`redactDeep` matched credential words as bare substrings**, redacting `author`, `authority` and `tokenize` out of the audit trail, and its depth cap of 8 replaced genuinely deep payloads (`mdms_create`, `tenant_bootstrap`) with a placeholder — discarding the arguments most worth auditing. Now word-segmented, depth 64, with a `WeakSet` cycle guard.
- **`MCP_ALLOWED_BASE_URLS` replaced the configured host instead of extending it**, so authorising one extra instance silently de-authorised `CRS_API_URL`.

### Deliberately not done
The introspection call still passes the token as `?access_token=`. That is egov-user's documented contract for `/user/_details` (a `@RequestParam`, which is how the gateway calls it too), so it does put the token in access logs — but moving it to a header on an unverified guess **fails closed**, i.e. every caller gets a 401. Documented in place rather than changed blind.

One review finding was **overstated and is recorded as such**: `base_url` needed no `..` guard, because the URL parser collapses those before the check runs. The test asserts that rather than adding dead code.

### CI

`DIGIT-MCP CI`, `Gatus coverage`, `Tilt CI` and the Flyway check are green; the infra-contract job that runs the new `local-setup` assertions passes 8/8. **`Local Setup CI` fails** — at the `Localization module check / egov-user not seeded` step, which fails identically on `develop` and is unrelated to this branch.

## Sixth commit (`bde20475`) — second review round

Two fresh reviewers (adversarial security, Node/TS correctness) re-ran against `f64120c5` with briefs that required each fix to come back CLOSED, PARTIALLY CLOSED or NOT CLOSED. They independently agreed on five items. Everything below was reproduced before being fixed.

**The redaction rewrite lost coverage it should have kept.** Moving from substring to word matching removed the intended false positives (`author`, `authority`, `tokenize`) and introduced false *negatives*: `access_key` was dropped from the word list outright, and the glue-fallback's length floor of 6 excluded `pwd`, `auth` and `token` — the three commonest glue components. Confirmed reaching `access.log` and `events.jsonl`: `access_key`, `aws_access_key`, `AWS_ACCESS_KEY_ID`, `authtoken`, `accesstoken`, `refreshtoken`, `idtoken`, `oauth`, `xauth`, `userpwd`, `newpwd`, `pwd2`. Now per-word matching with an explicit benign list, plus qualified words so `key` counts next to `access`/`secret`/`signing` but not in `sortKey`. 35 keys asserted in both directions.

**The `WeakSet` was traversal-scoped, not path-scoped.** A DAG is not a cycle — `{ source: t, target: t }` emitted the second occurrence as `[circular]`, deleting real audit data. The same over-redaction the rewrite existed to stop.

**REST calls still recorded nothing in the session store.** `f64120c5` claimed "all three dispatch paths now record"; only `mcpLogger` was actually wired. `recordToolCall` is a no-op until a session exists, and only `/mcp` ever called `ensureSession` — so on a REST-only deployment (what the ansible playbook and the xlsx runbook drive) `tenant_destroy` left an access-log line and **no** events row, no `/api/sessions/:id/events` entry, no Postgres record.

**Unauthenticated remote OOM.** `readBody` buffered without a cap, ahead of every gate — so one anonymous request killed the process on its memory limit, before `authenticateBearer` was ever reached. Pre-existing, but it defeated every gate in this PR. Capped at 16MB/413; verified a 400MB anonymous POST against `--max-old-space-size=192` now leaves the server serving.

**The tool-existence oracle was fixed on one route of two.** `dispatchTool` was reordered, `runBulk` was not — the sibling route still enumerated all 70 tools anonymously. The `C.7` test asserted only the non-bulk path, which is exactly why it survived.

**Trusted-proxy config failed OPEN on malformed entries.** `bits === 0` returned true before the range was validated, so `/0`, `garbage/0` and `10.0.0.1/-0` all meant "trust every peer". Separately the v4-mapped rebase was gated on the *normalised* range, which cannot tell `::ffff:10.0.0.0/104` from the typo `10.0.0.0/104` — the typo silently became a `/8`.

**`/v1/version` was an introspection amplifier.** Gating detail behind `authenticateBearer` meant any bearer, including junk, drove a full `/user/_details` round trip (only successes are cached) — a 1:1 amplifier against egov-user at the one endpoint documented as an open probe, blocking for the 10s timeout when that service hung. Now cache-hit only.

**`runStreamed` authenticated twice** — two OAuth logins per `body.auth` SSE request, the form ansible uses.

### Recorded as NOT defects
The previous round's IPv6 finding was **overstated**: `resolveClientIp` feeds only log attribution, never an authorization decision, so the worst case was a forged IP in the audit trail. Both reviewers also confirmed the `docs_get` SSRF fix, artifact-path confinement, introspection failure handling and `sensitiveOutput` as fully **CLOSED** against a wide set of attacks (DNS rebinding, userinfo, punycode, trailing-dot, IPv6-literal hosts, intermediate symlinks, TOCTOU, fail-open paths).

### DIFF
--- a/digit-mcp/src/index.ts
+++ b/digit-mcp/src/index.ts
@@ -17,6 +17,7 @@ import {
   checkToolAccess,
   getAuthMode,
   isAdminCaller,
+  isCachedCaller,
   parseBearer,
   resolveClientIp,
   type ValidatedCaller,
@@ -62,10 +63,40 @@ if (transportMode === 'stdio') {
     res.end(JSON.stringify(data));
   }
 
+  /**
+   * Maximum accepted request body. Bulk tenant payloads are the large case;
+   * 16 MB is well clear of those and well clear of the pod's memory limit.
+   */
+  const MAX_BODY_BYTES = 16 * 1024 * 1024;
+
+  class PayloadTooLargeError extends Error {
+    constructor() {
+      super(`Request body exceeds the ${MAX_BODY_BYTES} byte limit.`);
+      this.name = 'PayloadTooLargeError';
+    }
+  }
+
+  /**
+   * Read a request body, bounded.
+   *
+   * Unbounded, this defeated every gate in front of it: the body is buffered
+   * before any route can authenticate, so one anonymous request with a large
+   * chunked body killed the process on its memory limit — a reliable
+   * crash-loop from a peer holding no credentials.
+   */
   function readBody(req: IncomingMessage): Promise<string> {
     return new Promise((resolve, reject) => {
       const chunks: Buffer[] = [];
-      req.on('data', (chunk: Buffer) => chunks.push(chunk));
+      let total = 0;
+      req.on('data', (chunk: Buffer) => {
+        total += chunk.length;
+        if (total > MAX_BODY_BYTES) {
+          reject(new PayloadTooLargeError());
+          req.destroy();
+          return;
+        }
+        chunks.push(chunk);
+      });
       req.on('end', () => resolve(Buffer.concat(chunks).toString('utf8')));
       req.on('error', reject);
     });
@@ -299,6 +330,12 @@ if (transportMode === 'stdio') {
     toolArgs: Record<string, unknown>,
     onResult: (parsed: unknown) => T,
   ): Promise<T> {
+    // `recordToolCall` is a no-op until a session exists, and only the /mcp
+    // handler used to create one — so on a REST-only deployment (which is what
+    // the ansible playbook and the xlsx onboarding runbook actually drive)
+    // every call produced an access-log line and nothing in the session store,
+    // which is the half the operator console reads.
+    await sessionStore.ensureSession('http');
     const seq = sessionStore.recordToolCall(tool.name, toolArgs);
     mcpLogger.toolCall(tool.name, toolArgs);
     const start = Date.now();
@@ -399,12 +436,17 @@ if (transportMode === 'stdio') {
     // Authenticate and authorize before committing the 200. An SSE stream that
     // opens successfully and then emits `event: error` reads as success to any
     // client keying on the status code, which is every ordinary HTTP client.
+    // Authenticate once, up front, and carry the result into the streamed body.
+    // Re-running authenticateRest there would mean a second OAuth login for
+    // every body.auth caller — the form ansible and the onboarding runbook use.
+    let caller: ValidatedCaller | null = null;
     const pre = await withRestLock(async () => {
       const snap = digitApi.snapshotAuth();
       try {
         const authResult = await authenticateRest(req, args);
         if (authResult === null) return { status: 401, body: { success: false, error: 'Authentication required.' } };
         if (!authResult.ok) return { status: authResult.status, body: { success: false, error: authResult.error } };
+        caller = authResult.caller;
         return authorizeRest(toolName, tool.access, authResult.caller);
       } finally {
         digitApi.restoreAuth(snap);
@@ -432,20 +474,9 @@ if (transportMode === 'stdio') {
         const snap = digitApi.snapshotAuth();
         setProgressEmitter((e) => writeEvent('progress', e));
         try {
-          const authResult = await authenticateRest(req, args);
-          if (authResult === null) {
-            writeEvent('error', { success: false, error: 'Authentication required.' });
-            return;
-          }
-          if (!authResult.ok) {
-            writeEvent('error', { success: false, error: authResult.error });
-            return;
-          }
-          const denied = authorizeRest(toolName, tool.access, authResult.caller);
-          if (denied) {
-            writeEvent('error', denied.body);
-            return;
-          }
+          // Re-apply the identity resolved above; restoreAuth in the pre-flight
+          // block discarded it, and re-authenticating would double the cost.
+          if (caller) applyValidatedCaller(caller);
           const { auth: _ignored, ...toolArgs } = args;
           void _ignored;
           await auditedCall(tool, toolArgs as Record<string, unknown>, (parsed) => {
@@ -478,11 +509,9 @@ if (transportMode === 'stdio') {
     auth: Record<string, string> | undefined,
     req: IncomingMessage,
   ): Promise<{ status: number; body: unknown }> {
-    const tool = restRegistry.getTool(toolName);
-    if (!tool) {
-      return { status: 404, body: { success: false, error: `Unknown tool: ${toolName}` } };
-    }
-    const boundTool = tool;
+    // Resolved after authentication, like dispatchTool: 404-for-unknown vs
+    // 401-for-known enumerates the whole tool surface anonymously, and fixing
+    // only the non-bulk route left the sibling wide open.
     return withRestLock(async () => {
       const snap = digitApi.snapshotAuth();
       try {
@@ -496,6 +525,11 @@ if (transportMode === 'stdio') {
           return { status: authResult.status, body: { success: false, error: authResult.error } };
         }
 
+        const boundTool = restRegistry.getTool(toolName);
+        if (!boundTool) {
+          return { status: 404, body: { success: false, error: `Unknown tool: ${toolName}` } };
+        }
+
         const denied = authorizeRest(toolName, boundTool.access, authResult.caller);
         if (denied) return denied;
 
@@ -503,7 +537,10 @@ if (transportMode === 'stdio') {
         const results: ItemResult[] = new Array(items.length);
         let nextIdx = 0;
 
-        async function worker(): Promise<void> {
+        // An arrow const, not a hoisted `function` declaration: a declaration is
+        // hoisted above the `if (!boundTool)` guard, so TypeScript will not
+        // narrow the captured const inside it.
+        const worker = async (): Promise<void> => {
           while (true) {
             const i = nextIdx++;
             if (i >= items.length) return;
@@ -517,7 +554,7 @@ if (transportMode === 'stdio') {
               results[i] = { ok: false, status: 500, body: { success: false, ...normalizeError(msg) } };
             }
           }
-        }
+        };
 
         const workers = Array.from({ length: Math.min(concurrency, items.length) }, () => worker());
         await Promise.all(workers);
@@ -667,8 +704,13 @@ if (transportMode === 'stdio') {
       // gitSha / buildTime / nodeVersion tell an unauthenticated caller exactly
       // which build is running and therefore which CVEs apply. The service name
       // and uptime stay open so the endpoint remains usable as a liveness probe.
+      // Cache-hit only. Calling authenticateBearer here would drive a full
+      // /user/_details round trip per request from an ANONYMOUS caller (only
+      // successes are cached), turning the one endpoint documented as an open
+      // liveness probe into a 1:1 amplifier against egov-user — and blocking
+      // this response for the introspection timeout when that service hangs.
       const detailed = getAuthMode() !== 'token' ||
-        (await authenticateBearer(req.headers['authorization'] || req.headers['Authorization'])).ok;
+        isCachedCaller(req.headers['authorization'] || req.headers['Authorization']);
       jsonResponse(res, 200, {
         service: 'digit-mcp',
         version: detailed ? (process.env.MCP_VERSION || '1.0.0') : undefined,
@@ -708,7 +750,8 @@ if (transportMode === 'stdio') {
         const result = await dispatchTool('tenant_bootstrap', body, req);
         jsonResponse(res, result.status, result.body);
       } catch (err) {
-        jsonResponse(res, 400, { success: false, error: String(err) });
+        const tooLarge = err instanceof Error && err.name === 'PayloadTooLargeError';
+        jsonResponse(res, tooLarge ? 413 : 400, { success: false, error: String(err) });
       }
       return;
     }
@@ -731,7 +774,8 @@ if (transportMode === 'stdio') {
         const result = await dispatchTool('city_setup', body, req);
         jsonResponse(res, result.status, result.body);
       } catch (err) {
-        jsonResponse(res, 400, { success: false, error: String(err) });
+        const tooLarge = err instanceof Error && err.name === 'PayloadTooLargeError';
+        jsonResponse(res, tooLarge ? 413 : 400, { success: false, error: String(err) });
       }
       return;
     }
@@ -767,7 +811,8 @@ if (transportMode === 'stdio') {
         const result = await dispatchTool('tenant_cleanup', toolArgs, req);
         jsonResponse(res, result.status, result.body);
       } catch (err) {
-        jsonResponse(res, 400, { success: false, error: String(err) });
+        const tooLarge = err instanceof Error && err.name === 'PayloadTooLargeError';
+        jsonResponse(res, tooLarge ? 413 : 400, { success: false, error: String(err) });
       }
       return;
     }
@@ -810,7 +855,8 @@ if (transportMode === 'stdio') {
           }
         });
       } catch (err) {
-        jsonResponse(res, 400, { success: false, error: String(err) });
+        const tooLarge = err instanceof Error && err.name === 'PayloadTooLargeError';
+        jsonResponse(res, tooLarge ? 413 : 400, { success: false, error: String(err) });
       }
       return;
     }
@@ -868,7 +914,8 @@ if (transportMode === 'stdio') {
         const result = await runBulk(bulkMatch[1], items, concurrency, body.auth, req);
         jsonResponse(res, result.status, result.body);
       } catch (err) {
-        jsonResponse(res, 400, { success: false, error: String(err) });
+        const tooLarge = err instanceof Error && err.name === 'PayloadTooLargeError';
+        jsonResponse(res, tooLarge ? 413 : 400, { success: false, error: String(err) });
       }
       return;
     }
@@ -882,7 +929,8 @@ if (transportMode === 'stdio') {
         const result = await dispatchTool(toolMatch[1], body, req);
         jsonResponse(res, result.status, result.body);
       } catch (err) {
-        jsonResponse(res, 400, { success: false, error: String(err) });
+        const tooLarge = err instanceof Error && err.name === 'PayloadTooLargeError';
+        jsonResponse(res, tooLarge ? 413 : 400, { success: false, error: String(err) });
       }
       return;
     }
--- a/digit-mcp/src/index.ts
+++ b/digit-mcp/src/index.ts
@@ -18,11 +18,13 @@ import {
   getAuthMode,
   isAdminCaller,
   isCachedCaller,
+  isInside,
   parseBearer,
   resolveClientIp,
   type ValidatedCaller,
 } from './services/auth.js';
 import { runWithRequestContext, setContextUser } from './services/request-context.js';
+import { redactDeep } from './utils/redact.js';
 import { setProgressEmitter, type ProgressEvent } from './services/progress.js';
 import type { IncomingMessage, ServerResponse } from 'node:http';
 import { readFile, stat } from 'node:fs/promises';
@@ -117,9 +119,9 @@ if (transportMode === 'stdio') {
     let filePath = urlPath === '/' ? '/index.html' : urlPath;
     const resolved = resolve(join(UI_DIR, filePath));
 
-    // Path traversal protection. The trailing separator matters: without it a
-    // sibling directory sharing the prefix (`ui-evil` next to `ui`) passes.
-    if (resolved !== UI_DIR && !resolved.startsWith(UI_DIR + sep)) {
+    // Same containment helper as the artifact guard — one implementation, so a
+    // future edge-case fix cannot land on only one of them.
+    if (!isInside(resolved, UI_DIR)) {
       res.writeHead(403, { 'Content-Type': 'text/plain' });
       res.end('Forbidden');
       return;
@@ -210,9 +212,14 @@ if (transportMode === 'stdio') {
     if (auth && auth.username && auth.password && auth.tenant_id) {
       try {
         await digitApi.login(auth.username, auth.password, auth.tenant_id);
-        const user = digitApi.getAuthInfo().user;
-        setContextUser(user?.userName);
-        return { ok: true, caller: user ? { token: '', user } : null };
+        // Carry the token login actually minted. Returning '' here made
+        // `applyValidatedCaller` install an empty token that `isAuthenticated()`
+        // accepted while `encHeaders()` treated as absent — so a body.auth
+        // streamed call reached egov-enc-service with no Authorization header,
+        // silently reopening the decryption oracle H2 closed.
+        const info = digitApi.getAuthInfo();
+        setContextUser(info.user?.userName);
+        return { ok: true, caller: info.user && info.token ? { token: info.token, user: info.user } : null };
       } catch (err) {
         return { ok: false, status: 401, error: err instanceof Error ? err.message : String(err) };
       }
@@ -336,8 +343,12 @@ if (transportMode === 'stdio') {
     // every call produced an access-log line and nothing in the session store,
     // which is the half the operator console reads.
     await sessionStore.ensureSession('http');
-    const seq = sessionStore.recordToolCall(tool.name, toolArgs);
-    mcpLogger.toolCall(tool.name, toolArgs);
+    // Redact once and hand the same copy to both sinks: each used to run the
+    // full recursive walk over the same object, doubling the cost on every
+    // call for no benefit.
+    const redacted = redactDeep(toolArgs) as Record<string, unknown>;
+    const seq = sessionStore.recordToolCall(tool.name, toolArgs, redacted);
+    mcpLogger.toolCall(tool.name, toolArgs, redacted);
     const start = Date.now();
     try {
       const raw = await tool.handler(toolArgs);
@@ -428,11 +439,7 @@ if (transportMode === 'stdio') {
     req: IncomingMessage,
     res: ServerResponse,
   ): Promise<void> {
-    const tool = restRegistry.getTool(toolName);
-    if (!tool) {
-      jsonResponse(res, 404, { success: false, error: `Unknown tool: ${toolName}` });
-      return;
-    }
+
     // Authenticate and authorize before committing the 200. An SSE stream that
     // opens successfully and then emits `event: error` reads as success to any
     // client keying on the status code, which is every ordinary HTTP client.
@@ -440,6 +447,7 @@ if (transportMode === 'stdio') {
     // Re-running authenticateRest there would mean a second OAuth login for
     // every body.auth caller — the form ansible and the onboarding runbook use.
     let caller: ValidatedCaller | null = null;
+    let streamedTool: ToolMetadata | undefined;
     const pre = await withRestLock(async () => {
       const snap = digitApi.snapshotAuth();
       try {
@@ -447,7 +455,13 @@ if (transportMode === 'stdio') {
         if (authResult === null) return { status: 401, body: { success: false, error: 'Authentication required.' } };
         if (!authResult.ok) return { status: authResult.status, body: { success: false, error: authResult.error } };
         caller = authResult.caller;
-        return authorizeRest(toolName, tool.access, authResult.caller);
+        // Resolved after authentication, like dispatchTool and runBulk. Today
+        // every call site passes a hardcoded name, so this is not an oracle —
+        // it is consistency, so that a future caller passing user input does
+        // not silently reopen one.
+        streamedTool = restRegistry.getTool(toolName);
+        if (!streamedTool) return { status: 404, body: { success: false, error: `Unknown tool: ${toolName}` } };
+        return authorizeRest(toolName, streamedTool.access, authResult.caller);
       } finally {
         digitApi.restoreAuth(snap);
       }
@@ -479,7 +493,7 @@ if (transportMode === 'stdio') {
           if (caller) applyValidatedCaller(caller);
           const { auth: _ignored, ...toolArgs } = args;
           void _ignored;
-          await auditedCall(tool, toolArgs as Record<string, unknown>, (parsed) => {
+          await auditedCall(streamedTool!, toolArgs as Record<string, unknown>, (parsed) => {
             writeEvent('done', parsed);
           });
         } catch (err) {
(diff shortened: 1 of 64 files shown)


digit-mcp/src/index.ts
```diff
@@ -122,61 +167,120 @@ if (transportMode === 'stdio') {
...
 
     // Form 2: body.auth = { username, password, tenant_id }
     const auth = body.auth as Record<string, string> | undefined;
     if (auth && auth.username && auth.password && auth.tenant_id) {
       try {
         await digitApi.login(auth.username, auth.password, auth.tenant_id);
-        return { ok: true };
+        const user = digitApi.getAuthInfo().user;
+        setContextUser(user?.userName);
+        return { ok: true, caller: user ? { token: '', user } : null };
```
vinothrallapalli-eGov: **Bug (security):** Form-2 (`body.auth` username/password) authentication returns `caller.token` as an empty string instead of the real session token, and `runStreamed` later re-applies it via `applyToken('', user, ...)`, which `digitApi.isAuthenticated()` (`authToken !== null`) treats as authenticated and `encHeaders()` (`if (this.authToken)`) treats as absent.

**Scenario:** A body.auth-authenticated streamed call (the pattern the ansible playbook/xlsx onboarding runbook uses) passes MCP-level authorization normally, but every downstream call that relies on `encHeaders` (e.g. `decrypt_data`/`encrypt_data` to egov-enc-service) is sent with no `Authorization` header at all — reopening the 'unauthenticated decryption oracle' this same PR's commit message says was closed, silently, with no 401 surfaced.

digit-mcp/src/index.ts
```diff
@@ -280,6 +433,30 @@ if (transportMode === 'stdio') {
       jsonResponse(res, 404, { success: false, error: `Unknown tool: ${toolName}` });
```
vinothrallapalli-eGov: **Bug (security):** `runStreamed` resolves `restRegistry.getTool(toolName)` and returns 404 for an unknown tool (this block, lines 431-435) BEFORE calling `authenticateRest`/`authorizeRest` below, while `dispatchTool` and `runBulk` in this same file were explicitly fixed (per their own comments) to resolve the tool AFTER authentication, specifically to prevent tool-surface enumeration via 404-vs-401 status diffing.

**Scenario:** An anonymous caller diffs HTTP status codes across guessed tool names on the streaming endpoint (real tool -> 401/403, fake name -> 404) to enumerate the entire admin tool surface without presenting any credentials — exactly the oracle the sibling dispatch paths were patched to close, but `runStreamed` was missed.

digit-mcp/src/tools/discover-tools.ts
```diff
@@ -34,7 +35,11 @@ export function registerDiscoverTools(registry: ToolRegistry): void {
     name: 'enable_tools',
     group: 'core',
     category: 'discovery',
-    risk: 'read',
+    // Not public, and not a read. On /mcp the registry is per-request, but the
+    // REST shim shares one long-lived ToolRegistry across all callers, so this
+    // mutates state other callers observe.
+    access: 'employee',
```
vinothrallapalli-eGov: **Finding (security/DoS):** `enable_tools` is gated at `access: 'employee'`, but per the comment right above, its handler can disable any tool group except `core` in the REST shim's single long-lived, cross-caller `ToolRegistry`.

**Scenario:** Any low-privileged employee account (no admin role, from any tenant) can call `enable_tools` with `disable: [all non-core groups]` over `/v1` or `/api` and zero out tool availability for every other concurrent caller of the deployment until restart — a platform-wide DoS gated at the same tier as an ordinary read, despite a blast radius comparable to other admin-gated tools. Consider gating this at `admin`.

digit-mcp/src/services/auth.ts
```diff
@@ -0,0 +1,670 @@
...
+  if (!user) {
+    tokenCache.delete(token);
+    return {
+      ok: false,
+      status: 401,
+      error: 'Invalid or expired access token. Mint a new one from /user/oauth/token.',
+    };
+  }
+
+  if (tokenCache.size >= TOKEN_CACHE_MAX) tokenCache.clear();
```
vinothrallapalli-eGov: **Finding (reliability):** The 30s token-introspection cache evicts by calling `tokenCache.clear()` on the whole map once size reaches `TOKEN_CACHE_MAX`, rather than removing only the oldest/expired entries.

**Scenario:** Under sustained load with more distinct active callers than the cap within a 30s window, the cache repeatedly wipes itself entirely instead of evicting incrementally, forcing every recently-cached caller's next request to re-hit egov-user's `/user/_details` — a thundering-herd amplification against egov-user that can be triggered just by cycling through enough distinct (even legitimately issued) tokens. Consider an LRU/TTL-based partial eviction instead.

digit-mcp/src/services/auth.ts
```diff
@@ -0,0 +1,670 @@
...
+      `Remove the link at ${real}, or pass a different filename.`
+    );
+  }
+
+  // Only now, once the destination is known to be confined, create it.
+  mkdirSync(dirname(real), { recursive: true });
+  return real;
+}
+
+function isInside(candidate: string, base: string): boolean {
```
vinothrallapalli-eGov: **Cleanup:** This path-containment check (`isInside`) is independently re-implemented in `src/index.ts` (~line 122) for the static `UI_DIR` guard, rather than importing/reusing this helper.

Two independent implementations of the same security-critical 'is this path inside that directory' check, written in the same PR. A future edge-case fix (trailing slash, case-insensitive filesystem) applied to one copy could silently leave the other with the weaker behavior. Consider exporting this and importing it in index.ts.

digit-mcp/src/server.ts
```diff
@@ -94,6 +96,35 @@ export function createServer(options?: CreateServerOptions): Server {
...
+            ),
+          },
+        ],
+        isError: true,
+      };
+    }
+
     const start = Date.now();
     const sanitizedArgs = (args || {}) as Record<string, unknown>;
     mcpLogger.toolCall(name, sanitizedArgs);
```
vinothrallapalli-eGov: **Cleanup (perf):** `mcpLogger.toolCall()` here and `sessionStore.recordToolCall()` two lines below each independently run the new recursive `redactDeep()` over the same tool-call args object; the identical duplication exists in `index.ts`'s `auditedCall` for the REST/bulk path.

**Impact:** Every tool call now pays a full recursive redaction walk (regex key-splitting, cycle detection, depth tracking) twice instead of once; on large payloads (bulk `mdms_create`, `tenant_bootstrap` bodies up to this PR's new 16MB cap) this doubles avoidable CPU/latency on the request hot path purely for audit logging. Consider redacting once and passing the redacted copy to both sinks.

vinothrallapalli-eGov:
**Finding (security — not posted inline, pre-existing code untouched by this diff):** `digit-mcp/src/tools/user.ts` still defines and calls its own private `ensureAuthenticated()` (used by `user_search`, `user_create`, `user_role_add`, around line 393) instead of importing the centralized helper from `services/auth.ts` that every other of the original duplicate-implementation files was migrated to use in this PR. The local copy has no `getAuthMode()` check and unconditionally falls back to `CRS_USERNAME`/`CRS_PASSWORD`.

**Scenario:** Masked today because HTTP routes populate `digitApi`'s auth state before any handler runs, but any future code path (retry, test harness, refactor of dispatch order) that leaves `digitApi` unauthenticated when these handlers run in 'token' mode silently elevates an anonymous/under-authenticated caller to the container's own ADMIN credentials for `user_create`/`user_role_add` (which can grant SUPERUSER) — the exact class of bug this PR's headline fix eliminated everywhere else.

vinothrallapalli-eGov:
**Finding (security — not posted inline, pre-existing code untouched by this diff):** `filestore_get_urls`, `access_roles_search`, and `access_actions_search` (`filestore-acl.ts`) and `user_search` (`user.ts`) have no `access` field, defaulting to the weak 'employee' tier, while sibling write tools in the very same files (`filestore_upload`, `user_create`, `user_role_add`) were explicitly bumped to 'admin' in this PR.

**Scenario:** Any authenticated non-citizen employee (no admin role, any tenant) can call `filestore_get_urls` to mint signed download URLs for arbitrary `file_store_id`s on any tenant (citizen ID documents, employee documents), call `access_roles_search`/`access_actions_search` to map a tenant's full role/action topology, or call `user_search` to enumerate another tenant's mobile numbers/emails/roles — cross-tenant PII and access-control reconnaissance at the lowest non-public tier.

vinothrallapalli-eGov:
**Finding (security — not posted inline, pre-existing code untouched by this diff):** `getLocalDoc`/`getEngramDoc` (`digit-mcp/src/tools/docs.ts`, lines 138 and 213) still guard filesystem reads with a blacklist `filename.includes('..') || filename.includes('/')`, instead of the new realpath-based, symlink-safe `resolveArtifactPath()` this same PR introduced and uses in `snapshot.ts` for the identical class of bug.

**Scenario:** A single path segment with no '/' or '..' can still be a symlink inside `DOCS_DIR`/`ENGRAMS_DIR` pointing outside it; the blacklist doesn't catch it and `readFile` follows the symlink. The PR's own comment notes engram files are written by the agent from prior session content, making this a realistic escape path that the rest of the PR's traversal hardening didn't reach.

vinothrallapalli-eGov:
**Finding (security — not posted inline, pre-existing code untouched by this diff):** `persister_monitor`'s parity probe (`digit-mcp/src/tools/monitoring.ts`, ~line 443) reimplements the old insecure inline pattern (`if (!digitApi.isAuthenticated()) { ...CRS_USERNAME/CRS_PASSWORD...await digitApi.login(...) }`) with no `getAuthMode()` check, instead of calling the centralized `ensureAuthenticated()`.

**Scenario:** Currently low-risk only because the tool is `access: 'admin'` so `digitApi` is normally already authenticated by handler entry — but it is a second, unreviewed copy of exactly the fallback logic this PR centralized to prevent 'no credentials → act as ADMIN', and reintroduces that risk if this tool's access tier is ever loosened or its probe logic copied elsewhere.

singhalkarun:
Thanks — verified all ten. Nine were real and are fixed in `7fda973e` (CI green: 79 policy + 34 enforcement checks). Two were wrong on severity; details below.

### The two env-credential fallbacks — the most important ones

Both confirmed. `3364fd7`'s claim that it replaced "eight near-identical copies" was simply wrong: **`user.ts:393` was a ninth** (behind `user_search`/`user_create`/`user_role_add`, the last of which can grant SUPERUSER) and **`monitoring.ts:443` a tenth**. Neither consulted `getAuthMode()`.

Rather than fix the two you spotted, I swept for all of them — which found two more things you didn't:

- **`configure` fell back to env credentials in token mode.** The caller is already authenticated as themselves there, so the fallback let an admin on one tenant silently adopt the container's ADMIN account.
- **Seven more hardcoded `'eGov@123'` literals** in `mdms-tenant.ts` that `MCP_DEFAULT_PROVISIONING_PASSWORD` was supposed to have centralized — so that env var was partly inert.

Three structural tests now make this class un-reintroducible: no file under `src/tools/` may reference `CRS_USERNAME`/`CRS_PASSWORD` (except `mdms-tenant.ts`, where `configure` legitimately takes credentials as arguments), none may define its own `ensureAuthenticated`, none may hardcode the provisioning password. I verified they bite by reintroducing your `user.ts` copy — 3.6b and 3.6c both fail.

### Form-2 empty token — correct, and I introduced it

Confirmed exactly as described: `token: ''` → `isAuthenticated()` accepts (`!== null`) → `encHeaders()` omits (`if (this.authToken)`). Introduced two commits ago when I removed a duplicate auth pass from `runStreamed`. Form 2 now carries the token `login()` actually minted, and `isAuthenticated()` no longer counts an empty string as a credential.

### Tiers, docs symlink, cache, `isInside`, double redaction

All fixed as described. Tier census moves to **6 public / 25 employee / 39 admin**, pinned by test 3.3b. The `getLocalDoc`/`getEngramDoc` blacklist is replaced with the same realpath-and-reject-symlinks containment the artifact path uses — verified an `engram://` symlink to `/tmp` is now refused. Cache eviction drops expired entries first, then oldest, instead of `clear()`.

### Two I'm recording as not-defects

**`runStreamed` 404-before-auth is not an oracle.** All three call sites pass a hardcoded tool name (`tenant_bootstrap`, `city_setup`, `tenant_cleanup`) — there's no attacker-controlled `toolName` to enumerate with. I reordered it anyway so a future caller passing user input can't reopen one, but the stated attack doesn't work today.

**The `user.ts` fallback was latent, not live** — masked exactly as your own comment notes, because HTTP routes populate auth state before any handler runs. Fixed as a latent hazard; it was not exploitable on the current dispatch order.

Both were worth reporting — the fixes are cheap and the reasoning holds for the next refactor.

subhashini-egov (approved):

PR #1781, opened 2026-08-17
fix(dashboard): respect analytics batch budget

### LINKED ISSUE #1758: [Bug] Dashboard Analytics Request failed
# Description

What went wrong? Be specific.
When all KPIs are added via the '+ Add KPI' button, the request fails with a 400 error: 'Analytics request failed.'
---

# Steps to Reproduce

1. Login to ADMIN account
2. Add all the KPI's from the given list

Verified after recreating this twice. 

---
# Environment

- **Deployment**:  https://bometfeedbackhub.digit.org/dashboard-ui/employee/dashboard 
- **User role**: ADMIN 

---

# Screenshots / Logs

Attach screenshots, browser console errors, or service logs if available.

Image

---

KDwevedi:
## Summary

- advertise the backend analytics batch ceiling through authenticated and public pack responses
- split expanded dashboard query plans into sequential bounded requests and merge their result/error envelopes
- preserve successful chunks when another chunk fails, while surfacing base and companion-query errors on the affected tile
- add a production-catalog regression proving the PGR_ADMIN Add All plan expands from 28 tiles to 51 query refs and is sent as 50 + 1

## Root cause

The canonical PGR_ADMIN catalog contains 28 visible tiles, but comparison, time-series, and pin companions expand Add All into 51 named queries. The analytics endpoint atomically rejects any request above `MAX_BATCH_QUERIES = 50`, so the dashboard received HTTP 400 before any KPI was evaluated.

The client now consumes the server-advertised budget, with a fallback of 50 for older deployments. Chunks run sequentially so the fix does not increase concurrent database pressure. The response merger supports both deployed inline `{ error, message }` entries and an additive top-level `errors` map.

## Verification

- dashboard Node test suite: 247/247 passing
- targeted Java controller and batch-limit tests passing under JDK 17
- staged diff passes `git diff --check`

## Scope

This change applies only to `/dashboard-ui`. Handling the legacy `/dashboard` path is explicitly outside this effort. No dashboard layout or styling is changed.

Closes #1758

### DIFF
--- a/digit-ui-esbuild/products/dashboard/src/services/analyticsBatch.js
+++ b/digit-ui-esbuild/products/dashboard/src/services/analyticsBatch.js
@@ -1,8 +1,8 @@
+import { chunkValues, runSequentialChunks } from "./sequentialChunks";
+
 /** Backend fallback for deployments that predate the advertised pack budget. */
 export const DEFAULT_MAX_BATCH_QUERIES = 50;
 
-/** Companion query keys emitted by queryPlan.js for one rendered tile. */
-const COMPANION_SUFFIXES = ["__prior", "__series", "__pins"];
 const hasOwn = (value, key) => Object.prototype.hasOwnProperty.call(value, key);
 
 function objectMap(value) {
@@ -31,7 +31,11 @@ function canonicalError(value, fallback = null) {
     const fallbackCode = fallbackError?.code || fallbackError?.error;
     return {
       code: value,
-      message: String(fallbackCode === value && fallbackError?.message ? fallbackError.message : value),
+      message: String(
+        String(fallbackCode) === String(value) && fallbackError?.message
+          ? fallbackError.message
+          : value
+      ),
     };
   }
   if (value && typeof value === "object") {
@@ -77,11 +81,7 @@ export function normalizeAnalyticsBatch(payload) {
 export function chunkQueryRefs(refs, maxBatchQueries) {
   const entries = Object.entries(objectMap(refs));
   const limit = batchLimit(maxBatchQueries);
-  const chunks = [];
-  for (let offset = 0; offset < entries.length; offset += limit) {
-    chunks.push(Object.fromEntries(entries.slice(offset, offset + limit)));
-  }
-  return chunks;
+  return chunkValues(entries, limit).map((chunk) => Object.fromEntries(chunk));
 }
 
 /**
@@ -89,11 +89,17 @@ export function chunkQueryRefs(refs, maxBatchQueries) {
  * transport fails. Sequential execution preserves the backend cap's intent:
  * adding a second HTTP request must not double concurrent PostgreSQL pressure.
  */
-export async function runChunkedAnalyticsBatch(refs, maxBatchQueries, executeChunk) {
-  const chunks = chunkQueryRefs(refs, maxBatchQueries);
-  if (!chunks.length) {
-    return { results: {}, errors: null, partial: false, asOf: null, calendar: null, scope: null };
-  }
+export async function runChunkedAnalyticsBatch(
+  refs,
+  maxBatchQueries,
+  executeChunk,
+  options = {}
+) {
+  // Preserve the old transport contract for an empty/stale query plan: the
+  // backend still owns the authoritative asOf/calendar response and the call's
+  // telemetry. An empty plan is one request, not zero requests.
+  const boundedChunks = chunkQueryRefs(refs, maxBatchQueries);
+  const chunks = boundedChunks.length ? boundedChunks : [{}];
 
   const results = Object.create(null);
   const errors = Object.create(null);
@@ -104,9 +110,10 @@ export async function runChunkedAnalyticsBatch(refs, maxBatchQueries, executeChu
   let successfulChunks = 0;
   let firstFailure = null;
 
-  for (const chunk of chunks) {
-    try {
-      const response = normalizeAnalyticsBatch(await executeChunk(chunk));
+  const outcomes = await runSequentialChunks(chunks, executeChunk, options);
+  for (const { chunk, value, error } of outcomes) {
+    if (!error) {
+      const response = normalizeAnalyticsBatch(value);
       successfulChunks += 1;
       Object.assign(results, response.results);
       if (response.errors) Object.assign(errors, response.errors);
@@ -116,7 +123,7 @@ export async function runChunkedAnalyticsBatch(refs, maxBatchQueries, executeChu
       }
       if (calendar == null && response.calendar != null) calendar = response.calendar;
       if (scope == null && response.scope != null) scope = response.scope;
-    } catch (error) {
+    } else {
       firstFailure ||= error;
       partial = true;
       const message = errorMessage(error);
@@ -133,21 +140,16 @@ export async function runChunkedAnalyticsBatch(refs, maxBatchQueries, executeChu
   return {
     results,
     errors: Object.keys(errors).length ? errors : null,
-    partial: partial || Object.keys(errors).length > 0,
+    partial,
     asOf,
     calendar,
     scope,
+    roundTrips: chunks.length,
   };
 }
 
-/** Return the first failure that makes a rendered tile incomplete. */
+/** Return only the base-query failure that prevents a tile from rendering. */
 export function errorForTile(errors, kpiId) {
   if (!errors || !kpiId) return null;
-  if (hasOwn(errors, kpiId) && errors[kpiId]) return errors[kpiId];
-  for (const suffix of COMPANION_SUFFIXES) {
-    const key = `${kpiId}${suffix}`;
-    const companion = hasOwn(errors, key) ? errors[key] : null;
-    if (companion) return companion;
-  }
-  return null;
+  return hasOwn(errors, kpiId) && errors[kpiId] ? errors[kpiId] : null;
 }
--- a/digit-ui-esbuild/products/dashboard/src/services/analyticsBatch.js
+++ b/digit-ui-esbuild/products/dashboard/src/services/analyticsBatch.js
@@ -1,0 +1,153 @@
+/** Backend fallback for deployments that predate the advertised pack budget. */
+export const DEFAULT_MAX_BATCH_QUERIES = 50;
+
+/** Companion query keys emitted by queryPlan.js for one rendered tile. */
+const COMPANION_SUFFIXES = ["__prior", "__series", "__pins"];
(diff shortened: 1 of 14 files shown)


digit-ui-esbuild/products/dashboard/src/services/analyticsBatch.js
```diff
@@ -0,0 +1,153 @@
...
+    errors: Object.keys(errors).length ? errors : null,
+    partial: partial || Object.keys(errors).length > 0,
+    asOf,
+    calendar,
+    scope,
+  };
+}
+
+/** Return the first failure that makes a rendered tile incomplete. */
+export function errorForTile(errors, kpiId) {
```
vinothrallapalli-eGov, line 152: **Bug:** `errorForTile` now blanks the entire tile whenever only a companion query (`__prior`/`__series`/`__pins`) fails, even though the base query succeeded — the old lookup only matched the exact base kpiId key.

**Scenario:** A card's base value query succeeds but its `card__prior` comparison query fails. `errors` contains only `card__prior`; `errorForTile('card', ...)` falls through the COMPANION_SUFFIXES loop and returns that error. `AdminDashboard.jsx` passes this to `KpiTile`, whose unconditional early return on `error` discards `result` entirely — so a tile with valid base data renders only an error box instead of the metric (with maybe a missing comparison badge).

digit-ui-esbuild/products/dashboard/src/services/analyticsBatch.js
```diff
@@ -0,0 +1,153 @@
...
+  const results = Object.create(null);
+  const errors = Object.create(null);
+  let partial = false;
+  let asOf = null;
+  let calendar = null;
+  let scope = null;
+  let successfulChunks = 0;
+  let firstFailure = null;
+
+  for (const chunk of chunks) {
```
vinothrallapalli-eGov: **Bug:** Sequential chunk execution has no cancellation when a newer request supersedes it, so a superseded batch still runs all remaining chunk HTTP calls to completion.

**Scenario:** A user with a 51-ref 'Add All' plan (2 chunks) changes a filter mid-load. `AdminDashboard.jsx` only checks `reqId !== reqIdRef.current` after the whole promise settles, not between chunks inside this for-of loop. Every abandoned generation still issues both sequential chunk POSTs against the DB before being discarded — exactly the added DB pressure the sequential-chunking design was meant to avoid, now multiplied per chunk and extended in duration on every rapid filter change.

digit-ui-esbuild/products/dashboard/src/services/analyticsBatch.js
```diff
@@ -0,0 +1,153 @@
...
+}
+
+/**
+ * Execute bounded chunks sequentially, retaining successful chunks if a later
+ * transport fails. Sequential execution preserves the backend cap's intent:
+ * adding a second HTTP request must not double concurrent PostgreSQL pressure.
+ */
+export async function runChunkedAnalyticsBatch(refs, maxBatchQueries, executeChunk) {
+  const chunks = chunkQueryRefs(refs, maxBatchQueries);
+  if (!chunks.length) {
```
vinothrallapalli-eGov: **Bug:** When refs resolve to an empty map (e.g. tiles reference kpiIds missing from a stale/racing catalog), this skips the network call entirely, dropping the backend-echoed `asOf`/calendar and API telemetry the old single-request path always produced.

**Scenario:** `queryPlan.js`'s `buildRefs` skips any tile whose kpiId isn't in `kpis` (`if (!def) continue`), so `tiles.length > 0` with `refs = {}` is reachable on a stale layout/catalog race. `chunkQueryRefs` then returns 0 chunks, and this returns immediately without ever calling `executeChunk`/`postAnalytics` — unlike the old code, which always sent the request and got a live `asOf` back. 'Last updated' now shows nothing and `recordApiCall` telemetry for that render is silently lost.

digit-ui-esbuild/products/dashboard/src/services/analyticsBatch.js
```diff
@@ -0,0 +1,153 @@
...
+ * contracts may also publish a top-level errors map. Keep both readable while
+ * tenants roll independently.
+ */
+function canonicalError(value, fallback = null) {
+  const fallbackError = fallback && typeof fallback === "object" ? fallback : null;
+  if (typeof value === "string" && value) {
+    const fallbackCode = fallbackError?.code || fallbackError?.error;
+    return {
+      code: value,
+      message: String(fallbackCode === value && fallbackError?.message ? fallbackError.message : value),
```
vinothrallapalli-eGov: **Bug:** `canonicalError` uses a strict `fallbackCode === value` comparison in the string-value branch but a coerced `String(fallbackCode) === String(code)` comparison in the object-value branch — inconsistent equality rules in the same function.

**Scenario:** If a chunk's inline result `error` is a string while the additive top-level `errors[key]` entry supplies the same logical code as a non-string (or vice versa), the string branch fails to recognize them as matching where the object branch would, so the richer fallback `message` is silently dropped and the tile shows the bare code as its message instead of the more descriptive one.

digit-ui-esbuild/products/dashboard/src/services/analyticsBatch.js
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  // Preserve the existing global-banner behaviour only when no data request
+  // succeeded at all. A partial transport failure keeps the usable tile data.
+  if (successfulChunks === 0) throw firstFailure;
+
+  return {
+    results,
+    errors: Object.keys(errors).length ? errors : null,
+    partial: partial || Object.keys(errors).length > 0,
```
vinothrallapalli-eGov: **Cleanup:** The final `partial: partial || Object.keys(errors).length > 0` is dead/redundant — every path that adds an entry to `errors` (normalizeAnalyticsBatch's hasErrors-driven partial, or the catch block's explicit `partial = true`) already sets the accumulated `partial` flag in the same iteration. Not a crash, but a maintenance trap for future readers.

digit-ui-esbuild/products/dashboard/src/services/analyticsBatch.js
```diff
@@ -0,0 +1,153 @@
...
+  }
+  return chunks;
+}
+
+/**
+ * Execute bounded chunks sequentially, retaining successful chunks if a later
+ * transport fails. Sequential execution preserves the backend cap's intent:
+ * adding a second HTTP request must not double concurrent PostgreSQL pressure.
+ */
+export async function runChunkedAnalyticsBatch(refs, maxBatchQueries, executeChunk) {
```
vinothrallapalli-eGov: **Cleanup:** `runChunkedAnalyticsBatch` reimplements the same 'chunk requests, tolerate partial failure, rethrow only if everything failed' pattern already present in `boundaryService.js`'s `fetchBoundariesByCodes` (chunk size 100, lastError/rethrow-if-all-failed logic) instead of factoring a shared chunked-request-runner.

Two independently-maintained copies of the same partial-failure/throw-if-total-failure policy now exist in the same product; a future policy change has to be found and applied in both places, and they will silently drift if only one is updated.

vinothrallapalli-eGov:
**Cleanup (code review):** `markBatchStart`/`markAllWidgetsReady` (`AdminDashboard.jsx` ~L746) still treat a whole multi-chunk sequential batch as a single timing span, with nothing distinguishing 'N sequential round trips' from 'one genuinely slow query' in the emitted metric.

**Scenario:** For a catalog whose Add-All plan crosses 50 refs (2 sequential HTTP round trips via the new `runChunkedAnalyticsBatch`), `dashboard.all_widgets_ready.ms` now reflects the sum of both round trips. An ops alert tuned against this metric's historical single-request baseline could false-positive a performance regression purely because a tenant's catalog grew past the batch cap, not because anything got slower.

(Not posted inline — this code path isn't part of the diff.)

digit-ui-esbuild/products/dashboard/src/services/analyticsBatch.js
KDwevedi: Fixed in bca8680e2. errorForTile now returns only an exact base-query failure, so a failed __prior/__series/__pins companion no longer replaces a successful base result. The regression test now asserts that card__prior does not blank card.

digit-ui-esbuild/products/dashboard/src/services/analyticsBatch.js
KDwevedi: Fixed in bca8680e2. The dashboard now aborts the superseded fetch and invalidates its request generation during effect cleanup. The shared sequential runner also checks shouldContinue/signal before every chunk, with a regression test proving a 51-ref stale generation issues only its first call.

digit-ui-esbuild/products/dashboard/src/services/analyticsBatch.js
KDwevedi: Fixed in bca8680e2. An empty ref map is deliberately executed as one empty-query request, preserving backend asOf/calendar metadata and API telemetry. The new test verifies the call and freshness response.

digit-ui-esbuild/products/dashboard/src/services/analyticsBatch.js
KDwevedi: Fixed in bca8680e2. Both canonicalError branches now compare normalized string codes, and the regression test covers a string inline error with a logically identical fallback code/message.

digit-ui-esbuild/products/dashboard/src/services/analyticsBatch.js
KDwevedi: Fixed in bca8680e2. The returned partial flag now uses the accumulated partial value directly; the redundant Object.keys(errors) condition was removed.

digit-ui-esbuild/products/dashboard/src/services/analyticsBatch.js
KDwevedi: Fixed in bca8680e2. Analytics and boundary batching now use the shared sequentialChunks runner for ordered execution, partial-failure outcomes, and cancellation. Each caller retains only its domain-specific merge and total-failure handling.

KDwevedi:
Addressed the timing-metric concern in bca8680e2. Each settled batch now emits `dashboard.analytics_round_trips.count`, and the correlated `dashboard.load` record includes `analytics_round_trips`, so operators can distinguish a multi-round-trip catalog threshold from a genuinely slow single request while retaining the end-to-end `dashboard.all_widgets_ready.ms` meaning.

vinothrallapalli-eGov (approved):

Issue #1739, opened 2026-08-12
[Enhancement] Rework the O&M Runbook based on feedback


subhashini-egov:
Feedback has been provided on v1 of the O&M runbook. Please rework that and commit to master branch at the earliest.

subhashini-egov:
@dhruv-1001 Just pinging to surface this to the top. Please check on the dev server and update the runbook. Let's review and close out.

PR #1788, opened 2026-08-18
docs: rework the O&M runbook and document the localhost full/slim presets (#1739, #1598)

dhruv-1001:
Two documentation gaps plus one collection fix, each verified against a running
stack or against the config the claim describes rather than against memory.

## O&M runbook rework (#1739, follow-up to #1597)

Answering the review feedback, and correcting what had gone stale:

- **Gatus categories are now documented.** The catalogue holds 57 checks in 12
  groups, not "~50 in ten" — `Observability` and `Public Endpoint` were missing
  entirely, and the group the docs called "Sign-in / identity" is labelled
  `Keycloak` on the actual dashboard. Added per-group check counts (verified
  equal to the config, group by group) and a section on why a deployment sees
  fewer groups than the catalogue defines, since all but four are feature-gated.
- **Gatus history survives restarts now** (#1610 moved it to SQLite). Five
  places still said in-memory / lost on restart.
- **Container memory limits exist.** The runbook said no container has one,
  having grepped for `mem_limit`; the caps are written as
  `deploy.resources.limits.memory` and are set on 10 of 60 services. Listed
  them, and what a capped container does differently when it runs out.
- **Declared JVM heap is ~12.8 GB across 22 services**, not ~6.3 GB.
- **Added the L1/L2 line**, which the feedback asked for: anything with a screen
  is L1 — including creating users and master data — and backend, complexity and
  the server are L2. Reflected in both runbooks, and the "not covered" list no
  longer contradicts it.

Retention, log-file caps and the raise-CPU/memory procedure were already
covered; re-verified the numbers against loki-config, tempo-config, the
Prometheus flag and group_vars, and they are correct.

## Localhost full and slim (#1598)

- Commented every setting in `localhost-full.yml.example` and
  `localhost-slim.yml.example`: what it does, what values it takes, and why this
  preset picks the one it does. Comments only — parsed both before and after,
  no key or value changed.
- Added `local-setup/docs/LOCALHOST-FULL-AND-SLIM.md`: what the two profiles
  are, that they differ by exactly two flags, how to run one, the become-password
  gotcha (`deploy.sh` never passes -K, which only bites on a local target), the
  preflight warning both presets emit and why it is harmless, and how to switch.
- Both presets verified to pass `scripts/preflight.py`.

## digit-core-validation collection

One request in `digit-core-validation.postman_collection.json` returned 400: it
sent the boundary-hierarchy criteria as query parameters instead of in the body.
Fixed; all 13 requests now return 200 against a local stack.

## Not in this PR

The local-setup setup walkthroughs (`local-setup/README.md`,
`local-setup/docs/LOCAL-SETUP-GUIDE.md`, the root README) are handled
separately so one PR owns them.

### DIFF
--- a/docs/operations/l2-diagnosis.md
+++ b/docs/operations/l2-diagnosis.md
@@ -6,6 +6,14 @@ services, read logs across the whole stack, and change configuration.
 **Your job:** take what L1 captured and identify **which component is failing and why** —
 far enough that the fix is either yours to apply or clearly a product defect to send on.
 
+**What should reach you:** anything that needs the server, a log, the database, or a
+configuration file the UI does not expose — plus every question of *why*. Work that has a
+screen (creating users, editing master data, loading the onboarding spreadsheets) is L1's,
+even when it looks administrative; see
+[Who does what](README.md#who-does-what--the-l1--l2-line). If those requests are reaching
+you, the usual cause is that the service desk has no Configurator / HRMS admin login — fix
+that once rather than absorbing the queue.
+
 This assumes [L1 first response](l1-first-response.md) has been worked, or that you're
 starting cold and have done it yourself. Every query and panel name here is real and was run
 against a live deployment.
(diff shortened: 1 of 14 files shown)


docs/operations/l2-diagnosis.md
```diff
@@ -348,20 +356,43 @@ is a deployment change, not something to apply on the box**, because anything ed
 
 | Limit | Where it lives | Notes |
 |---|---|---|
-| **JVM heap** (`-Xmx`) | `JAVA_TOOL_OPTIONS` or `JAVA_OPTS` per service in the compose file | This is the ceiling that actually causes `OutOfMemoryError`. Currently ~6.3 GB of heap is allocated across the JVM services |
-| **Container memory** | not set today | The deployed compose has **no `mem_limit` on any container**, so a container can use as much host RAM as it asks for. Only the JVM heap bounds it |
-| **Container CPU** | not set today | No `cpus` limit either — services compete freely for host CPU |
+| **JVM heap** (`-Xmx`) | `JAVA_TOOL_OPTIONS` or `JAVA_OPTS` per service in the compose file | This is the ceiling that actually causes `OutOfMemoryError`. 22 services declare `-Xmx`, totalling **~12.8 GB** — more than a 16 GB host has, so they rely on not all peaking at once |
```
KDwevedi, line 359: So we don't have configured memory limits here @dhruv-1001 ?

KDwevedi (approved):
The documentation seems to be correct, if a little verbose and claude-flavored at various places. 
Approving, though maybe we should spend some human time reducing verbosity and/or repetition here and there.

Issue #1393, opened 2026-07-22
[Bug] MapConfig not seeded at state-level tenant on 0→1 install


priyanshu-egov:
On a fresh install, the default-data-handler seeds the `RAINMAKER-PGR.MapConfig`
**schema** from `utilities/default-data-handler/src/main/resources/schema/RAINMAKER-PGR.json`,
but no **data** record is inserted for the state-level tenant. The only path
that currently writes a `MapConfig` record is Configurator Phase 2
(`mdmsService.upsertMapConfig`, via `deriveMapPosition` in
`configurator/src/utils/mapConfigFromBoundaries.ts`), which fires per-city
during boundary onboarding — the state root itself never gets a base/default row.

KDwevedi:
## Scoped and code-verified

Verified against `upstream/develop` at `83330b047` on 2026-08-18.

### Root cause

- `utilities/default-data-handler/src/main/resources/schema/RAINMAKER-PGR.json` registers `RAINMAKER-PGR.MapConfig` with required/unique `code`.
- No `RAINMAKER-PGR.MapConfig.json` exists under either `mdmsData` or `mdmsData-dev`, so startup has no data resource to create.
- `StartupSchemaAndMasterDataInitializer` registers schemas and loads MDMS for `default.tenant.id`.
- Configurator Phase 2 derives geometry and calls `upsertMapConfig(tenantId, ...)` for the city being onboarded; it does not establish a state-root baseline.

The issue premise is therefore confirmed by the current code and resource set. This is static/code-path verification; no disposable fresh deployment was available for runtime verification.

### Important scope constraint

Do **not** fix this by only adding the row to the ordinary `mdmsData/**/*.json` bundle without isolating it. `DataHandlerService.loadNewTenantProductionData` reuses that bundle for every target tenant. That would give every newly onboarded city an owned default row, preventing it from cleanly inheriting later state-level map styling before Phase 2 creates its geometry-specific override.

A dev-only file is also insufficient because `mdmsData-dev` is conditional on `dev.enabled` and does not cover a production 0-to-1 install.

### Implementation scope

1. Add a root-only MDMS resource path for data loaded by `StartupSchemaAndMasterDataInitializer` at `default.tenant.id`, but not by `loadNewTenantProductionData`.
2. Add `RAINMAKER-PGR.MapConfig.json` to that root-only path with one minimal singleton row:
   - `code: DEFAULT`
   - `baseMapTheme: voyager`
[...]

PR #1792, opened 2026-08-18
fix(ddh): seed default MapConfig for every tenant

KDwevedi:
## Summary

- seed a minimal `RAINMAKER-PGR.MapConfig` `DEFAULT` row through the standard default-data-handler MDMS bundle
- apply the same neutral presentation defaults to the root and every tenant provisioned later
- keep geography fields absent until Configurator Phase 2 derives and merges them
- document that tenant-owned defaults intentionally shadow later state-root presentation changes
- add a resource/schema contract test and cover preservation of seeded presentation fields during Phase 2 upsert

Relates to #1393.

## Verification

- `mvn -q test` under Java 17 — passed
- focused Configurator MapConfig and boundary derivation tests — 15 passed
- ESLint on the changed Configurator test — passed
- `git diff --check` — passed

The repository-wide Configurator lint command still reports its existing unrelated lint backlog; the changed test itself is clean.


Lokendra-egov:
A couple of things worth a look before merging:

**1. The hardcoded seed will block inheriting a state-root customization made before a tenant is provisioned.** Every newly-provisioned tenant gets its own `DEFAULT` row with the literal constants (`baseMapTheme: "voyager"`, `wardHighlightColor: "#FFA74F"`) via `loadNewTenantProductionData`. `upsertMapConfig`'s own-vs-inherited logic (`configurator/src/api/services/mdms.ts:159-172`, unchanged here) always finds an "own" record once one exists, so it never falls through to inherit from the parent. Concrete case: `ke` root's `wardHighlightColor` gets set to a brand color, then `ke.newcity` is provisioned afterward — it gets the neutral default forever, not the brand color, and nothing in the documented Phase 2 merge (center/boundaryTenantId/zoom only) corrects it. Worth confirming this pre-existing-customization-at-creation case is actually intended, since the PR's own wording ("shadow later state-root presentation changes") only discusses changes made *after* seeding.

**2. The "schema contract test" doesn't validate against the schema.** It hardcodes the expected values (`assertEquals("voyager", ...)`, `assertEquals("#FFA74F", ...)`) and checks that the seed's field *names* exist in `properties` — it never runs the schema's own enum/pattern/bounds validation. A future edit setting `baseMapTheme` to an invalid value like `"Voyager"`, or a numeric field past its stated max, wouldn't be caught by this test as currently written.

Smaller: the same test also asserts on `required.get(0)`/`x-unique.get(0)` positionally, so a harmless reorder of those arrays would fail it for no real reason.

One thing I couldn't fully verify: a tenant that already holds some other/legacy-keyed MapConfig row would have the whole `DEFAULT` seed skipped for that schema (the bulk loader's idempotency check is per-schemaCode, not per-record). The docs suggest a Flyway migration re-keys such rows to `DEFAULT` first, which would make this a non-issue, but nothing in this PR exercises that interaction directly — might be worth a test if that migration ordering isn't airtight.

Data itself (field presence/values) checked out fine — geography fields are genuinely absent (not null), and the seeded values are schema-valid.

KDwevedi:
Addressed in 32d42ed4c.

**1. Pre-existing state-root customization is no longer shadowed.** You were right that the shared bundle was the wrong vehicle — `loadNewTenantProductionData` replays `mdmsData/**` for every tenant, so each new city got its own literal row and `upsertMapConfig` never fell through to the parent. The seed now lives in a root-only `stateMdmsData/` bundle loaded by `StartupSchemaAndMasterDataInitializer` for `default.tenant.id` alone (new `state.mdms.data.path`). A city provisioned later owns no MapConfig row, so mdms-v2 resolves it up the tree and it picks up whatever the root currently has, including changes made after the seed. When Phase 2 does write a city row, the existing shadow-create path merges the patch over the inherited values — covered by a new Configurator test using the exact root-seed shape. This restores the acceptance criteria in #1393; the docs and the v2.12-beta changelog now describe root-only seeding instead of per-tenant.

**2. The contract test now validates against the schema.** It builds a draft-07 validator from the checked-in `definition` node and validates every seed record, so enum/pattern/bounds/`additionalProperties` are enforced rather than restated. A companion negative-control test asserts the validator actually rejects `"Voyager"`, a non-hex ward colour, `defaultZoom: 99`, and an unknown field — otherwise the validation would be theatre. The remaining hand-written assertions are now about the seed's *shape* (single record, `code: DEFAULT`, no geography fields), not its literal presentation values.

**Smaller:** `required`/`x-unique` are asserted with a contains check, not by index.

**Legacy colour-keyed row:** the per-schemaCode idempotency skip is the intended behaviour here (#1393 asks for "no second active row"), and with root-only seeding it only applies to the root. I documented the ordering in `docs/map-config.md`: until the Flyway migration re-keys the legacy row to `DEFAULT`, that root gets no seed at all — which is the desired outcome rather than a duplicate singleton.

Verification: `mvn clean test` in `default-data-handler` — 13 passed (JDK 17). I mutation-checked the new startup test by removing the root-only load: it fails as expected, then passes again on restore. Configurator MapConfig suite — 5 passed; ESLint on the changed test — clean.

Lokendra-egov (approved):

Issue #1517, opened 2026-07-30
[Bug] Error encountered while creating the workflow in Novu


pradeepkumarcm-egov:
# Description
 The below variable declared in  bootstrap-novu-whatsapp.sh 

#   NOVU_SMS_BODY          (default: Complaint {{payload.complaintNo}} status is {{payload.status}})

This script creating problem when workflow created in novu,

<img width="1470" height="956" alt="Image" src="https://github.com/user-attachments/assets/2716b4a0-bd5e-4a1c-8064-3f72827b4d4e" />

The expected value in .env.novu is as below,

NOVU_SMS_BODY='Complaint {{payload.complaintNo}} status is {{payload.status}}'

rahuljose-egov:
**Process fix**

KDwevedi:
RCA and fix are ready in PR #1796.

The SMS body parsing reported here is a real bootstrap bug, but it is separate from opening the workflow creation UI. The broader deployed-instance failure was caused by missing browser-facing routes/URLs in the deployment contract: the dashboard navigates to `/env/:environmentSlug/workflows/create`, the browser API needs `/novu-api`, and Socket.IO connects at `/socket.io` when given an origin URL. The Helm deployment exposed only `/novu` and injected cluster-only service URLs.

The PR fixes both layers, adds regression contracts, and documents the required routes. I also validated read-only against the current Bomet and Mozambique deployments: the create form opens and survives a hard reload. The old `digit.mctd.gov.mz` hostname currently fails HTTPS; the configured Mozambique host is `cms-pilot.digit.org`.

The item remains **In progress** while PR #1796 is under development/review. After merge it should move to **To Review** for QA validation; QA owns Done/closure.

KDwevedi:
Deployment verification for merged PR #1796 is not yet clean, so this remains **In progress** rather than moving to **To Review**.

Checked the current Bomet and Mozambique public hosts on 2026-08-24:

- `/env/<slug>/workflows/create` serves the Novu dashboard on both hosts.
- `/novu-api/v1/health-check` returns 200 on both hosts.
- The required root Socket.IO transport route (`/socket.io/?EIO=4&transport=polling`) returns 404 on both hosts, while the legacy `/novu-ws/socket.io/` route responds.
[...]

PR #1796, opened 2026-08-18
Fix deployed Novu workflow creation and routing (#1517)

KDwevedi:
## Summary

- load Novu dotenv defaults safely so notification bodies containing spaces survive bootstrap
- expose the dashboard deep-link, API, and WebSocket routes required by the Helm deployment
- use an origin-only browser WebSocket URL and proxy the Socket.IO root path
- document the ingress contract and add static regression coverage

## Root cause

Issue #1517 identified a real bootstrap failure: shell-sourcing an unquoted message body splits it into commands. That occurs during workflow bootstrap, however, and cannot prevent the dashboard create form from opening.

The page-opening failure was in the deployment contract. The Novu 2.3 dashboard is built for root-relative routes: creation navigates to `/env/:environmentSlug/workflows/create`, while API and Socket.IO calls also need public browser routes. The Helm chart exposed only `/novu` and injected cluster-only service URLs. The Compose route had already gained `/env`, but its WebSocket URL still encoded `/novu-ws` as though Socket.IO treated it as a path.

## Verification

- Jest deployment and infrastructure contract suites: 20 tests passed
- Helm lint and template passed with existing-secret and TLS configuration
- rendered chart has four ingress resources, TLS on each, and `/env`, `/novu-api`, `/novu-ws`, and `/socket.io` routes
- dotenv parser checks cover quoted/unquoted spaces and caller environment precedence
- live read-only browser validation on the current Bomet and Mozambique deployments: create form opens and survives a hard reload

Refs #1517

QA should validate the deployment and owns moving the issue from To Review to Done and closing it.

### DIFF
--- a/backend/novu-bridge/config/.env.novu
+++ b/backend/novu-bridge/config/.env.novu
@@ -16,7 +16,7 @@ NOVU_INTEGRATION_NAME=twilio-whatsapp
 NOVU_INTEGRATION_ID=twilio-whatsapp
 NOVU_WORKFLOW_ID=complaints-whatsapp
 NOVU_WORKFLOW_NAME=ComplaintsWhatsAppWorkflow
-NOVU_SMS_BODY=Complaint {{payload.complaintNo}} status is {{payload.status}}.
+NOVU_SMS_BODY='Complaint {{payload.complaintNo}} status is {{payload.status}}'
 
 # Event workflows to create/check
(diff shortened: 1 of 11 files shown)


KDwevedi:
CI note: the two early failures are unrelated repository-baseline failures, not introduced by this PR:

- `gatus-coverage` reports the existing duplicate `grafana` Service name across `monitoring` and `digit`.
- `values keys must be read by their chart` reports the five existing stale `prometheus.*` entries in `KNOWN_UNREAD` and explicitly identifies their #1659 merge-order dependency. The same deploy-as-code workflow is already failing on recent `master` pushes.

The Novu-sensitive checks completed so far are green (`Config combination + infra contracts`, `Ansible static validation`, YAML duplicates, and Tilt Kubernetes load). The full render and remaining tests are still running.

Lokendra-egov (approved):

PR #1804, opened 2026-08-18
fix(digit-ui): match StateInfo by tenant code instead of picking index [0]

### LINKED ISSUE #1429: [Bug] State-level tenant onboarding: MDMS common-masters.StateInfo search returns multiple records, first one used by default
When onboarding a state-level tenant, common-masters.StateInfo in MDMS ends up with more than one row (e.g. one for the parent tenant pg and one for the child tenant mz). The MDMS search for a given tenantId returns both records in the array, and downstream code picks index [0] unconditionally instead of matching on the actual tenant — so the wrong StateInfo object (wrong logo/banner/config) can be applied.

Image

Image

Lokendra-egov:
## Summary
- State-level tenant onboarding can leave more than one `common-masters.StateInfo` row visible in a single MDMS search (e.g. one for the parent root tenant and one for the child state tenant).
- `digit-ui-esbuild/packages/libraries/src/services/molecules/Store/service.js` unconditionally picked `StateInfo?.[0]`, so a child tenant's citizens/employees could get the parent's logo, banner, and locale/localization config instead of their own.
- Added a `pickStateInfo(list, stateCode)` helper that matches the record whose `code` (the uppercase root tenant code) equals the tenant actually being loaded, falling back to `[0]` when nothing matches so single-record deployments are unaffected. Applied in both `getTenantConfig` and `digitInitData`.

Fixes #1429

## Test plan
- [x] Verified syntax with `node --check`
- [ ] Manual verification against a tenant with duplicate StateInfo rows (parent `pg` / child state) to confirm the child's branding now loads correctly

### DIFF
--- a/digit-ui-esbuild/packages/libraries/src/services/molecules/Store/service.js
+++ b/digit-ui-esbuild/packages/libraries/src/services/molecules/Store/service.js
@@ -37,6 +37,19 @@ const renderTenantLogos = (stateInfo, tenants) => {
   });
 };
 
+// State-level tenant onboarding can leave more than one common-masters.StateInfo
+// row visible for a given search (e.g. one for the parent root tenant and one
+// for the child state tenant). `code` is the uppercase root tenant code, so
+// match it against the tenant actually being loaded instead of assuming the
+// first array entry belongs to this tenant; fall back to [0] when nothing
+// matches so single-record deployments keep working unchanged.
+const pickStateInfo = (stateInfoList, stateCode) => {
+  const list = stateInfoList || [];
+  return (
+    list.find((info) => info?.code?.toUpperCase() === stateCode?.toUpperCase()) || list[0] || {}
+  );
+};
+
 export const StoreService = {
   getInitData: () => {
     return Storage.get("initData");
@@ -68,7 +81,7 @@ export const StoreService = {
       // return;
     }
     const { MdmsRes } = await MdmsService.init(stateCode);
-    const stateInfo = MdmsRes["common-masters"]?.StateInfo?.[0] || {};
+    const stateInfo = pickStateInfo(MdmsRes["common-masters"]?.StateInfo, stateCode);
     const uiHomePage = MdmsRes["common-masters"]?.uiHomePage?.[0] || {};
     return {
       languages: stateInfo.hasLocalisation ? stateInfo.languages : [{ label: "ENGLISH", value: Digit.Utils.getDefaultLanguage() }],
@@ -92,7 +105,7 @@ export const StoreService = {
   digitInitData: async (stateCode, enabledModules, modulePrefix) => {
 
     const { MdmsRes } = await MdmsService.init(stateCode);
-    const stateInfo = MdmsRes["common-masters"]?.StateInfo?.[0] || {};
+    const stateInfo = pickStateInfo(MdmsRes["common-masters"]?.StateInfo, stateCode);
     const uiHomePage = MdmsRes["common-masters"]?.uiHomePage?.[0] || {};
 
     const themeConfig = MdmsRes["common-masters"]?.ThemeConfig?.[0];


KDwevedi (commented):
Reviewed against the two live deployments (bomet, moz). Safe to merge — no regression on either — but it is a no-op on bomet, which is our largest instance of exactly this bug. Details below.

**Pagination is not a concern on this path — verified.** `MdmsService.init` posts to [`Urls.MDMS` = `/{MDMS_V1_CONTEXT_PATH}/v1/_search`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/develop/digit-ui-esbuild/packages/libraries/src/services/atoms/urls.js#L4) with no limit/offset, and the v1 server path builds its SQL in [`MdmsDataQueryBuilder`](https://github.com/egovernments/Digit-Core/blob/master/core-services/mdms-v2/src/main/java/org/egov/infra/mdms/repository/querybuilder/MdmsDataQueryBuilder.java#L12-L15), which has **no `LIMIT`/`OFFSET` clause at all** (only [`MdmsDataQueryBuilderV2`](https://github.com/egovernments/Digit-Core/blob/master/core-services/mdms-v2/src/main/java/org/egov/infra/mdms/repository/querybuilder/MdmsDataQueryBuilderV2.java#L93-L98) paginates). So `find` always scans the complete record set.

**`code` is the full lowercase tenant id, not "the uppercase root tenant code"** — the comment on `pickStateInfo` is misleading. Live values are `ke`, `mz`, `ke.bomet`, `pg`; the `toUpperCase()` on both sides makes it work anyway, but please fix the comment.

**The fix does not disambiguate bomet, where the duplicate is real and user-visible.** bomet boots at `stateTenantId = "ke"` and `tenantId=ke` returns two records **both with `code=ke`**, so `find` returns index 0 — bit-for-bit what `[0]` returns today:

```
# bomet: /opt/digit/nginx/globalConfigs.js:19  ->  var stateTenantId = "ke";
$ curl -s localhost:18000/egov-mdms-service/v1/_search -d '{"MdmsCriteria":{"tenantId":"ke",
    "moduleDetails":[{"moduleName":"common-masters","masterDetails":[{"name":"StateInfo"}]}]}}'
count 2
0  code=ke  name=Demo   ['en_IN', 'fr_FR', 'pt_PT']   <- picked today, and still picked after this PR
1  code=ke  name=Kenya  ['en_IN']
```

Index 0 is `order by data.createdtime desc`, so a re-write of the "Kenya" row silently drops `fr_FR`/`pt_PT` from the language switcher. The distinguishing field here is `name`, not `code` — two active rows sharing a `code` also violate the schema's own `x-unique: ["code"]`, so the real repair is deactivating the stale row; this PR can't cover it.

**Deployments booting at a child tenant get no match and fall through to `[0]`.** `tenantId=ke.bomet` returns rows whose `code` is `ke`, so the match fails. Worth a root-segment retry — `stateCode.split(".")[0]` — before the `list[0]` fallback.

**Micro-bug:** `info?.code?.toUpperCase() === stateCode?.toUpperCase()` evaluates `undefined === undefined` as true, so a record with no `code` matches a missing `stateCode`. Practically unreachable given [App.js's guard](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/develop/digit-ui-esbuild/src/App.js#L51-L57), but a leading `stateCode &&` is one token.

**Non-blocking scope note:** the same `StateInfo?.[0]` pick remains in workbench at [`LocalisationSearchConfig.js#L55`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/develop/digit-ui-esbuild/packages/modules/workbench/src/configs/LocalisationSearchConfig.js#L55) and [`LocalisationAdd.js#L43`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/develop/digit-ui-esbuild/packages/modules/workbench/src/pages/employee/LocalisationAdd.js#L43) — same class of bug, fine as a follow-up rather than widening this PR.

For the record, moz is clean: `stateTenantId = "mz"`, and `tenantId=mz` returns exactly one record (`code=mz`, `en_IN`+`pt_PT`) — the `pg` row never reaches the array because the query is `tenantid LIKE 'mz%'` and `FallbackUtil` narrows to ancestors of the requested tenant.

priyanshu-egov (approved):

PR #1805, opened 2026-08-18
fix(dashboard): make Reset also clear filters, not just KPI layout

### LINKED ISSUE #1471: [Bug] Dashboard: 'Reset' button does not clear the 'All types' filter
# Description
On the Dashboard page, after selecting a complaint type filter, clicking the 'Reset' button (top right, near Search/+Add KPI/Export) does not clear the selected type filter back to 'All types'.

---

# Steps to Reproduce
Step 1: Open the Dashboard page.
Step 2: Click the 'All types' dropdown and select 'Grievance' > 'All in Grievance'.
Step 3: Click the 'Reset' button.
Step 4: Check the value shown in the type filter.

---

# Expected Behavior
Clicking 'Reset' should clear all applied filters, returning the type filter to 'All types'.

---

# Actual Behavior
The type filter remains set to the previously selected value ('All in Grievance') after clicking 'Reset'. Note: the separate 'Clear' link next to the filters does correctly reset this filter, so 'Reset' and 'Clear' appear to be inconsistent.

Lokendra-egov:
## Summary
- The Dashboard header's "Reset" button and the filter bar's "Clear" link were two independent controls over two independent stores: Reset only called `resetLayout()` (rebuilds KPI widget grid geometry), never touching filter state, so the complaint-type tree filter (and every other filter) stayed exactly as set after clicking Reset, while Clear correctly restored it to "All types".
- Reset now calls both `resetLayout()` and `clearFilters()`, matching the reported expectation that "Reset" clears everything, not just widget positions.

Closes #1471

## Test plan
- [x] Full existing `products/dashboard` test suite (234 tests) passes unchanged
- [x] Syntax-checked the changed file via esbuild
- [ ] Manual repro per the issue's steps: select a complaint type filter (e.g. Grievance > All in Grievance), click Reset, confirm the type filter returns to "All types"

### DIFF
--- a/digit-ui-esbuild/products/dashboard/src/AdminDashboard.jsx
+++ b/digit-ui-esbuild/products/dashboard/src/AdminDashboard.jsx
@@ -900,6 +900,18 @@ const AdminDashboardInner = ({ onSignOut, embedded = false, publicMode = false, 
     },
     [clearFilters]
   );
+  // The header's "Reset" button and the filter bar's "Clear" link are two
+  // independent controls over two independent stores — resetLayout (from
+  // useCatalogLayout) only rebuilds the KPI grid geometry, it never touches
+  // filter state. Without also calling clearFilters here, "Reset" left the
+  // complaint-type tree filter (and every other filter) exactly as the user
+  // had set it, while "Clear" correctly restored it to "All types" — the
+  // exact inconsistency reported in egovernments/CCRS#1471.
+  const handleResetLayout = useCallback(() => {
+    dashboardMetrics.markInteraction("filter");
+    resetLayout();
+    clearFilters();
+  }, [resetLayout, clearFilters]);
 
   return (
     <DashboardLayout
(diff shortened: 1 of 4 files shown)


KDwevedi:
Since Reset now clears filters as well as layout, the tooltip at [`DashboardHeader.jsx:241`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/aadbc519a483ed57776bfcd63c6571607329e4a9/digit-ui-esbuild/products/dashboard/src/components/DashboardHeader.jsx#L241) should just say "Reset" (matching the visible button label) rather than "Reset layout".

Lokendra-egov:
@KDwevedi Fixed in e5b057b9 — tooltip now reuses the same `DASHBOARD_HEADER_RESET` key as the visible button label, so it can't drift from the button text again. Also dropped the now-unused `DASHBOARD_HEADER_RESET_LAYOUT` key from both locale seed files (en_IN, pt_PT).

KDwevedi (approved):

PR #1789, opened 2026-08-18
chore(local-setup): remove the Jupyter Lab service, keep the dataloader library

dhruv-1001:
Closes #1743.

> **Stacked on #1788** (the docs PR). Review that one first — this branch is based on it, so the diff here shows only the Jupyter removal. It rebases cleanly onto `master` once #1788 lands.

---

## The catch in the issue as written

The issue says "remove `local-setup/jupyter/`". That directory held two unrelated things:

1. the **Jupyter Lab container** — Dockerfile, notebooks, a compose service, a Kong route;
2. `dataloader/` — the **CRSLoader library**, which all five scripted dataloaders import via `../jupyter/dataloader`, which `tests/` imports, and which Ansible rsyncs for its CI phase.

The replacement lived inside the thing being replaced. Deleting the directory would have taken the scripted dataloaders down with it.

**So the service is gone and the library moved up one level:** `local-setup/jupyter/dataloader/` → **`local-setup/dataloader/`**. Git tracks it as a pure rename — no file contents changed.

---

## Removed

| | |
|---|---|
| `local-setup/jupyter/` | Dockerfile, `.dockerignore`, `requirements.txt`, `LocalSetup.ipynb`, `DataLoader_v2.ipynb` — all only reachable through Jupyter Lab |
| Compose | the `jupyter` service in all six files that declared it, plus the `profiles: [local-dev]` stanza in the core overlay |
| Kong | `jupyter-service` / `jupyter-route` (the `/jupyter` path) |
| Tilt | the `dc_resource` in both Tiltfiles and the "Start Jupyter" nav button |
| Misc | port 18888 in the devcontainer forwards; the `jupyter` cpu entry in all four performance profiles; the gatus-coverage waiver; the `egovio/tilt-demo-jupyter:latest` entry in the frozen `:latest` debt list |

## Repointed

Every consumer of the library: the five CI dataloaders, `setup-uitest-tenant.py`, two python test modules, the CI telemetry-user-agent grep, `performance/ansible/playbook-setup.yml`, and four places in `playbook-deploy.yml`.

One of those needed thought rather than sed: the Ansible rsync destination had to move as well. The copied scripts resolve the library as `SCRIPT_DIR/../dataloader`, so it has to stay a **sibling** of `ci-tests/scripts/` — `ci-tests/jupyter/dataloader/` → `ci-tests/dataloader/`.

## The coverage question the issue asks

> Confirm scripted dataloaders fully cover the jupyterbook's use cases

[...]

PR #1809, opened 2026-08-18
docs(local-setup): rewrite the Ansible setup path and the onboarding guide for first-time operators

dhruv-1001:
Rewrites the Ansible setup path and the tenant-onboarding guide for someone who
has never seen DIGIT before. Engineer-level detail is kept, but moved into
collapsible **under the hood** notes so the instructions read straight through.

Stacked on #1789, which is on #1788, so the diff against `master` shows those
commits too. This PR's own commits are the five after `fd5571e3`.

---

## `local-setup/README.md` — Option C rewritten

Six numbered steps, each ending in something you can check.

| Asked for | Done |
|---|---|
| Clone the latest release for a release deployment, with the command | Step 1, plus a `git ls-remote` one-liner for finding the current tag. The old text only described cloning the default branch |
| A script for the prerequisites | `local-setup/scripts/install-prereqs.sh`, linked from Step 2 |
| Explain the host_vars fields, which are necessary, with an example | Step 3 — a table of every field to change, and `quickstart.yml.example` to copy |
| Drop `--check --diff` | Gone from Option C, and from the three other places in the repo that still offered it |
| Drop the `rsync … mkdir "/opt/digit/docker/…"` workaround | Gone |
| Cover the credentials needed for the configurator | Step 6 is a credentials table: configurator, employee app, onboarded employees, citizen app and Grafana |
| End at *Next: onboard a tenant*, move the rest out | Everything after it is now `docs/STACK-REFERENCE.md`, linked from a short index |

Also new in Option C: a note on why there are five tenant variables and what
breaks when each is wrong; the `sudo` trap (`deploy.sh` never passes `-K`); the
preflight warning that is expected and harmless; and a table of the failures
that actually happen.

## `local-setup/docs/ONBOARDING-AND-ADDONS.md`

**New — the concepts, before the steps.** What a tenant is, root versus city
side by side, and the answer to the question that actually blocks people: which
of the two do I put this in. What master data is, and why the four phases are
in that order. What the configurator is — the admin console, its two modes, and
that it only calls public APIs.

**New — every column of every spreadsheet.** One table per sheet: column,
required or not, what it means in plain words, and a filled-in example. Phases
1, 2, 3.1, 3.2 and 4, read off the template generator and the parser.

**New — `scripts/make-onboarding-examples.py`**, which generates all five
workbooks filled in with a small but complete and internally consistent city:
four departments, seven boundaries, ten complaint types over a three-level
hierarchy, six employees. Every "Download Template" button gives you empty
columns; this gives you a worked example to edit.

**Corrected.** Phase 3 was documented as it worked two versions ago: complaint
types have not been part of the common-masters workbook for some time. Step 3.1
is departments and designations; complaint types are step 3.2, where you define
the hierarchy levels in the browser and the template is generated with one
column per level. Also fixed: the configurator login (the tenant is prefilled
from `state_tenant_id`, the account is `bootstrap_user`/`bootstrap_password`,
one of four roles is required); `dob` is required and enforces 18+; `userName`
is optional; `Designation` has a comma-separated `department` column; boundary
rows take lat/long; Phase 2 only writes `MapConfig` when boundaries carry
geometry; employees log in with the employee code, not `userName`; and the
phone rule is the tenant's, not a fixed Kenyan regex.

## `local-setup/scripts/install-prereqs.sh` (new)

git, curl, rsync, unzip, python3 + venv, Node 20, ansible + ansible-lint +
yamllint, then the galaxy collections. Debian/Ubuntu, RHEL/Fedora/Rocky/Alma/
CentOS, Arch and openSUSE; an unrecognised distro prints the package list and
exits 2 rather than guessing. Ansible goes into a private venv symlinked into
`~/.local/bin`, which sidesteps PEP 668 identically everywhere. Idempotent, and
only asks for sudo when a package is actually missing. `--check` reports without
changing anything.

## host_vars

- **`quickstart.yml.example` (new)** — a complete single-machine config with the
  settings grouped into "change these seven things" and "leave the rest alone".
- **`_example.yml`** — every finding from the variable audit. Wrong or stale:
  the header told you to edit `inventory/hosts.yml` (deploy.sh regenerates it);
  the REQUIRED claim (`map_center` has no default anywhere and aborts the
  deploy); `asset_s3_bucket` does not set the filestore bucket; `boot_tenant` is
  CI-only; `twilio_auth_token` reads bare. Invisible knobs now listed:
  `nginx_features.grafana` defaults to **true**, so the old block published
  `/grafana/` without the operator seeing the switch — added at its existing
  default with the rationale, plus `root_to_digit_ui` and a new "less-common
  knobs" section. Removed `build_novu_bridge` and `build_pgr_services`: neither
  ever had a task behind it.

## Verification

- `install-prereqs.sh` run end to end on Ubuntu 24.04 with an isolated `HOME` —
  creates the venv, installs ansible core 2.21.3 and both collections; second
  run is a no-op. Unknown-distro path exercised with a synthetic os-release.
- `quickstart.yml.example` and `_example.yml` both parse, pass `yamllint` and
  pass `scripts/preflight.py`. `_example.yml`'s parsed values were diffed
  against the original: only the two dead keys removed, and two
  `nginx_features` keys added at their existing effective defaults. No
  behaviour change.
- The five generated workbooks were parsed back and checked against a
  re-implementation of the configurator's own validation — tenant-code pattern,
  parents-before-children, level completeness, `serviceCode` uniqueness, the
  9–10 digit and Kenyan mobile rules, the dob range — plus every cross-sheet
  reference. Zero violations.
- Every internal anchor and relative link in the touched Markdown resolves, and
  nothing else in the repo linked to the sections that moved.

**Not verified:** I did not run `./deploy.sh` end to end for this PR. The Option
C prose is derived from `playbook-deploy.yml`, `deploy.sh`, `preflight.py` and
the nginx and globalConfigs templates, not from a live deployment.

**Out of scope:** Options A and B in `local-setup/README.md` are untouched.
`docs/notification-onboarding/install-{fresh,upgrade}.md` still list
`build_novu_bridge` / `build_pgr_services`; harmless, but worth a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added local and remote deployment guidance, including full and slim presets.
  * Added prerequisite installation support and onboarding spreadsheet generation.
  * Added a standalone Python DataLoader workflow with localization, validation, and workflow configuration support.
* **Improvements**
  * Retired the Jupyter Lab setup and replaced it with direct DataLoader usage.
  * Expanded onboarding, troubleshooting, resource, and operations guidance.
  * Clarified health dashboard coverage and persistent monitoring history.
* **Bug Fixes**
  * Corrected boundary hierarchy request handling for tenant-specific searches.


dhruv-1001:
@coderabbitai review

KDwevedi:
Reviewed the docs and the two new scripts, and spot-checked the load-bearing claims against the playbook — `map_center`, the six no-default `nginx_features` keys, `IS_BUCKET_FIXED`, `container_name: openbao`, the progress-file path and the `v2.12-beta` tag all check out, every relative link and heading anchor resolves, `--check --diff` is gone repo-wide, and the >3-level fix in `a5e33d7` produces correct paths and unique `serviceCode`s at 2/4/5 levels. Six things to fix before merge:

**1. `install-prereqs.sh` breaks when run as root.** [`install-prereqs.sh#L285`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/a5e33d7e9d43420a5e3738ef0b2b84e9d6f00116/local-setup/scripts/install-prereqs.sh#L285) and [`#L289`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/a5e33d7e9d43420a5e3738ef0b2b84e9d6f00116/local-setup/scripts/install-prereqs.sh#L289) pipe into `$SUDO -E bash -`, but `SUDO` is empty when `id -u` is 0 ([L90-L97](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/a5e33d7e9d43420a5e3738ef0b2b84e9d6f00116/local-setup/scripts/install-prereqs.sh#L90-L97)), so the command becomes `-E bash -` → `bash: -E: command not found`. Every other call site uses bare `$SUDO cmd` and is fine; this one hits the README's own "deploying to a remote server as root" path.

**2. `--check` never validates the ansible-core ceiling.** `install_ansible` returns early in check mode at [L309-L320](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/a5e33d7e9d43420a5e3738ef0b2b84e9d6f00116/local-setup/scripts/install-prereqs.sh#L309-L320), before `check_ansible_core`, and `verify()` is skipped entirely at [L464](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/a5e33d7e9d43420a5e3738ef0b2b84e9d6f00116/local-setup/scripts/install-prereqs.sh#L464). On a box that already has ansible-core 2.21, `./install-prereqs.sh --check` prints `[skip]` and exits 0, hiding the exact failure the `ANSIBLE_CORE_MAX` pin exists to prevent.

**3. `boot_tenant` is documented two contradictory ways, in two files this PR touches.** [`_example.yml#L76`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/a5e33d7e9d43420a5e3738ef0b2b84e9d6f00116/local-setup/ansible/inventory/host_vars/_example.yml#L76) now says it is inert and read only by the opt-in CI suite, while [`quickstart.yml.example#L81`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/a5e33d7e9d43420a5e3738ef0b2b84e9d6f00116/local-setup/ansible/inventory/host_vars/quickstart.yml.example#L81-L84) still says it is "the tenant the citizen app defaults to" — the old claim this PR is correcting elsewhere.

**4. The dead-knob removal is half done.** [`_example.yml#L420`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/a5e33d7e9d43420a5e3738ef0b2b84e9d6f00116/local-setup/ansible/inventory/host_vars/_example.yml#L420) drops `build_novu_bridge`/`build_pgr_services` on the grounds that no task ever read them (confirmed — no reference anywhere in the playbook), but both are still live lines in [`bomet.yml.example#L125-L126`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/a5e33d7e9d43420a5e3738ef0b2b84e9d6f00116/local-setup/ansible/inventory/host_vars/bomet.yml.example#L125-L126).

**5. The header-matching rule in the generator's docstring is not what the parser does.** [`make-onboarding-examples.py#L21-L24`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/a5e33d7e9d43420a5e3738ef0b2b84e9d6f00116/local-setup/scripts/make-onboarding-examples.py#L21-L24) claims headers are matched "case-insensitively, ignoring `*` and extra whitespace", but `excelParser.ts` matches per-field alias lists and `parseComplaintHierarchyExcel` matches level columns by the **exact** level code ([`excelParser.ts#L1017`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/a5e33d7e9d43420a5e3738ef0b2b84e9d6f00116/configurator/src/utils/excelParser.ts#L1017), only `_`→space). The generated files work because they emit the exact strings, but the stated rule will mislead anyone editing a header by hand.

**6. Two minor ones.** [`make-onboarding-examples.py#L73-L77`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/a5e33d7e9d43420a5e3738ef0b2b84e9d6f00116/local-setup/scripts/make-onboarding-examples.py#L73-L77) writes a row of four empty strings into "Tenant Branding Details", producing a blank data row rather than a header-only sheet; and `${#have[@]}`/`${#NOTES[@]}` on an empty array under `set -u` ([L194](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/a5e33d7e9d43420a5e3738ef0b2b84e9d6f00116/local-setup/scripts/install-prereqs.sh#L194), [L466](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/a5e33d7e9d43420a5e3738ef0b2b84e9d6f00116/local-setup/scripts/install-prereqs.sh#L466)) is an unbound-variable error on bash 4.2/4.3, which only matters if the advertised CentOS support is meant to include 7.

Unrelated to this PR: `gatus-coverage` and `test` both fail on `master` too (duplicate k8s Service named `grafana` in `devops/`, and `mdms-backend is unhealthy`).

subhashini-egov (approved):

PR #1810, opened 2026-08-19
feat(pgr): enforce access decisions instead of making them (#1050)

KDwevedi:
Part of #1050. **Depends on** Digit-Core [#1403](https://github.com/egovernments/Digit-Core/pull/1403), which adds the PDP endpoints this PR calls, and on the seed PR that registers the analytics actions.

## What changes

pgr-services stops being its own policy decision point. Complaint search and the analytics surface now ask `egov-accesscontrol` what is allowed and enforce the typed answer:

- bind the returned scope to parameterized SQL,
- short-circuit a `DENY` before touching the database,
- drop rows the PDP denied,
- apply the field obligations it returned.

Deleted rather than adapted: `org.egov.pgr.policy` (registry, evaluator, scope engine, field-visibility service), `PrincipalScopeResolver`, `MDMSUtils.fetchAccessControlActions`, `AnalyticsController.DASHBOARD_CONFIG_ROLES`, `AnalyticsService.OFFICER_PII_ROLES`, `KpiDefinition.rbac.visibleTo`, `DashboardPack.roles`, and the `json-logic` dependency itself.

Dashboard authorization used to be decided in five places that could disagree. It is now decided in one and consumed in all of them.

## Analytics

One `_resolve` call per request settles actions 2640-2648. A capability is an allowed action URL, never a role: `/_access` answers the dashboard's bootstrap, officer-PII projection rides on the officer capability, and tiles and packs are selected by `requiredActionUrl`. The public dashboard stays an isolated anonymous surface chosen by `public: true`, with no identity to resolve and no capability to hold.

## Row scope

Applied identically on both enforcement points:

- **SUBTREE tenant** matches the tenant itself or something under a `.` beneath it, so `pg.foo` can no longer read `pg.foobar`.
- **Jurisdiction** is segment equality on the `|`-joined `boundary_path`, so `WARD_1` cannot swallow `WARD_10`.
- Every scope value is a bind parameter.
- Rows correlate by `serviceRequestId`; a page that cannot be correlated fails closed.

401, 403 and 503 stay distinct — an expired session, a denial and an unreachable PDP send a user to three different remedies.

## Verification

`mvn -o test` on JDK 17: **317 tests, 0 failures**. New coverage includes search/count parity, exact vs subtree tenant boundaries, department+jurisdiction intersection, citizen self-scope, bulk row denial, post-enrichment PII redaction, and malformed response/obligation handling.

`NoLocalPolicyEvaluationTest` asserts the architecture against the source, because adding one role check back would compile and pass everything else.

Draft: do not merge before the Digit-Core PDP and the seed PR.


KDwevedi:

PR #1811, opened 2026-08-19
feat(mdms): register the analytics actions the dashboard will be authorized by (#1050)

KDwevedi:
Part of #1050. **First in the deploy order** — the canonical policy must exist before any enforcement point starts asking for it.

Adds actions 2640-2648 for `/pgr-services/v2/analytics/*` and their role mappings, so dashboard authorization has somewhere to live other than hard-coded role lists in the frontend and in PGR.

## scopeRef, not a copy

2641 (`_query`) and 2642 (`packs`) borrow their row scope from action 2008 rather than restating it:

```json
"resource": { "complaint": { "scopeRef": { "actionId": 2008 } } }
```

That is what keeps the dashboard and PGR search answering the same question. A copy of the policy drifts on the first edit; granting 2008 outright would hand every dashboard viewer the search API. Note the reference lives **inside** the resource-type block — the resolver does not read a top-level `"scopeRef": 2008`.

## Scope of the seed change

Action 2008 itself gains the authored complaint policy — the same scope and attribute rules in every source — and its bare Nairobi rows are replaced rather than left to shadow it.

Seeds land in all three authoritative places (default-data, the Nairobi tenant masters, the local full dump) for the **same seven tenants that already carried 2008**. No tenant set is widened. Stale `DGRO` is deliberately not mapped.

Verified across all nine sources (default-data, seven Nairobi tenants, full-dump): `url`, `method` and `resource` are identical for 2008 and 2640-2648, and every role mapping matches the intended set exactly.

Every action is disabled and POST, so **this changes nothing until an enforcement point starts asking** — it is safe to land ahead of its consumers.

Draft; do not merge before the Digit-Core PDP ([#1403](https://github.com/egovernments/Digit-Core/pull/1403)) is agreed, since it fixes the contract these records are written against.


KDwevedi:

PR #1812, opened 2026-08-19
feat(dashboard): render from the server's decision instead of a role list (#1050)

KDwevedi:
Part of #1050. **Last in the deploy order.**

This branch stacks on #1811 (the analytics action seeds) and so contains its commit as well — GitHub cannot base a PR on a branch that lives only on the fork. Merge #1811 first and this reduces to the single frontend commit. It also depends on #1810, which implements the `/_access` endpoint this consumes, and on Digit-Core [#1403](https://github.com/egovernments/Digit-Core/pull/1403).

## What changes

The dashboard stops deciding who may see it. The route guard and the home card resolve one capability bootstrap — `POST /pgr-services/v2/analytics/_access` — and the catalog gates every tile and pack on `requiredActionUrl` instead of a role array.

Gone with them: `DASHBOARD_ROLES`, `dss.DashboardConfig.allowedRoles`, `KpiDefinition.rbac.visibleTo`, `DashboardPack.roles`, and the localStorage role sniffing behind the persona metric tag.

## The gate fails closed

A network error, a non-2xx, or a body it cannot read all resolve to "not allowed". There is no client-side fallback list left to widen access when the server cannot be reached — which is exactly what the old MDMS-backed gate did whenever MDMS was down.

## Public is an audience, not a role

`public: true` on a tile or pack opts it into the anonymous dashboard, additively, and never grants an employee anything.

## UI is unchanged

Layout, styles and interaction are untouched — this is authorization plumbing, not a redesign. No CSS file appears in this diff.

## Verification

`node --test` across the dashboard: **256 tests, 0 failures**. `npm run build` succeeds.

The capability mapping reproduces the previous personas exactly — the PGR_ADMIN add-all plan still resolves to the same 28 definitions it did under `rbac.visibleTo`.

`dashboardCapabilityGate.test.js` pins the parts that would otherwise rot quietly: every published tile carries a gate the seeds actually define, packs are chosen by capability or by public and never both, the analytics actions are identical across every authoritative seed, 2641/2642 borrow action 2008's policy by nested reference, and no dashboard source carries a role code at all.

Not yet done: the visual smoke test of the loading / denied / allowed states against a live stack.


KDwevedi:

Issue #1361, opened 2026-07-21
[Blocked on ABAC] Employee dashboard shows only data in the employee's configurable geography


KDwevedi:
**Type:** User story. **Status: Blocked** on ABAC from @vinothrallapalli-eGov (#1083 / #1086).

## Acceptance criterion

**Every employee user sees only the data relevant to their configurable geography on the dashboard.**

Dashboard KPIs and charts aggregate only over complaints within the employee's HRMS-assigned jurisdiction (boundary subtree) — enforced at row level, server-side.

## Why this is not already covered

Unlike the inbox, the dashboard's row-scope seam is **fully built but switched off**.

`/v2/analytics` (CCRS `pgr-services` fork) resolves four scope axes in `PrincipalScopeResolver` → `AnalyticsPlanner.applyScope`:

| Axis | State |
|---|---|
| tenant | ON |
| citizen-self | ON |
| department | ON |
| **boundary** (`boundary_path LIKE prefix%`) | **commented out** |

In `PrincipalScopeResolver.java` the jurisdiction branch is explicitly disabled — `String boundaryPrefix = null;` with the HRMS-jurisdiction read commented out just below it, under a comment reading *"boundary / jurisdiction scope: DELIBERATELY SKIPPED for now"*. `applyScope` already injects `boundary_path LIKE prefix%` when a prefix is supplied, and the daily grain already does `LEFT JOIN bnd ON bnd.code = svc.locality_code`. **Enabling it is a resolver-only change of roughly five lines.**

Note also: card RBAC (`visibleTo` / roles, #963) is **orthogonal** — it hides cards, it does not restrict data. Hiding a card is not scoping.

So the code is not the blocker. The blockers are policy and data, below.

## Blocked on

**#1086** — ABAC/JBAC pgr-services Tier-2 PDP (@vinothrallapalli-eGov); its `PrincipalScopeResolver`/`AnalyticsScope` work adds exactly this jurisdiction resolution *("previously scaffolded/disabled")* alongside the existing department axis. Parent **#1083**. Uncommenting ahead of that would collide with his change — hence blocked rather than done now.

Dashboard scoping parts B/C under **#908** are checked off, but that covers department + the generalized attribute mechanism; the boundary axis is the part still dark.

## Prerequisites that are ours, not Vinoth's

[...]

Issue #1110, opened 2026-07-09
Dashboard: instrumentation to capture render lag in production


subhashini-egov:
## Summary

Add production instrumentation to the Configurable Dashboard so we can **continuously observe render lag** in the wild — not just during the one-off benchmark exercise in #1109. Without instrumentation we have no visibility into how dashboard performance degrades as a tenant's data grows, no early-warning signal when a code change regresses render time, and no data to prioritise the follow-up optimisation issues coming out of #1109.

## What to instrument

**Client-side (browser)** — emit a metric for every dashboard load and every filter re-render:

| Metric | Definition |
|---|---|
| `dashboard.ttfb.ms` | Navigation → first byte of dashboard HTML |
| `dashboard.first_widget_visible.ms` | Navigation → first widget paints something (not a spinner) |
| `dashboard.all_widgets_ready.ms` | Navigation → last widget resolves + paints (final "loaded" state) |
| `dashboard.filter_apply.ms` | Filter interaction → all affected widgets re-render |
| `dashboard.persona_switch.ms` | Persona/tab change → new layout fully rendered |
| `dashboard.slow_api_calls.count` | Count of API calls > 2s during the load |
| `dashboard.total_payload.bytes` | Sum of all response bytes during the load |
| `dashboard.error_widgets.count` | Widgets that failed to render or hit an API error |

Each metric tagged with: `tenant`, `persona`, `layout_id`, `record_count_tier` (bucketed: `<10k`, `10k-50k`, `50k-100k`, `>100k`), and `user_agent_family`.

**Server-side** — trace the slowest 3 backend queries feeding dashboard widgets:

- Emit query duration + result-row count for the top-N slowest MDMS/PGR/DSS queries per dashboard load
- Correlate to the client-side load via a trace ID passed from browser (`x-trace-id` header on the first API call, propagated)

## Delivery vehicle

Use the deployment's existing observability stack:

- Metrics → Prometheus histograms (bucketed for p50/p95/p99)
- Traces → whatever tracing pipeline the platform already runs (Jaeger / OTLP-collector — align with SREs)
- Client-side capture via the `web-vitals` library + a small `dashboard-metrics.ts` module that hooks the widget lifecycle events already in digit-ui-v2's dashboard code

## Dashboards / alerts (out of scope but noted)

- A Grafana board showing p95 `all_widgets_ready.ms` per tenant per persona over time
[...]

PR #1814, opened 2026-08-19
feat(dashboard): consume #1441's ABAC instead of the bespoke dashboard RBAC (#1050)

KDwevedi:
Closes the split-brain #1050 was opened for, by **calling** Vinoth's merged #1441 rather than extracting, relocating or re-implementing it. **No changes to egov-accesscontrol**, and no `scopeRef` — #1441 was deliberately built so the policy rides inside the opaque `resource` field on action 2008, and nothing here needs more.

Three commits, in deploy order.

---

## 1. The dashboard is scoped by the policy search already uses

Analytics resolved its own row scope, in its own resolver, from its own HRMS lookup and its own hard-coded list of tenant-wide roles. #1441 resolves the same question from action 2008's authored policy. Two resolvers reading two rule sets drift, and these had: a ward-scoped GRO was correctly restricted on `/v2/request/_search` and saw **every ward** on the dashboard, because `PrincipalScopeResolver` set `boundaryPrefix = null` on purpose.

So analytics stops resolving and starts asking. `AnalyticsRowScopeResolver` calls `SearchAccessPolicyService.resolveScope` and gets back the same `PgrSearchScope` the search path runs under. `PrincipalScopeResolver` and `AnalyticsScope` — PGR's second scope resolver and second scope type — are deleted. `isPureCitizen`, a different question shared with `EnrichmentService`, moves to `Principals` unchanged.

This is the migration #1441 planned for, in its own words:

> once this engine has proven itself on PGR search, Dashboard/Analytics can migrate to a ScopePolicy-driven resolve() here too and PrincipalScopeResolver's ScopeAxis machinery can retire

**Two deliberate behaviour changes**, both the content of the ticket:

- The dashboard applies the **jurisdiction axis it used to discard** (#1361).
- Roles are scoped by the authored policy rather than a hard-coded `TENANT_WIDE_ROLES` list, so a role absent from 2008's `roleScopes` takes its `default` block — on the dashboard exactly as on search. Widening one is now an MDMS edit, not a deploy.

`#1280`'s per-tenant department-scoping switch is preserved for deployments whose complaint data carries no departments. It is a second policy source and the only sanctioned divergence; folding it into 2008 is left for later.

**`DashboardSearchScopeParityTest` is the regression that matters**: one principal, one policy, both surfaces agree — asserted role by role, plus fail-closed and citizen self-scope, with the #1280 switch pinned as the only thing that can make them differ.

## 2. The dashboard is gated on granted actions, not role lists

Who may see the dashboard was decided in five places free to disagree: `DASHBOARD_ROLES` in the browser, `dss.DashboardConfig.allowedRoles`, `DashboardPack.roles`, `KpiDefinition.rbac.visibleTo`, and `DASHBOARD_CONFIG_ROLES`/`OFFICER_PII_ROLES` in this service. They become one grant in `ACCESSCONTROL-ROLEACTIONS`, where every other grant already lives.

Resolving actions 2640-2648 needs no new policy code: `/access/v1/actions/mdms/_get` is role-scoped, so *"is this action visible to the caller"* already **is** the capability question. `AccessPolicyRegistry.isActionVisible` is a six-line addition reusing the registry's fetch and its 15-minute per-(tenant, url, role set) cache.

These actions carry **no policy of their own** — no `resource`, no `condition`. Row scope stays on 2008 and is evaluated separately, so an endpoint grant and a row filter stay different questions.

The catalog moves with it: `requiredActionUrl` replaces the role arrays, `public: true` is an additive anonymous marker that never grants an employee anything, and a tile declaring neither is invisible rather than visible to everyone. The mapping **reproduces the existing personas** rather than redrawing them.

Seeds are additive in all three authoritative sources — **0 rows removed** — for exactly the seven tenants that already carried action 2008. `DGRO` is deliberately not mapped.

## 3. The frontend renders from the server's decision

One `_access` bootstrap; the browser holds no role codes. The gate **fails closed** — where the old MDMS-backed gate fell back to a hard-coded list, widening access precisely when MDMS was down. No CSS file appears in the diff.

---

## Verification

- pgr-services: **430 tests, 0 failures** on JDK 17 (the pinned `java.version`; the Mockito/Lombok breakage reported earlier was a Java 26 artifact).
- dashboard: **256 tests, 0 failures**; `npm run build` succeeds.
- `AnalyticsSeedParityTest` proves the capability actions are identical across default-data, the Nairobi tenant masters and the full dump, that they carry no policy, that action 2008 is untouched and unshadowed, and that **no seed mentions `scopeRef`**.

## Not done

- The 8c/E2E matrix, and the visual smoke test of the loading / denied / allowed states — both need a live stack.
- `#1280` folded into action 2008.
- A `develop`-facing port. This targets `master` because that is where #1441 lives; `develop` has no `org.egov.pgr.policy` at all.

Depends on nothing. Draft: merging needs approval, and the behaviour deltas above deserve a decision, not just a review.


KDwevedi:
## Review pass — two fail-open holes fixed, and one claim in the description above is wrong

An independent review of this branch found two authorization holes I introduced, plus a seed gap that inverts one of the behaviour claims I made. Fixed in `fab6c3aa0`; the description above needs the correction below.

### Fixed: the #1280 switch overturned denials

`AnalyticsRowScopeResolver` dropped the department axis whenever the tenant had `departmentScoping=disabled`. The engine expresses deny-all as a department list holding one sentinel (`PgrSearchScope.deniedAll` → `[__scope_denied__]`, with `jurisdictionCodes` null) — so on those tenants **every deny became a fully unrestricted tenant-wide scope**: a caller querying a tenant outside their own subtree, an unresolvable principal, and strict mode with no authored policy all stopped being denied and started reading every row.

The switch turns an axis off; it must not overturn a decision. A denied scope is now returned untouched.

### Fixed: the anonymous dashboard was dead

`doQuery` resolved the row scope through the ABAC engine unconditionally, including for the public surface. With no identity there is no role-scoped action lookup to make, so `MDMSUtils.fetchAccessControlActions` throws for an empty role set and **every `/public/_query` call would have returned 503**. `publicSurfaceScope` was written for precisely this and never wired to anything. Now wired.

Both were invisible to the 430-test suite for the same reason: controller tests mock `AnalyticsService`, service tests mock the scope resolver, so nothing exercised the seam. The new tests fail without their fixes — verified by reverting each. **433 tests, 0 failures.**

### Correction: on the live Nairobi seed this WIDENS the dashboard, it does not tighten it

The description above says roles become policy-scoped, and lists two tightenings. That is true of `default-data` and `full-dump`, whose action 2008 carries the authored `resource.complaint.scope`. It is **not** true of `ansible/nairobi-mdms`, where action 2008 has **no scope block on any of the seven tenants** — verified. Those deployments fall back to `SearchAccessPolicyService.DEFAULT_SCOPE_POLICY` (`department: ALL, jurisdiction: OWN`).

So on bomet/nairobi this PR **removes department scoping from the dashboard**, where the retired resolver derived it from HRMS and failed closed. Search already behaves that way on those tenants, so the parity claim holds — but the dashboard moves in the permissive direction, and I presented the opposite.

**This needs a decision before merge, and it is an MDMS edit, not a code change:** either author 2008's scope block in the Nairobi masters to match the other two sources (which also tightens complaint search on those tenants — #1441's domain, real user-visible impact), or accept the widening deliberately. I have not done either, because both are policy calls.

### Also worth knowing, not fixed here

- **403 vs 503 does not hold for real outages.** `ServiceRequestRepository.fetchResult` only rethrows 4xx; connection-refused, timeouts and 5xx are swallowed and return null, so an outage surfaces as "not granted" (403), not 503. Capability checks still fail closed, but my claim about the distinction is only true for 4xx. Pre-existing #1441 behaviour — and on the same path `getCondition` takes the backward-compatible allow branch, so Tier-2 fails *open* during an outage in non-strict mode. Not mine to flip inside this PR; worth its own issue alongside the `abac.strict-mode` rollout.
- **Persona narrowings I did not call out.** `MDMS_ADMIN` and `HRMS_ADMIN` were in the old `OFFICER_PII_ROLES` but hold no 2646 grant, so they lose officer-PII projection. `DGRO` had dashboard access via the old role lists and gets no capability grant at all, so it loses the dashboard entirely. Tiles with `visibleTo: []` (16 of them) meant "any authenticated role" and are now limited to the eight granted roles. All fail-closed, all deliberate in direction, but they were not stated.
- **Jurisdiction semantics differ between the surfaces.** Search exact-matches the complaint's locality; the dashboard matches any segment of the ancestor chain. For a caller whose HRMS jurisdiction is a non-leaf boundary the dashboard will count more than search returns. The dashboard side is arguably the more correct one, but "both surfaces agree" is not literally true for non-leaf jurisdictions, and the parity suite compares resolved scopes rather than query semantics, so it cannot catch this.
- **`_access` costs up to 9 accesscontrol round trips per request** and the registry does not cache misses, so ungranted actions refetch every time. It fires from the employee home card for every employee. One fetch of the role-scoped action list would answer all nine.

SQL correctness in `applyScope` was reviewed in detail and came back clean: valid PostgreSQL, all values bound, empty lists deny rather than drop the predicate, NULL columns excluded, parameter ordering correct.

KDwevedi:
@vinothrallapalli-eGov — one decision needed from you, and it is the only thing I think blocks this.

This PR puts the dashboard onto your #1441 engine rather than building a second one. Analytics no longer resolves its own scope; it calls `SearchAccessPolicyService.resolveScope` and applies the `PgrSearchScope` that `/v2/request/_search` already runs under. `PrincipalScopeResolver` and `AnalyticsScope` are deleted. That is the migration your own javadoc described — *"once this engine has proven itself on PGR search, Dashboard/Analytics can migrate to a ScopePolicy-driven resolve() here too and PrincipalScopeResolver's ScopeAxis machinery can retire"* — so I have tried to change none of your semantics, only the set of callers.

**The decision.** Action 2008 carries the authored `resource.complaint.scope` in `default-data-handler` and in `local-setup/db/full-dump.sql`, but **not in `ansible/nairobi-mdms`** — no `resource` block on any of the seven tenants there. So on bomet and nairobi, `SearchAccessPolicyService.DEFAULT_SCOPE_POLICY` applies: `department: ALL, jurisdiction: OWN`.

The consequence is that this PR *removes* department scoping from the dashboard on those deployments, where the retired resolver derived it from HRMS and failed closed. The two surfaces still agree — search already behaves that way there — but the dashboard moves in the permissive direction, and I would rather you decide which of these you want than pick one myself:

1. **Author 2008's scope block in the Nairobi masters** to match the other two sources. This tightens the dashboard, but it also changes `/v2/request/_search` for real users on those tenants — your call, not mine, which is why I have not touched it.
2. **Accept the widening**, on the grounds that the default policy is the intended one for those tenants and the previous dashboard behaviour was the outlier.

If it is (1), I am happy to write the seed change in this PR or a separate one, whichever you prefer.

**Two smaller things you may want to weigh in on, neither blocking.**

Jurisdiction is enforced differently on the two surfaces because the storage differs: search exact-matches the complaint's locality, while the analytics grains match any segment of the `|`-joined `boundary_path`. For a caller whose HRMS jurisdiction is a non-leaf boundary — a county or sub-county — the dashboard will count the whole subtree while search returns only exact-locality matches. I think the dashboard side is the more useful behaviour, but it means "one policy, both surfaces" is not literally true for non-leaf jurisdictions, and the parity suite compares resolved scopes rather than query semantics, so it cannot catch it.

Separately, and entirely pre-existing: `ServiceRequestRepository.fetchResult` only rethrows 4xx, so a connection refusal, timeout or 5xx from accesscontrol comes back as `null` and `getCondition` then takes the backward-compatible allow branch. Tier-2 fails **open** during an outage while `abac.strict-mode` is off. Capability checks in this PR fail closed on the same path, so nothing here regresses, but it seems worth a ticket alongside the strict-mode rollout.

Full context, including the two fail-open bugs an independent review found in my own branch and how they were fixed, is in the comment above.

KDwevedi:
## Rebased on master, and the two halves of this PR can ship independently

Merged current `master`, which brings in @vinothrallapalli-eGov's #1827 and #1826. **454 backend tests, 256 frontend tests, production build clean.**

### #1827 changes the fallback this PR sits on — followed, not worked around

An action that is visible but genuinely bare — no `scope` block, no legacy `condition` — used to fall back to `DEFAULT_SCOPE_POLICY` (jurisdiction at `OWN`), silently gating tenants that had never configured anything on HRMS data nobody had set up. It now resolves to zero axes: nobody restricted.

The parity fixture moves to the new `resolveScopeState` API, and there is a new test pinning that an unconfigured action leaves the **dashboard exactly as unrestricted as it leaves search**. That path is how a deployment runs with row-scoping effectively off, and a dashboard that only half-honoured it would be worse than one that ignored it.

This also enlarges the Nairobi question I raised above rather than closing it: those tenants' action 2008 is bare, so post-#1827 they get *no* axes rather than jurisdiction-`OWN`. The fallback is now deliberate and live-verified, so the remaining question is only whether Nairobi ought to have a policy authored at all.

### The capability half does not depend on Digit-Core; the scope half does

Worth separating, because it changes what can ship when.

Capability resolution reads `$.actions[*].url`, and `url` is on the stock `Action` model — so **actions 2640-2648 resolve correctly against stock egov-accesscontrol today**. Who may open the dashboard, which tiles they see, which pack they get, and the officer-PII gate all work with no Digit-Core change at all.

Row scope reads `resource.complaint.scope`, and `resource` is **not** on stock `Action` — the field is dropped from the `/v1/actions/mdms/_get` response, so the authored policy is invisible and (post-#1827) every tenant resolves to zero axes. That half needs the Digit-Core projection.

The intermediate state is coherent and safe rather than half-broken: dashboard authorization is enforced, row scoping is as unrestricted as it is on `/v2/request/_search` today, and the two surfaces still agree — which is the invariant this PR exists to establish.

I closed my own Digit-Core [#1403](https://github.com/egovernments/Digit-Core/pull/1403) in favour of `abac-changes`, which does the same job properly: fields on `Action` itself, a Flyway migration for the columns, row mappers and query builders carried through, and the null-`getTs()` guard that pgr-services currently works around. That branch has no PR open yet, and until it ships **CCRS #1441's authored policy is not being enforced anywhere** — which seems worth knowing independently of this PR.

### Also in this push

The `persona` metric tag (#1110) was the first matching role code and briefly became a full action URL. It is now the capability's short name (`_query`, `capabilities/officer`) — bounded exactly as before, and a sane OTEL attribute.

KDwevedi:
## CI triage after undrafting

All three red checks are repository-baseline/infrastructure failures, not failures introduced by this branch:

- **Gatus coverage:** fails because two existing Kubernetes Services are both named `grafana` but front different workloads. The checker explicitly asks for one to be renamed; no dashboard/ABAC file participates.
- **Local Setup `test`:** all services started, all 13 direct health checks passed, seed-data verification passed, and this PR's Ansible/static contract jobs passed. The job then failed because the baseline `egov-user` localization module has 0 messages. Current `master` fails at the exact same **Localization module check** in [run 32803739331](https://github.com/egovernments/Citizen-Complaint-Resolution-System/actions/runs/32803739331).
- **Tilt `tilt-test`:** `tilt ci` itself succeeded with “All workloads are healthy”; the post-run telemetry assertion received an empty telemetry log. Current `master` fails at the exact same **Verify telemetry events captured** step in [run 32803739291](https://github.com/egovernments/Citizen-Complaint-Resolution-System/actions/runs/32803739291).

Branch-relevant checks are green: Digit MCP build, Ansible validation, config/infra contracts, Flyway/dump alignment, Tiltfile Kubernetes load, and WIP/review-readiness. Combined with the local PGR/dashboard/configurator/bootstrap suites recorded in the description, there is no branch-specific CI fix to make before review.

KDwevedi:
Superseded by #1871, recreated from current `master` with only the intended change set.

PR #1823, opened 2026-08-20
fix(novu-bridge): explicitly target the WhatsApp Twilio integration

Lokendra-egov:
## Summary
- Novu resolves the PRIMARY integration for a channel unless a trigger names an explicit `overrides.<channel>.integrationIdentifier`. Twilio WhatsApp delivery is modeled in Novu as an `sms`-channel step, so a second, WhatsApp-registered Twilio integration living alongside the primary (plain SMS) integration was never actually selected — every WHATSAPP dispatch silently resolved to the primary SMS sender and was rejected by Twilio for a from/to channel mismatch.
- Adds `novu.bridge.integration.id.whatsapp` (env `NOVU_BRIDGE_INTEGRATION_ID_WHATSAPP`, default blank/no-op) and has `NovuClient.identifyThenTrigger` set `overrides.sms.integrationIdentifier` for WHATSAPP dispatches when configured. Backward compatible: unset, this is a no-op.
- Wires the new env var through `application.properties`, `local-setup/docker-compose.egov-digit.yaml`, and the Ansible `digit.env.j2` template.

## Context
Found while investigating why WhatsApp notifications weren't sending on the `ke.india` / `bometfeedbackhub.digit.org` deployment. Root cause was a chain of gaps (WHATSAPP channel disabled by flag, missing `complaints-whatsapp` Novu workflow, and this integration-selection issue) — the first two were fixed directly as config/data on that box; this PR is the code fix for the last one.

Verified live on `bometfeedbackhub.digit.org`: built and deployed this exact change, created a dedicated `twilio-whatsapp` Novu integration, and drove a real complaint through APPLY → ASSIGN → RESOLVE — WhatsApp delivered (Twilio status `read`) at every step to two different real Indian numbers.

## Test plan
- [x] `mvn test` in `backend/novu-bridge` — 92/92 passing, including 2 new tests covering the override (present when configured, absent when not) and the existing WhatsApp-gate tests
- [x] Live end-to-end verification on `bometfeedbackhub.digit.org` (see above)

### DIFF
--- a/backend/novu-bridge/src/main/java/org/egov/novubridge/service/NovuClient.java
+++ b/backend/novu-bridge/src/main/java/org/egov/novubridge/service/NovuClient.java
@@ -83,6 +83,7 @@ public class NovuClient {
         // sender/credentials; we only add the template. SMS/EMAIL keep the free-form path.
         if (StringUtils.hasText(templateId)) {
             Map<String, Object> overrides = buildProviderTemplateOverrides(templateId, contentVariables);
+            applyWhatsappIntegrationOverride(overrides, channel);
             return trigger(workflowId, scopedSubscriberId, phone, payload, transactionId, overrides, null);
         }
 
@@ -90,6 +91,24 @@ public class NovuClient {
     }
 
     private static final ObjectMapper CONTENT_VAR_MAPPER = new ObjectMapper();
+
+    /**
+     * Novu selects the PRIMARY integration for a channel unless the trigger names an
+     * explicit {@code overrides.<channel>.integrationIdentifier}. WhatsApp-via-Twilio is
+     * an "sms"-channel step in Novu, so without this override a WhatsApp send would
+     * silently resolve to the primary (plain SMS, non-WhatsApp-registered) Twilio
+     * integration and be rejected by Twilio for a from/to channel mismatch. Only applies
+     * when {@code novu.bridge.integration.id.whatsapp} is configured; otherwise this is a
+     * no-op so deployments without a dedicated WhatsApp integration are unaffected.
+     */
+    private void applyWhatsappIntegrationOverride(Map<String, Object> overrides, String channel) {
+        if (!"WHATSAPP".equalsIgnoreCase(channel) || !StringUtils.hasText(config.getWhatsappIntegrationId())) {
+            return;
+        }
+        Map<String, Object> smsOverride = new HashMap<>();
+        smsOverride.put("integrationIdentifier", config.getWhatsappIntegrationId());
+        overrides.put("sms", smsOverride);
+    }
 
     /**
      * The exact {@code {providers:{twilio:{_passthrough:{body:{contentSid, contentVariables}}}}}}
--- a/backend/novu-bridge/src/main/java/org/egov/novubridge/service/NovuClient.java
+++ b/backend/novu-bridge/src/main/java/org/egov/novubridge/service/NovuClient.java
@@ -81,9 +81,14 @@ public class NovuClient {
         // routing key. Pass it + positional contentVariables as a Twilio provider override so Novu
         // sends the approved template rather than the free-form body. The integration supplies the
         // sender/credentials; we only add the template. SMS/EMAIL keep the free-form path.
-        if (StringUtils.hasText(templateId)) {
-            Map<String, Object> overrides = buildProviderTemplateOverrides(templateId, contentVariables);
-            applyWhatsappIntegrationOverride(overrides, channel);
+        Map<String, Object> overrides = StringUtils.hasText(templateId)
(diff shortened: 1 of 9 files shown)


backend/novu-bridge/src/main/java/org/egov/novubridge/service/NovuClient.java
```diff
@@ -83,6 +83,7 @@ public NovuResponse identifyThenTrigger(String subscriberId, Contact contact, St
         // sender/credentials; we only add the template. SMS/EMAIL keep the free-form path.
         if (StringUtils.hasText(templateId)) {
             Map<String, Object> overrides = buildProviderTemplateOverrides(templateId, contentVariables);
+            applyWhatsappIntegrationOverride(overrides, channel);
             return trigger(workflowId, scopedSubscriberId, phone, payload, transactionId, overrides, null);
```
vinothrallapalli-eGov, line 92: identifyThenTrigger's free-form fallback path a few lines below (`return trigger(workflowId, scopedSubscriberId, phone, email, payload, transactionId);`, the no-templateId branch) calls the `trigger` overload with no `overrides` parameter at all, so `applyWhatsappIntegrationOverride` (added on this line) can never run on that branch.

Currently masked because `DispatchPipelineService` gates WHATSAPP dispatch on `templateId` being present before calling `identifyThenTrigger`. But any future/alternate caller that dispatches a WHATSAPP message via `identifyThenTrigger` without a `templateId` (e.g. a free-form Twilio session message) would silently reintroduce the exact primary-integration misroute this PR fixes — with no override applied and no error surfaced.

vinothrallapalli-eGov (commented):
## Review: WhatsApp integration override

**On the "could Novu's dashboard config replace this flag?" question:** Checked Novu's docs/roadmap — the dashboard only lets you mark one integration as *Primary* per channel. There's no dashboard mechanism to route a *specific trigger* to a non-primary integration when two integrations share a channel (exactly this case: plain-SMS Twilio + WhatsApp-registered Twilio, both under the `sms` channel). The only supported mechanism today is the trigger-time `overrides.<channel>.integrationIdentifier` payload field — i.e. code, not dashboard config. Conditional/rule-based provider selection is on Novu's roadmap but unshipped. So the env-var-driven override added here is currently the sanctioned approach, not a workaround for something Novu already solves.

Two related gaps found in files this PR doesn't touch, so I can't anchor inline comments on them — flagging here instead:

1. **`backend/novu-bridge/src/main/java/org/egov/novubridge/web/controllers/ProviderController.java:300`** — the `/providers/test-send` WHATSAPP path builds Twilio overrides via `buildWhatsappOverrides()` but never applies the new `overrides.sms.integrationIdentifier` override that `identifyThenTrigger` now applies. An admin using the "test send" button to verify a newly configured WhatsApp integration would still resolve to the primary SMS integration and get a false-negative Twilio rejection for the same from/to channel mismatch this PR fixes.

2. See inline comment on `NovuClient.java`.

3. **`local-setup/ansible/inventory/host_vars/_example.yml`** — the new `novu_bridge_integration_id_whatsapp` variable (wired into `digit.env.j2` in this PR) isn't documented next to its sibling `novu_bridge_workflow_id_whatsapp`. An operator onboarding a new tenant via the standard `_example.yml`-copy flow has no discoverable hint this knob exists, and may ship a new WhatsApp integration without ever setting it — reintroducing this same bug.

backend/novu-bridge/src/main/java/org/egov/novubridge/service/NovuClient.java
Lokendra-egov: 1. Fallback-path gap (inline comment on NovuClient.java): the override is now applied regardless of whether templateId is set, so it doesn't silently rely on the pipeline's gating logic to stay safe.
2. /providers/test-send gap: WhatsApp test-sends now apply the same integration override as live dispatch— previously, a test-send would've validated the wrong (primary SMS) integration and given a false pass/fail.
3. Docs gap: added novu_bridge_integration_id_whatsapp to _example.yml so it's documented for anyone onboarding a new tenant.

Added 2 new tests for these; the full suite is 94/94 passing.

KDwevedi (approved):
Approved. The routing matches Novu's supported multi-integration behavior: Twilio WhatsApp is an SMS-channel integration, and overrides.sms.integrationIdentifier is the correct way to select the non-primary active integration. The Twilio Content SID passthrough remains intact, and the live/test-send paths now behave consistently.

Non-blocking nudge: once WHATSAPP is enabled, consider failing fast (or persisting an explicit configuration error) when NOVU_BRIDGE_INTEGRATION_ID_WHATSAPP is blank. The backward-compatible no-op default is sensible for existing deployments, but on an enabled WhatsApp deployment it silently falls back to the primary SMS integration and can recreate the same misroute.

PR #1825, opened 2026-08-20
docs(operations): bring the handbook in line with nine dashboards and shipped alerting

### LINKED ISSUE #1739: [Enhancement] Rework the O&M Runbook based on feedback

dhruv-1001:
## Why

The deployment grew from four Grafana dashboards to nine, and the operations handbook did not follow. Three documents still told operators that Postgres, Kafka and Kong had **no metrics at all**, while their exporters, scrape jobs and dashboards were live and reporting:

```
up  =>  otel-collector, node, postgres-exporter, redpanda, kong    (all 1)
```

That is wrong in the direction that costs an incident. "Everything is slow but nothing is red" is answerable from the PostgreSQL dashboard; "gateway, or the service behind it?" from Kong's proxy-vs-upstream latency split; "filed but never indexed" from broker-side consumer lag. None of the three was mentioned anywhere an operator would look.

Closes #1739. Also the doc half of #1597.

> **Relationship to #1788.** That PR reworked the runbook against the first round of
> feedback and merged; #1739 stayed open because the deployment kept moving. This is the
> second pass: it brings the handbook in line with the monitoring that shipped after #1788
> was written, and corrects three statements in it that had since become false.

## What changed

| | |
|---|---|
| **New `docs/operations/dashboards.md`** | All nine dashboards, split out of `reference.md` (which had passed 600 lines). The five DIGIT-authored boards get full panel tables; the four imported ones (Node Exporter, PostgreSQL, Redpanda, Kong) get a **triage subset** plus an explicit instruction to ignore the rest |
| **Coverage matrix corrected** | `reference.md`, `l2-diagnosis.md`, `alert-channels.md`. Postgres, the broker and the gateway have metrics. Redis, Elasticsearch, MinIO and nginx still do not — that is now the whole gap |
| **Gatus alerting** | Described as what it has been since #1609: shipped, every endpoint already opted in at 3-fail/2-pass with send-on-resolved, off only because `gatus_slack_webhook_url` is empty. It previously said "alerting is commented out, send us the endpoints you want" |
| **`observability_level` documented** (#1657) | On `metrics` there is no Loki — which silently voids Step 4 of the L1 checklist, both memorised LogQL queries on the cheat sheet, and much of L2's method |
| **Grafana login corrected** | ac18db6c fixed README and alerts-setup and missed `l1-first-response.md` and `reference.md`, so half the handbook still promised no password |
| **L2 playbook rewired** | A "which half of the request is at fault" step using Kong's split; server-side Postgres beside the client-side pool metrics; broker-side lag as the primary |
| **14 new alert rules** | Database, broker and gateway. `ConsumerGroupEmpty` is the one that earns its place — zero consumers on a queue stops a pipeline dead while every tile stays green |
| **`enabling-monitoring.md`** | The "not merged yet" banner listed four items; three merged (#1657, #1652, #1675). Only #1673 has not |

### Why the imported dashboards get a triage subset

PostgreSQL, Redpanda and Kong are upstream community dashboards — roughly 35, 50 and 25 panels, written for someone tuning that software. A complete panel-by-panel table would be technically accurate and operationally useless. Each section names the 3–5 panels worth opening during an incident and says plainly that ignoring the remainder is the correct use of the page.

## Screenshots

Six images, 412 KB total, following the convention already used by
`docs/notification-onboarding/images/`. Captured with Playwright against panel-isolating URLs
(`?viewPanel=N&kiosk`), so each frame is one panel rather than a cropped page.

**Every image was reviewed before committing** — no citizen data, no domain, no URL bar, and
the Logs dashboard is clipped to its control row so log text cannot enter the frame.

Taking the pictures found two errors that reading the docs had not.

### The OOM panel counts its own observers

`OOM events (current range)` read **116, in red, on a healthy deployment**. The documented
Explore query returns zero over the same window — because they are not the same query:

```
panel:      {service_name=~".+"}                                       |~ "(?i)OutOfMemoryError|GC overhead limit"
documented: {compose_project="digit", compose_service!~"grafana|loki"} |~ ...
```

The panel does not exclude `grafana` and `loki`, which log the text of any search back into
the stream. Grouped by service, the 116 were **160 lines from loki, 3 from grafana, none from
any DIGIT service.** No crash had occurred.

The handbook warns about this echo in three places, but the panel it sends L1 to does not
filter it — and it is self-reinforcing: running the documented OOM query in Explore makes
L1's number go up. It climbs fastest exactly when someone is investigating.

Documented rather than papered over — L1 Step 3a now says to read the service names first and
treat an all-grafana/loki reading as zero; `dashboards.md` carries the caveat; `known-issues.md`
gains a row. The screenshot is included as the illustration of the trap, captioned with what it
really shows.

> **Follow-up, deliberately not in this PR.** Fixing the panel is a one-line change to
> `local-setup/otel/grafana/provisioning/dashboards/jvm-services.json`. That file is a
> deployment artifact and this PR is docs-only, so it is left out — but the documentation is a
> workaround, not the fix, and the query should be corrected separately.

### The Logs dashboard controls are not called what the docs called them

The docs told operators to set **service**, **level** and **q**. Those are the variable names
in the JSON; the labels rendered on screen are **Service**, **Level** and **Search regex**. An
operator following Step 4 would scan the page for a control called "q" and never find it.
Corrected in both files, and the screenshot shows the real labels.

## Verification

Every PromQL and LogQL query in the handbook was **extracted from the markdown and executed against a running deployment** — 107 queries, 97 returning data and 10 correctly empty (no OOMs, nothing died, no pool timeouts). That last category is why the cookbook now explains that "No data" is usually the healthy answer, and that `NaN` on a percentile means an empty window rather than a broken query.

Findings that came out of verifying rather than reading:

- **`/gatus/` 404s.** Only `/gatus` (no trailing slash) 301s to `/status/` — nginx matches `location = /gatus`. The docs pointed at the form that fails.
- **`HostSwapping` can never fire** on these boxes: `node_memory_SwapTotal_bytes` is `0`, so the documented expression evaluates to `NaN`. Noted in the catalogue rather than left as a rule that silently does nothing.
- **11 containers carry a memory cap, not 10** — `node-exporter` is capped in the monitoring overlay. `postgres-exporter` is not capped at all.
- **Loki exposes 8 labels, not the 5 listed.** `level` is among the missing ones, and the log dashboard is built on it.
- **15 instrumented Java services, not 16.**

All internal `.md` links and all heading anchors were checked to resolve.

## Not covered / follow-ups

- **Redis, Elasticsearch, MinIO and nginx still have no metrics.** Documented as the remaining gap rather than papered over.
- **Grafana alert provisioning (#1673) has not merged**, so a deployed Grafana still has zero alert rules and one placeholder contact point. The handbook says so, and points at Gatus as the thing that is available today.
- Commands not verifiable end-to-end here (`docker inspect` format strings, the log-truncate reclaim, `docker restart`) are unchanged from the shipped docs; the `docker ps` templates were validated.

## Scope

Docs only — no behaviour change, no code touched.


subhashini-egov (approved):
Has been reviewed over a call. Approving for merge.

PR #1826, opened 2026-08-20
fix: MDMS masters edit access blocked by truncated ACCESSCONTROL seed + over-broad configurator gating

### LINKED ISSUE #1815: [Bug] Configurator: ADMIN is unable to create or edit employee
# Description

ADMIN is unable to create or edit employee. User has only view access

---

# Actual Behavior

https://github.com/user-attachments/assets/45a4ea0f-7cfd-4a98-9add-3e276863e65a

vinothrallapalli-eGov:
## Summary

Investigating "ADMIN can't create/edit masters in the configurator despite correct role mapping" surfaced three distinct, compounding bugs:

1. **`tenant_bootstrap`'s MDMS copy step silently truncated at 500 rows** (no pagination) — `ACCESSCONTROL-ACTIONS-TEST.actions-test` routinely has 500+ rows on a real source tenant, so a bootstrapped tenant could end up missing write-action records (confirmed live: `common-masters.Department`/`Designation` create/update actions, ids 2582/2583/2613/2614, were entirely absent) even though roleactions correctly mapped `MDMS_ADMIN` to everything else. Fixed with a paginated fetch helper, used in both the essential-schema copy and the later accesscontrol actions-bridge step.
2. **Bootstrap ADMIN never got `MDMS_ADMIN`/`LOC_ADMIN`** — held `SUPERUSER` but not the roles the masters/localization access-policy actually gates edit on. Added to all 4 `standardRoles` literals in `digit-mcp`.
3. **Configurator's masters access-policy check applied to every resource, not just mdms-v2-backed ones** — Employees (hrms), Boundaries, Complaints (pgr), Localization, Users, Workflow, etc. have no `schema` to check against, so they permanently resolved to view-only for *every* role, including `SUPERUSER`/`MDMS_ADMIN`, with a code comment already flagging this as a known-but-unfixed risk. Scoped the check to `type: 'mdms'` resources, plus a small explicit allowlist for the two non-'mdms'-typed resources (Access Roles, Access Actions) that still edit real mdms-v2-backed data; everything else is unrestricted, matching pre-gating behavior.

Backfilled `local-setup/db/full-dump.sql`'s `pg` tenant seed with the missing rows so a fresh local-setup deploy gets the same coverage, instead of shipping the same truncated-copy gap forward.

### Follow-up fixes from review

A human/AI-assisted review caught real gaps in the above (thank you!):
- `access-roles`/`access-actions` were left completely ungated by the `type === 'mdms'` narrowing — fixed via an explicit allowlist (`isAccessControlGated`) instead of widening the type check.
- The pagination fix could still drop/duplicate rows: `ACCESSCONTROL-ACTIONS-TEST.actions-test` has ~329/330 rows sharing one `createdtime`, an unstable sort tie under offset pagination. `fetchAllMdmsV2Raw` now dedupes by `uniqueIdentifier` across pages and cross-checks the total against `/mdms-v2/v2/_count`.
- A later page's failure discarded already-fetched earlier pages (returned `[]` instead of partial results); a `Promise.all` schema-copy failure was only logged, not recorded in `results.data.failed`, so bootstrap could report success after silently skipping a schema. Both fixed.
- The `full-dump.sql` backfill imported 13 duplicate mdms-v2 action rows for URLs `pg` already had a working action id for, and 40 roleactions for role codes that don't exist in this seed's `ACCESSCONTROL-ROLES.roles`. Both cleaned up — see the dedicated `fix(db-seed)` commit for the verification detail.

## Verified live

Against a redeployed `ap` local-setup instance, logged in as `ADMIN`:
- Systematically checked `canEdit` for all 39 `type: 'mdms'` resources in `resourceRegistry.ts` — 31 now correctly resolve `true` for `MDMS_ADMIN` (up from ~2). The remaining 8 (Complaint Hierarchy, Theme Config, Tenant Boundary, Map Config, Notification Routing/Template/Provider Template) have never had a write-action defined anywhere in the platform's reference config — a separate, pre-existing gap needing a role-gating decision, not a data-sync fix.
- Non-mdms resources (Employees, Complaints, Localization) are now editable and confirmed to never even call `masters.canView`/`canEdit`.
- Access Roles/Access Actions confirmed to still call `masters.canView`/`canEdit` for real (regression test added).

## Test plan

- [x] `configurator` vitest suite passes, including regression tests proving mdms resources (and the access-role/access-action allowlist) still gate correctly while non-mdms resources are unconditionally open.
- [x] `configurator/packages/data-provider` node:test suite: same pre-existing failures as baseline (unrelated integration tests needing a live API), all `accessPolicy`/`resourceRegistry` tests pass.
- [x] `digit-mcp`: `tsc --noEmit` clean, `npm test` (tool-registration validator) passes.
- [x] `full-dump.sql`: reparsed end-to-end (JSON validity, column counts) and re-checked against the reviewer's own invariants (duplicate action ids, duplicate `(rolecode, actionid)` pairs, orphan role codes, duplicate row uuids) — all clean.
- [x] Live redeploy of `ap` via `sudo ./deploy.sh ap` — clean, all containers healthy, PGR search regression-checked (still returns correctly scoped results).

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


KDwevedi:
## Review — SQL backfill verified clean; two substantive issues in the gating change and the seed

Reviewed by Claude Opus 5 with an independent second pass by Codex (gpt-5.6-sol, high reasoning effort). The seed claims below come from parsing the `eg_mdms_data` COPY block on both branches, not from reading the diff.

### What checks out

| | master | this PR |
|---|---|---|
| actions-test rows / distinct `data.id` | 246 / 246 | 330 / 330 |
| roleactions | 370 | 515 |
| duplicate roleaction `data.id` | **1101×28, 745×3, 489×2** | none (7 nulls, pre-existing on both) |
| roleactions referencing a missing actionid | 0 | 0 |
| duplicate (actionid, rolecode, tenant) | 0 | 0 |
| duplicate row uuid | 0 | 0 |

+84 actions and +145 roleactions exactly as described, and the duplicate-id cleanup in `abd77ef` is real and complete — master genuinely shipped 33 colliding roleaction ids. COPY-block termination and column counts are intact across all 56 blocks.

### 1. The Access Roles / Access Actions screens are now completely ungated

[`useMastersCapability.tsx:97`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/abd77ef7f99276ff0bc662a2ab17f886c1e22bd5/configurator/src/hooks/useMastersCapability.tsx#L97)

`access-roles` and `access-actions` are registered as `type: 'access-role'` / `'access-action'` while carrying `schema: 'ACCESSCONTROL-ROLES.roles'` and `schema: 'ACCESSCONTROL-ACTIONS-TEST.actions-test'` ([`resourceRegistry.ts:103`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/abd77ef7f99276ff0bc662a2ab17f886c1e22bd5/configurator/packages/data-provider/src/providers/resourceRegistry.ts#L103), [`resourceRegistry.ts:114`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/abd77ef7f99276ff0bc662a2ab17f886c1e22bd5/configurator/packages/data-provider/src/providers/resourceRegistry.ts#L114)). With the check narrowed to `type === 'mdms'`, both `canViewResource` and `canEditResource` return `true` unconditionally for them.

**Failure scenario:** any authenticated employee reaching the configurator sees enabled Create/Save controls on the two screens that edit the permission system itself — the roles and actions that every other gate is derived from.

Note that the comment this PR deletes from `accessPolicy.ts` named exactly these two resources as the hazard case; the PR resolves that comment by opening them rather than by wiring them. Gating on `config?.schema` instead of `config?.type` would keep them checked, but their write path isn't a `/mdms-v2/v2/_create|_update/<schema>` URL so `WRITE_URL_RE` never matches and `canEdit` would be permanently `false` — the original hazard, relocated. An explicit `gated: true` flag on the registry entry (or a small allowlist) looks like the honest fix. At minimum this belongs in the PR description: it is a security-relevant widening currently presented as "matching pre-gating behavior".

### 2. The backfill duplicates 12 action URLs that already existed in the seed

URLs with more than one action row went from **3 on master to 15 on this branch**:

```
/mdms-v2/v2/_create/common-masters.Department            ids [17, 2583]
/mdms-v2/v2/_create/common-masters.Designation           ids [18, 2582]
/mdms-v2/v2/_create/common-masters.StateInfo             ids [21, 2580]
/mdms-v2/v2/_create/RAINMAKER-PGR.ServiceDefs            ids [19, 2593]
/mdms-v2/v2/_create/tenant.tenants                       ids [20, 2527, 2570]
/mdms-v2/v2/_update/ACCESSCONTROL-ROLEACTIONS.roleactions ids [2537, 2630]
… 6 more
```

On master, ids 17/18 already existed — they were simply not mapped to `MDMS_ADMIN`. This PR imports duplicate action rows from the source tenant and maps the new ids, leaving the originals unmapped.

**Failure scenario:** `AccessPolicyRegistry.fetch()` resolves an action with `actions.get(0)` — first row wins. Once anyone authors a `resource`/`condition` on one of a duplicated pair, which policy applies becomes arbitrary. The configurator is safe today because `canEdit` unions every matching id (`actionIds.some(...)`), but `canView` resolves the masters conditions with `actionRecords.find(r => r.data.url === '/mdms-v2/v2/_search')` — also first-wins, and one duplicate away from the same non-determinism.

Mapping `MDMS_ADMIN` onto the existing ids 17/18 achieves the same outcome without permanently duplicating the reference seed and propagating the duplicates to every tenant bootstrapped from `pg`.

### 3. Forty roleactions reference role codes the seed doesn't define

The backfill adds 37 `CMS_SCREENING_OFFICER` and 3 `CONFIG_ADMIN` mappings, neither of which appears in `ACCESSCONTROL-ROLES.roles`. Master had one such orphan (`PGR_ADMIN`), this branch has 41. They're inert, but it's source-tenant-specific data entering the reference seed — worth filtering the backfill to the roles the seed actually defines.

### 4. The pagination fix can still drop rows at exactly the sizes it targets

[`mdms-tenant.ts:1576`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/abd77ef7f99276ff0bc662a2ab17f886c1e22bd5/digit-mcp/src/tools/mdms-tenant.ts#L1576)

mdms-v2 paginates with `ORDER BY data.createdtime DESC` followed by `OFFSET ?` / `LIMIT ?` (`MdmsDataQueryBuilderV2`), and **329 of the 330 seeded actions-test rows share `createdtime = 1766039437780`**. A 329-row tie under an unstable sort means page boundaries are not reproducible across separate requests.

**Failure scenario:** on the ">500 rows on a real source tenant" case this helper exists for, page 2 can re-return rows from page 1 and omit others — the copy silently misses records, which is the original bug. `all.push(...rows)` never dedupes, so the duplicate half is silent too. Deduping by `uniqueIdentifier` and cross-checking the total against `/mdms-v2/v2/_count` would make the helper actually exhaustive.

Two smaller points on the same helper: `offset` advances by the requested `PAGE_SIZE` rather than `rows.length`, so a deployment that clamps `limit` server-side would stop after one short page (checked against Digit-Core — `mdms.default.limit=10` applies only when `limit` is absent and there is no clamp, so this is theoretical for this stack); and a dataset of exactly 100,000 rows fills all 200 pages and throws despite having been fetched completely.

### 5. Partial-page failures collapse to "this master is empty"

[`mdms-tenant.ts:2617`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/abd77ef7f99276ff0bc662a2ab17f886c1e22bd5/digit-mcp/src/tools/mdms-tenant.ts#L2617)

`.catch(() => [])` now discards successfully-fetched earlier pages when a later page rejects, so a department/designation scan that got page 1 and failed on page 2 returns `[]` and the ADMIN HRMS employee is skipped. Relatedly, a `Promise.all` rejection in the essential-schema copy is only logged — it isn't added to `results.data.failed`, so a bootstrap can report success having skipped an entire schema.

### 6. Minor

- `configure` is registered `risk: 'read'` ([`mdms-tenant.ts:445`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/abd77ef7f99276ff0bc662a2ab17f886c1e22bd5/digit-mcp/src/tools/mdms-tenant.ts#L445)) yet writes roles to an existing user ([`mdms-tenant.ts:613`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/abd77ef7f99276ff0bc662a2ab17f886c1e22bd5/digit-mcp/src/tools/mdms-tenant.ts#L613)). This is pre-existing and `SUPERUSER` was already in the list, so adding `MDMS_ADMIN`/`LOC_ADMIN` is a marginal increase over a grant that already exceeds them — but the classification is worth fixing separately.
- `LOC_ADMIN` still holds exactly one roleaction in the seed (`/localization/messages/v1/_upsert`, action 1734); `_create` (1529) and `_update` (1530) remain unmapped. Fine if upsert is the only path the configurator uses — worth confirming, since the PR grants ADMIN that role specifically to edit localization.
- The PR description credits a fix that isn't in this diff: "`mdms-backend`'s Docker healthcheck can never succeed (distroless image, no shell/wget)". The branch touches 5 files — `accessPolicy.ts`, `useMastersCapability.tsx`, `useMastersCapability.test.tsx`, `mdms-tenant.ts`, `full-dump.sql` — with no compose or k8s file among them. Either a commit was dropped or the description is stale.

KDwevedi (approved):

PR #1827, opened 2026-08-20
fix(pgr-abac): PGR search fully backward compatible when no scope/condition is authored

vinothrallapalli-eGov:
## Summary

`SearchAccessPolicyService.resolveScope()` fell back to `DEFAULT_SCOPE_POLICY` (jurisdiction required at level `OWN`) any time an action had no `resource.complaint.scope` block — including when the action has **never had any policy authored on it at all** (no scope block AND no legacy `condition`). That silently required HRMS jurisdiction data from every tenant that hasn't configured PGR search's ABAC, even though Tier-2 (`AccessPolicyRegistry#getCondition`) already treats that exact "genuinely unconfigured" case as backward-compatible-allow.

## Fix

Added `AccessPolicyRegistry#isPolicyUnconfigured` to distinguish two previously-conflated cases:
- **Action visible but truly bare** (no `scope`, no `condition`) → now gets a new `UNRESTRICTED_SCOPE_POLICY` (zero axes), so `ScopePolicyEngine` imposes no restriction on department or jurisdiction for anyone — matches Tier-2's backward-compatible allow instead of quietly gating on data nobody configured.
- **Legacy tenant with a hand-authored `condition` but no `scope` block** → still gets `DEFAULT_SCOPE_POLICY`, unchanged, since some restriction was already intended, just not in the new shape.

`pgr.abac.strict-mode` continues to fail closed uniformly for both sub-cases. Citizen self-scoping and tenant/subtree authorization in `PolicyDrivenScopeResolver` are untouched — those are hardcoded, not axis-driven.

## Verified live

Redeployed `pgr-services` to a running `ap` instance, then live-tested the actual regression:
1. Backed up `ap`'s action 2008 (currently has a configured `resource.complaint.scope`).
2. Stripped it to a bare shape (no `resource`, no `condition`) via the MDMS v2 API.
3. Confirmed search went from **0 results** (previously blocked — the admin test user's own HRMS jurisdiction value happens to be the tenant code itself, which matches no real complaint's locality) to **all 9 complaints, fully unrestricted** — log line confirms `departments=null jurisdictions=null (policy-driven)`.
4. Restored the original config from the backup and confirmed search returned to its prior (correctly scoped) behavior.

## Test plan

- [x] New unit tests at both layers: `AccessPolicyRegistryTest` (`isPolicyUnconfigured` — true for no-action/bare-action, false for scope-present/condition-present, propagates on malformed scope) and `SearchAccessPolicyServiceTest` (`resolveScope` resolves to zero axes for both "no action visible" and "action bare" cases).
- [x] Full `pgr-services` suite: 415 tests, 0 failures.
- [x] Live end-to-end verification on a redeployed `ap` instance (see above), including a full restore back to the pre-test live state.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/backend/pgr-services/src/main/java/org/egov/pgr/policy/AccessPolicyRegistry.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/policy/AccessPolicyRegistry.java
@@ -283,6 +283,29 @@ public class AccessPolicyRegistry {
     }
 
     /**
+     * True when this action's Tier-1 SQL scope has genuinely nothing authored for it at all — no
+     * {@code resource.<resourceType>.scope} block AND no legacy {@code condition} either — the SAME
+     * "not configured" case {@link #getCondition} treats as backward-compatible-allow (or no action
+     * visible at all). {@code SearchAccessPolicyService} uses this to decide its Tier-1 fallback: a
+     * legacy tenant with a hand-authored {@code condition} but no {@code scope} block still gets the
+     * structural {@code DEFAULT_SCOPE_POLICY} (some restriction was already intended, just not in the
+     * new shape), but an action with NEITHER must get a fully unrestricted Tier-1 scope — otherwise
+     * Tier-1 silently imposes a jurisdiction requirement nobody ever configured, while Tier-2 (this
+     * class' {@link #getCondition}) is already allowing everything for the exact same "unconfigured"
+     * action. Deliberately does NOT catch {@link MalformedScopePolicyException} — a present-but-broken
+     * {@code scope} block is an authoring mistake, not "unconfigured", and must propagate exactly like
+     * {@link #getScopePolicy} already does for the same case.
+     */
+    public boolean isPolicyUnconfigured(String actionUrl, RequestInfo requestInfo, String tenantId, String resourceType) {
+        Map<String, Object> action = getAction(actionUrl, requestInfo, tenantId);
+        if (action == null)
+            return true;
+        if (extractScopePolicy(action, resourceType).isPresent())
+            return false;
+        return action.get("condition") == null;
+    }
+
+    /**
      * Extracts and validates the field-visibility rules for {@code resourceType} from the same
      * Action's {@code resource} JSON object: {@code resource[resourceType].attributes} is a JSON
      * object keyed by field path, each value an independent {@code {condition, onDeny}} rule — any
--- a/backend/pgr-services/src/main/java/org/egov/pgr/policy/AccessPolicyRegistry.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/policy/AccessPolicyRegistry.java
@@ -295,14 +295,46 @@ public class AccessPolicyRegistry {
      * action. Deliberately does NOT catch {@link MalformedScopePolicyException} — a present-but-broken
      * {@code scope} block is an authoring mistake, not "unconfigured", and must propagate exactly like
      * {@link #getScopePolicy} already does for the same case.
+     *
+     * <p><b>{@code resource.<resourceType>.attributes} (field masking) alone does NOT count as
+     * "configured" here</b> — this method only looks at {@code scope} and {@code condition}. An
+     * action that declares field-masking rules (e.g. REDACTing a PII field) but no {@code scope}/
+     * {@code condition} still resolves {@code true} (fully unrestricted row-level access), by design:
+     * field masking and row-level scoping are independent axes of this policy, and an operator
+     * authoring one is not implicitly opting into the other. A caller who wants "any policy at all"
+     * to imply row restriction must check {@link #getFieldVisibilityRules} separately, not read that
+     * intent into this method.
      */
     public boolean isPolicyUnconfigured(String actionUrl, RequestInfo requestInfo, String tenantId, String resourceType) {
+        return resolveScopeState(actionUrl, requestInfo, tenantId, resourceType).unconfigured();
+    }
+
+    /**
+     * Combined form of {@link #getScopePolicy} + {@link #isPolicyUnconfigured} that fetches the
+     * action exactly once instead of twice. {@code SearchAccessPolicyService#resolveScope} used to
+     * call both separately — two independent {@link #getAction} calls that, on a cache miss (a
+     * "not found"/not-visible action is never cached — see this class' Javadoc), each triggered a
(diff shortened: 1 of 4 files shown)


KDwevedi:
## Review — verified against live bomet data and the Digit-Core sources

Reviewed by Claude Opus 5 with an independent second pass by Codex (gpt-5.6-sol, high reasoning effort). Every claim below was checked against the code rather than inferred from the diff.

### This does fix the regression it targets

On the bomet deployment, not one action row carries a `resource` or a `condition` key:

```
select count(*) filter (where data ? 'condition') as has_condition,
       count(*) filter (where data ? 'resource')  as has_resource,
       count(*)
from eg_mdms_data where schemacode='ACCESSCONTROL-ACTIONS-TEST.actions-test';
 → 0 | 0 | 6529
```

So `isPolicyUnconfigured` returns true and Tier-1 resolves to `UNRESTRICTED_SCOPE_POLICY`. I traced the zero-axis policy end to end: empty `getAxes()` in `ScopePolicyEngine.resolve` → empty result map → `resolveViaEngine` passes `null` for both department and jurisdiction → `PgrSearchScope(tenantId, stateLevel, null, null, null)`. The `__scope_denied__` sentinel can no longer be produced on this path. `expandDepartmentNames` handles the null input correctly, and `ScopePolicy.of(List.of(), Map.of())` performs no non-empty-axes validation, so the empty policy constructs cleanly.

Citizen self-scoping and the tenant/subtree authorization both run **before** the axis engine in `PolicyDrivenScopeResolver.resolve`, so they are genuinely unaffected as the Javadoc claims. Strict-mode symmetry with Tier-2 also holds — `allowIfBackwardCompatibilityEnabled` gates on the same `isAbacStrictMode()` flag, so both tiers fail closed together.

### 1. `isPolicyUnconfigured` reads only `scope` + `condition`, so a partially-authored action fails open

[`AccessPolicyRegistry.java:299`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/b77becb38b9db90f0965ad774791af62cf1dd7f5/backend/pgr-services/src/main/java/org/egov/pgr/policy/AccessPolicyRegistry.java#L299)

`attributes` and `scope` are siblings under `resource.complaint` (`ATTRIBUTES_KEY` / `SCOPE_KEY`, [`AccessPolicyRegistry.java:44`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/b77becb38b9db90f0965ad774791af62cf1dd7f5/backend/pgr-services/src/main/java/org/egov/pgr/policy/AccessPolicyRegistry.java#L44)). A tenant that authored field masking (`resource.complaint.attributes`) but no `scope` and no `condition` is now classified as "genuinely nothing authored" and gets fully unrestricted rows, where before it got `DEFAULT_SCOPE_POLICY` (jurisdiction `OWN`). The same applies to an action carrying a `scope` for a *different* resourceType.

**Failure scenario:** a tenant configures complaint field masking but never authors row scoping. Before this change an employee saw only their own jurisdiction's complaints with sensitive fields masked; after it they see every complaint in the tenant, still masked. That is a row-visibility widening driven by a config that only ever spoke about fields.

There is a defensible counter-argument — Tier-2 already returned allow for that action, so this makes the two tiers agree — but the PR body doesn't mention the case and no test covers it. Suggest either treating "a `resource.complaint` block exists at all" as configured, or stating the widening explicitly in the Javadoc and adding the test.

### 2. Two separate `getAction` calls — a policy found by the second is silently discarded

[`SearchAccessPolicyService.java:94`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/b77becb38b9db90f0965ad774791af62cf1dd7f5/backend/pgr-services/src/main/java/org/egov/pgr/policy/SearchAccessPolicyService.java#L94)

`getScopePolicy` fetches the action, then `isPolicyUnconfigured` fetches it again. Null results are deliberately not cached ([`AccessPolicyRegistry.java:394`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/b77becb38b9db90f0965ad774791af62cf1dd7f5/backend/pgr-services/src/main/java/org/egov/pgr/policy/AccessPolicyRegistry.java#L394) — `if (action != null) cache.put(...)`), so on the unconfigured path the second call is a real second round-trip, not a cache hit.

**Failure scenario:** the first call returns nothing (rollout window, replica lag, cache expiry between the two calls) and the second returns an action carrying a well-formed `scope`. `resolveScope` never applies that scope — the boolean only reports "configured", so the request falls to `DEFAULT_SCOPE_POLICY` instead. `_count` ([`PGRService.java:321`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/b77becb38b9db90f0965ad774791af62cf1dd7f5/backend/pgr-services/src/main/java/org/egov/pgr/service/PGRService.java#L321)) and `_search` ([`PGRService.java:169`](https://github.com/egovernments/Citizen-Complaint-Resolution-System/blob/b77becb38b9db90f0965ad774791af62cf1dd7f5/backend/pgr-services/src/main/java/org/egov/pgr/service/PGRService.java#L169)) resolve independently, so the two can disagree for the same request — the exact class of divergence this file's Javadoc says the design must prevent.

Small fix: fetch the action once and derive both the `Optional<ScopePolicy>` and the unconfigured flag from that single result.

### 3. Doubled access-control calls on the hot search path

Same root cause as finding 2, since null action results aren't cached. Per request in non-strict mode: `_count` makes 2 calls instead of 1; `_search` returning rows makes 4 instead of 3. Configured tenants are unaffected (positive results cache), and strict mode exits before `isPolicyUnconfigured`. Fixing finding 2 removes this too.

### 4. Test gap

The five new `AccessPolicyRegistryTest` cases cover no-action, bare-action, scope-present, condition-present and malformed-throws — but not the attributes-only action from finding 1, which is the case most likely to bite a tenant that has started adopting the feature.

### Not a problem (checked)

- `_count` and `_search` both route through `resolveScope`, so the tiers stay consistent by construction.
- Malformed-scope propagation is unreachable-but-harmless in `isPolicyUnconfigured`: `getScopePolicy` would already have thrown before that line is reached.
- `AccessControlUnavailableException` propagates from both `getScopePolicy` and `isPolicyUnconfigured`, so outage behaviour is unchanged.

KDwevedi (approved):

PR #1829, opened 2026-08-20
fix(pgr): keep the search context when opening a complaint's details

Hari-egov:
Two ways of reaching the complaint-details page lose the context the result list was built with. The details page then re-queries with a **narrower scope than the row the user just clicked**, and renders "No Results Found" on a complaint that was visible a moment earlier.

Frontend only — no API, workflow, MDMS or role changes.

## 1. Inbox → details dropped `createdBy`

The inbox scopes a reception officer to the complaints they created, and `PGRDetails` **already** honours a `createdBy` query param. The break was in the link, which re-derived the condition independently:

```js
// before — link decides from row data
const creatorUuid = row?.businessObject?.service?.auditDetails?.createdBy;
const query = creatorUuid && creatorUuid === currentUserUuid ? `?createdBy=…` : "";

// search decides from the role list
if (cmsRoles.includes("CMS_RECEPTION_OFFICER") && cmsRoles.every(c => c === "CMS_RECEPTION_OFFICER")) …
```

Two implementations of one rule. `auditDetails.createdBy` is not guaranteed to be present in the inbox payload, and when it was absent the param silently dropped and the details search ran unscoped.

Both sides now call one exported helper, `receptionOnlyCreatedByUuid()`, so the link and the search **agree by construction** rather than by happening to match. Applied to the table cell link and the mobile link.

## 2. Admin search → details queried the wrong endpoint

`AdminSearch` built a bare URL with no marker, so `PGRDetails` fell through to its default `usePGRSearch` → `/pgr-services/v2/request/_search`, which is **scoped to the viewer**. Admin search (`/v2/request/_admin/_search`) is not. Any result outside the viewer's own scope opened to an empty page.

`AdminSearch` now stamps `?src=admin` on the link, and `PGRDetails` queries `_admin/_search` when it sees it, keeping the normal hook for every other entry point. Both return the same `ServiceWrappers` shape — identical `Request` plumbing — so nothing downstream of the fetch changes.

## Notes for review

- `usePGRSearch`'s signature is `(searchparams, tenantId, filters, config)` — the react-query config is the **fourth** argument. Passing it third lands in `filters` and silently corrupts the query key. Called out in a comment at the call site because it is easy to get wrong.
- The two fetches are mutually exclusive via `enabled`, so only one request is ever in flight.
- Verified all three files parse under esbuild with the JSX loader. The duplicate-key warnings it reports in `UICustomizations.js` are pre-existing (lines 2645 / 2717 / 2726) and untouched here.

## Testing

Both defects are **code-proven**; I was not able to reproduce #2 end to end on `cms-pilot` because that environment currently holds only 17 complaints, all within one department, so there is no cross-department row to click. Reproducing it needs a viewer whose scope excludes a complaint that admin search returns.

Worth flagging separately while looking at this: `POST /v2/request/_admin/_search` returns **500 NullPointerException** when `RequestInfo.userInfo` is null, rather than 401 — same endpoint as the already-documented missing server-side authorization.

### DIFF
--- a/digit-ui-esbuild/products/pgr/src/configs/UICustomizations.js
+++ b/digit-ui-esbuild/products/pgr/src/configs/UICustomizations.js
@@ -110,6 +110,31 @@ function cleanObject(obj) {
   return obj;
 }
 
+/** The createdBy uuid the PGR inbox scopes itself to, or null when it does not.
+ *
+ *  Reception officers file complaints on citizens' behalf but are neither the
+ *  accountId (that's the citizen) nor ever an assignee, so an unscoped inbox
+ *  shows them nothing of their own work. Applied only when reception is the
+ *  user's SOLE CMS role — a multi-role user (e.g. reception + supervisor) keeps
+ *  the wider view their other role is entitled to.
+ *
+ *  Exported as one function so the inbox SEARCH and the row's DETAIL LINK agree
+ *  by construction. They used to decide independently: the search read the role
+ *  list, while the link compared `auditDetails.createdBy` on the row. When that
+ *  field was absent from the inbox payload the link silently dropped the param,
+ *  the details page then queried unscoped, and the officer got "No Results
+ *  Found" on a complaint they had just clicked.
+ */
+export const receptionOnlyCreatedByUuid = () => {
+  const info = Digit.UserService.getUser()?.info;
+  const cmsRoles = (info?.roles || [])
+    .map((r) => r?.code)
+    .filter((c) => c && (c.startsWith("CMS_") || c === "SUPERUSER" || c === "ADMIN"));
+  const receptionOnly =
+    cmsRoles.includes("CMS_RECEPTION_OFFICER") && cmsRoles.every((c) => c === "CMS_RECEPTION_OFFICER");
+  return receptionOnly ? info?.uuid || null : null;
+};
+
 export const UICustomizations = {
   businessServiceMap,
   updatePayload: (applicationDetails, data, action, businessService) => {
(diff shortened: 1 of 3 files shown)


Hari-egov:
## Added: "only my complaints" checkbox — and a correction to the earlier commit

### New

The inbox scoped a reception officer to the complaints they created with **no way to see anything else**. Added a checkbox to the left filter panel, **checked by default**, so the default view is unchanged and unchecking it widens the search.

- Rendered **only** when reception is the user's sole CMS role — the same condition under which the search scopes to `createdBy` at all, so the control never appears without something to control.
- An **unset** value is treated as checked, so an officer who never touches it keeps the previous behaviour.
- Label seeded as `ES_PGR_FILTER_ONLY_MY_COMPLAINTS` in `default`, `en_IN` and `pt_PT` ("Apenas as reclamações que registei").

### Correction — the detail-link change in the first commit is reverted

Implementing the checkbox showed the earlier link change to be wrong, so this commit reverts that half.

The first commit made **both** detail links carry `?createdBy=<own uuid>` for every row whenever the user was a reception-only officer. That was correct while the inbox could *only* ever show that officer's own complaints. It is **wrong as soon as the checkbox can be unchecked**: the details page applies the param as a search filter, so passing it for someone else's complaint scopes the lookup to this user and resolves to **nothing** — on a row visible in the list a moment earlier.

Both links now go back to deriving the param from the **row's creator**, which is the rule that holds whether the filter is on or off.

The shared `receptionOnlyCreatedByUuid()` helper stays — it is what the search gate and the checkbox's visibility condition agree on, so the control and the filter it drives cannot diverge.

### Also correcting the original description of this PR

I need to withdraw a claim I made when opening it. I wrote that `auditDetails.createdBy` "is not guaranteed to be present in the inbox payload, and when it was absent the param silently dropped". I did not verify that before asserting it, and testing on a local stack **disproves it**:

- `0 of 49` complaints lacked `auditDetails.createdBy`
- all `9 of 9` inbox rows carried `?createdBy=` on the pre-change bundle

So the first commit did not fix a dropped-param bug — there wasn't one — and its link change would have introduced one once this checkbox landed. What remains of that commit is worth keeping on its own terms: the shared helper removes a duplicated rule, and the admin-search fix (item 2, `?src=admin` → `_admin/_search`) is unaffected and still stands as described.

Verified: all four modified JS files parse under esbuild with the JSX loader; the three localization files are valid JSON.

PR #1834, opened 2026-08-21
fix(deploy): two deploy blockers — seed the configured bootstrap user; keep OpenBao secrets in .env across reruns

singhalkarun:
Fixes two independent deploy-killing bugs found while onboarding a new box whose host_vars override the defaults. Both were debugged and verified against a live deploy.

## Issue 1 — custom `bootstrap_user` kills every fresh deploy

Operators are allowed (and encouraged by `_example.yml`) to override `bootstrap_user`/`bootstrap_password` in host_vars to move off the well-known `ADMIN/eGov@123` defaults. But any fresh deploy that does so dies at:

```
TASK [validate — auth flow: ADMIN can mint a token]
HTTP Error 400: {"error":"invalid_request","error_description":"Invalid login credentials"}
```

Root cause (confirmed by decrypting the searched username from egov-user's logs via enc-service): the validate task mints the deploy's first token as `bootstrap_user`, but **nothing ever creates that user**:

- `seeds/user-seed.sh` hardcodes `ADMIN`/`GRO` with `eGov@123` and receives no override env.
- The playbook's own "ensure ADMIN exists" task runs ~400 lines *after* the validate, **and** is gated `when: state_root != 'pg'` — so pg-root deploys never run it at all.

### Fix (cfd55d43)

- **`digit.env.j2`** renders `BOOTSTRAP_USER`/`BOOTSTRAP_PASSWORD` into the compose `.env`.
- **`docker-compose.egov-digit.yaml`** passes them to the user-seed container as `SEED_ADMIN_USER`/`SEED_ADMIN_PASSWORD` (defaulted, so standalone compose is byte-identical to today).
- **`seeds/user-seed.sh`** creates the configured user with the full admin role set on every seed tenant, before anything validates. When the name differs from `ADMIN`, stock `ADMIN/eGov@123` is still seeded — the CI-gated tasks and Postman collections hardcode it. Distinct mobile number avoids collision with stock ADMIN; passwords are JSON-escaped.
- **`playbook-deploy.yml`** — the auth validate now fails through an assert with an actionable message for the one case seeding can't fix: the user already exists with a *different* password (creates are skipped on `DuplicateUserName`, and there is intentionally no unauthenticated password reset). Previously this surfaced as a bare HTTP 400 after 6 silent retries.
- **`_example.yml`** documents the fresh-box vs existing-box behavior.

## Issue 2 — `.env` regeneration wipes OpenBao secrets, killing every rerun

The `Write per-tenant compose .env` template task rewrites `.env` from scratch on every run, deleting the `TENANT SECRETS` / `KEYCLOAK SECRETS` blocks the OpenBao section appends later in the play. On a **rerun** the postgres role already has the vault password (from the previous run's `ALTER ROLE`), but the regenerated `.env` falls back to the `egov123` default — so `Start DIGIT stack` recreates the DB consumers with the wrong password and the play aborts at:

```
db-history-normalize | FATAL: password authentication failed for user "egov"
service "db-history-normalize" didn't complete successfully: exit 1
```

…long before the OpenBao section could re-append the block. Any box whose previous run got past the OpenBao section (and whose vault `postgres_password` ≠ `egov123`) hits this on **every subsequent deploy**.

### Fix (6155624b)

Slurp the previous `.env` before templating and restore the marked blocks verbatim immediately after. The OpenBao section still refreshes them from the vault later in the play, so secret rotation propagates exactly as before.

## Verification

- `sh -n` / `bash -n` on the seed script (it runs under busybox `sh` in `egovio/curl`); `docker compose config` renders correct env with and without overrides; YAML-parsed the playbook and compose file.
- **Issue 1, live end-to-end** on a real deploy box: ran the modified script with `SEED_ADMIN_USER=CLAUDETEST` → user created with admin roles; replayed the exact validate-task request (`/user/oauth/token`, same headers/body) → **HTTP 200 with access_token**. Test user deactivated afterwards.
- **Issue 2, functional ansible run** of the slurp → template → restore sequence: rerun restores the block byte-for-byte (special characters intact, mode 0600); fresh box with no prior `.env` is byte-identical to today's behavior.

https://claude.ai/code/session_01JkN4YEJXf3vXPP822KpdU4

> Note: a third blocker (distroless mdms-v2 image vs CMD-SHELL healthcheck) was hit while validating this PR; it was already fixed upstream by #1832 (image bump to `maven-jdk21-f3a7c13`), which is merged into this branch instead of carrying a healthcheck workaround here.


singhalkarun:
Second rerun-only bug found while validating on the live box, fixed in 6155624b:

**`.env` regeneration wipes the OpenBao secret blocks.** The `Write per-tenant compose .env` template task rewrites `.env` from scratch every run, deleting the `TENANT SECRETS` / `KEYCLOAK SECRETS` blocks the OpenBao section appends later in the play. On a rerun the postgres role already has the vault password (from the previous run's `ALTER ROLE`), but the regenerated `.env` falls back to the `egov123` default — so `Start DIGIT stack` recreates the DB consumers with the wrong password and `db-history-normalize` aborts the play with `FATAL: password authentication failed for user \"egov\"` before the OpenBao section can re-append the block. Any box whose previous run got past the OpenBao section (and whose vault `postgres_password` ≠ `egov123`) hits this on every subsequent deploy.

Fix: slurp the previous `.env` before templating and restore the marked blocks verbatim immediately after. The OpenBao section still refreshes them from the vault later, so rotation behaves exactly as before. Verified both paths with a local ansible run: rerun restores the block (special characters intact, mode 0600); fresh box with no prior `.env` is unchanged.

singhalkarun:
Third deploy blocker fixed in 078cd269 — this one predates the PR (it was the very first failure on the box) and resurfaces on every deploy because the fix people apply by hand gets overwritten when Ansible re-syncs the compose files:

**Distroless `mdms-v2` image vs CMD-SHELL healthcheck.** `egovio/mdms-v2:mdms-count-api-3ac4f8e` was republished on 2026-08-13 as a distroless image (no `/bin/sh`, no `wget`), so the `CMD-SHELL wget` healthcheck can never pass — every probe fails with `exec: \"/bin/sh\": no such file or directory` — and `dependency failed to start: container digit-mdms-backend-1 is unhealthy` kills the deploy while the app itself starts fine. It's the same trap that already cost `egov-accesscontrol` its healthcheck (see the comment there), and bomet only works via an uncommitted box-local override.

Fix: a one-shot `hc-init` service copies the **statically linked** busybox from the official `busybox:1.36.1-musl` image into an `hc_tools` volume; `mdms-backend` mounts it read-only at `/hc` and probes via exec-form `["/hc/busybox", "wget", …]`. Self-contained for both the Ansible tier and standalone compose — no binary in git, no host-path mount that breaks when the file is absent. (An alpine busybox wouldn't work: it's dynamically linked against musl, which distroless lacks.)

Verified live on a deploy that was failing exactly this way: after applying, `digit-mdms-backend-1` reports `(healthy)`.

Follow-up candidate (not in this PR): the same `hc_tools` mount would let `egov-accesscontrol` get a real healthcheck back.

singhalkarun:
Dropped 078cd269 (busybox healthcheck workaround) — the mdms-backend unhealthy failure was already properly fixed by #1832 (image bump to `maven-jdk21-f3a7c13`, which has a shell + wget again). Merged current master into this branch instead; verified on the live box that the new image passes the existing CMD-SHELL healthcheck.

vinothrallapalli-eGov (approved):

Issue #1514, opened 2026-07-30
Decide: who may call the /v2/analytics endpoints (auth posture + roles)


singhalkarun:
## The question

`/pgr-services/v2/analytics/_query`, `catalog/_search`, and `packs` currently answer **without any login**; protection relies on the ABAC scope resolver degrading anonymous callers to the PUBLIC-floor catalog. That may be a legitimate product choice (a public-transparency floor) — but it has never been decided.

The decision becomes forcing once gateway RBAC is enforced: these APIs need either a permissions-table entry (who can call them) or an explicit place on the no-login whitelist.

## Where this is tracked

This is **decision D5** in the gateway auth/RBAC enablement pack (PR #1255, `docs/gateway-rbac/`). Proposed default there: superuser + employee roles per the `dss.DashboardConfig` role gate — i.e. *not* anonymous.

## History

Originally filed as #1279, closed unresolved into an off-repo personal backlog (KDwevedi#27). Re-filed here so the decision lives with the org repo and blocks the enforce flip visibly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1837, opened 2026-08-21
Enable Kong gateway RBAC, add ACCOUNT_ADMIN role, fix deploy ordering bugs

vinothrallapalli-eGov:
## Summary
- Enables `ENFORCE_UNAUTH`/`ENFORCE_RBAC` in `local-setup/kong/kong.yml` (previously audit-only) and closes the `actiontest`/`roleaction` coverage gaps that would otherwise deny legitimate traffic once enforced: PGR actions (migration, inbox search/count, dashboard, analytics), `ACCESSCONTROL-ACTIONS.actions`, `common-masters.UserValidation`/`MobileNumberValidation`, `ThemeConfig`, `adminSchema`, `DashboardConfig`, both `ComplaintHierarchy*` schemas, and `novu-bridge/*` (mapped to `ACCOUNT_ADMIN`, `SUPERUSER`, `MDMS_ADMIN`, `CSR`, `GRO`, `PGR_LME`).
  - **Known accepted gap:** `/common-persist` has no `actiontest` coverage yet — deliberately left for a follow-up, since persister isn't required for the flows this change targets.
- Adds a new `ACCOUNT_ADMIN` role (238 action mappings covering the Configurator's full resource surface plus tenant-bootstrap actions), granted to the bootstrap `ADMIN` user *alongside* its existing roles (replacing them breaks `digit-mcp`'s own hardcoded bootstrap-verification check, which does a subset check against specific role codes).
- Fixes a real `.env` templating ordering bug in `playbook-deploy.yml`: OpenBao secrets were read and written into `.env` *after* the first `docker compose up`, so redeploying an already-provisioned tenant started Postgres/MCP-Postgres with a password-less `.env` while the volume already held the real password — failing `db-history-normalize`/`digit-mcp` on every redeploy. `Secrets prep` now runs before the stack ever starts.
- `mdms-v2`/`egov-user` image bumps; restored a working healthcheck now that the new `mdms-v2` image has a real shell (the previous pin was a distroless build with no shell at all).
- Kong's `AUTH_OPTIONAL` must never include `_createnovalidate` (unauthenticated arbitrary user/role creation) — the two bootstrap callers that used it now hit `egov-user` directly over the docker network instead, matching the existing k8s reference deploy's pattern.
- Fixed a Redis cache-key bug in the tenant-bootstrap mobile-validation cache-bust: it deleted a key (`validationRules`) that doesn't exist; the real key is `egov-user:mobile-val:<tenant>:<countryCode|default>`.
- `digit-mcp` holds long-lived connections to `egov-user`/`egov-hrms` and needs to be recreated after they restart during post-bootstrap, or its own tenant-bootstrap calls fail.
- Converted two `ansible.builtin.uri` + `until`/`retries` tasks to plain shell retry loops after hitting an ansible-core edge case where the registered result's own data showed success but `failed_when` still fired.

## Test plan
- [x] `sudo ./deploy.sh <tenant>` runs end-to-end to completion (`failed=0`) on a redeploy of an already-provisioned tenant, including the mobile-validation and INTERNAL_USER/ADMIN re-provisioning steps.
- [x] Verified `egov-accesscontrol`'s `/access/v1/actions/_authorize` returns 200 for `ACCOUNT_ADMIN`, `SUPERUSER`, `MDMS_ADMIN`, `CSR`, `GRO`, `PGR_LME` against a `novu-bridge` endpoint.
- [x] Verified the Configurator UI (`/configurator/login`) loads correctly end-to-end through Kong once `nginx_features.configurator` is enabled for a tenant (tenant-local config, not part of this diff).
- [x] `ansible-playbook --syntax-check` and a plain YAML parse pass on `playbook-deploy.yml`.
- [ ] `/common-persist` RBAC coverage — tracked as a follow-up, not required for this change.


singhalkarun:
Checked this against the enablement pack in #1255, and re-ran the Appendix B log audit.

**Audit corpus:** `bomet` kong-gateway log, 2026-08-17 15:44 → 2026-08-24 23:51 UTC (7.3 days, no overlap with prior windows). 12,083 `RBAC-audit` lines — 12,019 would-403, 64 would-401, 67 distinct URIs. Replayed against this branch's `full-dump.sql`:

- **8,506 of 12,019 would-403s clear (71%)**, across 15 URIs. Largest: `/mdms-v2/v2/_count` 5,651 (47% of the corpus, and absent from #1255's sheet — this PR adds it), `cache-bust` 1,120, `ComplaintHierarchy{,Definition}` 1,201, analytics 254.
- **3,513 remain**, across 48 URIs.

## R3

2,786 of the 3,513 residual denials come from one principal: `AUTO_ESCALATE, CITIZEN, CSR, DGRO, EMPLOYEE, GRO, PGR_LME, PGR_VIEWER, SUPERUSER, SUPERVISOR`. It holds `SUPERUSER` but not `MDMS_ADMIN`; this PR grants `MDMS_ADMIN + ACCOUNT_ADMIN` on the MDMS write paths, so it still 403s.

**If R3 (`SUPERUSER` passes everything) were implemented, 3,441 of the 3,513 clear — 98%**, leaving 72 hits across 4 URIs. R3 is not implemented here: `SUPERUSER` covers 152 of 358 actions in the dump.

`ACCOUNT_ADMIN` appears in 0 requests in the window — it lands at tenant bootstrap/re-provision, so it can't unblock traffic on an already-provisioned tenant.

## Endpoints with no permission row

28 protected POST paths from the sheet still have no `actiontest` row. Under `ENFORCE_RBAC=true` the gate fails closed — 403 for everyone including `SUPERUSER`:

- D4 (7): boundary create/update/delete + the four `egov-bndry-mgmnt` endpoints
- D14 (3): `/xstate-chatbot/{message,reminder,status}`
- D15 (3): workflow `businessservice/v2/*` · D10 (3): novu dispatch dry-run/test-trigger/validate
- D9 (2): `/user-preference/v1/{_search,_upsert}`
- D8 `/inbox/v1/_search` · D11 `/localization/messages/v1/_delete` (46 live denials) · D12 `/egov-url-shortening/shortener` · D13 `_nearingslacount`
- E1 `/egov-enc-service/crypto/v1/{_decrypt,_generatekey}` — `_generatekey` has 295 live denials
- Not in the sheet: `/mdms-v2/v2/_create/RAINMAKER-PGR.MapConfig` (181 live denials)

Most produced no warnings only because bomet doesn't call them; per Appendix B, a quiet log is not proof.

## Pack items still open

| item | status |
|---|---|
| R1 (53 APIs), R2 (widen 20) | already on `master` before this PR — 0 added here; 8 R2 rows still lack `SUPERUSER` |
| R3 | not implemented |
| D1–D16 (44 decisions) | ~17 seeded; sheet's DECISION column is empty, so chosen implicitly — including D5/analytics (#1514) |
| E1 (9) | 6 of 9 seeded |
| E2 (POST-only check) | untouched — `if kong.request.get_method() ~= "POST" then return end` |
| E3 (`/egov-location`) | untouched — route-scoped `pre-function` at kong.yml:427 still shadows the global one |
| E4 (`AUTH_OPTIONAL` entries) | untouched — byte-identical; `["/user/_search"]=true` still present |
| Appendix B re-verification | not run |

## Three specifics

1. `/kc` and `/auth` are not in `AUTH_OPTIONAL`. The `pre-function` is global, so a form-encoded POST there fails `cjson.decode` and hits `deny_unauth()` — SSO 401s when `ENFORCE_UNAUTH` flips. (Raised by @KDwevedi on #1255.)
2. The Helm seed copy wasn't updated: `devops/deploy-as-code/charts/backbone-services/db-seed/files/full-dump.sql.gz` is at 199 URLs / 369 grants vs `local-setup`'s 309 / 823. R1 requires both.
3. The new kong.yml comment says `PGR_ADMIN` is missing from `ACCESSCONTROL-ROLES.roles` — it's present in `full-dump.sql` on both `master` and this branch.

## Suggestion

Land the `ACCOUNT_ADMIN` role, the 402 seed rows, the `_createnovalidate` removal, the `.env`/OpenBao ordering fix, the Redis cache-key fix and the image bumps. Hold the two `ENFORCE_*` booleans until R3 lands, the 28 endpoints are seeded, `/kc` is classified, the Helm dump is synced, and Appendix B has been run.

vinothrallapalli-eGov:
@singhalkarun, the MDMS ADMIN role will be mapped only to APIs related to MDMS masters. Other APIs, such as Boundary Management (create/update) and Localisation (upsert/search), will not be mapped to this role.

That’s why I have introduced a new ACCOUNT_ADMIN role. Users with this role will have access to all configurator-enabled features, including onboarding and other screens.

So, please try using a user with the ACCOUNT_ADMIN role. I will also re-verify the above-mentioned APIs that are currently missing for the ACCOUNT_ADMIN role.

The other existing roles will continue to have limited access based on the permissions assigned to those roles. For ex: CSR, LME, CITIZEN, GRO, etc..

singhalkarun:
@vinothrallapalli-eGov you're right — correction on my side. I re-ran the audit across all three Kong deployments (bomet, naipepea, helsinki; 54,535 would-403s, 106 would-401s). **With `ACCOUNT_ADMIN` on the caller, 98.85% of denials clear.** My earlier "71% / R3 is the blocker" framing came from bomet alone and was not representative — the seed rows alone clear 16% across all three, and R3 lands within 556 requests of your approach out of 54,535. Dropping R3 as a blocker; your design holds.

I also withdraw my claim that the `egov-bndry-mgmnt` rows are on the wrong path — see item 3, it's unresolved rather than wrong.

Here's what's left, all verified against the three deployments' live seed and traffic:

### 1. Seed rows missing entirely (628 denials, no `actiontest` row so no role can reach them)

```
298  /egov-enc-service/crypto/v1/_generatekey                  (sheet E1)
181  /mdms-v2/v2/_create/RAINMAKER-PGR.MapConfig
 80  /mdms-v2/v2/_create/dss.KpiDefinition
 46  /localization/messages/v1/_delete                         (sheet D11)
  4  /mdms-v2/v2/_search/Workflow.BusinessServiceMasterConfig
  4  /mdms-v2/v2/_create/dss.DashboardPack
  3  /mdms-v2/v2/_update/dss.KpiDefinition
  3  /mdms-v2/v2/_create/common-masters.FormValidations
  2  /mdms-v2/v2/_search/common-masters.FormValidations
  2  /mdms-v2/v2/_search/common-masters.MobileNumberValidation
  1  /egov-mdms-service/v2/_search
  1  /egov-mdms-service/v2/_count
  1  /access/v1/actions/search
  1  /mdms-v2/v2/_search/ke
  1  /mdms-v2/v2/_search/RAINMAKER-PGR.MapConfig
```

### 2. Configurator surface gaps for `ACCOUNT_ADMIN`

Localisation is fully covered except `_delete` (above). Boundary is partial — `boundary/_create` and `boundary-hierarchy-definition/_create` have `ACCOUNT_ADMIN`, but these have **no row on any deployment**:

```
/boundary-service/boundary/_update
/boundary-service/boundary-relationships/_update
/boundary-service/boundary-relationships/_delete
```

### 3. Boundary Management — needs a decision, then a grant

The live seed carries `/egov-bndry-mgmnt/v1/_{generate,generate-search,process,process-search}` granted to `MDMS_ADMIN` only. #1255's source-derived sheet says the gateway path is `/egov-bndry-mgmnt/boundary-management/v1/_*` and marks the `/v1/` form as matching no real endpoint. Kong matches exactly, so at most one works — and there are **zero `/egov-bndry-mgmnt` requests in any Kong log on any deployment**, with the service returning 400 for every path including invented ones, so neither traffic nor probing settles it. Worth confirming from the service source, then adding `ACCOUNT_ADMIN` to whichever path is real.

### 4. `AUTH_OPTIONAL` additions

- `/auth/callback` — **11 observed would-401s on naipepea** (08-20, 08-21). Keycloak SSO callback carries no `RequestInfo` envelope, so it 401s the day `ENFORCE_UNAUTH` flips.
- `/otp/v1/_send` — 1 would-401 on naipepea. The table has `/user-otp/v1/_send` and `/otp/v1/_validate` but not this.
- `/kc` — unexercised on these boxes but structurally identical to `/auth/callback`.

### 5. Two notes for the PR description

- `ACCOUNT_ADMIN` has **0 role rows and 0 `eg_userrole_v1` grants on all three deployments** today. Existing tenants only benefit after their operator is re-provisioned — worth saying explicitly so nobody flips the flags on an existing tenant first.
- The Helm `db-seed` dump is still at 199 URLs / 369 grants vs `local-setup`'s 309 / 823.

E2 (POST-only enforcement), E3 (`/egov-location` route-scoped plugin shadowing the global one) and E4 are untouched here and belong in a separate change.

<sub>Method: `RBAC-audit` lines from each deployment's kong-gateway log incl. rotated files, parsed for `uri=`, caller `roles=[]`, `tenantId`, replayed against this branch's seed. Seed and role state read from each `docker-postgres`.</sub>

vinothrallapalli-eGov:
@singhalkarun The following are not valid API's, the mdms search API is one and schemacode will send in body unlike create and update where it send in URL
  1  /mdms-v2/v2/_search/ke
  1  /mdms-v2/v2/_search/RAINMAKER-PGR.MapConfig


/egov-bndry-mgmnt/v1/* -- these are valid APIs

singhalkarun (approved):
LGTM. Re-verified `d05cd06` against all three deployments — the seed side is closed.

**Residual denials, replaying 54,535 would-403s against the new dump with `ACCOUNT_ADMIN` on the caller:**

| deployment | would-403 | residual before | residual now |
|---|---:|---:|---:|
| bomet | 12,019 | 530 | 1 |
| naipepea | 40,105 | 11 | 0 |
| helsinki | 2,411 | 87 | 0 |

628 → 1. You covered everything on the list plus `boundary/_delete`, novu dispatch, `user-preference` and `_decrypt`. Integrity checks pass: no duplicate uniquenames, action ids or `(role, action)` pairs, no orphaned roleactions, consistent 10-column format.

The single remaining hit is `/mdms-v2/v2/_search/ke` — a caller passing a tenantId where a schemaCode belongs. That's a client bug, not a seed gap; **don't add a row for it.**

Two corrections to things I raised earlier, both checked against gateway traffic and both non-issues:

- **`/egov-enc-service/crypto/v1/_encrypt` has no role grant, and that's fine.** Zero `_encrypt` requests cross Kong on any deployment — only `_generatekey` (298 bomet / 7 helsinki) and health checks do. It's a service-to-service call over the docker network. Worth a grant for consistency someday, not a blocker.
- **`/auth/callback` is not an SSO break here.** All 11 hits on naipepea returned 503 because naipepea and helsinki run no `keycloak` or `token-exchange-svc` container — the route has no upstream. They appear alongside `/auth/.env`, `/auth.yml?v=…` and `/auth1.js` probes, so they're scanners. bomet, the one box actually running Keycloak, shows no `/auth/callback` traffic through Kong. @KDwevedi's structural point still stands for whenever SSO is fronted by the gateway, but it doesn't block this change.

Also withdrawing my earlier request for an E-row on `/mdms-v2/v2/_search/<suffix>`: enumerating the real suffixes across all three boxes, they're schema codes, so seeding them per-schema — what you did — is the right pattern.

Non-blocking follow-ups, unchanged from before: the Helm `db-seed` dump is still at 199 URLs / 369 grants vs `local-setup`'s 333; the `egov-bndry-mgmnt` path question (`/v1/_*` vs `/boundary-management/v1/_*`) is unexercised on every deployment and wants settling from the service source; and worth a line in the PR description that `ACCOUNT_ADMIN` has 0 grants on existing tenants until their operator is re-provisioned.

KDwevedi (approved):
approving given @singhalkarun has reviewed

PR #1838, opened 2026-08-21
feat(dashboard): public page gets filters, Add KPI/Reset and a language switcher (#1797)

### LINKED ISSUE #1797: [Bug] Public Dashboard missing context and filters
# Description

What went wrong? Be specific.
In the public dashboard (https://bometfeedbackhub.digit.org/digit-ui/public-dashboard.html), the top bar with filters (Start Date, End Date, Ward, Complaint Type etc, as in the employee dashboard) are not present. 

---

# Expected Behavior

The top bar should be with with the filters, Clear Button and as well as the Add KPI button and Reset Button above it (Screenshot below from the employee dashboard)

The provision to change language is also not present

Image

---

dhruv-1001:
## Summary

Closes #1797 (P0). The public dashboard (`/digit-ui/public-dashboard`) now has the same top bar as the employee dashboard — **Start/End date, Ward, Complaint type, Clear**, plus **Add KPI** and **Reset** — and an in-page **language switcher**.

The public page is the employee dashboard mounted with `mode="public"`; #1703 deliberately stripped it to a zero-input page in three places (UI hidden behind `readOnly=publicMode`, bare `{kpiId}` refs, and a backend that rejected any `params`). This PR re-opens each layer behind a fixed allow-list rather than removing the boundary.

## Backend — public APIs (`pgr-services`)

| endpoint | change |
|---|---|
| `POST /v2/analytics/public/_query` | accepts `{kpiId[, params]}`; `params` rebuilt from the allow-list **`dateFrom`, `dateTo`, `ward`, `serviceCode`, `complaintPath`** (scalar, ≤128 chars, dates ISO) — anything else is `400 invalid_param`. Reachable KPIs = every published def tagged `PUBLIC` (the pack remains the fail-closed enablement gate). No `compare`/`series`/`hierLevel`/`window`. |
| `POST /v2/analytics/public/catalog/_search` | **new** — PUBLIC catalog as safe descriptors; feeds the Add KPI menu. |
| `POST /v2/analytics/public/_options` | **new** — ward / complaint-type codes that carry complaints, from two server-owned distinct queries under the anonymous scope; counts stripped; caller sends only `tenantId`. |

RequestInfo is still discarded by construction on all four aliases; all fail closed when `publicDashboardEnabled` is not `true`.

**Gateway:** both new paths are Kong auth-optional with the same 60/min/IP + 1 MB limits as the existing two. `check-gateway-whitelist-parity.py` and `infra-contracts.test.ts` updated; the K8s Spring gateway deliberately stays closed (it does not strip spoofed `userInfo`).

## Frontend (`products/dashboard`)

- Filter bar, Clear, Add KPI, Reset, drag/resize/remove render in public mode; `buildPublicRefs` carries `globalParams(filters)`; `useCatalog` reads the public catalog; `useFilterOptions` reads `/public/_options` (same envelope shape, so the option builder is shared); the complaint-type tree loads anonymously the way `boundaryService` already did.
- Visitor layout/filters persist under **public-only keys** (`…catalog-layout.v1.<tenant>.public`, `…-filters-v4-public`) — disjoint from every employee slot.
- `LanguageMenu` in the standalone/public header, fed by `common-masters.StateInfo.languages` (same master as the TopBar). `localeRuntime.setLanguage()` switches in place and remembers the choice under `ccrs.dashboard.public-locale.v1`.
- Two latent bugs fixed on the way: the public page was reading `Employee.locale` (inheriting a co-hosted employee session's language), and the standalone boundary bundle was hardcoded to `rainmaker-boundary-admin` instead of following `HIERARCHY_TYPE`.
- Header shows the `<ward> · <period>` context line in every mode; "Public view" survives as a pill.

## Seeds & docs

- `DASHBOARD_HEADER_LANGUAGE` added to the generated l10n packs (`dashboard-l10n-seed.ts` → `export-l10n.mjs --check` clean).
- `docs/dashboard-configuration/20-packs-and-rbac.md`: the four public aliases, the param allow-list, and "tag `PUBLIC` → Add KPI menu; list in `public-default` → on by default".
- `docs/dashboard-configuration/90-localization.md`: the switcher, its storage key, and the **bomet caveat** — `ke` declares `en_IN` + `sw_KE`, but only `en_IN`/`pt_PT` dashboard packs exist, so Swahili renders the dashboard chrome in the English fallback until an `sw_KE` pack is upserted (data, not code; `enable-dashboard.sh` already warns on this).

## Out of scope (not in the issue)

Prior-period deltas / sparklines on public cards, per-widget Group-by, the employee sidebar. Also pre-existing and untouched: the public page computes default dates in the Nairobi fallback zone because it cannot read `dss.DashboardConfig.timeZone` — worth its own ticket.

## Verification

- `pgr-services`: 168 analytics tests green (`AnalyticsController*`, `AnalyticsService*`, `KpiCatalogService*`, `PrincipalScopeResolver*`, `KpiQueryComposer*`) — `AnalyticsControllerPublicTest` rewritten (17), new `AnalyticsServicePublicOptionsTest` (3).
- dashboard: 255 Node tests green; `npm run build` emits both entries.
- `check-gateway-whitelist-parity.py` (+ `--self-test`), `infra-contracts.test.ts`, `export-l10n.mjs --check`, `git diff --check` all clean.
- Mutation-checked: re-hiding the filter bar, dropping params from the public plan, leaking `Employee.locale`, and removing `ward` from the backend allow-list each turn the corresponding test red.
- **Not yet done:** a live browser pass on a composed environment (local stack was down). Suggested smoke: open the public page → filter bar present, changing Ward/dates re-queries; Add KPI lists the PUBLIC tiles; language menu lists StateInfo languages and switches in place; a hand-crafted `/public/_query` with `params:{hierLevel:"1"}` or a non-PUBLIC kpiId returns 400; DevTools shows no `authToken`/`userInfo` on any request.

### DIFF
--- a/.github/scripts/check-gateway-whitelist-parity.py
+++ b/.github/scripts/check-gateway-whitelist-parity.py
@@ -38,6 +38,8 @@ _ENTRY = re.compile(r'\["(/[^"]+)"\]\s*=\s*true')
 KONG_ONLY_AUTH_OPTIONAL = {
     "/pgr-services/v2/analytics/public/packs",
     "/pgr-services/v2/analytics/public/_query",
+    "/pgr-services/v2/analytics/public/catalog/_search",
+    "/pgr-services/v2/analytics/public/_options",
 }
 
 
--- a/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsController.java
+++ b/backend/pgr-services/src/main/java/org/egov/pgr/analytics/AnalyticsController.java
@@ -2,6 +2,7 @@ package org.egov.pgr.analytics;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import lombok.extern.slf4j.Slf4j;
 import org.egov.common.contract.request.RequestInfo;
@@ -27,6 +28,13 @@ import java.util.stream.Collectors;
  *   POST /v2/analytics/packs           — return best-match DashboardPack + safe tile descriptors
  *   POST /v2/analytics/catalog/_search — return all visible KpiDefinition tiles (no query/rbac)
  *
+ * Anonymous twins (Kong-only auth-optional aliases; RequestInfo is discarded by construction):
+ *   POST /v2/analytics/public/packs           — curated PUBLIC pack, fail-closed when disabled
+ *   POST /v2/analytics/public/catalog/_search — every PUBLIC-tagged published tile (Add KPI menu)
+ *   POST /v2/analytics/public/_options        — filter-bar option codes (wards / complaint types)
+ *   POST /v2/analytics/public/_query          — {kpiId[, params]} refs over PUBLIC tiles, with
+ *                                               params restricted to PUBLIC_QUERY_PARAMS
+ *
  * Body (single):  { "RequestInfo": {...}, "tenantId": "ke", "query": { ...grammar... } }
  * Body (batch):   { "RequestInfo": {...}, "tenantId": "ke", "queries": { "name": {...}, ... } }
  */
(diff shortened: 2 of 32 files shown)


dhruv-1001:
@claude review

KDwevedi (changes requested):
Second-pass review against the current head (`357b3a59`). Two merge-blocking issues remain:

1. **[P1] Public baked filters can still be displaced by declared defaults.** In `resolveKpiRef`, `withoutBakedNarrowings(...)` runs before `withDeclaredDefaults(...)`. If a public KPI bakes (for example) `filters.service_code` and also declares a non-empty `serviceCode` default, the guard removes the caller value, the default is then reintroduced, and `KpiQueryComposer.applyEqFilter` overwrites the baked `eq`. Please apply the baked-filter protection to the effective params (including defaults), and add a regression covering a baked predicate plus a conflicting declared default.

2. **[P2] One-sided public date ranges are accepted and silently ignored.** `sanitizePublicParams` validates each date only when present, while `KpiQueryComposer` applies a range only when both are present. Thus `dateFrom`-only or `dateTo`-only requests execute the unbounded/base KPI instead of returning `invalid_param`. Please require the two fields as a pair and cover both one-sided cases.

The long complaint-path behavior is tracked in #1862 and the broader privacy/filter-granularity question in #1863; those are follow-ups and are not additional blockers in this review.

The currently failing Gatus/Tilt/local-stack checks appear infrastructure-related and are not the reason for this request-changes decision.

Issue #1836, opened 2026-08-21
[Bug] Powered by DIGIT missing or broken across multiple pages


rahuljose-egov:
Description

The "Powered by DIGIT" logo is broken across multiple pages — it either renders as a broken-image icon with only the alt text ("Powered by DIGIT") visible, instead of the actual logo or is not available at all. Confirmed visually on the Employee Login screen (e.g. Bomet County tenant) across all of its states (blank, city selected, filled), the configurator, employee dashboard and the public dashboard. 
-----

Expected Behavior

The DIGIT logo should render correctly next to "Powered by DIGIT" text on **every page** (Across the CMS sites, Dashboard and Configurator), for every role and user.

---

Ask

- Audit all pages pertaining to CMS, across all active tenants, to confirm which ones are actually affected (not just the login screen already confirmed).
- Add the "Powered by DIGIT" footer along with the logo. 

@dhruv-1001 Check if an automated test for the  "Powered by DIGIT" footer can be added against all future PR's.

<img width="1415" height="870" alt="Image" src="https://github.com/user-attachments/assets/c6c145c1-80a8-45db-a617-6a5f45559d8c" />

<img width="1470" height="882" alt="Image" src="https://github.com/user-attachments/assets/a7d872f9-173e-411c-87ef-7b32e4967289" />

<img width="1464" height="874" alt="Image" src="https://github.com/user-attachments/assets/e81de3fd-1ffe-4a97-9386-179f141da55e" />

---

## Split into (added by @KDwevedi during triage — original report above unchanged)

Three surfaces, three different root causes, so this is tracked as three pieces. Details in [the triage comment](https://github.com/egovernments/Citizen-Complaint-Resolution-System/issues/1836#issuecomment-5370056324).

| | Scope | Cause | State |
|---|---|---|---|
| #1839 | **CMS** (citizen + employee) | `footer_logo_url` defaulted to `""` on every Ansible deployment → `<img src="">` | PR open |
| #1840 | **Dashboard** | `DashboardFooter` never ported from the legacy tree to the shipping esbuild product | PR open, stacked on #1839 |
| #1841 | **Configurator** | No attribution anywhere; separate codebase | Ticket open, @prtkjakhar |

This issue stays open until all three land. #1839 is the prerequisite — it commits the logo in-repo and supplies the non-empty default, so #1840 and #1841 render nothing without it.

@dhruv-1001's automated-test ask is covered in the triage comment.

[...]

PR #1839, opened 2026-08-21
fix(ui): ship the "Powered by DIGIT" asset in-repo so the footer renders on a stock install

KDwevedi:
Fixes the broken/missing **Powered by DIGIT** attribution reported in #1836, for the CMS surfaces (citizen + employee). Dashboard and configurator are tracked separately — see "Not in scope" below.

## Root cause

`footer_logo_url` / `footer_bw_logo_url` defaulted to `""` in `group_vars/digit.yml`, and **all six** shipped host_vars examples re-set them to `""`. So the Ansible-rendered `globalConfigs.js` carried `var footerLogoURL = "";` on every deployment, and every consumer got ``.

Verified live before the fix:

```
$ curl -s https://<bomet>/digit-ui/globalConfigs.js | grep footer
24:  var footerBWLogoURL = "";
25:  var footerLogoURL = "";
```

Naipepea is identical. One cause, two symptoms — which is why the report describes both:

- **Broken-image icon + alt text** on pages that render the `` unconditionally: `Login`, `Otp`, `ForgotPassword`, `ChangePassword`, `SignUp`, `SignUp-v2`, `Login-v2`, `LanguageSelection`, `ViewUrl`, `common/Login`, `citizen/index`.
- **Nothing at all** on the employee shell (`employee/index.js:159`), which already guards on the value.

The checked-in `globalConfigs` files *did* carry a value, but pointed at `s3.ap-south-1.amazonaws.com/egov-uat-assets/` — a UAT asset bucket that no production tenant should be depending on.

## Fix

1. **Commit the logo in-repo** under `digit-ui-esbuild/public/brand/` — the colour variant plus the white-on-transparent one used on the dark login banner. The build copies `public/` into the docroot, so it serves at `/digit-ui/brand/…`.
2. **Default every config to that root-relative path** — the Ansible `group_vars`, plus the five checked-in `globalConfigs` variants. Nothing deployment-specific, no external asset host, renders on a fresh install with zero configuration, works on any domain.
3. **Stop the examples from disabling it.** The six host_vars examples now carry a commented-out rebrand line instead of an `""` that silently switched the footer off. Documented explicitly: `""` means *no footer*, not *default*.
4. **Guard `ImageComponent` against an empty `src`.** An `` with no src is always a broken-image icon and never a useful render, so an unset config should render nothing rather than the broken glyph. One guard covers all 29 call sites, so this failure mode cannot come back through a different config key.

## Verification

- Ran the build's **actual** public-asset copy step (sliced verbatim out of `esbuild.build.js`, not reimplemented): emits `build/brand/digit-footer.png` (13317 B) and `build/brand/digit-footer-bw.png` (7778 B), byte-identical to source.
- Rendered `globalConfigs.js.j2` through Jinja with the new `group_vars` → `var footerLogoURL = "/digit-ui/brand/digit-footer.png";`, and `node --check` passes on the output.
- `node --check` clean on all five edited `globalConfigs` files; `ImageComponent.js` compiles clean under esbuild with the guard present in the output.
- All six host_vars examples parse and now inherit the default rather than overriding it.

## Not in scope (separate work off #1836)

- **Dashboard** — `DashboardFooter.jsx` exists only in the legacy `frontend/micro-ui/web/src/dashboard/` tree and was never ported to the shipping `digit-ui-esbuild/products/dashboard/`. Deployed bundle contains `dashboard-footer` 0×. Follow-up PR.
- **Configurator** — no attribution anywhere; the wizard footer is text-only and the management shell and login page have none. Deployed bundle contains `Powered by DIGIT` 0×. Separate ticket.

Refs #1836


KDwevedi:
Superseded by #1870, recreated from current `master` with only the intended change set.

